PR #7, opened 2016-04-10
Added round and limit to zoom steps

jespino:


src/uxbox/data/workspace.cljs
```diff
@@ -249,7 +249,7 @@
 (defrecord IncreaseZoom []
   rs/UpdateEvent
   (-apply-update [_ state]
-    (let [increase #(* % 1.05)]
+    (let [increase #(min (/ (Math/round (* % 105)) 100) 3)]
```
niwinz: Please use the `uxbox.util.math` namespace for math interop. It already contains functions for rounding and similar stuff.

PR #10, opened 2016-04-11
Insert shapes alwais on top

jespino:


src/uxbox/state/shapes.cljs
```diff
@@ -10,7 +10,7 @@
   (let [sid (random-uuid)
         shape (merge shape {:id sid :page page})]
     (as-> state $
-      (update-in $ [:pages-by-id page :shapes] conj sid)
+      (update-in $ [:pages-by-id page :shapes] #(cons sid %))
```
niwinz: `cons` does not return vector but `:shapes` property should be always a verctor, maybe `#(into [] (cons sid %))` ?

jespino: fixed! :smile:

PR #14, opened 2016-04-15
Fix blocked groups selection

jespino:


src/uxbox/data/shapes.cljs
```diff
@@ -305,7 +304,8 @@
             xf (comp
                 (filter #(= (:page %) pageid))
                 (remove :hidden)
-                (remove :blocked)
+                (remove #(and (not (:blocked %)) (= (:type %) :builtin/group)))
+                (remove #(and (not= % (sh/resolve-parent %)) (:blocked (sh/resolve-parent %))))
```
niwinz: Resolve parent is a "expensive" operation so I think executing it twice for shape is not good approach. I think the best approach here is define the transducer out of the event, define that concrete functions as private predicates and later use that transducer in the event.

PR #15, opened 2016-04-15
Adding group/degroup functionality with shortcuts

jespino:


niwinz:
In general it seems nice!
But, would be nice also put the main logic of that event into a function inside uxbox/state/project.clj and make this event use that function. This function has a lot of testeable logic and having it as a pure function that receives state and some parameters is a big win!

PR #21, opened 2016-04-17
Draw tools shortcuts

jespino:


src/uxbox/ui/workspace/shortcuts.cljs
```diff
@@ -33,6 +33,22 @@
    :ctrl+c #(rs/emit! (dw/copy-to-clipboard))
    :ctrl+v #(rs/emit! (dw/paste-from-clipboard))
    :ctrl+z #(rs/emit! (udh/backwards-to-previous-version))
+   :ctrl+b #(rs/emit! (dw/select-for-drawing
+     {:type :builtin/rect
+      :name "Rect"
+      :stroke "#000000"}))
+   :ctrl+e #(rs/emit! (dw/select-for-drawing
+     {:type :builtin/circle
+      :name "Circle"}))
```
niwinz: This seems like a not pretty good approach, having this hardcoded here and in other location (sidebar/drawtools.clj). Maybe declaring some constants in `sidebar/drawtools.clj` and then, use them here can be better approach.

jespino: Completely agree, I will change it

jespino:
Fixed! :smile:

PR #78, opened 2017-03-09
Add text outline to ruler dimensions.

### LINKED ISSUE #75: Measure tool needs text background
Measurements cannot be seen if the background and measurement text are the same colour. A (semi-transparent) text background would fix this.
ruler4

studiospring:
First time, so I hope I'm doing this right ;-)
I have fixed #75 in a different way to @niwinz, by adding a text outline. I think this is a cleaner design. Sorry I couldn't do this before you got your solution finished. Let me know what you think.
outline

Signed-off-by: Sean Loughman <sean@studiospring.com.au>

### DIFF
--- a/frontend/src/uxbox/main/ui/workspace/ruler.cljs
+++ b/frontend/src/uxbox/main/ui/workspace/ruler.cljs
@@ -32,22 +32,22 @@
                      (mth/precision 2))
         angle (-> (gpt/angle pt center)
                   (mth/precision 2))
-        transform1 (str "translate(" (+ (:x pt) 35) "," (- (:y pt) 10) ")")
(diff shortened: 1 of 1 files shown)


studiospring:
Now uses canvas colour for outline.
outline2

niwinz:
What do you think about this @elhombretecla 

From my point of view it looks nice, but the final decision is yours.

niwinz:
In any case, don't merge it. I want make some code style tweaks before ;)

elhombretecla:
Thank you so much for this issue and for your work @studiospring

I'm agree with you. We need to improve this design. Your proposal is good but I think that we have to try to continue with the visual hierachy. Thats why I think this tooltip design is more apropiate with the other ones.

uxbox ruler2

studiospring:
I see what you mean @elhombretecla. It would be nice if the background expanded with the text though!

PR #97, opened 2019-05-21
Docker containers, bug fix and translations

madmath03:
This PR is a compilation of all the work we did on our side while waiting for the developments here to restart.
Everything is not finished but at least nothing that was previously working has been broken so far.
This was originally created from the `master` branch but I'm sending this to your `wip` branch if you want to review/change it on your side before merging to master

Content of this PR (might not be exhaustive): 
* :sparkles: Introduction of "production-ready" docker containers for backend and front end. Each container contains a released version of each component. Containers are using a common base image we developed: https://github.com/Monogramm/docker-uxbox-builder. Containers are currently available on our DockerHub ([monogramm/uxbox](https://hub.docker.com/r/monogramm/uxbox/)) and links in README will need to be updated once you create your own public docker images;
* :sparkles: Introduction of a docker-compose file with all the needed services and using the "production-ready" containers (will close #95);
* :sparkles: Rework of manage.sh to execute unit tests (backend tests still missing because need running pgsql), build of previously mentioned production containers and run of docker-compose;
* :construction_worker: Introduction of CI mechanisms (Travis CI) to close #94 : CI is setup to execute all manage.sh targets and ensure they end successfully.
* :globe_with_meridians: Added several missing translations and replaced some hard coded texts with new translations (added French translations while at it)
* :wrench: Added a few configuration properties (issue #92 for instance), making possible to setup  almost all the application through (docker) environment variables
* :bug: Fixed a really nasty bug in copy/paste and undoable actions management
* :memo: Updated the README with some information for easier setup of development environment docker configuration
* :iphone: Fixing CSS and some actions to make the app compatible with touch screens (mostly for tablet, smartphones would require a whole lot of work to dynamically hide uneeded sections)
* :lipstick: Fixed some SCSS variables and improved the (hidden but still found) UXBOX dark theme (looks absolutely great, by the way)
* :bento: Added some default colors in the store : UXBOX colors (I like the idea of providing the users the very colors / icons / images that make UXBOX beautiful), Bootstrap colors set, Material Design Colors by "theme", Social Media colors
* :construction: Tried (but failed to finish) to restore the defaults Material Design icons. I believe it is a great feature to provide users with well known set of icons like Material / FontAwesome / IcoMoon, and if possible UXBOX own icons
[...]

PR #98, opened 2019-06-12
Add CI for build refactor

### LINKED ISSUE #94: Add Travis CI
:construction_worker: Add automated build of docker image with Travis CI

madmath03:
:construction: **Still in progress**

Objective is to close #94 by adding to CI:
- [x] release build
- [x] release run
- [x] unit tests placeholders


madmath03:
Hi @niwinz,

For information, I'm currently building on the side the docker repositories we discussed in #97:
* https://github.com/Monogramm/docker-uxbox-frontend
* https://github.com/Monogramm/docker-uxbox-backend

Frontend is working fine, but backend is giving me a bit more trouble so far, even though everything seems to work fine here...
Anyway, I will make it work, it's just a matter of time.

What I was wondering though was about the change on how the backend is delivered. I'm really not used to Clojure so I was wondering : what's the benefit of deploying the backend as source instead of packaging the backend as an uberjar ?
The latest seemed a little faster / safer to me since all the libraries were already in archive, but, again, there is a lot I do not know about Clojure 😅 

In the meantime, I should be able to finish this PR with the unit tests soon.

niwinz:
hi @madmath03 
The reason to deploy as source instead of uberjar, is because some dependencies (bouncycastle) does not works properly under "fatjar" (bacause they are cryptographically signed, and when it is repackaged, the signature is no longer valid). Additionally, deploying as source will allow in future things like deploy with 0 downtime or patch live code (in the same way as erlang can do it). If this were proprietary project deployed on third party, i can understand the no show the source, but this is an open source project, so this is not a problem. 

On the end deploying a source is very similar to previous jar file, with exception that is not packaged into single jar.

niwinz:
@madmath03 if you need some help, just tell me, maybe we can join some chat, for to have more direct contact?

niwinz:
@madmath03 i think that the best direction is leave in this PR the unit test part as "placeholders" and let me finish the `back/api-refactor` PR because i'm changing a lot of stuff there and tests are affected.

madmath03:
@niwinz regarding adding Unit Tests as "placeholders", that's also what I was thinking and started doing in the latest commit of this PR.

I wanted to have a try at calling the existing tests by myself, mostly for training / learning more about Clojure UT, but I might not have the time for that, and as you said, probably better to leave it to you.

I will check if everything is clean in this PR and send it to you.

Also, regarding your offer for chat, thanks a lot. I might send you an email if I'm stuck at some point 😉

madmath03:
Added a script in backend to prepare the release. The idea is to keep the inteeligence out of the manage.sh script and delegate to another script as much as possible. We can then use this script to more easily build a Docker container.

@niwinz All good for me here. I let review this and merge it if that's fine for you.

manage.sh
```diff
@@ -46,14 +49,74 @@ function run-devenv {
...
+
 function build-release-frontend-local {
+    if ! $(docker images | grep $IMGNAME | grep -q $REV); then
+        build-devenv
+    fi
+
+    mkdir -p $HOME/.m2
+    rm -rf ./frontend/node_modules
+    # FIXME Ugly... should be a better way
+    chmod 777 ./frontend
```
niwinz, line 107: why this `chmod 777` ? do you have permissions related problem with that mounted directories?

Consider building the image with `--build-arg EXTERNAL_UID=$(id -u) ` (i will add it to the default build-devenv command. This parameter will create the image user with the same id as in your machine, so your permission problems will disappear.

madmath03: Indeed I've had permissions issues when building the frontend (don't remember which files exactly though) so I added the FIXME as a reminder.

Your idea should probably do the trick. I will try it a bit later this week and let you know.

PR #104, opened 2019-07-19
Fix db config

madmath03:
:bug: Add missing DB username and password
```
Execution error (PSQLException) at org.postgresql.core.v3.ConnectionFactoryImpl/doAuthentication (ConnectionFactoryImpl.java:540).
The server requested password-based authentication, but no password was provided.
```
:sparkles: Add some script in backend entrypoint to init JDBC URI from base DB parameters


niwinz:
This is something that i haven't finished, so many thanks! I probably in a future will move the db patrams -> db-uri handling into the clojure code instead of having it on the entrypoint

madmath03:
You're welcome 😉 
I'm very impatient to test the develop branch so I wanted to help by fixing its runtime issues.

With this I will be able to quickly build and test the latest features and maybe also help bug fixing as well.

PR #113, opened 2019-08-10
Translations of main components

madmath03:
:globe_with_meridians: Translations of main components

PR #117, opened 2019-09-14
Translations of main components for multicanvas branch

madmath03:
🌐 Translations of main components

@niwinz This is basically a duplicate PR of #113 that I did for the wip/multicanvas branch because the develop branch does not work (several issues while opening uxbox front).
They are both valid and contain pretty much the same. I just create this because I do not know if or when each branch is going to be stable and merged.


niwinz:
Sorry for the confusion with the branches. I started working on multicanvas support but i found myself working in many different stuff in this branch... i have plan to merge it to develop as soon as i think is more or less usable.

niwinz:
Merged, i close the other pr!

PR #118, opened 2019-09-20
Docker/Devenv Improvements

niwinz:
This starts using docker compose for starting development environment. The main ideas behind this change:

- split services in different containers (make the development environment more similar to deployment environment)
- persistence across reboots/stops of the container (database, shell history, ...)
- you can execute multiple stuff with single (devenv) container up.

cc: @madmath03 tellme that do you think about this change


madmath03:
@niwinz I like the idea of having the devenv split into several services like the prod does. I will try to give you my help on this 😉

PR #154, opened 2020-03-29
Migrate from figwheel to shadow-cljs.

niwinz:
Relevant improvements:
- Improved development build times.
- nrepl support out of the box for cljs.
- faster watch build stop/start.
- powerful integration with npm, we no longer need this so complex way to incorporate third party libraries.
- enables future code splitting and dynamic code loading (different bundles: one for dashboard one for workspace and one for preview app for example).

Relevant downsides:
- Incompatible right now with rebel-readline, so the CLI repl does not has autocomplete.


docker/devenv/files/start-tmux.sh
```diff
@@ -11,13 +11,18 @@ popd
 
 tmux -2 new-session -d -s uxbox
 
-tmux new-window -t uxbox:1 -n 'figwheel'
+tmux new-window -t uxbox:1 -n 'shadow server'
 tmux select-window -t uxbox:1
 tmux send-keys -t uxbox 'cd uxbox/frontend' enter C-l
-tmux send-keys -t uxbox 'clojure -Adev tools.clj figwheel' enter
+# tmux send-keys -t uxbox 'shadow-cljs server' enter
```
Alotor: If this is commented then the tmux window won't do anything.

frontend/src/uxbox/main/data/workspace.cljs
```diff
@@ -39,7 +39,7 @@
    [uxbox.util.transit :as t]
    [uxbox.util.uuid :as uuid]
    [uxbox.util.webapi :as wapi]
-   [vendor.randomcolor])
+   #_[vendor.randomcolor])
```
Alotor, line 42: Why is this commented?

Alotor (approved):

PR #159, opened 2020-04-02
123/zoom improvements

hirunatan:


frontend/src/uxbox/main/ui/keyboard.cljs
```diff
@@ -5,6 +5,10 @@
   (fn [e]
     (= (.-keyCode e) keycode)))
 
+(defn ^boolean alt?
+  [event]
+  (.-altKey event))
+
```
hirunatan, line 11: I added this in one of the tries, currently not needed but think does not harm leaving it here.

PR #169, opened 2020-04-09
✨ Light and dark themes support

wizzytod:
:sparkles: Light and dark themes support:

- Themes are 2 scss files `resources/styles/main-light.scss` and `resources/styles/main-dark.scss`
- Default theme is light (Managed by `UXBOX_THEME` env variable)
- User can change theme in profile settings


frontend/src/uxbox/config.cljs
```diff
@@ -18,6 +18,7 @@
 (let [config (gobj/get goog.global "uxboxConfig")
       public-url (gobj/get config "publicURL" "http://localhost:6060")]
 
-  (def default-language "en")
+  (def default-lang "en")
```
madmath03: Just a detail I think, but why rename this field ?

frontend/src/uxbox/util/i18n.cljs
```diff
@@ -16,7 +16,7 @@
    [uxbox.util.transit :as t]
    [uxbox.util.storage :refer [storage]]))
 
-(defonce locale (get storage ::locale cfg/default-language))
+(defonce locale (get storage ::locale cfg/default-lang))
```
madmath03: Just a detail but, why rename this field ?
It might be better if we keep PRs strictly related to one feature, and renaming language fields are not related to theme management, right ?

madmath03 (commented):
@wizzytod

frontend/src/uxbox/config.cljs
wizzytod: Yeah, I think your are right. I will revert to `default-language`.

frontend/src/uxbox/util/i18n.cljs
wizzytod: Yeah. I just can not pass by an inconsistent lang and language.

niwinz:
First of all, thanks!

I think there has been some confusion about the themes. Our idea with themes is not to offer "customization" but rather accessibility. If you look, currently the dashboard is light, but workspace is dark. It is inconsistent with the names of the themes (light, dark).

The light and dark theme concept is legacy code :P 

So, our idea is to merge this PR, since it adds all the "infrastructure" to have themes (good job); with some additional changes: we will remove the concept of dark and light and we will add theme "default" and we will leave the window open to future themes. 

Since we are making many changes, maintaining themes right now would add an additional load to the team; that's why we think now is not a good time to include additional themes because they can quickly become obsolete.

madmath03:
Thanks @niwinz.

I kind of figured that the light and dark themes were legacy due to the recent changes in the design, but since we had this idea (and started the developments) of allowing users to choose the themes for quite some time now, we wanted to finish it as it provides _infrastructure_ for that.

Let us know if there any other things we need to change on our side before merging.

niwinz:
@madmath03 nice!

This PR is already merged (manually). I squashed all the commits and add an other one with the renaming of themes and removing the dark one. I think that when we normalize the styles, in a near future we can think on experiment adding a new theme.

Thank you very much for this great work!

PR #172, opened 2020-04-10
Fixes copy/paste with groups

Alotor:


frontend/src/uxbox/main/data/workspace.cljs
```diff
@@ -2178,28 +2178,52 @@
...
-               :id id
-               :frame-id frame-id
-               :obj reframed-obj}))
+
+                  prepare-child
+                  (fn [child-id]
+                    (prepare-shape-change objects (get objects child-id) delta frame-id))
+
+                  children-changes (mapcat prepare-child (:shapes obj))
+                  is-children? (set (:shapes obj))
```
hirunatan: Esta línea es confusa... estás cogiendo los hijos de un obj, lo conviertes en un set y preguntas si tiene datos? y entonces es un children (en plural)? No entiendo la mecánica.

Alotor: Try to comment in english :)

The sets in clojure are functions. So with a set i'm creating a function that checks if the argument is inside the children of the object.

A longer version of the same code could be...

```clojure
is-children? (fn [id] (contains? (set (:shapes obj)) id))
```
But it's much easier to just use the set as a function.

Alotor: Rereading the code, maybe a better name would be `is-child?`

hirunatan: A set can be used as a function that asks if some item is in the set? I'ts a very obscure idiom of Clojure, but well, if it's common use...

But yes, rename it to is-child? Thanks.

niwinz: yeah is very common

PR #177, opened 2020-04-14
:sparkles: Use relative coords in options panel

hirunatan:


frontend/src/uxbox/main/ui/workspace/sidebar/options/measures.cljs
```diff
@@ -23,6 +24,17 @@
   (let [options (or options #{:size :position :rotation :radius})
         locale (i18n/use-locale)
 
+        data (deref refs/workspace-data)
+        parent (get-in data [:objects (:frame-id shape)])
```
Alotor, line 28: Better to use a referenc only over the shape that you're checking. Otherwise any change in the data will provoke a change.

There is already a reference for objects given their ids. You either use that or create a new one. Using the one that exists already you can do:

```clojure
parent (-> shape :frame-id vector refs/objects-by-id deref first)
```

Alotor: `vector` and `first` are there because now needs a colection but if you make a similar one that only watches for a single object you can do...

```clojure
parent (-> shape :frame-id refs/object deref)
```

niwinz: in this case we dont need to use mf/deref, just use default deref, when yo change the shape, the parent frame is not going to change.

PR #187, opened 2020-04-24
245/shapes transforms improvements

Alotor:


frontend/src/uxbox/util/geom/point.cljs
```diff
@@ -9,13 +9,17 @@
...
+  (:require
+   [cuerdas.core :as str]
+   [uxbox.util.math :as mth]
+   [cognitect.transit :as t]))
 
 ;; --- Point Impl
 
 (defrecord Point [x y])
 
+(defn s [{:keys [x y]}] (str "(" x "," y ")"))
```
niwinz, line 21: Maybe we can implement the toString method to point, and use the point as is... (minor, don't change right now, we can experiment with this later).

PR #189, opened 2020-04-28
:bug: Fixes problems with rotation handlers

Alotor:
Currently there are two issues with the rotation handlers:

- When a shape is rotated the handler is not rotated with it so it's quite strange for the user to rotate an already rotated shape
- They don't scale with the zoom so when the workspace is zoom-in it's difficult to rotate.

https://tree.taiga.io/project/uxboxproject/issue/312


Alotor:
https://tree.taiga.io/project/uxboxproject/issue/312

PR #191, opened 2020-04-30
Other/fixes and refactors

niwinz:


frontend/src/uxbox/util/storage.cljs
```diff
@@ -9,24 +9,27 @@
 
 (defn- persist
   [alias value]
-  (let [key (name alias)
-        value (t/encode value)]
-    (.setItem js/localStorage key value)))
+  (when-not (or (= *target* "nodejs")
```
Alotor: maybe we should extract this condition to it's own function `is-worker?`

niwinz: Agree

Alotor (approved):

PR #200, opened 2020-05-07
238/infinite viewport (part 1)

niwinz:


frontend/src/uxbox/main/ui/workspace/rules.cljs
```diff
@@ -2,156 +2,129 @@
...
-        zoom (mf/deref refs/selected-zoom)
-        translate-x (- (- scroll-padding) (:x scroll))]
-    [:svg.horizontal-rule
-     {:width c/viewport-width
-      :height 20}
-     [:rect {:height 20
-             :width c/viewport-width}]
-     [:g {:transform (str "translate(" translate-x ", 0)")}
-      [:& horizontal-rule-ticks {:zoom zoom}]]]))
+  [{:keys [zoom size]}]
```
Alotor, line 18: Both horizontal and vertical rules are almos identical. I think we should try to "merge" both or at leas get some common functions so we don't have to maintain through copy/paste :smile:

Alotor: We'll refactor this part in a following MR

PR #205, opened 2020-05-12
:tada: Add interactions with the mouse

hirunatan:


frontend/src/uxbox/main/ui/workspace/shapes/interactions.cljs
```diff
@@ -149,9 +144,12 @@
 (mf/defc interactions
   [{:keys [selected] :as props}]
   (let [data (mf/deref refs/workspace-data)
+        local (mf/deref refs/workspace-local)
```
hirunatan, line 147: This new dependency adds some overhead to the main drawing loop, but I don't see any good way of avoiding it.

PR #219, opened 2020-05-19
Grids

Alotor:


frontend/resources/styles/main/partials/sidebar-element-options.scss
```diff
@@ -263,6 +263,11 @@
     padding: $x-small $big $x-small $x-small;
     position: relative;
 
+    & hr {
+        margin: 0;
+        border-color: $color-gray-20;
+    }
+
```
hirunatan: We are using indentation of 2 columns in general, please be consistent.

Alotor: done

PR #220, opened 2020-05-20
Color picker fixes

Alotor:
- Refactor color pickers in workspace to use a new component
- Allows opacity to be changed inside the color picker
- Some minor bug fixes in grid functionality

### DIFF
--- a/frontend/src/uxbox/util/color.cljs
+++ b/frontend/src/uxbox/util/color.cljs
@@ -31,7 +31,9 @@
 
 (defn hex->rgb
   [v]
-  (into [] (gcolor/hexToRgb v)))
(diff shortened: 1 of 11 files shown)


frontend/src/uxbox/util/color.cljs
```diff
@@ -31,7 +31,9 @@
 
 (defn hex->rgb
   [v]
-  (into [] (gcolor/hexToRgb v)))
+  (try
+    (into [] (gcolor/hexToRgb v))
+    (catch js/Object e [0 0 0])))
```
niwinz, line 36: instead of js/Object you should use `:default`

PR #225, opened 2020-05-22
23/user profile

niwinz:


backend/src/uxbox/emails.clj
```diff
@@ -62,3 +62,11 @@
 (def password-recovery
   "A password recovery notification email."
   (emails/build ::password-recovery default-context))
+
+(s/def ::pending-email ::us/string)
```
hirunatan: don't we have a schema for a valid email address?

frontend/src/uxbox/main/ui/settings/change_email.cljs
```diff
@@ -0,0 +1,102 @@
...
+    (st/emit! (du/request-email-change data))))
+
+(mf/defc change-email-form
+  [{:keys [locale profile] :as props}]
+  [:section.modal-content.generic-form
+   [:h2 (t locale "settings.change-email-title")]
+
+   [:span.featured-note
+    [:span.text
+     [:span "We’ll send you an email to your current email "]
```
hirunatan, line 60: There are some untranslated texts here

frontend/src/uxbox/main/ui/settings/header.cljs
```diff
@@ -16,22 +16,60 @@
...
+;; [:section.main-bar
+;;  [:nav
+;;   [:a.nav-item
+;;    {:class (when profile? "current")
+;;     :on-click #(st/emit! (rt/nav :settings-profile))}
+;;    (t locale "settings.profile")]
+;;   [:a.nav-item
+;;    {:class (when password? "current")
+;;     :on-click #(st/emit! (rt/nav :settings-password))}
+;;    (t locale "settings.password")]]]]))
```
hirunatan: May this be deleted?

frontend/src/uxbox/main/ui/auth.cljs
```diff
@@ -0,0 +1,93 @@
...
+
+(mf/defc auth
+  [{:keys [route] :as props}]
+  (let [section (get-in route [:data :name])
+        locale (mf/deref i18n/locale)]
+    [:*
+     [:& messages]
+     [:div.auth
+      [:section.auth-sidebar
+       [:a.logo i/logo]
```
hirunatan: Should have a href here, pointing to home for example.

frontend/src/uxbox/main/ui/settings/change_email.cljs
niwinz: yes, becasue this is peding to be refactored in the notifications US

frontend/src/uxbox/main/ui/settings/header.cljs
niwinz: true

frontend/src/uxbox/main/ui/auth.cljs
niwinz: What is the home page in this example?

hirunatan: I think the dashboard if the user is logged in, or the root if not. In fact, I think the / address should always redirect to dashboard when the user is logged in.

hirunatan: But this is not an issue of this MR. What I'd do here is that the logo directs to /, and then see if we want this to go dashboard in other issue.

niwinz: ok

PR #235, opened 2020-06-02
Duplicate move

Alotor:


frontend/src/uxbox/main/data/workspace/transforms.cljs
```diff
@@ -206,33 +208,53 @@
...
-  (ptk/reify ::start-move
-    ptk/UpdateEvent
-    (update [_ state]
-      (-> state
-          (assoc-in [:workspace-local :transform] :move)))
-
+             (rx/with-latest vector ms/mouse-position-alt)
+             (rx/flat-map
+              (fn [[_ alt?]]
+                (prn "ALT" alt?)
```
niwinz: :fire:

frontend/src/uxbox/main/ui/utils.cljs
```diff
@@ -0,0 +1,20 @@
...
+;; defined by the Mozilla Public License, v. 2.0.
+;;
+;; Copyright (c) 2020 UXBOX Labs SL
+
+(ns uxbox.main.ui.utils
+  (:require
+   [rumext.alpha :as mf]
+   [beicon.core :as rx]))
+
+(defn use-stream
```
niwinz: I think we can add this function to uxbox.main.ui.hooks ns that already has many of them.

frontend/src/uxbox/main/data/workspace/transforms.cljs
Alotor: Done

frontend/src/uxbox/main/ui/utils.cljs
Alotor: You're right

PR #240, opened 2020-06-03
Allows user to change canvas background color

Alotor:


frontend/src/uxbox/main/data/workspace.cljs
```diff
@@ -1329,6 +1329,24 @@
...
+  (ptk/reify ::change-canvas-color
+    ptk/WatchEvent
+    (watch [_ state stream]
+      (rx/of (dwc/commit-changes
+              [{:type :set-option
+                :option :background
+                :value color}]
+              [{:type :set-option
+                :option :background
+                :value nil}]
```
hirunatan: Shouldn't you remember the current color and put it in the undo command?

Alotor: You're right. Fixed

PR #254, opened 2020-06-09
Issues related to groups & layers

niwinz:


common/uxbox/common/geom/shapes.cljc
```diff
@@ -247,30 +242,16 @@
 (declare group->rect-shape)
 (declare rect->rect-shape)
 
+;; TODO: completly remove
+
```
hirunatan, line 243: when will this TODO be cleared?

niwinz: on my next PR

PR #272, opened 2020-06-29
:tada: Allow change fill color of multiple objects

hirunatan:


frontend/src/uxbox/main/refs.cljs
```diff
@@ -119,6 +119,14 @@
 (def selected-shapes
   (l/derived :selected workspace-local))
 
+(def selected-objects
+  (letfn [(selector [state]
+           (let [selected (get-in state [:workspace-local :selected])
+                 page-id (get-in state [:workspace-page :id])
+                 objects (get-in state [:workspace-data page-id :objects])]
+             (->> selected (map #(get objects %)))))]
```
niwinz: `mapv` instead of `map` should be used. `map` is lazy, then the `=` should be used for checking equality because the result of the selector is always a new list instance (that makes the default `identical?` checking useless).

frontend/src/uxbox/main/ui/workspace/sidebar/options/circle.cljs
```diff
@@ -10,14 +10,16 @@
...
 (mf/defc options
   [{:keys [shape] :as props}]
-  [:div
-   [:& measures-menu {:shape shape
-                      :options #{:size :position :rotation}}]
-   [:& fill-menu {:shape shape}]
-   [:& stroke-menu {:shape shape}]])
+  (let [ids [(:id shape)]
+        stroke-values (select-keys shape stroke-attrs)]
+    [:div
```
niwinz: this `div` is necessary? if not, you can replace it with `[:*` that uses React fragment and does not creates additional DOM nesting.

frontend/src/uxbox/main/refs.cljs
hirunatan: done

frontend/src/uxbox/main/ui/workspace/sidebar/options/circle.cljs
hirunatan: This was here before, but you are true. Changed in all shapes.

frontend/src/uxbox/main/refs.cljs
```diff
@@ -119,6 +119,14 @@
 (def selected-shapes
   (l/derived :selected workspace-local))
 
+(def selected-objects
+  (letfn [(selector [state]
+           (let [selected (get-in state [:workspace-local :selected])
+                 page-id (get-in state [:workspace-page :id])
+                 objects (get-in state [:workspace-data page-id :objects])]
+             (->> selected (mapv #(get objects %)))))]
```
niwinz, line 127: when you have only one chain, the `->>` should be avoided.

PR #279, opened 2020-07-20
:sparkles: Trim spaces in profile input fields

hirunatan:


frontend/src/uxbox/main/ui/settings/profile.cljs
```diff
@@ -58,7 +58,8 @@
      [:& input
       {:type "text"
        :name :fullname
-       :label (t locale "settings.fullname-label")}]
+       :label (t locale "settings.fullname-label")
+       :trim "true"}]
```
niwinz: why this `true` is text?

hirunatan: solved

PR #308, opened 2020-09-07
New color picker

Alotor:
https://tree.taiga.io/project/uxboxproject/task/589


niwinz:
A note about `package-lock.json`: as we're using yarn, the `package-lock.json` is no longer needed and is recommended does not have `yarn.lock` and npm's `package-lock.json` together.

frontend/src/app/main/refs.cljs
```diff
@@ -99,6 +99,12 @@
                  (get-in file [:data :colors])))
              st/state))
 
+(def workspace-recent-colors
+  (l/derived (fn [state]
+               (when-let [file (:workspace-file state)]
```
niwinz: Don't use `:workspace-file` here. Prefer use the `:workspace-data` because it is updated more early on commiting changes locally. The `:workspace-file` serves as "reference" data (or previous data) on comparison operations and is updated always after server roundtrip.

frontend/src/app/main/ui/workspace/colorpicker.cljs
```diff
@@ -2,27 +2,333 @@
...
+   [app.util.color :as uc]
+   [app.main.ui.icons :as i]
+   [app.common.math :as math]
+   [app.common.uuid :refer [uuid]]
+   [app.main.data.workspace.libraries :as dwl]
+   [app.main.ui.modal :as modal]
+   [okulary.core :as l]
+   [app.main.refs :as refs]))
+
+(def recent-colors
```
niwinz: Seems like this ref is replaced with `refs/workspace-recent-colors`

frontend/src/app/util/dom.cljs
```diff
@@ -203,3 +203,12 @@
   [b]
   {:pre [(blob? b)]}
   (js/URL.createObjectURL b))
+
+(defn set-css-property [node property value]
+  (.setProperty (.-style node) property value))
```
niwinz: Even if shadow-cljs does not emit warnings on interop calls, it is recommended to hint the JS objects with `^js`  to avoid unexpected exception on production builds.

Alotor: OK

frontend/src/app/main/ui/workspace/colorpicker.cljs
Alotor: Ups, I forgot to remove that code. Done

frontend/src/app/main/refs.cljs
Alotor: OK

PR #312, opened 2020-09-10
Pixel-level color picker

Alotor:
https://tree.taiga.io/project/uxboxproject/us/563?milestone=272498


frontend/src/app/main/ui/modal.cljs
```diff
@@ -1,61 +1,126 @@
...
 
-(defonce state (atom nil))
+(defonce components (atom {}))
+
+(defn show-modal [id type props]
+  (ptk/reify ::show-modal
+    ptk/UpdateEvent
+    (update [_ state]
+      (-> state
+          (assoc-in [:workspace-local :modal] {:id id
```
niwinz: I think the modal should be on the root, becuase we can have modals outside of workspace, and workspace local is a local state of the page...

frontend/src/app/main/ui/shapes/text.cljs
```diff
@@ -83,33 +89,70 @@
...
-          (let [style #js {:display "inline-block"
-                           :width "100%"}]
-            [:div.paragraphs {:key index :style style} children])
-
-          "paragraph"
-          (let [style (generate-paragraph-styles (clj->js node))]
-            [:p {:key index :style style} children])
-
-          nil))))))
+    (let [embed-resources? (mf/use-ctx muc/embed-ctx)
```
niwinz: i think this let (and the following effects) shold be outside the `mf/html`

frontend/src/app/main/ui/shapes/text.cljs
```diff
@@ -83,33 +89,70 @@
...
+        (let [children (map-indexed render-text-node children)]
+          (case type
+            "root"
+            (let [style (generate-root-styles (clj->js node))]
+              
+              [:div.root.rich-text
+               {:key index
+                :style style
+                :xmlns "http://www.w3.org/1999/xhtml"}
+               [:*
```
niwinz: This is not neceessary.

frontend/src/app/main/ui/workspace/colorpicker.cljs
```diff
@@ -102,6 +102,11 @@
         shared-libs (mf/deref refs/workspace-libraries)
         recent-colors (mf/deref refs/workspace-recent-colors)
 
+        picking-color? (mf/deref refs/picking-color?)
+        picked-color (mf/deref refs/picked-color)
+        picked-color-select (mf/deref refs/picked-color-select)
+        picked-shift? (mf/deref refs/picked-shift?)
+
```
niwinz, line 124: Seems like the picker refs are only used here, in this case i think they should be defined here in this namespace instead of on refs. Refs is for commonly used refs used in many sites (as a general rule). What do you think about this?

frontend/src/app/main/ui/modal.cljs
Alotor: OK

frontend/src/app/main/ui/shapes/text.cljs
Alotor: OK

frontend/src/app/main/ui/shapes/text.cljs
Alotor: You're right. I'll change it

frontend/src/app/main/ui/workspace/colorpicker.cljs
Alotor: Yes, I think it's better. I've changed it

PR #336, opened 2020-09-24
:tada: Synchronize library colors

hirunatan:

> Screenshots/GIFs

### DIFF
--- a/frontend/src/app/main/data/workspace/libraries.cljs
+++ b/frontend/src/app/main/data/workspace/libraries.cljs
@@ -30,6 +30,8 @@
    [cljs.spec.alpha :as s]
    [potok.core :as ptk]))
 
+(declare sync-file)
+
 (defn add-color
   [color]
   (us/assert ::us/string color)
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/data/workspace/libraries.cljs
```diff
@@ -698,3 +707,82 @@
...
+    (loop [shapes (seq linked-shapes)
+           rchanges []
+           uchanges []]
+      (let [shape (first shapes)]
+        (if (nil? shape)
+          [rchanges uchanges]
+          (let [[shape-rchanges shape-uchanges]
+                (generate-sync-colors-shape shape page colors)]
+            (recur (next shapes)
+                   (concat rchanges shape-rchanges)
```
niwinz, line 750: Don't use concat in this case, is much better using `into` or `app.common.data/concat`. The builtin clojure concat work with seqs (not with vectors) so it is lazy an incurr additional overhead (is worth in other cases but in loop and similar procedimens you just want `into` or `d/concat`).

PR #344, opened 2020-10-01
448/typographic styles

Alotor:


frontend/src/app/main/data/workspace.cljs
```diff
@@ -73,9 +73,10 @@
...
 
 (def default-layout
-  #{:sitemap
-    :sitemap-pages
-    :layers
+  #{;; :sitemap
+    ;; :sitemap-pages
+    ;; :layers
+    :assets
     :element-options
```
hirunatan: should you restore this?

frontend/src/app/main/data/workspace/persistence.cljs
```diff
@@ -300,8 +300,7 @@
                   (rx/mapcat
                     #(rx/zip (rp/query :file-library {:file-id library-id})
                              (rp/query :media-objects {:file-id library-id
-                                                       :is-local false})
-                             (rp/query :colors {:file-id library-id}))))
+                                                       :is-local false}))))
```
hirunatan, line 303: Why removing colors? Shouldn't we instead add typographies?

common/app/common/pages.cljc
```diff
@@ -833,6 +873,22 @@
   [data {:keys [id]}]
   (d/dissoc-in data [:components id]))
 
+;; -- Typography
+
+(defmethod process-change :add-typography
+  [data {:keys [typography]}]
+  (update data :typography assoc (:id typography) typography))
```
hirunatan: Wouldn't be better that the name of the collection of tipography objects be :tipographies ?

frontend/src/app/main/data/workspace/persistence.cljs
Alotor: No, that query is not necessary anymore because the color data comes inside the file.

frontend/src/app/main/data/workspace.cljs
Alotor: It's done in a later commit

common/app/common/pages.cljc
Alotor: I prefer a singular word because the other one is too long

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
```diff
@@ -403,14 +418,17 @@
                  [(t locale "workspace.assets.delete") handle-delete-typography]]}]
      (when open?
        [:div.group-list
-        (for [typography (sort-by (comp - :ts) typographies)]
+        (for [typography (sort-by :ts typographies)]
```
hirunatan, line 422: Is it ok to sort this by time? Other assets are sorted by name (see apply-filters function). Also the filter in "Search assets"  is not working, it does not find any typography.

Alotor: The sorting of the library assets is something we'll need to address further down the line. As it is right now, I think it's better to sort by time to improve the user experience.

common/app/common/pages.cljc
hirunatan: eh.. but this is inconsistent with other assets, that are called "pages", "colors" and "components". And may cause confusion, perhaps we don't know if a variable points to the whole collection of typographies or to one particular one; or else we need to name a variable "tipographies" that refers to the "tipography" element. And it's only two characters longer...

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
hirunatan: It's possible; although it should be consistent for all kinds of assets. It seems strange that colors or graphics are sorted by name, and typographies by time.

PR #345, opened 2020-10-02
Dynamic alignment improvements

Alotor:
Fixes some issues with dynamic alignment

### DIFF
--- a/frontend/src/app/main/ui/workspace/snap_distances.cljs
+++ b/frontend/src/app/main/ui/workspace/snap_distances.cljs
@@ -176,7 +176,7 @@
         check-in-set
         (fn [value number-set]
           (->> number-set
-               (some #(<= (mth/abs (- value %)) 0.5))))
+               (some #(<= (mth/abs (- value %)) 1))))
 
         ;; Left/Top shapes and right/bottom shapes (depends on `coord` parameter
         [lt-shapes gt-shapes] @to-measure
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/ui/workspace/snap_distances.cljs
```diff
@@ -185,19 +185,24 @@
...
+
         ;; Show the distances that either match one of the distances from the selrect
         ;; or are from the selrect and go to a shape on the left and to the right
-        show-distance? #(check-in-set % (into #{} (concat
-                                                   (map first distance-coincidences)
-                                                   (set/intersection lt-distances gt-distances))))
+        show-distance?
+        (fn [dist]
+          (let [distances-to-show
+                (->> (d/concat (mapv first distance-coincidences)
```
niwinz: You can done all this in one step just using `#{}` as a first value of d/concat. 
Also, the use of `filterv` should be replace `filter` instead I think (because you are converting seq to vec and later to seq (inside d/concat).

frontend/src/app/main/ui/workspace/snap_points.cljs
```diff
@@ -65,11 +110,12 @@
...
         ;; can cause problems with react keys
-        snap-points (into #{} (mapcat (fn [[point snaps coord]]
-                                        (cons point snaps))
-                                      @state))
-        snap-lines (into #{} (mapcat (fn [[point snaps coord]]
-                                       (when (not-empty snaps) (map #(vector point %) snaps))) @state))]
+        snap-points (->> @state
+                         (mapcat add-point-to-snaps)
+                         (into #{}))
+
```
niwinz, line 114: You can avoid creating intermediate seqs with `(into #{} (mapcat add-point-to-snaps) @state)`

frontend/src/app/main/ui/workspace/snap_distances.cljs
Alotor: OK

frontend/src/app/main/ui/workspace/snap_points.cljs
Alotor: OK

PR #356, opened 2020-10-14
:sparkles: Change copy texts and style of library dialogs

hirunatan:

> Screenshots/GIFs

### DIFF
--- a/frontend/resources/locales.json
+++ b/frontend/resources/locales.json
@@ -1263,13 +1263,13 @@
   "modals.delete-project-confirm.message" : {
     "used-in" : [ "src/app/main/ui/dashboard/files.cljs:57" ],
     "translations" : {
-      "en" : "Are you sure you wan't to delete this project?"
+      "en" : "Are you sure you want to delete this project?"
     }
   },
   "modals.delete-project-confirm.title" : {
     "used-in" : [ "src/app/main/ui/dashboard/files.cljs:56" ],
     "translations" : {
-      "en" : "Deleting project"
+      "en" : "Delete project"
     }
   },
   "modals.delete-team-confirm.accept" : {
--- a/frontend/resources/locales.json
+++ b/frontend/resources/locales.json
@@ -1155,12 +1155,6 @@
       "es" : "Añadir “%s” como Biblioteca Compartida"
     }
   },
-  "modals.add-shared-confirm.title" : {
-    "used-in" : [ "src/app/main/ui/workspace/header.cljs:111", "src/app/main/ui/dashboard/grid.cljs:113" ],
-    "translations" : {
-      "en" : "Adding shared library"
-    }
-  },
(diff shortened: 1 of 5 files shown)


frontend/resources/locales.json
```diff
@@ -1269,13 +1263,13 @@
...
     "translations" : {
-      "en" : "Are you sure you wan't to delete this project?"
+      "en" : "Are you sure you want to delete this project?"
     }
   },
   "modals.delete-project-confirm.title" : {
     "used-in" : [ "src/app/main/ui/dashboard/files.cljs:56" ],
     "translations" : {
-      "en" : "Deleting project"
+      "en" : "Delete project"
```
niwinz, line 1272: Why this change? The design/prototypes expectes the `Deleting project` title.

frontend/src/app/main/ui/dashboard/grid.cljs
```diff
@@ -109,10 +109,12 @@
            (dom/stop-propagation event)
            (st/emit! (modal/show
                       {:type :confirm
-                       :message (t locale "modals.add-shared-confirm.message" (:name file))
-                       :title (t locale "modals.add-shared-confirm.title")
+                       :message ""
+                       :title (t locale "modals.add-shared-confirm.message" (:name file))
                        :hint (t locale "modals.add-shared-confirm.hint")
+                       :cancel-label ""
```
niwinz: why not just omit the :cancel-label in this case? or use `nil` instead of `""`

frontend/resources/locales.json
hirunatan: It's asked for in this issue https://tree.taiga.io/project/uxboxproject/issue/759

frontend/src/app/main/ui/dashboard/grid.cljs
hirunatan: If you omit cancel-label, then the default "Cancel" label is used (see confirm.cljs). We can use other system, but there should be a way to distinguish between "the default label" or "no label and no button".

hirunatan: See https://tree.taiga.io/project/uxboxproject/issue/760

frontend/resources/locales.json
niwinz: I think, they should first agree before fill issues, the prototypes says "Deleting projects". But is ok.

frontend/src/app/main/ui/dashboard/grid.cljs
niwinz: Hmm in case we want to have the ability to omit cancel button, i think is better have `:default` value for default and omit it by default. Empty string is very confusing as option for ommit it.

hirunatan: fixed!

PR #366, opened 2020-10-26
:sparkles: Rework nested components to avoid indirect references

hirunatan:

> Screenshots/GIFs

### DIFF
--- a/frontend/src/app/main/data/workspace/libraries.cljs
+++ b/frontend/src/app/main/data/workspace/libraries.cljs
@@ -27,9 +27,12 @@
    [app.util.i18n :refer [tr]]
    [app.util.router :as rt]
    [app.util.time :as dt]
+   [app.util.logging :as log]
    [beicon.core :as rx]
    [cljs.spec.alpha :as s]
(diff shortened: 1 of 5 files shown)


frontend/src/app/main/data/workspace/libraries.cljs
```diff
@@ -27,10 +27,13 @@
    [app.util.i18n :refer [tr]]
    [app.util.router :as rt]
    [app.util.time :as dt]
+   [app.util.logging :as log]
    [beicon.core :as rx]
    [cljs.spec.alpha :as s]
    [potok.core :as ptk]))
 
+(log/set-level! :warn)
```
Alotor, line 35: Why is there a top-level statement here?

hirunatan: It's explained in the [logging framework manual](https://github.com/penpot/penpot/blob/develop/docs/02-Frontend-Developer-Guide.md#logging-framework). This call sets the logging level for the current namespace (in this case app.main.data.workspace.libraries).

PR #383, opened 2020-11-05
:bug: Take into account indirect libraries

hirunatan:

> Screenshots/GIFs

### DIFF
--- a/frontend/src/app/main/ui/workspace/libraries.cljs
+++ b/frontend/src/app/main/ui/workspace/libraries.cljs
@@ -7,6 +7,7 @@
 
 (ns app.main.ui.workspace.libraries
   (:require
+   [app.common.data :as d]
    [rumext.alpha :as mf]
    [cuerdas.core :as str]
    [okulary.core :as l]
@@ -142,7 +143,6 @@
                                 :value (tr "workspace.libraries.update")
                                 :on-click #(update-library (:id library))}]])]])]))
 
-
 (mf/defc libraries-dialog
   {::mf/register modal/components
    ::mf/register-as :libraries-dialog}
@@ -152,7 +152,9 @@
         locale       (mf/deref i18n/locale)
         project      (mf/deref refs/workspace-project)
         file         (mf/deref workspace-file)
-        libraries    (mf/deref refs/workspace-libraries)
+        libraries    (->> (mf/deref refs/workspace-libraries)
+                          (d/removem (fn [[key val]]
+                                       (:is-indirect val))))
         shared-files (mf/deref refs/workspace-shared-files)
 
         change-tab   #(reset! selected-tab %)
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/ui/workspace/libraries.cljs
```diff
@@ -152,7 +152,9 @@
         locale       (mf/deref i18n/locale)
         project      (mf/deref refs/workspace-project)
         file         (mf/deref workspace-file)
-        libraries    (mf/deref refs/workspace-libraries)
+        libraries    (d/filterm
+                       (fn [[key val]] (:is-direct val))
+                       (mf/deref refs/workspace-libraries))
```
Alotor: I think this is more legible using the "pipe" macro like:

```clojure
(->> (mf/deref refs/workspace-libraries)
     (d/filterm (fn [[key val]] (:is-direct val)))
```

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
```diff
@@ -631,7 +631,8 @@
 
 (mf/defc assets-toolbox
   []
-  (let [libraries (mf/deref refs/workspace-libraries)
+  (let [libraries (filter :is-direct
+                          (vals (mf/deref refs/workspace-libraries)))
```
Alotor: Same here, the pipe improves legibility
```clojure
(->> (mf/deref refs/workspace-libraries)
     (vals)
     (filter :is-direct))

File: frontend/src/app/main/ui/workspace/sidebar/assets.cljs [hunk 3]
```diff
@@ -697,8 +698,7 @@
         :open? true
         :filters @filters}]
 
-      (for [file (->> (vals libraries)
-                      (sort-by #(str/lower (:name %))))]
+      (for [file (sort-by #(str/lower (:name %)) libraries)]
```
Alotor: Same here. You undid the threading macro but I think the problem is the anonymous function

frontend/src/app/main/ui/workspace/libraries.cljs
hirunatan: Done. I have also inverted the condition, because I want nils to pass the filter.

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
hirunatan: Done

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
hirunatan: Done. Removed the thread because I don't need (vals) here, but it's still useful for the sort.

Alotor (approved):

PR #396, opened 2020-11-24
Bezier curves

Alotor:


frontend/src/app/main/data/workspace/drawing/path.cljs
```diff
@@ -0,0 +1,697 @@
...
+
+;; EVENTS
+
+(defn init-path [id]
+  (ptk/reify ::init-path
+    ptk/UpdateEvent
+    (update [_ state]
+      (-> state
+          #_(assoc-in [:workspace-drawing :object :initialized?] true)
+          #_(assoc-in [:workspace-local :edit-path :last-point] nil)))))
```
niwinz: This event looks **noop**

frontend/src/app/main/data/workspace/drawing/path.cljs
```diff
@@ -0,0 +1,697 @@
...
+                 (rx/with-latest merge ms/mouse-position)
+
+                 ;; We change to the stream that emits the first event
+                 (rx/switch-map
+                  #(rx/race (make-node-events-stream stream)
+                            (make-click-stream stream %)
+                            (make-drag-stream stream %)
+                            (make-dbl-click-stream stream %))))]
+
+        (->> (rx/concat
```
niwinz: The `->>` is not necessary here i think.

Alotor: You're right. That's some code I forgot to remove :smile:

PR #427, opened 2020-12-07
Fixed problem with import SVG image size

Alotor:


Alotor:
@niwinz I need to check with you the implications in the environments.

niwinz:
This is included in my bugfixes mr

PR #443, opened 2021-01-07
Upload SVG as shapes

Alotor:


frontend/src/app/main/data/workspace/persistence.cljs
```diff
@@ -376,22 +409,54 @@
...
+                    (rx/mapcat #(rp/mutation! :create-file-media-object-from-url %))
+                    (rx/do on-image))
+
+               (->> (rx/from uris)
+                    (rx/filter svg-url?)
+                    (rx/merge-map fetch-svg)
+                    (rx/merge-map parse-svg)
+                    (rx/with-latest vector uris)
+                    (rx/map #(assoc (first %) :name (or name (url-name (second %)))))
+                    (rx/do on-svg)))
```
niwinz: As far as I understand it, this stream is just for "side-effect" (executing the `do`). In this case I think the `rx/ignore` is missing here. This will avoid emiting the result of this stream to the STORE stream.

frontend/src/app/main/exports.cljs
```diff
@@ -77,26 +78,42 @@
...
-             :text   [:> text/text-shape opts]
-             :rect   [:> rect/rect-shape opts]
-             :path   [:> path/path-shape opts]
-             :image  [:> image/image-shape opts]
-             :circle [:> circle/circle-shape opts]
-             :frame  [:> frame-wrapper {:shape shape}]
-             :group  [:> group-wrapper {:shape shape :frame frame}]
-             nil)])))))
+          (if (and (= :svg-raw (:type shape))
+                   (not= :svg (get-in shape [:content :tag])))
```
niwinz: Why we have 2 ways to identify the raw-svg shapes?

frontend/src/app/main/ui/context.cljs
```diff
@@ -13,6 +13,7 @@
 
 (def embed-ctx (mf/create-context false))
 (def render-ctx (mf/create-context nil))
+(def def-ctx (mf/create-context false))
```
niwinz, line 16: I think this context need some explanation note because it is not obvious the purpose of this ctx on the first/quick look on the code.

frontend/src/app/main/exports.cljs
Alotor: The issue with the SVG is that certain elements break if they are wrapped inside a group (that what shape-container create). So, only the svg-raw in the root (<svg> tag) will be wrapped. The other elements will not be wrapped inside a group

frontend/src/app/main/ui/context.cljs
Alotor: It's explained in the place where it's used. 
https://github.com/penpot/penpot/pull/443/files#diff-a1591439ff86a209f08e8db2535263b0405739726587a1dc03fe4b5d6416d8b1R81

```clojure
          ;; We cannot wrap inside groups the shapes that go inside the defs tag
          ;; we use the context so we know when we should not render the container
```

PR #451, opened 2021-01-12
Changes to the selection in workspace and layers

Alotor:
https://tree.taiga.io/project/uxboxproject/issue/415


frontend/src/app/main/data/workspace/selection.cljs
```diff
@@ -184,24 +198,51 @@
...
+(defn select-last-layer
+  ([position]
+   (ptk/reify ::select-last-layer
+     ptk/WatchEvent
+     (watch [_ state stream]
+       (let [page-id  (:current-page-id state)
+             objects (dwc/lookup-page-objects state page-id)
+             find-shape
+             (fn [selection]
+               (when (seq selection)
```
niwinz, line 229: minor: probably this check can be done in the rx stream transformations; maybe with `(rx/filter #(boolean (seq %))` before the `rx/map`.

PR #461, opened 2021-01-15
Frontend tests

hirunatan:

> Screenshots/GIFs

### DIFF
--- a/frontend/src/app/util/globals.js
+++ b/frontend/src/app/util/globals.js
@@ -1,0 +1,42 @@
+/**
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * This Source Code Form is "Incompatible With Secondary Licenses", as
+ * defined by the Mozilla Public License, v. 2.0.
+ *
+ * Copyright (c) 2020 UXBOX Labs SL
+ */
+
+/*
+ * Provide instances of some javascript global objects, by looking for
+ * them in the browser, the current worker or creating a dummy for test
+ * environment.
+ */
+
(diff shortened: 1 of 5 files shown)


frontend/src/app/util/globals.js
```diff
@@ -0,0 +1,49 @@
...
+ * Provide instances of some javascript global objects, by looking for
+ * them in the browser, the current worker or creating a dummy for test
+ * environment.
+ */
+
+"use strict";
+
+goog.provide("app.util.globals");
+goog.provide("app.util.globals.window");
+goog.provide("app.util.globals.location");
```
niwinz: this line and the previous one, is not necessary (window and location), only the first one is mandatory

frontend/src/app/util/globals.js
```diff
@@ -0,0 +1,49 @@
...
+
+goog.scope(function() {
+  var win;
+  if (typeof window !== "undefined") {
```
niwinz: don't acces window **as-is**; use `goog.global.window`

frontend/src/app/util/globals.js
```diff
@@ -0,0 +1,49 @@
...
+  } else if (typeof self !== "undefined"){
+    win = self;
+  } else {
+    win = {};
+  }
+
+  app.util.globals.window = win;
+
+  var loc;
+  if (typeof location !== "undefined") {
```
niwinz: the same as for `window`

frontend/src/app/util/globals.js
hirunatan: done

frontend/src/app/util/globals.js
hirunatan: done

frontend/src/app/util/globals.js
hirunatan: done

PR #474, opened 2021-01-20
Snap distances performance improvements

Alotor:


frontend/src/app/main/snap.cljs
```diff
@@ -132,13 +148,13 @@
...
-        gt-cand (inner-distance shapes-gt)
+        lt-cand (inner-distance (mapv :selrect shapes-lt))
+        gt-cand (inner-distance (mapv :selrect shapes-gt))
 
         ;; Distance between the elements to either side and the current shape
         ;; this is the distance that will "snap"
-        lt-dist (mapv dist-lt shapes-lt)
-        gt-dist (mapv dist-gt shapes-gt)
+        lt-dist (->> (map dist-lt shapes-lt) (into #{}))
+        gt-dist (->> (map dist-gt shapes-gt) (into #{}))
```
niwinz: In this case `(into #{} (map dist-gt) shapes-gt)` will avoid additional seq creation (by `map`) thanks to transducers

frontend/src/app/main/ui/workspace/snap_distances.cljs
```diff
@@ -146,13 +163,14 @@
         ;; Gets the distance to the current selection
         lt-distances (->> lt-shapes (map distance-to-selrect) (filter pos?) (into #{}))
```
niwinz: did't see before, but here you can share the same transducer for two operations and avoid create 4 seqs

PR #502, opened 2021-01-29
Bugfixing

Alotor:


frontend/src/app/main/ui/workspace/viewport.cljs
```diff
@@ -727,7 +730,8 @@
        :width (:width vport 0)
        :height (:height vport 0)
        :view-box (format-viewbox vbox)
-       :ref viewport-ref
+       :ref #(do (mf/set-ref-val! viewport-ref %)
+                 (reset! viewport-node %))
```
niwinz, line 734: I'm not very convinced with this solution of passing dom node just as a prop, but right now i don't have better alternatives. Maybe we need to give more time on this in a future.

PR #503, opened 2021-01-29
Refcounts for Storage

niwinz:


backend/src/app/db.clj
```diff
@@ -186,17 +184,17 @@
...
 (defn insert-multi!
   [ds table param-list]
   (doseq [params param-list]
-      (insert! ds table params))
+    (insert! ds table params))
   ;; FIXME: Won't work
   #_(let [keys (->> param-list first keys (into []))
-        params (->> param-list (mapv #(->> keys (select-values %) (into []))) )]
-    (jdbc-sql/insert-multi! ds table keys params default-options)))
+          params (->> param-list (mapv #(->> keys (select-values %) (into []))) )]
```
Alotor: Don't use the clojure one. They do different things. The currently used if a key is not present in a map will return nil for that key. Clojure removes that value. Is necesary otherwise will fail because could have different number of rows.

niwinz: This change only implies indentation change, no behavior/code is changed.

PR #510, opened 2021-02-01
:tada: Add frontend tests to circleci.

niwinz:


.circleci/config.yml
```diff
@@ -21,37 +21,47 @@ jobs:
...
             PENPOT_DATABASE_PASSWORD: penpot_test
             PENPOT_REDIS_URI: "redis://localhost/1"
 
+      - run:
+          working_directory: "./frontend"
+          name: frontend tests
+          command: |
+            yarn install
+            npx shadow-cljs compile tests
+            node target/tests.js
```
hirunatan: This way, the tests execute twice, don't they? Is it necessary the second line?

niwinz: hmm, i think you are right

PR #552, opened 2021-02-03
Recursion in components

Alotor:


common/app/common/pages/helpers.cljc
```diff
@@ -376,3 +376,25 @@
...
     ;; The first id will be the top-most
     (get objects (first sorted-ids))))
+
+(defn is-parent?
+  "Check if `parent-candidate` is parent of `shape-id`"
+  [objects shape-id parent-candidate]
+
+  (loop [current (get objects parent-candidate)
+         done #{}
+         pending (get-in objects [parent-candidate :shapes]) ]
```
niwinz: Can be this replaced with `pending (:shapes current)` ?

Alotor: You're right. I'll change it

PR #554, opened 2021-02-03
🐛 FIX: js variable prefix app->penpot

violoncelloCH:

while setting up and configuring penpot I realized that the js var changes don't take effect, so I compared with design.penpot.app and found that the actual var prefix is `penpot` and not `app`

> Screenshots/GIFs

not applicable as only doc change


niwinz:
Merged manually, thanks :heart:

PR #585, opened 2021-02-05
:lipstick: Update UXBOX name in emails

kdrag0n:
When registering for a new account, I noticed that the HTML emails had the new Penpot name but the plain-text versions were still using the old UXBOX name. This commit fixes the discrepancy.


hirunatan:
Thanks very much! We missed this...

PR #636, opened 2021-02-12
Deep selection improvements

Alotor:
- Disable groups interactions when holding "Ctrl" key (deep selection)
- New action in context menu to "edit" some shapes (binded to key "Enter")


frontend/src/app/main/streams.cljs
```diff
@@ -120,6 +120,21 @@
...
                  (rx/dedupe))]
         (rx/subscribe-with ob sub)
+        sub))
+
+(defonce keyboard-ctrl
+  (let [sub (rx/behavior-subject nil)
+        ob  (->> (rx/merge
+                  (->> st/stream
+                       (rx/filter keyboard-event?)
+                       (rx/map :ctrl))
```
hirunatan: This doesn't work. Apparently when you release ctrl key, an on-key-up is fired, but the :ctrl flag is still set.

PR #680, opened 2021-02-22
 :sparkles: Add catalan translation (partial) 

NatachaMenjibar:
Update locales.json file to add catalan translation (partial)


niwinz:
Merged

PR #739, opened 2021-03-08
Opacity and blending modes

Alotor:
- Add layer opacity and blend mode to shapes


frontend/src/app/main/ui/shapes/attrs.cljs
```diff
@@ -120,6 +120,17 @@
         (obj/merge! attrs (clj->js stroke-attrs)))
       attrs)))
 
+(defn add-layer-props [attrs shape]
+  (let [layer-attrs
+        (cond-> {}
```
niwinz: maybe here a combination of `obj/merge!` + `when` will result in a more efficient code

niwinz: or better: `(obj/new)` + `cond->` + `obj/set!`

PR #747, opened 2021-03-11
:tada: Drag and drop files in the dashboard

### LINKED ISSUE #547: Allow to move files between projects (and teams)
It's really simple to create files and start prototyping, but now I've got those files all in the "Drafts" section. Only then I'm realizing that I can create projects to organize those files. However it doesn't seem to be possible to move files between Projects

Allow for files to be moved between projects (and possibly even teams). Ideally this should be possible both, over the context/three-dot menu and via drag'n'drop.

I don't really see any alternative here.

**Additional context**
Nothing for now. But Penpot looks and feels great!

hirunatan:

> Screenshots/GIFs


frontend/src/app/main/data/dashboard.cljs
```diff
@@ -171,8 +179,10 @@
     (ptk/reify ::fetch-shared-files
       ptk/WatchEvent
       (watch [_ state stream]
-        (->> (rp/query :shared-files {:team-id team-id})
-             (rx/map #(partial fetched %)))))))
+        (rx/concat
+          (->> (rp/query :shared-files {:team-id team-id})
+               (rx/map #(partial fetched %)))
+          (rx/of (clear-selected-files)))))))
```
niwinz: I think that clear-selected-files is a side effect that should be done by the caller and not on this event. I'mean, this event is for fetching data, not clearing selection.

hirunatan: Done. You can squash the commit into the first one when merging.

PR #759, opened 2021-03-17
Add some Turkish translation

gizembln:

Add some Turkish translations.

There was no Turkish translation and I wanted to add some for the Turkish community.


niwinz:
Looks awesome! Thank you very much. Looks everything correct.

We understand that is not a very accessible way, but right now is that we have.
We will try to priorize to improve this in a near future.

gizembln:
Thank you! It is okay to translate it right now, but you can consider it in the future.

niwinz:
Thank you, merged manually.

PR #794, opened 2021-03-26
added 98 new Turkish translation strings.

mbrksntrk:
added 98 new Turkish translation strings. but it is not complete.


niwinz:
Merged manually, Thanks!

PR #840, opened 2021-04-13
Generic openid-connect auth subsystem.

niwinz:


frontend/translations/en.po
```diff
@@ -68,6 +68,12 @@ msgstr "Great to see you again!"
 msgid "auth.login-with-github-submit"
 msgstr "Login with Github"
 
+msgid "auth.login-with-google-submit"
+msgstr "Login with Google"
+
+msgid "auth.login-with-oidc-submit"
+msgstr "Login with OpenID (SSO)"
+
```
hirunatan: We should write always spanish translation at least. It's makes not much sense to add it later.

backend/src/app/http/oauth.clj
```diff
@@ -0,0 +1,278 @@
...
+
+(defn- initialize-oidc-provider
+  [cfg]
+  (let [opts {:base-uri      (cf/get :oidc-base-uri)
+              :client-id     (cf/get :oidc-client-id)
+              :client-secret (cf/get :oidc-client-secret)
+              :token-uri     (cf/get :oidc-token-uri)
+              :auth-uri      (cf/get :oidc-auth-uri)
+              :user-uri      (cf/get :oidc-user-uri)
+              :scope         "openid profile email name"
```
burnoutberni, line 201: This should also be configurable

niwinz: Can you explain, why? Penpot need mandatory the "email" and "name" fields. So the default scopes requires the minumum, (probably the `profile` scope is not necessary here); and allowing user set a wrong config, seems bad idea, also more scopes are also not necessary.

burnoutberni: You're right, you could also make `openid profile email name` the default and only make optional scopes configurable, that's probably the better way to do it. Please check out my comment on the issue for more info.

PR #865, opened 2021-04-23
Bugfix assets

hirunatan:

> Screenshots/GIFs

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/assets.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/assets.cljs
@@ -149,7 +149,7 @@
 ;; ---- Components box ----
 
 (mf/defc components-box
-  [{:keys [file-id local? components listing-thumbs? open?] :as props}]
+  [{:keys [file-id local? components listing-thumbs? open? change-selected] :as props}]
   (let [state (mf/use-state {:menu-open false
                              :renaming nil
                              :top nil
@@ -236,10 +236,7 @@
                          (toggle-select selected component-id)
 
                          (kbd/shift? event)
-                         (extend-select selected component-id groups)
-
-                         :default
-                         (replace-select selected component-id)))))))
+                         (extend-select selected component-id groups)))))))
 
         create-group
         (mf/use-callback
@@ -278,6 +275,10 @@
            (dnd/set-data! event "penpot/component" {:file-id file-id
                                                     :component component})
            (dnd/set-allowed-effect! event "move")))]
+
+    (mf/use-effect
+      (mf/deps [change-selected selected])
+      #(change-selected (count selected)))
 
     [:div.asset-section {:on-click unselect-all}
      [:div.asset-title {:class (when (not open?) "closed")}
@@ -351,7 +352,7 @@
 ;; ---- Graphics box ----
 
 (mf/defc graphics-box
-  [{:keys [file-id local? objects listing-thumbs? open?] :as props}]
+  [{:keys [file-id local? objects listing-thumbs? open? change-selected] :as props}]
   (let [input-ref  (mf/use-ref nil)
         state      (mf/use-state {:menu-open false
                                   :renaming nil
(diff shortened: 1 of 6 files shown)


frontend/src/app/main/ui/workspace/sidebar/assets.cljs
```diff
@@ -579,6 +581,8 @@
 
         click-color
         (fn [event]
+          (when on-select
+            ((on-select (:id color)) event))
```
Alotor, line 581: this is very strange. Why don't use a callback with two arguments? `on-select (:id color) event`

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
```diff
@@ -696,8 +712,33 @@
...
+                         :position :right})))
+
+        unselect-all
+        (mf/use-callback
+          (fn [event]
+            (swap! state assoc :selected empty-selection)))
+
+        on-select
+        (mf/use-callback
+          (mf/deps state)
```
Alotor: do you need to depend on `state`? you're not reading nothing. I think you don't need dependencies here

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
```diff
@@ -715,21 +756,24 @@
...
   (let [state (mf/use-state {:detail-open? false
                              :menu-open? false
                              :top nil
-                             :left nil})
+                             :left nil
+                             :selected empty-selection})
 
-        selected (mf/deref refs/selected-shapes)
-        local (deref refs/workspace-local)
+        local    (deref refs/workspace-local)
```
Alotor, line 770: You're using normal `deref` instead of `mf/deref`

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
```diff
@@ -744,13 +788,25 @@
...
+          (swap! state update :selected
+                 (fn [selected]
+                   (cond
+                     (kbd/ctrl? event)
+                     (toggle-select selected (:id typography))
+
+                     (kbd/shift? event)
+                     (extend-select selected (:id typography) {"" typographies}))))
+
+          (let [ids (get-in @st/state [:workspace-local :selected])
```
Alotor, line 798: why access the state directly and not create a ref?

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
hirunatan: I read the state to make a swap and update

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
hirunatan: I added this to an existing function. I think this will change when I make the seleccion global to all asset types.

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
hirunatan: This was already so before this merge.

I think this is because I do not want the component to repaint every time anything changes in :workspace-local state. Perhaps should be good idea to make a more specific lens.

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
hirunatan: Don't know, this also was already so before this branch :)

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
niwinz: The state on deps is not necessary

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
niwinz: true, this is explicit usage of normal `deref`

PR #901, opened 2021-05-04
Events/Analytics/Exporter

niwinz:


backend/src/app/util/logging.clj
```diff
@@ -60,17 +60,21 @@
...
        (if (enabled? ~logger-sym ~level-sym)
          ~(if async
-            `(send-off logging-agent (fn [_#] (write-log! ~logger-sym ~level-sym ~cause (build-map-message ~props))))
-            `(write-log! ~logger-sym ~level-sym ~cause (build-map-message ~props)))))))
+            `(send-off logging-agent
+                       (fn [_#]
+                         (let [message# (or ~raw (build-map-message ~props))]
+                           (write-log! ~logger-sym ~level-sym ~cause message#))))
+            `(let [message# (or ~raw (build-map-message ~props))]
+               (write-log! ~logger-sym ~level-sym ~cause message#)))))))
```
hirunatan, line 77: I think you can move the `(let` to before the `(if async`, this way you only need to write it once

frontend/src/app/main/data/events.cljs
```diff
@@ -0,0 +1,238 @@
...
+                                       (when (fn? impl-fn)
+                                         (some-> (impl-fn event)
+                                                 (update :profile-id #(or % profile-id)))))))
+                           (rx/filter :profile-id)
+                           (rx/map (fn [event]
+                                     (let [session* (or @session (dt/now))
+                                           context  (-> @context
+                                                        (d/merge (:context event))
+                                                        (assoc :session session*))]
+                                       (swap! session (constantly session*))
```
hirunatan: Could it be `(reset! session session*)`?

backend/src/app/util/logging.clj
niwinz: No, this makes the message building synchronous, and the purpose of async is that message building is offlodead to an other thread.

frontend/src/app/main/data/events.cljs
niwinz: you are right, the swap! here is unnecesary. I'll change it.

backend/src/app/util/logging.clj
hirunatan: Okkk i didn't notice that  `(fn [_#])`

PR #905, opened 2021-05-05
Remove interactions on delete

hirunatan:

> Screenshots/GIFs

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 
 ### :sparkles: New features
 ### :bug: Bugs fixed
+
+- Remove interactions when the destination artboard is deleted [Taiga #1656](https://tree.taiga.io/project/penpot/issue/1656)
+
 ### :arrow_up: Deps updates
 ### :boom: Breaking changes
 ### :heart: Community contributions by (Thank you!)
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/data/workspace/common.cljs
```diff
@@ -372,6 +371,12 @@
                     #{}
                     ids)
 
+            interacting-shapes
+            (filter (fn [shape]
+                      (let [interactions (:interactions shape)]
+                        (seq (filter #(contains? ids (:destination %)) interactions))))
```
niwinz: I think `some` is a better idiom here (instead of filter + seq)

niwinz: And if  ids is a **set**, this can expressed as `(some ids (map :destination interactions))` if I'm not wrong

hirunatan: good trick, very compact!

CHANGES.md
```diff
@@ -28,6 +28,7 @@
 - Fix problem with pan and space [#811](https://github.com/penpot/penpot/issues/811)
 - Fix issue when parsing exponential numbers in paths
 - Remove legacy system user and team [#843](https://github.com/penpot/penpot/issues/843)
+- Remove interactions when the destination artboard is deleted [Taiga #1656](https://tree.taiga.io/project/penpot/issue/1656)
```
niwinz: I think it should be on the `next` release section.

hirunatan: true!

PR #918, opened 2021-05-06
:sparkles: Preserve layer order when copying shapes to the clipboard

hirunatan:
https://tree.taiga.io/project/penpot/issue/1617


niwinz:
I'll merge it but pls, update changelog with this.

PR #935, opened 2021-05-11
Frame thumbnails

Alotor:

> Screenshots/GIFs


frontend/src/app/main/data/workspace/persistence.cljs
```diff
@@ -572,3 +551,101 @@
...
+            :add-obj [(:id change)]
+            :mod-obj [(:id change)]
+            :del-obj [(:id change)]
+            :reg-objects (:shapes change)
+            :mov-objects (:shapes change)
+            []))
+
+        get-frame-id
+        (fn [id]
+          (or (and (= :frame (get-in objects [id :type])) id)
```
niwinz, line 588: is more efficient extract the shape from object and then check for type and or extract `frame-id`

PR #938, opened 2021-05-12
assets improvements

hirunatan:
https://tree.taiga.io/project/penpot/issue/1671

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/assets.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/assets.cljs
@@ -47,52 +47,108 @@
    [okulary.core :as l]
    [rumext.alpha :as mf]))
 
+; TODO: refactor to remove duplicate code and less parameter passing.
+;  - Move all state to [:workspace-local :assets-bar file-id :open-boxes {}
+;                                                            :open-groups {}
+;                                                            :reverse-sort?
+;                                                            :listing-thumbs?
+;                                                            :selected-assets {}]
+;  - Move selection code to independent functions that receive the state as a parameter.
+;
+; TODO: change update operations to admit multiple ids, thus avoiding the need of
+;       emitting many events and opening an undo transaction. Also move the logic
+;       of grouping, deleting, etc. to events in the data module, since now the
+;       selection info is in the global state.
+
 ;; ---- Group assets management ----
 
+(defn group-assets
+  "Convert a list of assets in a nested structure like this:
+
+    {'': [{assetA} {assetB}]
+     'group1': {'': [{asset1A} {asset1B}]
+                'subgroup11': {'': [{asset11A} {asset11B} {asset11C}]}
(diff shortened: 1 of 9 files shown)


frontend/src/app/main/ui/workspace/sidebar/assets.cljs
```diff
@@ -115,6 +116,53 @@
...
+          (fn [path]
+            (fn [event]
+              (dom/stop-propagation event)
+              (on-change-folded-groups
+                (toggle-folded-group folded-groups path)))))]
+    [:*
+     (when-not (empty? path)
+       (let [[other-path last-path truncated] (cp/compact-path path 35)]
+         [:div.group-title {:class (when-not group-open? "closed")
+                            :on-click (on-fold-group path)}
```
Alotor: This is strange. Why are you passing it as argument when it's a local function that can read it from the local?
Also it's strange the use-callback here because you're always calling it on render... 
I think you can create a use-callback that depends on `path` and don't need to pass it as argument, you can read it from the scope

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
```diff
@@ -200,22 +247,16 @@
             (st/emit! (dwu/start-undo-transaction))
             (apply st/emit!
                    (->> components
-                        (filter #(contains? selected-components (:id %)))
+                        (filter #(if multi-components?
```
Alotor, line 422: I think you should send just 1 event that changes all components instead of one per item. The issue with 1 event per component is that every event will change the objects and "force" a re-render. If you group them together and only change the state once you can have only 1 re-render. Also I think it's cleaner to move this filter to the `data` package.

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
hirunatan: This has been changed in the refactor.

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
hirunatan: This may be interesting. But it requires that the selected assets info is in the global state, now it's in a state local to the components. I'll do this after all state is moved to global space (I've added a TODO note for me not to forget it).

common/app/common/pages/helpers.cljc
```diff
@@ -408,12 +408,18 @@
     parent-idx))
 
 (defn split-path
-  [path]
   "Decompose a string in the form 'one / two / three' into
-  an array of strings, normalizing spaces."
+  a vector of strings, normalizing spaces."
+  [path]
   (->> (str/split path "/")
```
niwinz: This can be done using transducers for better efficiency.

hirunatan: One day I'll definitively learn what transducers are, and what are they useful for :P :P :P :P

PR #1005, opened 2021-06-02
Add resize constraints to shapes #1004

hirunatan:

> Screenshots/GIFs

### DIFF
--- a/frontend/src/app/main/data/workspace/transforms.cljs
+++ b/frontend/src/app/main/data/workspace/transforms.cljs
@@ -151,8 +151,7 @@
                                      :resize-origin origin
                                      :resize-transform shape-transform
                                      :resize-scale-text scale-text
-                                     :resize-transform-inverse shape-transform-inverse}
-                                    false))))
+                                     :resize-transform-inverse shape-transform-inverse}))))
 
           ;; Unifies the instantaneous proportion lock modifier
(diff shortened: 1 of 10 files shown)


frontend/src/app/main/data/workspace/transforms.cljs
```diff
@@ -426,10 +425,34 @@
 
 ;; -- Apply modifiers
 
+(defn- set-modifiers-recursive
+  [modif-tree objects shape modifiers]
+  (let [children (->> (get shape :shapes [])
+                      (map #(get objects %)))
+
+        transformed-shape (when (seq children)
```
Alotor: This could leave the `transformed-shape` to `nil` and you don't have any checks in place. What happens if this is nil?

hirunatan: This may occur when there are no children. But in this case the function set-child will never be executed.

Alotor: Then remove the `when` statement. Or add an assert at the beginning so we're sure this won't happen. If you're adding a condition is because this could happen.

hirunatan: The `when` is there so that if the value is not needed (because the shape has no children) the transformation is not calculated, thus saving a bit of time. I've added a comment for this.

PR #1020, opened 2021-06-08
:recycle: Reimplement GC mechanism for penpot database objects.

niwinz:


backend/src/app/rpc/mutations/profile.clj
```diff
@@ -181,7 +180,7 @@
 (defn create-profile
   "Create the profile entry on the database with limited input
   filling all the other fields with defaults."
-  [conn {:keys [id fullname email password is-active is-muted is-demo opts]
+  [conn {:keys [id fullname email password is-active is-muted is-demo opts deleted-at]
```
hirunatan, line 183: It should be useful to comment why do you accept a "deleted-at" attribute when creating a new profile (it's for demo accounts I understand). Without explanation this could be confusing for our future selves :P

backend/src/app/rpc/queries/profile.clj
```diff
@@ -92,14 +93,13 @@
 
 (def sql:retrieve-profile-by-email
   "select p.* from profile as p
-    where p.email = lower(?)
-      and p.deleted_at is null")
+    where p.email = lower(?)")
```
hirunatan: This is now unused, isn't it? May it be deleted?

backend/src/app/rpc/mutations/profile.clj
niwinz: This is a private function, and I think it is pretty clear, is a factory for profile row creation and deleted-at is a profile table field.

PR #1031, opened 2021-06-14
Add mechanism for offload the file data to external storage.

niwinz:


backend/src/app/config.clj
```diff
@@ -58,11 +58,8 @@
    :srepl-host "127.0.0.1"
    :srepl-port 6062
 
-   :storage-backend :fs
-
-   :storage-fs-directory "assets"
-   :storage-s3-region :eu-central-1
-   :storage-s3-bucket "penpot-devenv-assets-pre"
+   :assets-storage-backend :fs
+   :storage-assets-fs-directory "assets"
```
hirunatan, line 62: I would prefer to use always "assets-storage" or always "storage-assets", to make clear these settings are closely related.

hirunatan: I think I prefer "storage-assets". This way they will be also close to "storage-fdata".

PR #1046, opened 2021-06-17
Export-Import improvements

Alotor:

> Screenshots/GIFs

### DIFF
--- a/frontend/src/app/worker/import.cljs
+++ b/frontend/src/app/worker/import.cljs
@@ -118,7 +118,7 @@
 (defn resolve-images
   [file-id node]
   (if (and (not (cip/close? node))
-           (cip/has-image? type node))
+           (cip/has-image? node))
     (let [name     (cip/get-image-name node)
           data-uri (cip/get-image-data node)]
       (->> (upload-media-files file-id name data-uri)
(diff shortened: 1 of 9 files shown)


frontend/src/app/worker/import.cljs
```diff
@@ -162,25 +162,26 @@
...
-         (rx/flat-map #(process-page file-id zip %))
+         (rx/mapcat #(process-page file-id zip %))
          (merge-reduce import-page file)
          (rx/flat-map send-changes)
          (rx/ignore))))
 
+(defn process-package
+  [project-id zip-file]
+  (->> (uz/get-file zip-file "manifest.json")
+       (rx/flat-map (comp :files json/decode :content))
```
hirunatan, line 173: If the manifest file does not exist, or has an incorrect value, does this have a controlled behavior?

Alotor: Yes, will return error at the end of the stream

Alotor: In any case I'll handle error management in a future PR

hirunatan: Ok. We must be extremely cautious with all this, it may be a main security hole if not well protected.

For example, check the case of manifest.json exists and is a valid json file, but does not contain any :name or has a malicious file name in it...

PR #1066, opened 2021-06-29
Add library linking to export/import

Alotor:

> Screenshots/GIFs

### DIFF
--- a/backend/src/app/rpc/mutations/media.clj
+++ b/backend/src/app/rpc/mutations/media.clj
@@ -48,7 +48,7 @@
           :opt-un [::id]))
 
 (sv/defmethod ::upload-file-media-object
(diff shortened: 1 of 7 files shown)


backend/src/app/rpc/mutations/media.clj
```diff
@@ -48,7 +48,7 @@
           :opt-un [::id]))
 
 (sv/defmethod ::upload-file-media-object
-  [{:keys [pool] :as cfg} {:keys [profile-id file-id] :as params}]
+  [{:keys [pool] :as cfg} {:keys [profile-id file-id id] :as params}]
```
hirunatan: this id is not used?

Alotor: no, i've removed id

PR #1074, opened 2021-07-06
Import/drag drop

Alotor:
- Drag/drop on dashboard to import files
- Changed export progress modal


frontend/src/app/main/ui/dashboard/export.cljs
```diff
@@ -11,19 +11,68 @@
    [app.main.store :as st]
    [app.main.ui.icons :as i]
    [app.main.worker :as uw]
+   [app.util.data :refer [classnames]]
```
niwinz: app.util.data is deprecated and should not be used, use `dom/classnames` instead.

PR #1113, opened 2021-07-26
:sparkles: Ensure cleanining all nils from shapes reference field.

niwinz:


common/src/app/common/pages/changes.cljc
```diff
@@ -15,6 +15,20 @@
    [app.common.pages.spec :as spec]
    [app.common.spec :as us]))
 
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
hirunatan, line 19: Aren't those abstract enough to be in common/data?

niwinz: Yes and no, right now this is only useful in this ns, until now we didn't needed it on other namespaces, so I prefer keeping them here until we really need as general purpose.

An other reason is because we already have without-nils that can be changed to be polymorphic, but in this case want to ensure that it always returns a vector and not a coll of the same type as the input.

common/src/app/common/pages/changes.cljc
```diff
@@ -54,45 +68,50 @@
...
+                     (or (nil? frame-id) (contains? objects frame-id)))
+              (-> objects
+                  (assoc id (-> obj
+                                (assoc :frame-id frame-id)
+                                (assoc :parent-id parent-id)
+                                (assoc :id id)))
+                  (update parent-id update-parent))
+              objects))
+
+          (update-page [data]
```
hirunatan: This should be `(update-container [container]`.

Container is the abstraction we made for structures that may be a page or a component.

niwinz: you're right

common/src/app/common/pages/changes.cljc
```diff
@@ -107,32 +126,27 @@
...
-  (letfn [(delete-object [objects id]
+  (letfn [(delete-from-parent [parent]
+            (let [parent (update parent :shapes without-obj id)]
+              (cond-> parent
+                (and (:shape-ref parent)
+                     (not ignore-touched))
+                (-> (update :touched cph/set-touched-group :shapes-group)
+                    (dissoc :remote-synced?)))))
+
+          (delete-from-page [objects]
```
hirunatan: better `delete-from-objects`?

PR #1247, opened 2021-09-27
Bool shapes

Alotor:

> Screenshots/GIFs


common/src/app/common/path/shapes_to_path.cljc
```diff
@@ -4,22 +4,52 @@
...
-                   :medata])
-(def allowed-transform-types #{:rect
-                               :circle
-                               :image})
+   [app.common.path.bool :as pb]
+   [app.common.path.commands :as pc]))
+
+(def ^:const bezier-circle-c 0.551915024494)
+
+(def ^:const dissoc-attrs
```
niwinz: I don't think is a good idea declare **const** for not primitive JS values. I mean, it is OK, but the compiler will use the const values as inline values, and if this vars are used more than one time... it will increase the size of the final bundle without any real additional benefit.

PR #1265, opened 2021-10-05
Bugfixes

niwinz:


frontend/src/app/util/router.cljs
```diff
@@ -110,21 +110,6 @@
...
-  ptk/EffectEvent
-  (effect [_ state _]
-    (let [router (:router state)
-          path   (resolve router id path-params query-params)
-          uri    (-> (u/uri cfg/public-uri)
-                     (assoc :fragment path))
-          name   (str (name id) "-" (:file-id path-params))]
-      (dom/open-new-window (str uri) name))))
-
-(defn nav-new-window
```
hirunatan, line 116: If this one does not exist, couldn't we change the other event to nav-new-window (without the *)?

niwinz: Yes, but right now, I don't want to enlarge the scope of this MR, this can be done in other MR.

hirunatan: good

PR #1424, opened 2021-12-27
Performance improvements

Alotor:


common/src/app/common/geom/shapes/path.cljc
```diff
@@ -279,11 +279,18 @@
...
-         ykey (keyword (str prefix "y"))
+  ([command]
+   (command->point command nil))
+
+  ([command coord]
+   (let [params (:params command)
+         xkey (cond (= :c1 coord) :c1x
+                    (= :c2 coord) :c2x
+                    :else         :x)
+         ykey (cond (= :c1 coord) :c1y
```
niwinz: this `cond` can be a `case` that is much efficient

common/src/app/common/geom/shapes/transforms.cljc
```diff
@@ -14,20 +14,23 @@
...
+(defn- move-selrect [selrect pt]
+  (let [dx (.-x pt)
+        dy (.-y pt)]
+    (-> selrect
+        (update :x + dx)
+        (update :y + dy)
+        (update :x1 + dx)
+        (update :y1 + dy)
+        (update :x2 + dx)
+        (update :y2 + dy))))
```
niwinz: If the selrect is sensitive to performance, maybe we need to define a new type for selrect and this case, replace this update chain (that in this case will create at least 6 garbage objects). With a record we probably just create a single one with updated modifiers.

This is just a thinking, nothing definitive, we can talk about it and do it in other MR.

Alotor: I'll change the update for a new object creation so it has less objects

common/src/app/common/geom/shapes/bool.cljc
```diff
@@ -6,21 +6,27 @@
...
+   [app.common.data :as d]
    [app.common.geom.shapes.path :as gsp]
    [app.common.geom.shapes.rect :as gpr]
    [app.common.geom.shapes.transforms :as gtr]
    [app.common.path.bool :as pb]
    [app.common.path.shapes-to-path :as stp]))
 
+(defn calc-bool-content
+  [shape objects]
+  (->> (:shapes shape)
```
niwinz: I think this can be implemented using transducers

frontend/src/app/main/refs.cljs
```diff
@@ -232,23 +232,13 @@
...
-                           (gsh/merge-modifiers modifiers))
-                 xform (comp (map (d/getf objects))
-                             (remove nil?))]
-             (into [] xform ids)))]
-     (l/derived selector st/state =))))
+  [ids]
+  (let [selector
+        (fn [state]
+          (let [objects (wsh/lookup-page-objects state)
+                xform (comp (map (d/getf objects)) (remove nil?))]
```
niwinz: this can be simplified by using `keep` instead of map + filter nil?

frontend/src/app/main/store.cljs
```diff
@@ -46,17 +46,10 @@
...
-   nil)
-  ([event & events]
-   (apply ptk/emit! state (cons event events))
-   nil))
+(def emit! (partial ptk/emit! state))
 
 (defn emitf
   [& events]
-  #(apply ptk/emit! state events))
+  #(ptk/emit! state events))
```
niwinz: This is not the same code, on the new code you executing `ptk/emit!` with state and always a seq of events but this function is variadic one. In fact, this probable breaks some other code that uses emitf.

common/src/app/common/geom/shapes/bool.cljc
Alotor: OK

frontend/src/app/main/refs.cljs
Alotor: OK

frontend/src/app/main/store.cljs
Alotor: OK, I'll revert this change

frontend/src/app/main/ui/workspace/viewport.cljs
```diff
@@ -94,25 +92,22 @@
         drawing-tool      (:tool drawing)
         drawing-obj       (:object drawing)
 
-        selected-shapes   (into []
-                                (comp (map #(get objects %))
-                                      (filter some?))
-                                selected)
+        xf-select-shape   (comp (map (d/getf objects-modified)) (filter some?))
```
niwinz: map+filter => keep

Alotor: OK

PR #1461, opened 2022-01-12
:sparkles: Upload/dismiss all custom fonts at once

superalex:
Related to https://tree.taiga.io/project/penpot/us/1931

### DIFF
--- a/frontend/src/app/main/ui/dashboard/fonts.cljs
+++ b/frontend/src/app/main/ui/dashboard/fonts.cljs
@@ -103,6 +103,10 @@
                          (fn [error]
                            (js/console.log "error" error))))))
 
+        on-upload-all
+        (fn [items]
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/ui/dashboard/fonts.cljs
```diff
@@ -103,6 +103,10 @@
                          (fn [error]
                            (js/console.log "error" error))))))
 
+        on-upload-all
+        (fn [items]
+          (dorun (map on-upload items)))
```
niwinz: For this purpose is much better use just `run!` (replace the `dorun + map` with `run!`)

superalex: Thank you very much!, I had never used run! before :)

PR #1472, opened 2022-01-14
:sparkles: Pressing enter to exit the fields

superalex:
Related to https://tree.taiga.io/project/penpot/us/2447


frontend/src/app/main/ui/components/numeric_input.cljs
```diff
@@ -148,18 +148,20 @@
 
         handle-key-down
         (mf/use-callback
-         (mf/deps set-delta apply-value update-input)
+         (mf/deps ref set-delta apply-value update-input)
```
niwinz: you don't need to add refs to the deps, the refs are mutable state exposed from react, and they are stable (i mean they does not change pointer)

PR #1488, opened 2022-01-17
:bug: Disable embed images in export

hirunatan:


frontend/src/app/main/ui/render.cljs
```diff
@@ -107,7 +107,7 @@
      #(dom/set-page-style {:size (str (mth/ceil width) "px "
                                       (mth/ceil height) "px")}))
 
-    [:& (mf/provider embed/context) {:value true}
```
Alotor, line 110: This will break import/export penpot files

niwinz: I don't think it breaks the export, export uses its own `app.main.render` ns and passes the embedd parameter explicitly.
This change does not affects the export process.

niwinz:
I will cherry-pick the commit to my branch.

PR #1489, opened 2022-01-18
:sparkles: Graphic tablet use improvements: add scroll bars

superalex:
Related to https://tree.taiga.io/project/penpot/task/2550

### DIFF
--- a/frontend/src/app/main/ui/workspace/viewport/scroll_bars.cljs
+++ b/frontend/src/app/main/ui/workspace/viewport/scroll_bars.cljs
@@ -1,0 +1,212 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) UXBOX Labs SL
+
+(ns app.main.ui.workspace.viewport.scroll-bars
+  (:require
+   [app.common.geom.shapes :as gsh]
+   [app.common.geom.shapes.rect :as gpr]
+   [app.common.pages.helpers :as cph]
+   [app.main.data.workspace :as dw]
+   [app.main.store :as st]
+   [app.main.ui.workspace.viewport.utils :as utils]
+   [app.util.dom :as dom]
+   [rumext.alpha :as mf]))
+
+(mf/defc viewport-scrollbars
+  {::mf/wrap [mf/memo]}
+  [{:keys [objects viewport-ref zoom vbox]}]
+
+  (let [v-scrolling?              (mf/use-state false)
+        h-scrolling?              (mf/use-state false)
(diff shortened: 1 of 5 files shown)


frontend/src/app/main/ui/workspace/viewport/scroll_bars.cljs
```diff
@@ -0,0 +1,139 @@
...
+        scrollbar-height-stored (mf/ref-val scrollbar-height-ref)
+        height-factor-ref       (mf/use-ref nil)
+        vbox-y-ref              (mf/use-ref nil)
+
+        base-objects            (mf/deref refs/workspace-page-objects)
+        base-objects-rect
+        (mf/use-memo
+         (mf/deps base-objects)
+         (fn []
+           (let [root-shapes             (get-in base-objects [uuid/zero :shapes])
```
niwinz: this can be written using `(-> base-objects cph/get-top-frame :shapes)`
It also requires importing the `app.common.page.helpers`
**very minor**, uses a builtin helper instead of raw access

frontend/src/app/main/ui/workspace/viewport/scroll_bars.cljs
```diff
@@ -0,0 +1,139 @@
...
+
+        ;; bottom space hidden because of the scroll
+        bottom-offset           (-> (- (:y2 base-objects-rect) (+ (:y vbox) vbox-height))
+                                    (max 0)
+                                    (* vbox-height)
+                                    (/ (:height base-objects-rect)))
+
+        show-vertical-scroll?   (or @scrolling? (> top-offset 0) (> bottom-offset 0))
+
+        scrollbar-x             (+ (:x vbox) (:width vbox) (* inv-zoom -32))
```
niwinz: for performance reasons, consider extract the commonly used attrs from `vbox` such that `x` and `y` (they are used many times as `(:x vbox)...`), this improves performance because the access to a map has always more overhead than just access a constant value in scope.

frontend/src/app/main/ui/workspace/viewport/scroll_bars.cljs
```diff
@@ -0,0 +1,139 @@
...
+                                    (max (+ (:y vbox) (* inv-zoom 30))))
+
+        scrollbar-y             (if (> (+ scrollbar-y scrollbar-height) (+ (:y vbox) vbox-height)) ;; the scroll bar is stick to the bottom
+                                  (-> (+ (:y vbox) vbox-height)
+                                      (- scrollbar-height))
+                                  scrollbar-y)
+
+        on-mouse-move
+        (mf/use-callback
+         (mf/deps viewport-ref zoom height-factor-ref scrolling?)
```
niwinz: don't put react refs inside `deps`, refas are stable mutable pointer, I mean the ref instance does not change only change the pointed value, so putting them in deps is NOOP.

PR #1490, opened 2022-01-18
Enhancements and fixes

niwinz:


backend/src/app/http/middleware.clj
```diff
@@ -78,7 +102,8 @@
       (coll? body)
       (-> response
           (update :headers assoc "content-type" "application/transit+json")
-          (assoc :body (t/encode body opts)))
+          (assoc :body (transit-streamable-body body opts))
+          #_(assoc :body (t/encode body opts)))
```
Alotor: Remove comment

frontend/src/app/main/render.cljs
```diff
@@ -41,7 +370,7 @@
 (defn populate-fonts-cache [objects]
   (let [texts (->> objects
                    (vals)
-                   (filterv text?)
+                   (filterv #(= (:type %) :text))
```
Alotor, line 373: Is there any reason for this change?

niwinz: yes and no, text? predicate is not used anywhere, so for so small function I considered remove it from toplevel.

PR #1494, opened 2022-01-19
Improvements & Fixes (part 7)

niwinz:


exporter/src/app/browser.cljs
```diff
@@ -67,25 +67,21 @@
...
 
 (defn pdf
   ([page] (pdf page nil))
-  ([page {:keys [viewport omit-background? prefer-css-page-size? save-path]
-          :or {viewport {}
-               omit-background? true
-               prefer-css-page-size? true
-               save-path nil}}]
-   (let [viewport (d/merge default-viewport viewport)]
+  ([page {:keys [viewport omit-background? prefer-css-page-size? save-path]}]
```
hirunatan: If not used, perhaps the variables omit-background? and prefer-css-page-size may be removed.

backend/src/app/rpc/queries/files.clj
```diff
@@ -215,6 +217,39 @@
...
+(declare trim-file-data)
+
+(s/def ::page-id ::us/uuid)
+(s/def ::object-id ::us/uuid)
+
+(s/def ::trimmed-file
+  (s/keys :req-un [::profile-id ::id ::object-id ::page-id]))
+
+(sv/defmethod ::trimmed-file
+  "Retrieve a file by its ID and trims all unnecesary content from it."
```
hirunatan: Explain a bit more what is trimmed here? e.g. that we remove cached content to render one specific object.

PR #1498, opened 2022-01-21
Mainly cosmetic changes

niwinz:


common/src/app/common/pages/changes.cljc
```diff
@@ -160,9 +160,7 @@
                   update-fn #(d/update-when %1 %2 update-group %1)
                   xform     (comp
                              (mapcat #(cons % (cph/get-parents % objects)))
-                             (map lookup)
-                             (filter #(contains? #{:group :bool} (:type %)))
-                             (map :id)
```
hirunatan, line 165: the map to :id is removed?

niwinz: yes, because the original collection is a list of ID's. previusly we have lookup that retrieves the object, filter and then again extract the `:id`. NOW: we just do the lookup on filter.

hirunatan: okk

PR #1506, opened 2022-01-24
:sparkles: Add select layer to contest menu

EvaMarco:
This pr resolves [Taiga #2474](https://tree.taiga.io/project/penpot/us/2474)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,10 +4,11 @@
 
 ### :sparkles: New features
 
+- Add select layer option to context menu [Taiga #2474](https://tree.taiga.io/project/penpot/us/2474).
 - Guides [Taiga #290](https://tree.taiga.io/project/penpot/us/290?milestone=307334)
 - Improve file menu by adding semantically groups [Github #1203](https://github.com/penpot/penpot/issues/1203).
 - Add update components in bulk option in context menu [Taiga #1975](https://tree.taiga.io/project/penpot/us/1975).
-- Create e2e tests for drawing basic fors [Taiga #2608](https://tree.taiga.io/project/penpot/task/2608).
+- Create e2e tests for drawing basic shapes [Taiga #2608](https://tree.taiga.io/project/penpot/task/2608).
 - Create firsts e2e test [Taiga #2608](https://tree.taiga.io/project/penpot/task/2608).
 
 ## 1.11.0-beta
--- a/frontend/src/app/main/ui/workspace/context_menu.cljs
+++ b/frontend/src/app/main/ui/workspace/context_menu.cljs
@@ -13,6 +13,7 @@
    [app.main.data.workspace :as dw]
    [app.main.data.workspace.interactions :as dwi]
    [app.main.data.workspace.libraries :as dwl]
+   [app.main.data.workspace.selection :as dws]
    [app.main.data.workspace.shortcuts :as sc]
    [app.main.refs :as refs]
    [app.main.store :as st]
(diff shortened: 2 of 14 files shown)


frontend/src/app/main/ui/workspace/context_menu.cljs
```diff
@@ -64,22 +90,33 @@
...
-       [:span.submenu-icon i/arrow-slide])
-
-     (when (> (count children) 1)
-       [:ul.workspace-context-menu
-        {:ref submenu-ref
-         :style {:display "none" :left 250}
-         :on-context-menu prevent-default}
-        children])]))
+    (if has-icon?
+      [:li.sub-menu-item {:ref set-dom-node
```
hirunatan: Perhaps is better to name this class .icon-menu-item? Because I think it does not apply to all sub menu items, but only those with icons.

frontend/src/app/main/ui/workspace/context_menu.cljs
```diff
@@ -64,22 +90,33 @@
...
+                          :on-click on-click
+                          :on-pointer-enter on-pointer-enter
+                          :on-pointer-leave on-pointer-leave}
+       (when has-icon?
```
hirunatan: If we reached here, has-icon? has been already checked to be true

frontend/src/app/main/ui/workspace/context_menu.cljs
```diff
@@ -33,8 +34,33 @@
...
+            :exclude      i/bool-exclude
+            :intersection i/bool-intersection
+            #_:default    i/bool-union)
+    :svg-raw i/file-svg
+    nil))
+
+
 (mf/defc menu-entry
-  [{:keys [title shortcut on-click children] :as props}]
+  [{:keys [title shortcut on-click children selected has-icon? shape] :as props}]
```
hirunatan: `selected` should be `selected?`

hirunatan: Not sure if it's possible, but I'll prefer than, instead of passing a shape and calculating the icon here, to pass directly the icon, and do the calculation outside. So this component does not need to know anything about shapes. Furthermore, the has-icon? parameter would be unnecesary, it'd be enough to check icon for null.

frontend/src/app/main/ui/workspace/context_menu.cljs
```diff
@@ -33,8 +34,33 @@
...
+             (if (:masked-group? shape)
+               i/mask
+               i/folder))
+    :bool (case (:bool-type shape)
+            :difference   i/bool-difference
+            :exclude      i/bool-exclude
+            :intersection i/bool-intersection
+            #_:default    i/bool-union)
+    :svg-raw i/file-svg
+    nil))
```
hirunatan: I think this could be a component in main/ui/components named shape-icon. This way is not duplicated and is clearer to import it from layers menu and sidebars.

frontend/src/app/main/ui/workspace/context_menu.cljs
EvaMarco: For sure¡¡

frontend/src/app/main/ui/workspace/context_menu.cljs
EvaMarco: true, changed

frontend/src/app/main/ui/workspace/context_menu.cljs
EvaMarco: Great idea¡¡

frontend/src/app/main/ui/workspace/context_menu.cljs
EvaMarco: The has-icon? naming was really bad, I change it and move the icon check to its parent as you suggest. Tell me if it's not clear enough yet.

CHANGES.md
```diff
@@ -4,7 +4,9 @@
 
 ### :sparkles: New features
 
+- Add select layer option to context menu [Taiga #2474](https://tree.taiga.io/project/penpot/us/2474).
 - Guides [Taiga #290](https://tree.taiga.io/project/penpot/us/290?milestone=307334)
+- Create e2e tests for drawing basic fors [Taiga #2608](https://tree.taiga.io/project/penpot/task/2608).
```
hirunatan: ?

PR #1508, opened 2022-01-24
Guides

Alotor:
https://tree.taiga.io/project/penpot/us/290


frontend/src/app/main/data/workspace/guides.cljs
```diff
@@ -0,0 +1,101 @@
...
+  (us/verify ::tpo/guide guide)
+  (ptk/reify ::update-guides
+    ptk/WatchEvent
+    (watch [it state _]
+      (let [page-id (:current-page-id state)
+            guides (-> state wsh/lookup-page-options (:guides {}))
+
+            new-guides (assoc guides (:id guide) guide)
+            
+            rch [{:type :set-option
```
niwinz: maybe we can do this using the same approach as in `app.common.pages.changes-builder` ?

frontend/src/app/main/data/workspace/guides.cljs
```diff
@@ -0,0 +1,101 @@
...
+(defn move-frame-guides
+  "Move guides that are inside a frame when that frame is moved"
+  [ids]
+  (us/verify (s/coll-of uuid?) ids)
+
+  (ptk/reify ::move-frame-guides
+    ptk/WatchEvent
+    (watch [_ state _]
+      (let [objects (wsh/lookup-page-objects state)
+            frame-ids? (->> ids (filter #(= :frame (get-in objects [% :type]))) (into #{}))
```
niwinz: you can retrieve the frames using `app.common.pages.helpers/select-frames`. Right now this function returns always a vector but maybe it can be defined like this to accept different collection type:

```clojure
(defn select-frames
  ([objects] (select-frames objects []))
  ([objects initial]
   (let [lookup #(get objects %)
         frame? #(= :frame (:type %))
         xform  (comp (map lookup)
                      (filter frame?))]
     (->> (:shapes (lookup uuid/zero))
          (into initial xform)))))
```

This is just a suggestion, I'm not sure if this change to the select-frames function is worth.

frontend/src/app/main/ui/workspace/viewport/snap_points.cljs
```diff
@@ -152,29 +152,26 @@
   {::mf/wrap [mf/memo]}
   [{:keys [layout zoom objects selected page-id drawing transform modifiers] :as props}]
 
-  (let [;; shapes        (mf/deref (refs/objects-by-id selected))
-        ;; filter-shapes (mf/deref refs/selected-shapes-with-children)
+  (let [shapes
+        (->> selected
+             (map #(get objects %))
+             (filterv (comp not nil?)))
```
niwinz: maybe replace with `(into [] (keep #(get objects %)) selected)` ?

frontend/src/app/util/snap_data.cljs
```diff
@@ -0,0 +1,245 @@
...
+
+(def snap-attrs [:frame-id :x :y :width :height :hidden :selrect :grids])
+
+;; PRIVATE FUNCTIONS
+
+(defn- make-insert-tree-data
+  [shape-data axis]
+  (fn [tree]
+    (let [tree (or tree (rt/make-tree))]
+      (as-> tree $
```
niwinz: for a single operation I think `as->` is useless here and in the following functions

frontend/src/app/util/snap_data.cljs
```diff
@@ -0,0 +1,245 @@
...
+    (-> page-data
+        (assoc-in [frame-id :x] (rt/make-tree))
+        (assoc-in [frame-id :y] (rt/make-tree)))))
+
+(defn- add-frame
+  [page-data frame]
+  (let [frame-id (:id frame)
+        parent-id (:parent-id frame)
+        frame-data (->> (snap/shape-snap-points frame)
+                        (map #(hash-map :type :shape
```
niwinz: forcing use hash-map for maps less than 8 keys is inefficient. My suggestion here is use the `array-map` instead of `hash-map` or let the cljs compiler decide using `#(-> {:type :shape, ...})` or `#(do {:type :shape, ...})`.

niwinz: Also, you using `map` and I think you here want to use `mapv`, because frame-data (on difference with other props) is also associated directly to state (the others are concatenated using concat-vec...)

frontend/src/app/util/snap_data.cljs
```diff
@@ -0,0 +1,245 @@
...
+        (update-in [frame-id :y] #(or % (rt/make-tree)))
+        (update-in [frame-id :x] (make-insert-tree-data (d/concat-vec frame-data grid-x-data) :x))
+        (update-in [frame-id :y] (make-insert-tree-data (d/concat-vec frame-data grid-y-data) :y)))))
+
+(defn- add-shape
+  [page-data shape]
+  (let [frame-id    (:frame-id shape)
+        snap-points (snap/shape-snap-points shape)
+        shape-data  (->> snap-points
+                         (mapv #(hash-map
```
niwinz: the same here about the use of `hash-map`

frontend/src/app/worker/snaps.cljs
```diff
@@ -6,179 +6,31 @@
...
-                (rt/range-query from to)))]
-    (->> ranges
-         (mapcat calculate-range)
-         set ;; unique
-         (into []))))
+  [{:keys [page-id frame-id axis ranges] :as message}]
+  (->> ranges
+       (mapcat #(sd/query @state page-id frame-id axis %))
+       (set) ;; unique
+       (into [])))
```
niwinz: I think this can be implemented using transducers:

```clojure
(into [] (comp #(sd/query @state page-id frame-id axis %) (distinct)) ranges)
```

frontend/src/app/main/data/workspace/guides.cljs
Alotor: OK

frontend/src/app/main/data/workspace/guides.cljs
Alotor: I'm going to change it to this:
```clojure
            is-frame? (fn [id] (= :frame (get-in objects [id :type])))
            frame-ids? (into #{} (filter is-frame?) ids)
```

frontend/src/app/main/ui/workspace/viewport/snap_points.cljs
Alotor: ok

frontend/src/app/util/snap_data.cljs
Alotor: OK

frontend/src/app/util/snap_data.cljs
Alotor: OK.

frontend/src/app/worker/snaps.cljs
Alotor: ok

PR #1511, opened 2022-01-25
Refactor Specs

niwinz:


common/src/app/common/pages/helpers.cljc
```diff
@@ -231,13 +231,14 @@
...
 (defn select-frames
-  [objects]
-  (let [lookup #(get objects %)
-        frame? #(= :frame (:type %))
-        xform  (comp (map lookup)
-                     (filter frame?))]
-    (->> (:shapes (lookup uuid/zero))
-         (into [] xform))))
+  ([objects] (select-frames objects []))
+  ([objects initial]
```
hirunatan: why this?

niwinz: oh, this should not be here, i revert it, thanks

PR #1512, opened 2022-01-25
Authorized cookie

hirunatan:
This MR sets a cookie that we can use to check from other sites of the same domain if a user is registered. Is not intended for on premise installations, although nothing prevents using it if some one wants to.

### DIFF
--- a/backend/src/app/config.clj
+++ b/backend/src/app/config.clj
@@ -48,6 +48,7 @@
    :database-uri "postgresql://postgres/penpot"
    :database-username "penpot"
    :database-password "penpot"
+   :authenticated-cookie-domain "penpot.app"
 
    :default-blob-version 3
    :loggers-zmq-uri "tcp://localhost:45556"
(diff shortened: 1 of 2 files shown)


backend/src/app/http/session.clj
```diff
@@ -48,22 +49,34 @@
...
+                                          :http-only true
+                                          :value id
+                                          :same-site (if cors? :none :lax)
+                                          :secure secure?})
+
+                (some? authenticated-cookie-domain)
+                (assoc authenticated-cookie-name {:domain authenticated-cookie-domain
+                                                  :value true
+                                                  :same-site :strict
+                                                  :secure true}))))))
```
niwinz: should be used the `secure?` here I think

hirunatan: done

backend/src/app/config.clj
```diff
@@ -48,6 +48,7 @@
    :database-uri "postgresql://postgres/penpot"
    :database-username "penpot"
    :database-password "penpot"
+   :authenticated-cookie-domain "penpot.app"
```
niwinz, line 51: I think we should not have it defaults, this feature is mainly "internal"

hirunatan: This is a provisonal commit I made for local testing, will be removed prior to merge.

hirunatan:
Replaced by other PR

PR #1527, opened 2022-02-01
:bug: Fix ungroup a component leaves asterisk in layers

hirunatan:
https://tree.taiga.io/project/penpot/issue/2694

### DIFF
--- a/frontend/src/app/main/data/workspace/groups.cljs
+++ b/frontend/src/app/main/data/workspace/groups.cljs
@@ -9,6 +9,7 @@
    [app.common.data :as d]
    [app.common.geom.shapes :as gsh]
    [app.common.pages :as cp]
+   [app.common.pages.changes-builder :as cb]
    [app.common.spec :as us]
    [app.main.data.workspace.changes :as dch]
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/data/workspace/groups.cljs
```diff
@@ -147,29 +148,25 @@
...
+                            :shape-ref
+                            :touched))]
+
+    (cond-> (-> (cb/empty-changes it page-id)
+                (cb/with-objects objects)
+                (cb/change-parent parent-id children index-in-parent)
+                (cb/remove-objects [(:id group)]))
+
+      (some? ids-to-detach)
+      (cb/update-shapes ids-to-detach detach-fn))))
```
niwinz, line 169: This can be written in this way that I think is easier to read:

```clojure
(-> (cb/empty-changes it page-id)
    (cb/with-objects objects)
    (cb/change-parent parent-id children index-in-parent)
    (cb/remove-objects [(:id group)])
    (cond-> (some? ids-to-detach)
      (cb/update-shapes ids-to-detach detach-fn)))
```

PR #1529, opened 2022-02-01
:sparkles: On user settings, hide the theme selector as long as we only have one theme

pabloalba:
Related to https://tree.taiga.io/project/penpot/us/2610

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 ### :sparkles: New features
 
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -23,6 +23,7 @@
 
 ### :sparkles: New features
 
+- On user settings, hide the theme selector as long as we only have one theme [Taiga #2610](https://tree.taiga.io/project/penpot/us/2610)
```
hirunatan: If this MR goes to develop, the line should be added in the top section, not in release 1.11

pabloalba: fixed

PR #1534, opened 2022-02-02
Toolbars Redesign

Alotor:
https://tree.taiga.io/project/penpot/us/2319

### DIFF
--- a/frontend/src/app/main/ui/hooks/resize.cljs
+++ b/frontend/src/app/main/ui/hooks/resize.cljs
@@ -32,34 +32,38 @@
         start-ref (mf/use-ref nil)
 
         on-pointer-down
-        (fn [event]
-          (dom/capture-pointer event)
-          (mf/set-ref-val! start-size-ref @size-state)
-          (mf/set-ref-val! dragging-ref true)
-          (mf/set-ref-val! start-ref (dom/get-client-position event))
-          (set! last-resize-type resize-type))
+        (mf/use-callback
+         (mf/deps @size-state)
+         (fn [event]
+           (dom/capture-pointer event)
+           (mf/set-ref-val! start-size-ref @size-state)
+           (mf/set-ref-val! dragging-ref true)
+           (mf/set-ref-val! start-ref (dom/get-client-position event))
+           (set! last-resize-type resize-type)))
 
         on-lost-pointer-capture
-        (fn [event]
-          (dom/release-pointer event)
-          (mf/set-ref-val! start-size-ref nil)
-          (mf/set-ref-val! dragging-ref false)
-          (mf/set-ref-val! start-ref nil)
-          (set! last-resize-type nil))
+        (mf/use-callback
+         (fn [event]
+           (dom/release-pointer event)
+           (mf/set-ref-val! start-size-ref nil)
+           (mf/set-ref-val! dragging-ref false)
+           (mf/set-ref-val! start-ref nil)
+           (set! last-resize-type nil)))
 
         on-mouse-move
-        (fn [event]
-          (when (mf/ref-val dragging-ref)
-            (let [start (mf/ref-val start-ref)
-                  pos (dom/get-client-position event)
-                  delta (-> (gpt/to-vec start pos)
-                            (cond-> negate? gpt/negate)
-                            (get axis))
-                  start-size (mf/ref-val start-size-ref)
-                  new-size (-> (+ start-size delta) (max min-val) (min max-val))]
-              
-              (reset! size-state new-size)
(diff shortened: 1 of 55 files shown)


frontend/src/app/main/ui/hooks/resize.cljs
```diff
@@ -0,0 +1,105 @@
...
+  (let [current-file-id (mf/deref refs/current-file-id)
+        size-state (mf/use-state (or (get-in @storage [::saved-resize current-file-id key]) initial))
+        parent-ref (mf/use-ref nil)
+
+        dragging-ref (mf/use-ref false)
+        start-size-ref (mf/use-ref nil)
+        start-ref (mf/use-ref nil)
+
+        on-pointer-down
+        (fn [event]
```
niwinz: You probably want use here the `mf/use-callback` or `mf/use-fn` (short alias of the first one)
The same as  for the next function.

frontend/src/app/main/ui/hooks/resize.cljs
```diff
@@ -0,0 +1,105 @@
...
+        node-ref
+        (mf/use-callback
+         (mf/deps callback)
+         (fn [node]
+           (let [current-observer (mf/ref-val current-observer-ref)
+                 prev-val         (mf/ref-val prev-val-ref)]
+
+             (when (and (not= prev-val node) (some? current-observer))
+               (log/debug :action "disconnect" :js/prev-val prev-val :js/node node)
+               (.disconnect current-observer)
```
niwinz, line 90: Remember to use `^js` hint on interop calls on unknown objects, in this case current-observer is clearly unknown to the compiler, and this can cause unexpected issues in advanced compilations

niwinz: The same applies to all interop calls that has unknown type for the compiler on this file.

frontend/src/app/main/ui/workspace/textpalette.cljs
```diff
@@ -0,0 +1,153 @@
...
+  (l/derived #(select-keys % [:editors]) refs/workspace-local =))
+
+(mf/defc textpalette
+  {::mf/wrap [mf/memo]}
+  []
+  (let [local             (mf/deref local-data)
+        selected-ids      (mf/deref refs/selected-shapes)
+        file-typographies (mf/deref refs/workspace-file-typography)
+        shared-libs       (mf/deref refs/workspace-libraries)
+        current-file-id   (mf/deref refs/current-file-id)]
```
niwinz: you don't need this deref. You already can use `(mf/use-ctx app.main.ui.context/current-file-id)`

Because workspace already provides this values using react context:

```clojure
    [:& (mf/provider ctx/current-file-id) {:value (:id file)}
     [:& (mf/provider ctx/current-team-id) {:value (:team-id project)}
      [:& (mf/provider ctx/current-project-id) {:value (:id project)}
       [:& (mf/provider ctx/current-page-id) {:value page-id}
       ...
```

frontend/src/app/main/ui/workspace/textpalette.cljs
```diff
@@ -0,0 +1,153 @@
...
+    [:& palette {:local local
```
niwinz: You passing local all down to the `typography-item` when you only need the `:editors` key.

frontend/src/app/main/ui/hooks/resize.cljs
Alotor: Ok

frontend/src/app/main/ui/hooks/resize.cljs
Alotor: ok, good catch

frontend/src/app/main/ui/workspace/textpalette.cljs
Alotor: OK

frontend/src/app/main/ui/workspace/textpalette.cljs
Alotor: Local only has the :editors key. It's a custom ref defined just above

```clojure
(def local-data
  (l/derived #(select-keys % [:editors]) refs/workspace-local =))
  ```

niwinz: Ok, You are right. In this case, two things:
This lense can be defined as `(l/derived :editors refs/workspace-local =)`.

And thinking a bit more on it... This is the case where I guess the better approach is passing the editors value from the parent (the parent component already derefs the refs/workspace-local) so you don't need to derefs it here with a ad-hoc lense).

PR #1539, opened 2022-02-04
:sparkles: Disallow using same password as user email

pabloalba:
Related to https://tree.taiga.io/project/penpot/us/2454

### DIFF
--- a/frontend/src/app/main/ui/auth/register.cljs
+++ b/frontend/src/app/main/ui/auth/register.cljs
@@ -60,6 +60,10 @@
     :email-already-exists
     (swap! form assoc-in [:errors :email]
            {:message "errors.email-already-exists"})
+    
+    :email-as-password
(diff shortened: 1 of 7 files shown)


niwinz:
Looks nice. I think we should also apply the same policy on the "password change" section.

frontend/src/app/main/ui/auth/register.cljs
```diff
@@ -64,6 +64,10 @@
     :email-already-exists
     (swap! form assoc-in [:errors :email]
            {:message "errors.email-already-exists"})
+    
+    :email-as-password
+    (swap! form assoc-in [:errors :password]
+       {:message "errors.email-as-password"})
```
niwinz: The `{:message...}` should be indented in the same way as previous swap!

pabloalba: fixed

PR #1545, opened 2022-02-04
Refactor page helpers

niwinz:


backend/dev/user.clj
```diff
@@ -6,6 +6,7 @@
 
 (ns user
   (:require
+   [app.common.data :as d]
```
hirunatan, line 9: is this unused? does not fire a linter warning?

niwinz: This is `user` namespace, it intended for REPL usage, user namespace is excluded from linter.

hirunatan: Understood!

PR #1571, opened 2022-02-09
:bug: fix error when posting an empty comment

EvaMarco:
This PR fixes https://tree.taiga.io/project/penpot/issue/2603


frontend/src/app/main/ui/comments.cljs
```diff
@@ -113,6 +113,7 @@
         content  (:content draft)
         pos-x    (* (:x position) zoom)
         pos-y    (* (:y position) zoom)
+        has-content? (and (not= content "") (some? content))
```
hirunatan: We have a `str/empty?` function for this.

You need to import it as
`[cuerdas.core :as str]`

PR #1580, opened 2022-02-12
Refactor transaction handling on storage related RPC methods

niwinz:
The refactor commits removes the transaction wrapping from two rpc calls that do storage operations. Storage operations (especially with S3 backend), can have unexpected latencies because they depend on network operations; holding a transaction on the probably slow operations causes that on large burst of requests to this endpoints the db connection pool exhaust, and possible cascade failure. The refactor implements it in a manner that no transaction at all is needed.

The main change consists on refactoring two RPC methods that can hold a long running transaction because of storage operations.

And additionally add support for:
- all AWS regions (previously only `eu-central-1` was supported).
- alternative S3 service providers (allowing setting up an endpoint URI)

Also adds MINIO to our devenv (S3 compatible opensource server).

### DIFF
--- a/frontend/src/app/main/data/workspace/libraries.cljs
+++ b/frontend/src/app/main/data/workspace/libraries.cljs
@@ -321,7 +321,7 @@
       ;; are small posibilities of race conditions with component
       ;; deletion.
       (when-let [component (get-in state [:workspace-data :components id])]
-        (let [[path name] (cp/parse-path-name new-name)
+        (let [[path name] (cph/parse-path-name new-name)
               objects     (get component :objects)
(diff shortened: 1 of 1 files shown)


backend/src/app/rpc/mutations/media.clj
```diff
@@ -77,6 +72,9 @@
                 :code :unable-to-access-to-url
                 :cause e))))
 
+;; TODO: we need to check the size before fetch resource, if not we
+;; can start downloading very big object and cause OOM errors.
+
```
hirunatan, line 77: This TODO could be dangerous? Perhaps someone uses this to make a DOS attack?

PR #1589, opened 2022-02-14
Redesign Team members

EvaMarco:
This PR solve this [task](https://tree.taiga.io/project/penpot/task/2324) of this [US](https://tree.taiga.io/project/penpot/us/2283)

Accidentally this PR solves this [issue](https://tree.taiga.io/project/penpot/us/2760)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@
 ### :boom: Breaking changes
 
 ### :sparkles: New features
+
+- Team members redesign [Taiga #2283](https://tree.taiga.io/project/penpot/us/2283)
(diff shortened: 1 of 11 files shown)


superalex:
- Logged as an owner if I try to change my own role to admin or editor there is an error shown. Also If I try to change my role to the same one (owner) there is a lightbox appearing that shouldn't be shown (as I'm not really doing any change)

EvaMarco:
> 

I've removed that option, it's not logical to remove the ownership like that. If you want to quit being an owner, set this role to someone and you will be admin.

superalex:
Thanks @EvaMarco it works great

Now I've detected one more situation. If I create a new team and then leave it (without adding any more members) I receive a "Unexpected validation error" error

EvaMarco:
Thanks, @superalex I could find the bug and fix it.

PR #1592, opened 2022-02-15
:sparkles: Team invitations

pabloalba:
Related to https://tree.taiga.io/project/penpot/us/2283

### DIFF
--- a/backend/src/app/migrations/sql/0067-add-team-invitation-table.sql
+++ b/backend/src/app/migrations/sql/0067-add-team-invitation-table.sql
@@ -1,0 +1,15 @@
+CREATE TABLE team_invitation (  
+  team_id uuid NOT NULL REFERENCES team(id) ON DELETE CASCADE,
+  email_to text NOT NULL,
+  role text NOT NULL,
+  valid_until timestamptz NOT NULL,
+  created_at timestamptz NOT NULL DEFAULT now(),
+  updated_at timestamptz NOT NULL DEFAULT now(),
+
+  PRIMARY KEY(team_id, email_to)
+);
+
+
+ALTER TABLE team_invitation
+  ALTER COLUMN email_to SET STORAGE external,
+  ALTER COLUMN role SET STORAGE external;
--- a/backend/src/app/rpc/mutations/teams.clj
+++ b/backend/src/app/rpc/mutations/teams.clj
@@ -23,6 +23,7 @@
    [app.util.services :as sv]
    [app.util.time :as dt]
    [clojure.spec.alpha :as s]
+   [cuerdas.core :as str]
    [datoteka.core :as fs]))
 
 ;; --- Helpers & Specs
@@ -359,12 +360,19 @@
               :role role))
       nil)))
 
+(def sql:upsert-team-invitation
+  "insert into team_invitation(team_id, email_to, role, valid_until)
+   values (?, ?, ?, ?)
+       on conflict(team_id, email_to) do
+          update set role = ?, valid_until = ?, updated_at = now();")
+
 (defn- create-team-invitation
   [{:keys [conn tokens team profile role email] :as cfg}]
   (let [member   (profile/retrieve-profile-data-by-email conn email)
+        token-exp (dt/in-future "48h")
         itoken   (tokens :generate
                          {:iss :team-invitation
-                          :exp (dt/in-future "48h")
+                          :exp token-exp
                           :profile-id (:id profile)
                           :role role
                           :team-id (:id team)
@@ -385,6 +393,10 @@
       (ex/raise :type :validation
                 :code :email-has-permanent-bounces
                 :hint "looks like the email you invite has been repeatedly reported as spam or permanent bounce"))
+
+
+    (db/exec-one! conn [sql:upsert-team-invitation
+                        (:id team) (str/lower email) (name role) token-exp (name role) token-exp])
 
     (eml/send! {::eml/conn conn
                 ::eml/factory eml/invite-to-team
@@ -418,3 +430,24 @@
                 :email email
                 :role role)))
       team)))
+
+
+;; --- Mutation: Update invitation role
+
+(s/def ::update-team-invitation-role
+  (s/keys :req-un [::profile-id ::team-id ::email ::role]))
+
+(sv/defmethod ::update-team-invitation-role
+  [{:keys [pool] :as cfg} {:keys [profile-id team-id email role] :as params}]
+  (db/with-atomic [conn pool]
+    (let [perms    (teams/get-permissions conn profile-id team-id)
+          team     (db/get-by-id conn :team team-id)]
+
+      (when-not (:is-admin perms)
+        (ex/raise :type :validation
+                  :code :insufficient-permissions))
+
+      (db/update! conn :team-invitation
+                  {:role (name role) :updated-at (dt/now)}
+                  {:team-id (:id team) :email-to (str/lower email)})
+      nil)))
--- a/backend/src/app/rpc/mutations/verify_token.clj
+++ b/backend/src/app/rpc/mutations/verify_token.clj
@@ -13,8 +13,9 @@
    [app.metrics :as mtx]
    [app.rpc.mutations.teams :as teams]
    [app.rpc.queries.profile :as profile]
-   [app.util.services :as sv]
-   [clojure.spec.alpha :as s]))
+   [app.util.services :as sv]   
+   [clojure.spec.alpha :as s]
+   [cuerdas.core :as str]))
 
 (defmulti process-token (fn [_ _ claims] (:iss claims)))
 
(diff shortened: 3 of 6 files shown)


backend/src/app/migrations/sql/0067-add-team-invitation-table.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE team_invitation (  
+  team_id uuid NOT NULL REFERENCES team(id) ON DELETE CASCADE,
+  email_to text NOT NULL,
+  role varchar(255) NOT NULL,
```
niwinz: don't use varchar, should be text, also remember to mark text fields as external.

example from audit_log table:
```sql
ALTER TABLE audit_log
  ALTER COLUMN name SET STORAGE external,
  ALTER COLUMN type SET STORAGE external,
  ALTER COLUMN props SET STORAGE external;
  ```

File: backend/src/app/rpc/mutations/teams.clj [hunk 2]
```diff
@@ -359,12 +360,19 @@
               :role role))
       nil)))
 
+(def sql:upsert-team-invitation
+  "insert into team_invitation(team_id, email_to, role, valid_until)
+   values (?, ?, ?, ?)
+       on conflict(team_id, email_to) do
+          update set role = ?, valid_until = ?, updated_at = clock_timestamp();")
```
niwinz: in this case, is better to use `now()` I think.

The difference between `now()` and `clock_timestamp()` is that first is stable across transaction, and the second one always return the latest timestamp.

backend/src/app/migrations/sql/0067-add-team-invitation-table.sql
```diff
@@ -0,0 +1,12 @@
...
+  valid_until timestamptz NOT NULL,
+  created_at timestamptz NOT NULL DEFAULT clock_timestamp(),
+  updated_at timestamptz NOT NULL DEFAULT clock_timestamp(),
```
niwinz: I think is better use `now()` here. I know that we have used the `clock_timestamp()` previosly in other sites and it is inconsistent across the codebase, but for now, the correct approach is use the `now()` unless we really need the `clock_timestamp()` behavior.

backend/src/app/rpc/mutations/teams.clj
```diff
@@ -359,12 +360,19 @@
...
+
 (defn- create-team-invitation
   [{:keys [conn tokens team profile role email] :as cfg}]
   (let [member   (profile/retrieve-profile-data-by-email conn email)
+        token_exp (dt/in-future "48h")
```
niwinz: Prefer use `-` as separator instead of `_`

backend/src/app/rpc/mutations/teams.clj
```diff
@@ -418,3 +430,24 @@
...
+    (let [perms    (teams/get-permissions conn profile-id team-id)
+          team     (db/get-by-id conn :team team-id)]
+
+      (when-not (:is-admin perms)
+        (ex/raise :type :validation
+                  :code :insufficient-permissions))
+
+      (db/update! conn :team_invitation
+                  {:role (name role) :updated_at (dt/now)}
+                  {:team_id (:id team) :email_to (str/lower email)})
```
niwinz: The same here: use `-` instead of `_`, they will be translated to `_` automatically for the database

backend/src/app/rpc/mutations/verify_token.clj
```diff
@@ -100,11 +101,21 @@
           :opt-un [:internal.tokens.team-invitation/member-id]))
 
 (defn- accept-invitation
-  [{:keys [conn] :as cfg} {:keys [member-id team-id role] :as claims}]
-  (let [params (merge {:team-id team-id
+  [{:keys [conn] :as cfg} {:keys [member-id team-id role member-email] :as claims}]
+  (let [
+        member (profile/retrieve-profile conn member-id)
+        invitation (db/exec-one!
```
niwinz: Prefer use here the `db/get-by-param`
Approx example:

```clojure
(db/get-by-param conn :team-invitation 
                 {:team-id ... :email-to ...}
                 {:check-not-found false})
;; or

(-> (db/query conn :team-invitation {:team-id ... :email-to ...}) first)

```

backend/src/app/rpc/mutations/verify_token.clj
```diff
@@ -100,11 +101,21 @@
...
+                    conn
+                    ["select role from team_invitation where team_id = ? and email_to = ?"
+                     team-id (str/lower member-email)])
+        ;; Update the role if there is an invitation
+        role (if (nil? (:role invitation)) 
+               role 
+               (keyword (:role invitation)))
```
niwinz: This can be expressed as `(some-> invitation :role keyword)`

backend/src/app/rpc/mutations/verify_token.clj
```diff
@@ -115,7 +126,13 @@
...
                   {:is-active true}
                   {:id member-id}))
-    (assoc member :is-active true)))
+    (assoc member :is-active true)
+    
+    ;; Delete the invitation
+    (db/exec-one!
+     conn
+     ["delete from team_invitation where team_id = ? and email_to = ?"
+      team-id (str/lower member-email)])))
```
niwinz: Prefer use the helper here:

```clojure
(db/delete! conn :team-invitation
            {:team-id ... :email-to ...})
```

backend/src/app/rpc/queries/teams.clj
```diff
@@ -229,3 +230,20 @@
...
+
+(def sql:team-invitations
+  "select email_to as email, role, (valid_until < ?) as expired from team_invitation where team_id = ?")
+
+
+(sv/defmethod ::team-invitations
+  [{:keys [pool] :as cfg} {:keys [profile-id team-id]}]
+  (with-open [conn (db/open pool)]
+    (check-read-permissions! conn profile-id team-id)
+    (db/exec! conn [sql:team-invitations (dt/now) team-id])))
```
niwinz: I think is better use the `now()` directly in SQL, so you don't need to pass the `(dt/now)` as parameter. But maybe I'm wrong.

backend/test/app/services_teams_test.clj
```diff
@@ -159,4 +166,60 @@
...
+              :team-id (:id team)}]
+
+    ;;insert an entry on the database with an enabled invitation
+    (db/exec-one! th/*pool*
+                  ["insert into team_invitation(team_id, email_to, role, valid_until) values (?, ?, ?, ?)"
+                   (:team-id data) "test1@mail.com" "editor", (dt/in-future "48h")])
+
+    ;;insert an entry on the database with an expired invitation
+    (db/exec-one! th/*pool*
+                  ["insert into team_invitation(team_id, email_to, role, valid_until) values (?, ?, ?, ?)"
```
niwinz: This simple instert can be done with `db/insert!`

```clojure
(db/insert th/*pool* :team-invitation
           {:team-id ...
            :email-to ...
            ...})
```

backend/src/app/rpc/queries/teams.clj
pabloalba: I think it is more consistent to always calculate the dates on the same place (database or back code). The token expiration date is set on back as (dt/in-future "48h"). I think calculating the date on the same place can avoid time discrepanciesand/or timezones issues...

backend/src/app/migrations/sql/0067-add-team-invitation-table.sql
pabloalba: fixed

backend/src/app/rpc/mutations/teams.clj
pabloalba: fixed

backend/src/app/migrations/sql/0067-add-team-invitation-table.sql
pabloalba: fixed

backend/src/app/rpc/mutations/teams.clj
pabloalba: fixed

backend/src/app/rpc/mutations/verify_token.clj
pabloalba: fixed

backend/src/app/rpc/mutations/verify_token.clj
pabloalba: fixed

backend/test/app/services_teams_test.clj
pabloalba: fixed

backend/src/app/rpc/mutations/teams.clj
pabloalba: fixed

backend/src/app/rpc/mutations/verify_token.clj
pabloalba: fixed

backend/src/app/rpc/queries/teams.clj
pabloalba: Finally changed to now()

PR #1606, opened 2022-02-18
✨ Ability to add multiple strokes to a shape

superalex:
Related to https://tree.taiga.io/project/penpot/us/2778

### DIFF
--- a/frontend/src/app/main/ui/shapes/custom_stroke.cljs
+++ b/frontend/src/app/main/ui/shapes/custom_stroke.cljs
@@ -323,7 +323,6 @@
                          (> (count (:fills shape)) 1)
                          (some :fill-color-gradient (:fills shape)))
         one-fill?    (= (count (:fills shape)) 1)
-        no-fills?    (= (count (:fills shape)) 0)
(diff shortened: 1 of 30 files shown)


frontend/src/app/main/ui/shapes/custom_stroke.cljs
```diff
@@ -286,5 +311,56 @@
        child]
 
       :else
-      child)))
-
+      [:*
```
Alotor: You can remove `[:*` here

frontend/src/app/main/ui/shapes/custom_stroke.cljs
```diff
@@ -286,5 +311,56 @@
...
+                   (obj/get (attrs/extract-stroke-attrs value position) "style")))]
+    props))
+
+(mf/defc shape-custom-strokes
+  {::mf/wrap-props false}
+  [props]
+  (let [child (obj/get props "children")
+        shape (obj/get props "shape")
+        elem-name (obj/get child "type")]
+    [:*
```
Alotor: This `[:*` is only necessary in the `for` branch. I think it's better to move it together with the nested for.

PR #1616, opened 2022-02-21
:sparkles: Persist color palette and color picker across refresh

### LINKED ISSUE #910: Color toolbar state resets / is not remembered
Every time you refresh or load up Penpot newly, the color bar on the bottom is hidden and also reset in the set of colors it shows.

This is how I’d like to have it, with our component colors in there:
Color component bar resets after reload

What I always need to do is:
1. Show the color bar again
2. Go to the menu
3. Pick "Nextcloud core components" again
need to change to pick core components colors

It would be great if the state of the color bar would just be remembered. Shown/hidden, as well as which color set was selected. :)

**Additional context**
On a related note, the same issue happens with the color picker in the sidebar – it always resets to "Recent colors", which results in non-linked colors being used:
Color picker always resets to recent colors instead of staying on shared library or last selection

pabloalba:
Related to https://tree.taiga.io/project/penpot/issue/1660

### DIFF
--- a/frontend/src/app/main/data/workspace.cljs
+++ b/frontend/src/app/main/data/workspace.cljs
@@ -52,6 +52,7 @@
    [app.util.http :as http]
    [app.util.i18n :as i18n]
    [app.util.router :as rt]
+   [app.util.storage :refer [storage]]
    [app.util.timers :as tm]
    [app.util.webapi :as wapi]
    [beicon.core :as rx]
@@ -245,6 +246,8 @@
             (rx/observe-on :async))))))
 
 (declare go-to-page)
+(declare load-flag)
+
 (defn initialize-page
   [page-id]
   (us/assert ::us/uuid page-id)
@@ -268,6 +271,12 @@
               (assoc :current-page-id id)
               (assoc :trimmed-page (dm/select-keys page [:id :name]))
               (assoc :workspace-local local)
+              (update :workspace-layout
+                      #(if (load-flag :colorpalette false)
+                         (conj % :colorpalette)
+                         (disj % :colorpalette)))
+              (assoc-in [:workspace-local :selected-palette] (load-flag :selected-palette :recent))
+              (assoc-in [:workspace-local :selected-palette-colorpicker] (load-flag :selected-palette-colorpicker :recent))
               (update :workspace-global assoc :background-color (-> page :options :background))
               (update-in [:route :params :query] assoc :page-id (dm/str id))))
         state))))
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/data/workspace.cljs
```diff
@@ -2005,6 +2014,20 @@
              (dwu/commit-undo-transaction))))))))
 
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Colorpalette
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn save-colorpalete-status
```
hirunatan: Don't like very much to have a separated function for this. I think it would be better to modify functions like toggle-layout-flags and remove-layout-flags functions, to check the flags, and store them in local storage when needed.

This would allow to easily extend the behavior to other layout flags. For example, to text palettes, that are not mentioned in the issue, but are very likely to be also needed.

frontend/src/app/main/data/workspace.cljs
```diff
@@ -2005,6 +2014,20 @@
...
+  [status]
+  (ptk/reify ::save-colorpalete-status
```
hirunatan: Note: is "palette", not "palete".

frontend/src/app/main/data/workspace/colors.cljs
```diff
@@ -327,3 +334,27 @@
     (update [_ state]
       (-> state
           (assoc-in [:workspace-local :editing-stop] spot)))))
+
+(defn set-current-colorpalette-show!
```
hirunatan: This is only a suggestion, but I think it'll be good.

We normally use "get" and "set" for events that access the global state. You could use here "load" and "save" or "store" to make clearer that you are using the local storage and not the in memory state.

frontend/src/app/main/data/workspace.cljs
```diff
@@ -412,6 +421,28 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Local storage Flags Manipulation
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(def storeable-layout-flags #{:colorpalette})
+(def storeable-workspace-local-flags #{:selected-palette :selected-palette-colorpicker})
+
+(defn store-layout-flags!
+  [state flags]
+  (doseq [item (filter #(contains? storeable-layout-flags %) flags)]
```
hirunatan: There is a clojure idiom that we use a lot: when you call a set as if it was a function, it returns the equivalent to using `contains?` on it. So you can replace this line with this one:
```clojure
(doseq [item (filter storeable-layout-flags flags)]
```

pabloalba: Oooh, cool

pabloalba: done!

PR #1623, opened 2022-02-21
Render Text as native SVG elements

Alotor:
https://tree.taiga.io/project/penpot/us/2759?milestone=309423


common/src/app/common/spec/shape.cljc
```diff
@@ -202,12 +202,52 @@
...
+
+(s/def :internal.shape.text.position-data/fill-color ::fill-color)
+(s/def :internal.shape.text.position-data/fill-opacity ::fill-opacity)
+(s/def :internal.shape.text.position-data/fill-color-gradient ::fill-color-gradient)
+
+(s/def :internal.shape.text.position-data/font-family string?)
+(s/def :internal.shape.text.position-data/font-size string?)
+(s/def :internal.shape.text.position-data/font-style string?)
+(s/def :internal.shape.text.position-data/font-weight string?)
+(s/def :internal.shape.text.position-data/rtl? boolean?)
```
niwinz: I think we have decided not use the `?` ending on props on the file data structures. (I know that we already have some of them hehe)

frontend/src/app/main/data/workspace/texts.cljs
```diff
@@ -208,12 +215,57 @@
...
               shape-ids (cond
                           (cph/text-shape? shape)  [id]
                           (cph/group-shape? shape) (cph/get-children-ids objects id))]
-          (rx/of (dch/update-shapes shape-ids update-fn)))))))
+          (rx/of (dch/update-shapes shape-ids #(update-text-content % update-node? attrs/merge attrs))))))))
+
+(defn migrate-content
+  [content]
+  (txt/transform-nodes
+   #(or (txt/is-text-node? %) (txt/is-paragraph-node? %))
```
niwinz: **minor**: use `some-fn` here. 

I mean: 
```clojure
(txt/transform-nodes
 (some-fn txt/is-text-node? txt/is-paragraph-node?)
 ...)
```

frontend/src/app/main/ui/hooks/mutable_observer.cljs
```diff
@@ -0,0 +1,58 @@
...
+             (when (some? node)
+               (let [options #js {:attributes true
+                                  :childList true
+                                  :subtree true
+                                  :characterData true}
+                     mutation-obs (js/MutationObserver. on-mutation)]
+                 (mf/set-ref-val! prev-obs-ref mutation-obs)
+                 (.observe mutation-obs node options))))))]
+
+    (mf/use-effect
```
niwinz: also **minor**: use `mf/with-effect`

example:
```clojure
(mf/with-effect
  (fn []
    (when-let [^js prev-obs (mf/ref-val prev-obs-ref)]
      (.disconnect prev-obs)
      (mf/set-ref-val! prev-obs-ref nil)))))
```
will reduce one level of indentation, because with-effect body is already part of an implicit function.

frontend/src/app/main/ui/render.cljs
```diff
@@ -92,7 +95,9 @@
         text-shapes
         (->> objects
              (filter (fn [[_ shape]] (= :text (:type shape))))
-             (mapv second))]
+             (mapv second))
```
niwinz: no vector need here so `map` is better here. also maybe use `(sequence <transducer> objects)` instead

frontend/src/app/main/ui/shapes/attrs.cljs
```diff
@@ -80,34 +80,38 @@
...
+         (cond
+           (contains? shape :fill-image)
+           (let [fill-image-id (str "fill-image-" render-id)]
+             {:fill (str/format "url(#%s)" fill-image-id)})
 
-                     (contains? shape :fill-color)
-                     {:fill (:fill-color shape)}
+           (contains? shape :fill-color-gradient)
+           (let [fill-color-gradient-id (str "fill-color-gradient_" render-id (if index (str "_" index) ""))]
+             {:fill (str/format "url(#%s)" fill-color-gradient-id)})
```
niwinz: the str/format is not used here because it has performance implications, we should still use plain `str` here I think.

frontend/src/app/main/ui/shapes/attrs.cljs
```diff
@@ -80,34 +80,38 @@
...
-                     (assoc :fillOpacity (:fill-opacity shape)))]
+           :else
+           {})
 
-    (obj/merge! attrs (clj->js fill-attrs))))
+         fill-attrs (cond-> fill-attrs
+                      (contains? shape :fill-opacity)
+                      (assoc :fillOpacity (:fill-opacity shape)))]
+
+     (obj/merge! attrs (clj->js fill-attrs)))))
```
niwinz, line 114: As a note for the future, I think we can build the js object directly here instead of build clojure object and then translate it to JS with clj->js, for the same performance reasons. This can be done a separated task/pr

frontend/src/app/main/ui/shapes/fills.cljs
```diff
@@ -20,45 +20,54 @@
...
+                            (obj/set! "patternTransform" transform))]
 
-         (when has-image
-           [:image {:xlinkHref (get embed uri uri)
-                    :width width
-                    :height height}])]]]))
+        [:*
+         (for [[_shape-index shape] (d/enumerate (or (:position-data shape) [shape]))]
+           (for [[fill-index value] (-> (d/enumerate (:fills shape [])) reverse)]
+             (cond (some? (:fill-color-gradient value))
```
niwinz: I think this cond with a single condition can be  just a `when`.

frontend/src/app/main/ui/shapes/fills.cljs
```diff
@@ -20,45 +20,54 @@
...
+                   (case (d/name (:type (:fill-color-gradient value)))
+                     "linear" [:> grad/linear-gradient #js {:id (str "fill-color-gradient_" render-id "_" fill-index)
+                                                            :gradient (:fill-color-gradient value)
+                                                            :shape shape}]
+                     "radial" [:> grad/radial-gradient #js {:id (str "fill-color-gradient_" render-id "_" fill-index)
```
niwinz: I'm pretty sure that in this case, where the data is known statically, the `#js` prefix is not necessary, but check it before just trusting me xD

frontend/src/app/main/ui/shapes/text/svg_text.cljs
```diff
@@ -0,0 +1,67 @@
...
+     (when (d/seek :fill-color-gradient position-data)
+       [:defs
+        (for [[index data] (d/enumerate position-data)]
+          (when (some? (:fill-color-gradient data))
+            [:& grad/gradient {:id (str "fill-color-gradient_" (get-gradient-id index))
+                               :attr :fill-color-gradient
+                               :shape data}]))])
+
+     [:> :g group-props
+      (for [[index data] (d/enumerate position-data)]
```
niwinz, line 50: I think the same enumerate can be shared in both loops.

frontend/src/app/main/ui/workspace/shapes/text.cljs
```diff
@@ -99,24 +113,114 @@
...
         edition-ref (mf/use-memo (mf/deps id) #(l/derived (fn [o] (= id (:edition o))) refs/workspace-local))
-        edition?    (mf/deref edition-ref)]
+        edition?    (mf/deref edition-ref)
+
+        local-position-data (mf/use-state nil)
+
+        sid-ref (mf/use-ref nil)
+
+        handle-change-foreign-object
+        (fn [node]
```
niwinz: this probably need to be wrapped on use-callback (or `use-fn` alias of the same).

frontend/src/app/util/text_svg_position.cljs
```diff
@@ -0,0 +1,123 @@
...
+          (conj result entry))
+        
+        (let [rects (get-range-rects text-node from-i (inc to-i))]
+          ;; If the rects increase means we're in a new paragraph
+          (if (> (.-length rects) 1)
+            (let [entry {:node parent-node
+                         :position (dom/bounding-rect->rect (if rtl? (second rects) (first rects)))
+                         :text current}]
+              (recur to-i to-i "" (conj result entry)))
+            (recur from-i (inc to-i) (str current (nth content to-i)) result)))))))
```
niwinz, line 52: Just a note for the future, I think this functions are good candidate to be written directly in JS, without involving immutable data structures until the end of process. What do you think?

common/src/app/common/spec/shape.cljc
Alotor: OK. I'll change it

frontend/src/app/main/data/workspace/texts.cljs
Alotor: Oh, I didn't know this function. Thanks! I'll change it

frontend/src/app/main/ui/hooks/mutable_observer.cljs
Alotor: Ok

frontend/src/app/main/ui/render.cljs
Alotor: ok

frontend/src/app/main/ui/shapes/attrs.cljs
Alotor: ok

frontend/src/app/main/ui/shapes/fills.cljs
Alotor: ok

frontend/src/app/main/ui/shapes/fills.cljs
Alotor: I think this was supposed to be `:>` so the props can be reused and not repeat code. I'll change so both component use the same props

frontend/src/app/main/ui/shapes/text/svg_text.cljs
Alotor: Yes, I think you're right

frontend/src/app/main/ui/workspace/shapes/text.cljs
Alotor: ok

frontend/src/app/util/text_svg_position.cljs
Alotor: I think it's really good idea. I'll add a comment and add the task to do it afterwards

PR #1638, opened 2022-02-25
Workspace refactor & other performance improvements

niwinz:


frontend/src/app/main/ui/workspace.cljs
```diff
@@ -70,6 +71,7 @@
 
        [:& viewport {:file file
                      :local local
+                     :wstate wstate
```
Alotor, line 74: It has the old name `wstate` I think better to call it `global` to be more coherent with its name in the state

Alotor: We will do it in a different MR

PR #1647, opened 2022-02-28
:recycle: Refactor backend: make backend almost async friendly (part 2)

niwinz:
This is the second part of the refactor that fixes many scalability issues related to the synchronous mode that was used from the beginning of the project development. The first part just fixes the big issues related to long queries and connection pool exhaustion, the urgent one. This one takes the next level changes:

1. Refactors the storage interface and storage backends, making them have asynchronous API. This benefits backend like S3 because they already offered async API. This makes using considerably less thread resources for the same amount of work thanks to the async + underlying netty impl that uses AWS SDK.
2. Refactor how storage object references are handling (remove the pending-recheck concept completely and expand the touched-gc concept). The touched-gc task initially has been introduced only for handle the file-media-objects and storage-objects reference counting. Now it is generalized and extended to all objects. This change makes more simpler RPC methods that previously needed a big long transaction around heavy weight methods that manages files, now they are transaction less, delegating to the touched-gc task handle possible aborted transactions and leaked resources. It some kind of soft reference counting system.
3. Add storage deduplication. Now storage has the concept of bucket and concept of deduplication. This helps reduce the unnecesary space usage on repeated uploads of same shared files. This is done calculating the hash (blake2b algorithm) of the file in the bucket. The deduplication happens only per bucket & backend.
4. Adapt all RPC methods that worked with files and/or susceptible to be improved with asynchronous flow.
5. Refactor on how executors are used from the PART 1: 
6. ... make all the heavy operations (like font conversion, image thumbnails, etc..) use a single, constrained executor called BLOCKING, its size it is adjusted to be a smaller one for not exhaust the resources of the server; 
7. ... use the DEFAULT executor for all the RPC methods, it allows have more threads.
8. ... for the concrete case of file-update operations we use an async queue/semaphore for avoid pool exhaustion (becuase of nature of this operations: potentially long, transaction block that allocates a connection for all the transaction duration).

Missing, that will be coming in the PART 3 PR:
1. We still have many HTTP stuff, processing in a blocking way, in JETTY threads. It is ok for now but it is not ideal.
2. Additionally, JETTY has the problem to upgrade websocket connection from asynchronous flow, so right now we are unable to move BLOCKING stuff from JETTY threads. I have spoken with jetty developers, and it does not has easy solution for our use case (we still need to know the user profile for know if we can upgrade websocket connection or not).

I have some time evaluated Undertow as a jetty replacement, and this looks the proper moment to do it. As we using the Ring interface for all our operations, this only consists to rewrite yetti adapter to use undertow instead of JETTY. This has many benefits for us. Because undertow programming model is more natural on how we are using it on penpot and with Ring. Also it allows us leave undertow the HTTP stuff, and move to out own execution model all our logic. And finally it will use less resources for the same work because we will no need to spawn too many threads on http part and only control our own executors (we will reduce it from 200 -> CPU*2 that can be 4,6,or 8 depending on instances).

Also can facilitate probably run multiple servers in the same instance that will make it easy separate the websocket part from HTTP part (our plans for improve current websockets solution).


backend/src/app/db.clj
```diff
@@ -247,7 +247,7 @@
   ([ds table params where opts]
    (exec-one! ds
               (sql/update table params where opts)
-              (assoc opts :return-keys true))))
+              (merge {:return-keys true} opts))))
```
Alotor, line 250: This is something we have to do from time to time. Maybe we can create a function in common.data like `with-defaults` that adds some default values when there is no value?

```
(with-defaults opts {:return-keys true})
```

Just an idea.

backend/src/app/http/assets.clj
```diff
@@ -35,71 +35,81 @@
...
+  (px/with-dispatch executor
+    (let [id   (coerce-id id)
+          mobj (db/exec-one! pool ["select * from file_media_object where id=?" id])]
+      (when-not mobj
+        (ex/raise :type :not-found
+                  :hint "object does not found"))
+      mobj)))
 
 (defn- serve-object
+  "Helper function that returns the appropriate responde depending on
```
Alotor: Typo in the comment :) "reponde"

backend/src/app/rpc/mutations/files.clj
```diff
@@ -353,7 +354,7 @@
 
     ;; We need to delete the data from external storage backend
     (when-not (nil? (:data-backend file))
-      (delete-from-storage cfg file))
+      @(delete-from-storage cfg file))
```
Alotor, line 357: I think it's more readable to use `deref` or storing the value in a let... at least for me having `@(..)` it's quite ugly

backend/src/app/rpc/mutations/profile.clj
```diff
@@ -410,32 +412,33 @@
...
-    (media/validate-media-type (:content-type file) #{"image/jpeg" "image/png" "image/webp"})
-    (media/run {:cmd :info :input {:path (:tempfile file)
-                                   :mtype (:content-type file)}})
+  [{:keys [pool storage executors] :as cfg} {:keys [profile-id file] :as params}]
 
+  ;; Validate incoming mime type
+  (media/validate-media-type! (:content-type file) #{"image/jpeg" "image/png" "image/webp"})
+
+  ;; Perform file validation
+  @(px/with-dispatch (:blocking executors)
```
Alotor: why is the deref needed here? you don't use the value. Will block until it's resolved? Maybe a comment here it's good. Also I think it's better to extract the result into a let and then use the `@`

backend/src/app/rpc/mutations/profile.clj
```diff
@@ -410,32 +412,33 @@
...
-          storage (media/configure-assets-storage storage conn)
-          cfg     (assoc cfg :storage storage)
-          photo   (teams/upload-photo cfg params)]
+          cfg     (update cfg :storage media/configure-assets-storage conn)
+          photo   @(teams/upload-photo cfg params)]
 
       ;; Schedule deletion of old photo
       (when-let [id (:photo-id profile)]
-        (sto/del-object storage id))
+        @(sto/touch-object! storage id))
```
Alotor: Same as above

backend/src/app/rpc/mutations/teams.clj
```diff
@@ -289,21 +290,25 @@
...
 (sv/defmethod ::update-team-photo
   {::rlimit/permits (cf/get :rlimit-image)}
-  [{:keys [pool storage] :as cfg} {:keys [profile-id file team-id] :as params}]
+  [{:keys [pool storage executors] :as cfg} {:keys [profile-id file team-id] :as params}]
+
+  ;; Validate incoming mime type
+  (media/validate-media-type! (:content-type file) #{"image/jpeg" "image/png" "image/webp"})
+
+  ;; Perform file validation
+  @(px/with-dispatch (:blocking executors)
```
Alotor: Same as above

backend/src/app/storage.clj
```diff
@@ -341,6 +347,9 @@
           (has-file-media-object-nrefs? [conn id]
             (-> (db/exec-one! conn [sql:retrieve-file-media-object-nrefs id id]) :nrefs pos?))
 
+          (has-profile-nrefs? [conn id]
+            (-> (db/exec-one! conn [sql:retrieve-profile-nrefs id id]) :nrefs pos?))
```
Alotor, line 351: why the `->` if there is only one statement?

backend/src/app/db.clj
niwinz: Agree, will think on it for the next PR

backend/src/app/rpc/mutations/files.clj
niwinz: Here we just need to "join" or block until this finishes, the unique alternative is `(deref (delete-from-storage cfg file))`
But i think is the same...

backend/src/app/storage.clj
niwinz: hmm there are 3 -> `:nrefs` and `pos?`

backend/src/app/db.clj
niwinz: Agree, I will think on it for the next PR.

PR #1658, opened 2022-03-04
:bug: Fix multiple edition

hirunatan:
Now multi edition of rotation and border radius works ok. Even if the selection contains mixed shapes that have or not these features, or if there are old artboards that do not have :rx attribute.

### DIFF
--- a/common/src/app/common/spec/radius.cljc
+++ b/common/src/app/common/spec/radius.cljc
@@ -6,8 +6,9 @@
 
 (ns app.common.spec.radius
   (:require
-   [app.common.spec :as us]
-   [clojure.spec.alpha :as s]))
+    [app.common.pages.common :refer [editable-attrs]]
+    [app.common.spec :as us]
+    [clojure.spec.alpha :as s]))
 
 (s/def ::rx ::us/safe-number)
 (s/def ::ry ::us/safe-number)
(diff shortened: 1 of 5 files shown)


common/src/app/common/spec/radius.cljc
```diff
@@ -30,6 +31,10 @@
 ;; shapes that has border radius, and so it hasn't :rx nor :r1. 
 ;; In this case operations must leave shape untouched.
 
+(defn has-radius?
+  [shape]
+  ((get editable-attrs (:type shape)) :rx))
```
niwinz: I think is better to use `contains?` here.

frontend/src/app/main/ui/workspace/sidebar/options/shapes/multiple.cljs
```diff
@@ -157,36 +157,47 @@
...
+          (let [read-mode      (get-in type->read-mode [type attr-group])
+                editable-attrs (filter (get cpc/editable-attrs (:type shape)) attrs)]
+            (case read-mode
               :ignore   [ids values]
-              :shape    (let [editable-attrs (filter (get cpc/editable-attrs (:type shape)) attrs)]
+
+              :shape    (let [;; Get the editable attrs from the shape, ensuring that all attributes
+                              ;; are present, with value nil if they are not present in the shape.
+                              shape-values (merge
+                                             (into {} (map #(hash-map % nil) editable-attrs))
```
niwinz: its better to use `(into {} (map #(vector % nil)) editable-attrs)`

common/src/app/common/spec/radius.cljc
hirunatan: done

frontend/src/app/main/ui/workspace/sidebar/options/shapes/multiple.cljs
hirunatan: done

PR #1672, opened 2022-03-09
:sparkles: Set an artboard as the file thumbnail

pabloalba:
Related with https://tree.taiga.io/project/penpot/us/1526

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ### :boom: Breaking changes
 ### :sparkles: New features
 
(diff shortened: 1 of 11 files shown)


niwinz:
Everything is ok, i just thinking that the method name should be more concrete. I think that `::file-data-for-thumbnail` can be a better alternative.

pabloalba:
> Everything is ok, i just thinking that the method name should be more concrete. I think that `::file-data-for-thumbnail` can be a better alternative.

Done!

PR #1673, opened 2022-03-09
Increase nginx timeout for exporting large artboards to pdf

ae-ae:
It can take more than 60 seconds to export large artboards (nginx default timeout is 60 seconds).


niwinz:
It is ok, for this change, but we are working in a proper solution for this that will not need to hold active http connection for it. So this wil be solved in the next release.

ae-ae:
Great news!

PR #1700, opened 2022-03-21
:zap: Minor performance improvement on file thumbnail handling

niwinz:


backend/src/app/http/middleware.clj
```diff
@@ -45,23 +46,17 @@
...
-              (yrs/response :status 400
-                            :headers {"content-type" "application/transit+json"}
-                            :body (t/encode-str data {:type :json-verbose}))))]
+                request)))]
 
     (fn [request respond raise]
       (try
         (let [request (process-request request)]
           (handler request respond raise))
         (catch Exception cause
```
Alotor: This catch should be surrounding `process-request` otherwise we can have "validation" errors for unexpected errors

PR #1701, opened 2022-03-21
Library changes builder

hirunatan:
NOTE: there are more refactors possible in `libraries_helpers.cljs`, but the refactor done was already too complex.

- Helpers like remove-shape, change-touched, etc. could be moved to changes-builder, or use functions defined there.
- They still use some reg-objects calls, that could be replaced by resize-parents.

But all those changes depend on the changes tree metadata working in multi container and multi file, a thing that requires further investigation.

### DIFF
--- a/common/src/app/common/pages/changes_builder.cljc
+++ b/common/src/app/common/pages/changes_builder.cljc
@@ -30,17 +30,29 @@
   [changes save-undo?]
   (assoc changes :save-undo? save-undo?))
 
-(defn with-page [changes page]
+(defn with-page
+  [changes page]
   (vary-meta changes assoc
              ::page page
-             ::page-id (:id page)
-             ::objects (:objects page)))
+             ::page-id (:id page)))
 
-(defn with-objects [changes objects]
+(defn with-container
+  [changes container]
+  (if (cph/page? container)
+    (vary-meta changes assoc ::page-id (:id container))
+    (vary-meta changes assoc ::component-id (:id container))))
+
+(defn with-objects
+  [changes objects]
   (let [file-data (-> (cp/make-file-data (uuid/next) uuid/zero)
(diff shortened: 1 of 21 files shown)


common/src/app/common/pages/changes_builder.cljc
```diff
@@ -224,10 +253,13 @@
   ([changes ids update-fn]
    (update-shapes changes ids update-fn nil))
 
-  ([changes ids update-fn {:keys [attrs ignore-geometry?] :or {attrs nil ignore-geometry? false}}]
-   (assert-page-id changes)
+  ([changes ids update-fn {:keys [attrs ignore-geometry? ignore-touched]
+                           :or {attrs nil ignore-geometry? false ignore-touched false}}]
```
niwinz: yo don't need `nil` as default value, because it is already will be `nil`.

niwinz: the same for the rest, of falses

hirunatan: Done!

It was for documentation. But you are true, actually i'ts not needed.

frontend/test/app/components_basic_test.cljs
```diff
@@ -84,127 +88,157 @@
...
-     (->> state
-          (the/do-update (dw/select-shapes (lks/set
-                                            (thp/id :shape1)
-                                            (thp/id :shape2))))
-          (the/do-watch-update dwl/add-component)
-          (rx/do
-            (fn [new-state]
+                                     {:name "Rect-2"}))
+         store (the/prepare-store state done
+           (fn [new-state]
```
niwinz: This indentation looks wrong

hirunatan: Solved. Compacted the commits to not have a very ugly diff.

PR #1702, opened 2022-03-21
:paperclip: Add more events instrumentation

niwinz:


frontend/src/app/main/ui/auth/register.cljs
```diff
@@ -160,13 +161,6 @@
 (defn- handle-register-error
   [form error]
   (case (:code error)
-    :registration-disabled
-    (st/emit! (dm/error (tr "errors.registration-disabled")))
-
-    :email-has-permanent-bounces
-    (let [email (get @form [:data :email])]
-      (st/emit! (dm/error (tr "errors.email-has-permanent-bounces" email))))
-
```
hirunatan, line 169: Why are those removed?

niwinz: because they are not handled here, so this is technically dead code.

Issue #1532, opened 2022-02-02
Import SVG with text


der-On:
When I import / drag & drop a SVG onto the canvas that contains text, the text is not beeing imported.

I would like text in SVGs to also be imported.

hirunatan:
Mmm.. this seems a bug on penpot. I have created an issue in Taiga, we'll try to fix it as soon as possible. Thanks for reporting.

https://tree.taiga.io/project/penpot/issue/2720

mlgarchery:
I have the same problem. Here is a svg file from Inkscape. Importing it result in weirdness (no text + not the same shapes).
Same thing when exporting a project, the resulting SVG misses the text.
This issue should probably be tagged as Bug.

```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!-- Created with Inkscape (http://www.inkscape.org/) -->

<svg
   width="51.807117mm"
   height="16.447977mm"
   viewBox="0 0 51.807117 16.447977"
   version="1.1"
   id="svg5"
   sodipodi:docname="logo.svg"
   inkscape:version="1.1.2 (b8e25be833, 2022-02-05)"
   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
   xmlns="http://www.w3.org/2000/svg"
   xmlns:svg="http://www.w3.org/2000/svg">
  <sodipodi:namedview
     id="namedview7"
     pagecolor="#ffffff"
     bordercolor="#666666"
     borderopacity="1.0"
     inkscape:pageshadow="2"
     inkscape:pageopacity="0.0"
     inkscape:pagecheckerboard="0"
     inkscape:document-units="mm"
     showgrid="false"
     inkscape:zoom="2.5620932"
     inkscape:cx="78.256327"
     inkscape:cy="19.124988"
     inkscape:window-width="1920"
     inkscape:window-height="1043"
     inkscape:window-x="1920"
     inkscape:window-y="0"
     inkscape:window-maximized="1"
     inkscape:current-layer="layer1"
     fit-margin-top="0"
     fit-margin-left="0"
     fit-margin-right="0"
     fit-margin-bottom="0" />
  <defs
     id="defs2">
    <rect
       x="117.84912"
       y="88.186447"
       width="506.15164"
       height="133.77667"
       id="rect2805" />
  </defs>
  <g
     inkscape:label="Layer 1"
[...]

PR #1708, opened 2022-03-22
Alotor/bugfixes

### LINKED ISSUE #1522: Prototype view not accessible from iOS -IPv4 related?
**Users invited to see a prototype are not able to open the link on their iOS device**
Instead of the prototype users see „internal error. If problem persists please contact support.“
We assume that PenPot runs on iPv4. This would explain iOS blocking the connection. 

**To Reproduce**
1.  Chose a page 
2.  Click „Share Prototype“
3. Choose „Zuschauer“ (probably „View only“ in the English version? 
4.  Send to Person with iOS device.

**Expected behavior**
The recipient of the link will not be able to open the prototype. 

**Screenshots**
C64EFBED-9246-400D-A904-24638FF1F393

**Smartphone (please complete the following information):**

-   Device: 5 different iPhones
-   OS: at least one of them: iOS 15.3
-   Browser  Safari and Chrome

Motivational Cat for you: 😻

### LINKED ISSUE #1531: Import SVG with images stored externally
When I import / drag & drop a SVG onto the canvas that contains images stored outside the SVG file penpot aborts due to a validation error. The error tells me that the URL to the image could not be found.

Instead I would like penpot to import the SVG and the images but gracefully ignore unreachable URLs so I can replace those missing images in penpot myself.

An alternative would be to replace images with unreachable URLs with a placeholder image that displays e.g. a red "X" to indicate that there is a missing image.

### LINKED ISSUE #1587: Rounded corner implementation improvements
Rounded rectangles should be constrained to the correct shape, and currently when the rectangle rounded value exceeds the short side value, the shape is displayed as an oval corner until the rectangle eventually becomes a full oval. This is inconsistent with the actual CSS code presentation style, where rounded corners are supposed to be positively rounded at all times and constrained by the length of the short side (perhaps the oval corners are a special design? But this is not the case in most design software, and it is also rare to see this element in actual interface design), and secondly, when I transpose the shape of the elliptical corners, his shape is wrong, and this phenomenon exists in Boolean operations (sorry, I can't describe it clearly, please see the video), I think this way of rounding corners should be improved. 

The rounded display should always be positive rounded and confined to 1/2 of the length of the short edge, but should also accept a larger rounded value. For example, if you add a 100px rounded corner to a 10 × 20 rectangle, the rounded corner will appear as 10/2=5px. If you adjust the rectangle to 100 × 200, the rounded corner will appear as 100/2=50px because it is still less than the input value of 100px (see CSS). When converting a shape to a curve,  should use the actual size of the rounded corners displayed instead of entering a value to ensure a consistent display.

If oval rounded corners are an implementation or style, maybe add a full rounded/oval corner toggle switch?

**Additional context**
image_2017
image_2018
The CSS display is not the same as penpot.

https://user-images.githubusercontent.com/40300919/153803893-a57560cd-3700-451b-9d52-2dc63b3dcb59.mp4

https://user-images.githubusercontent.com/40300919/153803900-e4d282ed-e03e-429b-abd7-f15b710b4319.mp4

The video shows the ellipse Angle being broken when converted to curve/in booleans

### LINKED ISSUE #1680: Faster access to renaming objects
## Problem

It's easy to miss this area for double-clicking:  
image

Also, the absence of keyboard shortcut for "Rename" is creating a massive dependency on the mouse input for seemingly simple quality-of-life operation.

## Expectation

Expected to have intuitively faster access to renaming objects:
- by double-clicking anywhere near by the name: preferably in range from the icon to the hover controls on the right;
- by selecting object and just pressing <kbd>F2</kbd>.

Alotor:
- Don't stop SVG import when an image cannot be imported [#1531](https://github.com/penpot/penpot/issues/1531)
- Fix crash on iOS when displaying viewer [#1522](https://github.com/penpot/penpot/issues/1522)
- Fix problem when importing a SVG with text [#1532](https://github.com/penpot/penpot/issues/1532)
- Fix problem when adding shadows to imported text [#Taiga 3057](https://tree.taiga.io/project/penpot/issue/3057)
- Fix problem when importing SVG's with uses with overriding properties [#Taiga 2884](https://tree.taiga.io/project/penpot/issue/2884)
- Fix inconsistency with radius in SVG an CSS [#1587](https://github.com/penpot/penpot/issues/1587)
- Fix clickable area in layers [#1680](https://github.com/penpot/penpot/issues/1680)

### DIFF
--- a/frontend/src/app/main/ui/shapes/attrs.cljs
+++ b/frontend/src/app/main/ui/shapes/attrs.cljs
@@ -98,14 +98,8 @@
            (contains? shape :fill-color)
            {:fill (:fill-color shape)}
 
-           ;; If contains svg-attrs the origin is svg. If it's not svg origin
-           ;; we setup the default fill as transparent (instead of black)
-           (and (not (contains? shape :svg-attrs))
-                (not (#{:svg-raw :group} (:type shape))))
-           {:fill "none"}
-
            :else
-           {})
+           {:fill "none"})
 
          fill-attrs (cond-> fill-attrs
                       (contains? shape :fill-opacity)
@@ -211,6 +205,13 @@
                       (-> styles
                           (obj/set! "fill" (obj/get svg-styles "fill"))
                           (obj/set! "fillOpacity" (obj/get svg-styles "fillOpacity")))
+
+                      ;; If contains svg-attrs the origin is svg. If it's not svg origin
+                      ;; we setup the default fill as transparent (instead of black)
+                      (and (contains? shape :svg-attrs)
+                           (#{:svg-raw :group} (:type shape))
+                           (empty? (:fills shape)))
+                      styles
 
                       :else
                       (add-fill styles (d/without-nils (get-in shape [:fills 0])) render-id 0))]
--- a/frontend/src/app/main/ui/shapes/attrs.cljs
+++ b/frontend/src/app/main/ui/shapes/attrs.cljs
@@ -7,6 +7,8 @@
 (ns app.main.ui.shapes.attrs
   (:require
    [app.common.data :as d]
+   [app.common.data.macros :as dm]
+   [app.common.geom.shapes :as gsh]
    [app.common.spec.radius :as ctr]
    [app.common.spec.shape :refer [stroke-caps-line stroke-caps-marker]]
    [app.main.ui.context :as muc]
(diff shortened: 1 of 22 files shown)


frontend/src/app/main/ui/shapes/attrs.cljs
```diff
@@ -26,49 +27,20 @@
...
-          right  (- (:height shape) r2 r3)
-          bottom (- (:width shape) r3 r4)
-          left   (- (:height shape) r4 r1)]
-      (obj/merge! attrs #js {:d (str "M" (+ (:x shape) r1) "," (:y shape) " "
+    (let [[r1 r2 r3 r4] (gsh/shape-corners-4 shape)
+          top    (- width r1 r2)
+          right  (- height r2 r3)
+          bottom (- width r3 r4)
+          left   (- height r4 r1)]
+      (obj/merge! attrs #js {:d (str "M" (+ x r1) "," y " "
```
niwinz: maybe use `dm/str` for this case

frontend/src/app/util/dom.cljs
```diff
@@ -306,8 +305,8 @@
     (boolean (.-fullscreenElement globals/document))
 
     :else
-    (ex/raise :type :not-supported
-              :hint "seems like the current browser does not support fullscreen api.")))
+    (do (js/console.error "Seems like the current browser does not support fullscreen api.")
```
niwinz: I think is better using the logging api, here instead of console log.

frontend/src/app/util/webapi.cljs
```diff
@@ -114,8 +113,8 @@
     (.webkitRequestFullscreen el)
 
     :else
-    (ex/raise :type :not-supported
-              :hint "seems like the current browser does not support fullscreen api.")))
+    (do (js/console.error "Seems like the current browser does not support fullscreen api.")
```
niwinz: the same here, console.* -> logging api

frontend/src/app/util/webapi.cljs
```diff
@@ -127,8 +126,8 @@
     (.webkitExitFullscreen js/document)
 
     :else
-    (ex/raise :type :not-supported
-              :hint "seems like the current browser does not support fullscreen api.")))
+    (do (js/console.error "Seems like the current browser does not support fullscreen api.")
```
niwinz: The same here :D

PR #1752, opened 2022-03-28
:bug: Avoid rotating when scrolling

EvaMarco:
This PR fixes a problem with numeric inputs. https://tree.taiga.io/project/penpot/issue/3140

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,6 +42,7 @@
 
 ### :bug: Bugs fixed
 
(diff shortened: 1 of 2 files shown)


niwinz:
As a suggestion only: the commit message should be shorter and if you need to explain something, you can use the "body" part of for larger explanation.

EvaMarco:
@niwinz thanks, I've have change it already :)

PR #1762, opened 2022-03-29
:bug: Fix problem when alt+drag duplicate frames

Alotor:
https://tree.taiga.io/project/penpot/issue/3141

### DIFF
--- a/frontend/src/app/main/data/workspace/persistence.cljs
+++ b/frontend/src/app/main/data/workspace/persistence.cljs
@@ -14,6 +14,7 @@
    [app.common.spec.change :as spec.change]
    [app.common.spec.file :as spec.file]
    [app.common.uuid :as uuid]
+   [app.config :as cfg]
    [app.main.data.dashboard :as dd]
    [app.main.data.events :as ev]
    [app.main.data.fonts :as df]
@@ -653,6 +654,9 @@
 
             frame-changes (->> stream
                                (rx/filter dch/commit-changes?)
+
+                               ;; Async so we wait for additional side-effects of commit-changes
+                               (rx/observe-on :async)
                                (rx/filter (comp not thumbnail-change?))
                                (rx/with-latest-from objects-stream)
                                (rx/map extract-frame-changes)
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/data/workspace/persistence.cljs
```diff
@@ -678,3 +682,26 @@
...
+              (cond
+                (some? metadata)
+                [(cfg/resolve-file-media metadata)]
+
+                (some? fill-image)
+                [(cfg/resolve-file-media fill-image)]))
+
+            uris (into #{}
+                       (comp (mapcat extract-urls)
+                             (filter some?))
```
niwinz, line 703: I think there are no need for mapcat, and everything can be simplified to `keep`. I mean, extract-urls, can return a single item without wrapping it into vector, and instead of mapcat + filter, use `keep`.

On the other hand, I think your intention was handle the case where you have the metadata and and fill-image, but in this case, the `cond` construct i think it is not adequate for this case... maybe `cond->> nil` + `cons`:

```clojure
(cond->> nil
  (some? metadata)
  (cons (cfg/resolve-file-media metadata))
  
  (some? fill-image)
  (cons (cfg/resolve-file-media fill-image)))
```

and then leave the xform as it is right now.

frontend/src/app/main/ui/shapes/text/fontfaces.cljs
```diff
@@ -76,14 +76,10 @@
...
-                     (filterv #(= :text (:type %))))
-
-        content (->> shapes (mapv :content))
-
-        ;; Retrieve the fonts ids used by the text shapes
-        fonts (->> content
-                   (mapv fonts/get-content-fonts)
+  (let [;; Retrieve the fonts ids used by the text shapes
+        fonts (->> (obj/get props "shapes")
+                   (filterv #(= :text (:type %)))
```
niwinz: `cph/test-shape?` predicate should be used here (i know that is not new, but I think we should change it gradually to named predicates that facilitates the readability)

frontend/src/app/main/data/workspace/persistence.cljs
Alotor: I think it's better to use the mapcat thinking in the future when you could potentialy have different images in the multiple fills.

frontend/src/app/main/ui/shapes/text/fontfaces.cljs
Alotor: ok

PR #1763, opened 2022-03-29
Migration fixes

niwinz:


common/src/app/common/pages/migrations.cljc
```diff
@@ -291,83 +282,91 @@
...
+              ;; First: delete the old unused props from image shapes
+              (cph/image-shape? object)
+              (dissoc :fill-color :fill-opacity)
+
+              ;; Second: fix already migrated shapes (only applicable
+              ;; if you are using nightly builds)
+              (and (cph/image-shape? object)
+                   (contains? object :fills)
+                   (= 1 (count (:fills object)))
+                   (= (:fill-color object) (get-in object [:fills 0 :fill-color])))
```
superalex: I would also check the fill-color and the opacity are the default values (1 and #B1B2B5)

niwinz: the main idea here is checking if the current :fills has the same value as prev and no more fills are added. I think we don't need more checks here, because previously we only have a single hidden fill-color (and fill-opacity) on images, and the previous migration creates the :fills with single element, so checking the color matches and has a single element is enough

backend/src/app/srepl/main.clj
```diff
@@ -66,86 +67,48 @@
...
 
-(defn verify-files
-  [system {:keys [age sleep chunk-size max-chunks stop-on-error? verbose?]
-           :or {sleep 1000
-                age "72h"
-                chunk-size 10
-                verbose? false
-                stop-on-error? true
-                max-chunks ##Inf}}]
+;; (defn check-image-shapes
```
niwinz, line 70: This function is commented just as example, I prefer have this code commented here instead of removing it.

hirunatan: this may be explained in a comment, to avoid someone deleting this in the future

niwinz: No one is going to touch this file unless it is needed, and when it needed. I mean, this is the namespace where we leave some functions that need to be executed on the server REPL (scripts). This is not so important to leave comments right now I think, I just saying that I don't want to remove it right now, in future it can be removed without any problem.

PR #1773, opened 2022-03-31
:bug: Fix add fill to artboard modify children

EvaMarco:
This PR solves [Taiga #3151](https://tree.taiga.io/project/penpot/issue/3151)

### DIFF
--- a/frontend/src/app/main/data/workspace/colors.cljs
+++ b/frontend/src/app/main/data/workspace/colors.cljs
@@ -8,12 +8,14 @@
   (:require
    [app.common.colors :as clr]
    [app.common.data :as d]
+   [app.common.pages.helpers :as cph]
    [app.main.data.modal :as md]
    [app.main.data.workspace.changes :as dch]
    [app.main.data.workspace.layout :as layout]
    [app.main.data.workspace.state-helpers :as wsh]
    [app.main.data.workspace.texts :as dwt]
    [app.main.repo :as rp]
+   [app.util.color :as uc]
    [beicon.core :as rx]
    [potok.core :as ptk]))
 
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/data/workspace/colors.cljs
```diff
@@ -390,3 +392,22 @@
...
+  (ptk/reify ::manage-color-in-palette
+    ptk/WatchEvent
+    (watch [_ state _]
+      (let [objects  (wsh/lookup-page-objects state)
+            selected (->> (wsh/lookup-selected state)
+                          (cph/clean-loops objects))
+            ids      (into []
+                           (comp (keep (d/getf objects))
+                                 (mapcat (fn [shape]
+                                           (cons (:id shape)
```
hirunatan: This causes that the fill or stroke is applied to the group shapes (it shouldn't, only to children).

frontend/src/app/main/data/workspace/colors.cljs
```diff
@@ -390,3 +392,26 @@
...
+(defn manage-color-shapes
+  [shape objects]
+  (let [shapes (case (:type shape)
+                 :group (cph/get-children objects (:id shape))
+                 [shape])]
+    (->> shapes
+         (remove cph/group-shape?)
+         (map :id))))
+
+(defn manage-color-in-palette
```
hirunatan: How about `apply-color-from-palette`? I find it more descriptive.

frontend/src/app/main/data/workspace/colors.cljs
```diff
@@ -390,3 +392,26 @@
     (update [_ state]
       (-> state
           (assoc-in [:workspace-global :editing-stop] spot)))))
+
+(defn manage-color-shapes
```
hirunatan: Perhaps move this function inside the event, instead of being a global function. And the name could be `select-shapes-for-color` or similar.

frontend/src/app/main/data/workspace/colors.cljs
EvaMarco: Good point¡¡, changed¡

frontend/src/app/main/data/workspace/colors.cljs
EvaMarco: Ok, I see. Let me know if now it's better

PR #1778, opened 2022-04-04
Fix undo

hirunatan:
Perhaps is the reason of this https://tree.taiga.io/project/penpot/issue/3178


niwinz:
Should be to staging

PR #1790, opened 2022-04-06
Enhancements on email console logging

niwinz:


luwol03:
Why was that pr closed?

niwinz:
It is closed because it was replaced with other that already included the changes #1792

luwol03:
Oh ok. Thank you for your very quick answer. Then I should be able to have this changes in the next release? Is that already planned?

niwinz:
yes, we will bump a minor release soon (next week we have holidays here in spain) so probably first days of the following week

PR #1834, opened 2022-04-22
Synchronize library colors in all parts of a shape

hirunatan:
https://tree.taiga.io/project/penpot/issue/3182

### DIFF
--- a/common/src/app/common/pages/changes_builder.cljc
+++ b/common/src/app/common/pages/changes_builder.cljc
@@ -286,37 +286,38 @@
          update-shape
          (fn [changes id]
            (let [old-obj (get objects id)
-                 new-obj (update-fn old-obj)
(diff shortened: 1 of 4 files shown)


common/src/app/common/pages/changes_builder.cljc
```diff
@@ -286,37 +286,38 @@
          update-shape
          (fn [changes id]
            (let [old-obj (get objects id)
-                 new-obj (update-fn old-obj)
+                 new-obj (update-fn old-obj)]
+             (if (= old-obj new-obj)
+               changes
```
hirunatan, line 291: Here the only change is adding this `(if`. The rest is the diff algorithm getting mad.

superalex:
I've tested the feature and it's working perfectly except for the shadows

(All the colors used in the image except the gray are the same library color)

PR #1835, opened 2022-04-22
:recycle: Persistence flow changes

niwinz:
This new approach intends to leave less room to frustration of spend too many time working and lost all the work because of ignore some unexpected error that happened on the half of work session. 
Instead of having a simple denounce (that can lead to accumulate too many chages) and the wrong approach on don't reset the state after error leaving the "wrong" data on the queue all the time...

The new approach uses more sophisticated trigger (a debounced changes plus a maximum interval of 5 seconds) and resets the state if an error happens.


frontend/src/app/util/webapi.cljs
```diff
@@ -37,7 +37,20 @@
...
+;;   (let [[mtype b64-data] (str/split data-uri ";base64,")
+;;         mtype   (subs mtype (inc (str/index-of mtype ":")))
+;;         decoded (.atob js/window b64-data)
+;;         size    (.-length ^js decoded)
+;;         content (js/Uint8Array. size)]
+
+;;     (doseq [i (range 0 size)]
+;;       (obj/set! content i (.charCodeAt decoded i)))
+
+;;     (js/Blob. #js [content] #js {"type" mtype})))
```
hirunatan: delete this?

frontend/src/app/main/data/workspace/persistence.cljs
```diff
@@ -71,93 +70,89 @@
...
+              (rx/observe-on :async)
+              (rx/tap #(swap! commits conj %))
+              (rx/take-until (rx/delay 100 stoper))
+              (rx/finalize (fn []
+                             (log/debug :hint "finalize persistence: changes watcher"))))
+
+         (->> (rx/from-atom commits)
+              (rx/filter (complement empty?))
+              (rx/sample-when (rx/merge
+                               (rx/interval 5000)
```
hirunatan, line 94: Why is this necessary? Won't it trigger a save process even when there is no commits to save?

frontend/src/app/util/webapi.cljs
hirunatan: i'll do it myself

PR #1900, opened 2022-05-13
Hirunatan fix mixed measures

hirunatan:
https://tree.taiga.io/project/penpot/issue/3366

### DIFF
--- a/common/src/app/common/attrs.cljc
+++ b/common/src/app/common/attrs.cljc
@@ -5,7 +5,49 @@
 ;; Copyright (c) UXBOX Labs SL
 
 (ns app.common.attrs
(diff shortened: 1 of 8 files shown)


common/src/app/common/attrs.cljc
```diff
@@ -37,47 +39,70 @@
...
 
-(defn merge
-  "Attrs specific merge function."
-  [obj attrs]
-  (reduce-kv (fn [obj k v]
-               (if (nil? v)
-                 (dissoc obj k)
-                 (assoc obj k v)))
-             obj
-             attrs))
```
hirunatan, line 83: Moved to avoid a circular dependency with geom.shapes.transform

PR #1912, opened 2022-05-17
Fix text synchronization

hirunatan:
Fixes this one https://tree.taiga.io/project/penpot/issue/3372

But also a bunch of other things when synchronizing text shapes.

### DIFF
--- a/common/src/app/common/geom/shapes.cljc
+++ b/common/src/app/common/geom/shapes.cljc
@@ -6,6 +6,7 @@
 
 (ns app.common.geom.shapes
   (:require
+   [app.common.data :as d]
    [app.common.data.macros :as dm]
    [app.common.geom.point :as gpt]
    [app.common.geom.shapes.bool :as gsb]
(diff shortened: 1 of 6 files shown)


common/src/app/common/geom/shapes.cljc
```diff
@@ -38,6 +39,14 @@
 
 ;; --- Helpers
 
+(defn left-bound
+  [shape]
+  (get shape :x (:x (:selrect shape)))) ; Paths don't have :x attribute
```
Alotor, line 44: using the selrect as "left bound" won't take into account transformations. Maybe it's better to do points->selrect->x. The same goes for top bound

hirunatan: I thought so. But the original code just checked for the untranslated :x and :y

This is used to check if a component has moved with all its inner shapes in place (to calculate the touched flags). When there is a rotation, the touched are already activated by other attributes. I think it's not necessary to check the transformation here (and we save some CPU). This function occurs in the critical path while moving shapes.

The only thing that I don't like is the name: "bound" seems to refer to the containing rectangle, and it's easy to think of the rotated shape. But can't find a better name.

PR #1955, opened 2022-05-26
:bug: Fix github auth without public email

superalex:
Related to https://tree.taiga.io/project/penpot/issue/3457

### DIFF
--- a/backend/src/app/http/oauth.clj
+++ b/backend/src/app/http/oauth.clj
@@ -81,14 +81,28 @@
                           :timeout 6000
                           :method :get}))
 
-          (validate-response [{:keys [status body] :as res}]
-            (when-not (= 200 status)
+          (retrieve-emails []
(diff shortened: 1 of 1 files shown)


backend/src/app/http/oauth.clj
```diff
@@ -81,14 +81,28 @@
...
+          (retrieve-emails []
+            (if (some? (:emails-uri provider))
+              (http-client {:uri (:emails-uri provider)
+                            :headers {"Authorization" (str (:type tdata) " " (:token tdata))}
+                            :timeout 6000
+                            :method :get})
+              (p/resolved {:status 200})))
+
+          (validate-response [[retrieve-res emails-res]]
+            (when-not (= 200 (:status retrieve-res))
```
Alotor: I think it's better to check the status to be in a range. 20X codes are all "OK" i think it's better to be safe and check if it's in the range [200, 300)

PR #1961, opened 2022-05-30
:bug: Fix old texts with opacity and no fill

superalex:
Related to https://tree.taiga.io/project/penpot/issue/3470

### DIFF
--- a/frontend/src/app/main/ui/shapes/attrs.cljs
+++ b/frontend/src/app/main/ui/shapes/attrs.cljs
@@ -56,49 +56,53 @@
     attrs))
 
 (defn add-fill
-  ([attrs shape render-id]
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/ui/shapes/attrs.cljs
```diff
@@ -56,10 +56,10 @@
     attrs))
 
 (defn add-fill
-  ([attrs shape render-id]
-   (add-fill attrs shape render-id nil))
+  ([attrs shape render-id type]
```
Alotor: Why add the type here when you have it already inside `(:type shape)`

superalex: I've renamed shape to fill-date because it only contains the :fills attribute of the shape

PR #1976, opened 2022-06-03
:bug: Fix auto scroll layers panel in firefox

hirunatan:
https://tree.taiga.io/project/penpot/issue/3531


superalex:
@hirunatan I think this should be a hotfix for main, isn't it?

PR #2002, opened 2022-06-16
:tada: Improvements on view mode

pabloalba:
Related to https://tree.taiga.io/project/penpot/us/3023


frontend/src/app/main/ui/viewer.cljs
```diff
@@ -60,6 +61,112 @@
              :height (* height zoom)
              :vbox   (str "0 0 " width " " height)})))
 
+
+(mf/defc viewer-pagination
+  [{:keys [index num-frames left-bar right-bar] :as props}]
+  [:*
+   (when (> index 0)
```
niwinz: condition can be simplified with `pos?` predicate

frontend/src/app/main/ui/viewer.cljs
```diff
@@ -60,6 +61,112 @@
...
+                           :visible (:background-overlay overlay))
+                   :style {:width (:width wrapper-size)
+                           :height (:height wrapper-size)
+                           :position "absolute"
+                           :left 0
+                           :top 0}
+                   :on-click #(when (:close-click-outside overlay)
+                                (close-overlay (:frame overlay)))}])
+               [:div.viewport-container.viewer-overlay
+                {:id (str "overlay-" (str (:id (:frame overlay))))
```
niwinz: You can replace this with `(str "overlay-" (-> overlay :frame :id))` (removed the nested `str` call)

frontend/src/app/main/ui/workspace/comments.cljs
```diff
@@ -57,20 +57,22 @@
       [:span.label (tr "labels.hide-resolved-comments")]]]))
 
 (mf/defc comments-sidebar
-  []
+  [{:keys [users threads page-id]}]
   (let [threads-map (mf/deref refs/threads-ref)
         profile     (mf/deref refs/profile)
-        users       (mf/deref refs/users)
+        users       (or users (mf/deref refs/users))
```
niwinz: From one render to other, `users` value can change from nil to something non-nil and backwards? if YES, this will break the react hooks rules.
The correct use here is `mf/deref` first, then do the OR, ensuring strict order of execution of hooks independently if `users` value.

PR #2064, opened 2022-06-30
:sparkles: Signin/Signup from shared link

pabloalba:
Related to https://tree.taiga.io/project/penpot/us/3472

### DIFF
--- a/frontend/src/app/main/ui/auth/recovery_request.cljs
+++ b/frontend/src/app/main/ui/auth/recovery_request.cljs
@@ -22,15 +22,19 @@
 (s/def ::recovery-request-form (s/keys :req-un [::email]))
 
 (mf/defc recovery-form
-  []
+  [{:keys [on-success-callback] :as props}]
   (let [form      (fm/use-form :spec ::recovery-request-form :initial {})
         submitted (mf/use-state false)
 
+        default-success-finish #(st/emit! (dm/info (tr "auth.notifications.recovery-token-sent")))
+
         on-success
         (mf/use-callback
-         (fn [_ _]
+         (fn [cdata _]
            (reset! submitted false)
-           (st/emit! (dm/info (tr "auth.notifications.recovery-token-sent")))))
+           (if (nil? on-success-callback)
+             (default-success-finish)
+             (on-success-callback (:email cdata)))))
 
         on-error
         (mf/use-callback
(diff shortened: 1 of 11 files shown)


frontend/src/app/main/ui/auth/recovery_request.cljs
```diff
@@ -22,15 +22,21 @@
...
-  []
+  [{:keys [on-success-callback] :as props}]
   (let [form      (fm/use-form :spec ::recovery-request-form :initial {})
         submitted (mf/use-state false)
 
+        default-success-finish #(st/emit! (dm/info (tr "auth.notifications.recovery-token-sent")))
+
+        success-finish #(if (nil? on-success-callback)
+                   (default-success-finish)
+                   (on-success-callback (:email %)))
```
niwinz: I think that this anonymos function can be inlined directly in the on-success callback, it is a single use anonymous function of 2 lines.

frontend/src/app/main/ui/viewer.cljs
```diff
@@ -255,6 +257,8 @@
     (when (not allowed)
       (st/emit! (dv/go-to-section :interactions)))
 
+    (st/emit! (modal/hide))
```
niwinz: Do not emit events on body of event, you want perform side effects, it is better to wrap it in an effect hook:

```clojure
(mf/with-effect []
  (st/emit! (modal/hide)))
```

niwinz: The same for the previous code that is not part of this PR.

frontend/src/app/main/ui/viewer/login.cljs
```diff
@@ -0,0 +1,103 @@
...
+        success-email-sent
+        (fn [email]
+          (reset! user-email email)
+          (set-current-section :email-sent))
+        success-register
+        (fn [data]
+          (reset! register-token (:token data))
+          (set-current-section :register-validate))
+        ]
+    (swap! storage assoc :redirect-url uri)
```
niwinz: The same, no side effects on render, wrap it with effect hook

superalex:
Great!,

Some stuff I found:

- The header need some space right and bottom:

image

- After login this button doesn't change and keeps saying me to login (if I refresh it shows I'm logged)

PR #2066, opened 2022-07-01
:sparkles: Update the shared link properties instead of create multiples links

pabloalba:
Related to https://tree.taiga.io/project/penpot/us/3685

### DIFF
--- a/backend/src/app/migrations/sql/0076-one-shared-link-per-file.sql
+++ b/backend/src/app/migrations/sql/0076-one-shared-link-per-file.sql
@@ -1,0 +1,3 @@
+DELETE FROM share_link 
+  WHERE created_at not in 
+    (SELECT max(created_at) FROM share_link GROUP BY file_id);
--- a/backend/src/app/rpc/mutations/share_link.clj
+++ b/backend/src/app/rpc/mutations/share_link.clj
@@ -51,8 +51,7 @@
                            :who-inspect who-inspect
                            :pages pages
                            :owner-id profile-id})]
-    (-> slink
-        (update :pages db/decode-pgarray #{}))))
+    (update slink :pages db/decode-pgarray #{})))
 
 ;; --- Mutation: Delete Share Link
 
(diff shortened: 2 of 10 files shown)


backend/src/app/rpc/mutations/share_link.clj
```diff
@@ -68,3 +68,29 @@
...
+(defn update-share-link
+  [conn {:keys [id profile-id pages who-comment who-inspect]}]
+  (let [pages (db/create-array conn "uuid" pages)
+        slink (db/update! conn :share-link
+                          {:who-comment who-comment
+                           :who-inspect who-inspect
+                           :pages pages
+                           :owner-id profile-id}
+                          {:id id})]
+    (-> slink
```
niwinz, line 94: No `->` is needed here,when we have only one chained transformation.

backend/src/app/rpc/queries/viewer.clj
```diff
@@ -29,8 +29,9 @@
           libs    (files/retrieve-file-libraries cfg false file-id)
           users   (comments/retrieve-file-comments-users pool file-id profile-id)
 
-          links   (->> (db/query pool :share-link {:file-id file-id})
-                       (mapv slnk/decode-share-link-row))
+          link   (->> (db/query pool :share-link {:file-id file-id} {:order-by [:created-at]})
+                      (mapv slnk/decode-share-link-row)
```
niwinz: O think this `mapv` should be a `map` (the normal one)

backend/src/app/migrations/sql/0076-one-shared-link-per-file.sql
```diff
@@ -0,0 +1,3 @@
+DELETE FROM share_link 
+  WHERE created_at not in 
+    (SELECT max(created_at) FROM share_link GROUP BY file_id);
```
niwinz, line 3: We should be careful with this kind of migrations, they are not backward compatible.

PR #2084, opened 2022-07-11
:sparkle: Comments positioning

superalex:
Related to: https://tree.taiga.io/project/penpot/task/3635

### DIFF
--- a/frontend/src/app/main/data/comments.cljs
+++ b/frontend/src/app/main/data/comments.cljs
@@ -8,7 +8,10 @@
   (:require
    [app.common.data :as d]
    [app.common.geom.point :as gpt]
+   [app.common.pages.helpers :as cph]
    [app.common.spec :as us]
+   [app.common.uuid :as uuid]
+   [app.main.data.workspace.state-helpers :as wsh]
    [app.main.repo :as rp]
    [beicon.core :as rx]
    [cljs.spec.alpha :as s]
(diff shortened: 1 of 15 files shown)


backend/src/app/worker.clj
```diff
@@ -497,7 +497,7 @@
   [{:keys [pool entries]}]
   (db/with-atomic [conn pool]
     (doseq [{:keys [id cron]} entries]
-      (l/trace :hint "register cron task" :id id :cron (str cron))
+      (l/info :hint "register cron task" :id id :cron (str cron))
```
niwinz: For consistency it should be trace I think.

superalex: You are right :)

frontend/src/app/main/data/comments.cljs
```diff
@@ -195,7 +257,16 @@
...
+                   (let [set-comment-threds (fn set-comment-threds [st comment-thread]
+                                              (let [path [:workspace-data :pages-index (:page-id comment-thread) :options :comment-threads-position (:id comment-thread)]
+                                                    thread-position (get-in state path)]
+                                                (cond-> st
+                                                  (nil? thread-position)
+                                                  (->
+                                                   (assoc-in (conj path :position) (:position comment-thread))
+                                                   (assoc-in (conj path :frame-id) (:frame-id comment-thread))))))
+                         state (assoc state :comment-threads (d/index-by :id data))]
+                     (reduce set-comment-threds state data)))]
```
niwinz: I think this that the reduce function can be extracted to a letfn binding:

```clojure
(letfn [(set-comment-threads [state thread]
          (let [path [:workspace-data :pages-index (:page-id thread) :options :comment-threads-position (:id thread)]
                pos  (get-in state path)]
            (cond-> state
              (nil? (get-in state path))
              (-> (assoc-in (conj path :position) (:position thread))
                  (assoc-in (conj path :frame-id) (:frame-id thread))))))

        (fetched [data state]
          (reduce set-comment-threads
                  (assoc state :comment-threads (d/index-by :id data))
                  data))]
     [...]
```

frontend/src/app/main/data/comments.cljs
```diff
@@ -338,3 +409,43 @@
...
+      ptk/UpdateEvent
+      (update [_ state]
+        (let [thread-id (:id thread)]
+          (assoc-in state [:comment-threads thread-id :frame-id] frame-id)))
+
+      ptk/WatchEvent
+      (watch [_ _ _]
+        (let [thread-id (:id thread)]
+          (->> (rp/mutation :update-comment-thread-frame {:id thread-id  :frame-id frame-id})
+               (rx/catch #(rx/throw {:type :update-comment-thread-frame}))
```
niwinz: The `:type` should be part of a general keyword part of some category. Something like `:comment-error` and the `:code` can be something that describes the concrete cause.

frontend/src/app/main/data/comments.cljs
```diff
@@ -338,3 +409,43 @@
...
+  [ids]
+  (us/verify (s/coll-of uuid?) ids)
+
+  (ptk/reify ::detach-comment-thread
+    ptk/WatchEvent
+    (watch [_ state _]
+      (let [objects (wsh/lookup-page-objects state)
+            is-frame? (fn [id] (= :frame (get-in objects [id :type])))
+            frame-ids? (into #{} (filter is-frame?) ids)
+            detach-comment (fn [comment-threads]
```
niwinz: This function looks like an alias, why not calling directly to `update-comment-thread-frame` ?

frontend/src/app/main/data/workspace/comments.cljs
```diff
@@ -95,8 +104,76 @@
...
+
+  (ptk/reify ::move-frame-comment-threads
+    ptk/WatchEvent
+    (watch [_ state _]
+      (let [objects (wsh/lookup-page-objects state)
+
+            is-frame? (fn [id] (= :frame (get-in objects [id :type])))
+            frame-ids? (into #{} (filter is-frame?) ids)
+
+            object-modifiers  (get state :workspace-modifiers)
```
niwinz: if the keyword is known at compile time, just use it as function instead of `get`

frontend/src/app/main/data/workspace/comments.cljs
```diff
@@ -95,8 +104,76 @@
...
+                    moved (gpt/to-vec (gpt/point (:x frame) (:y frame))
+                                      (gpt/point (:x frame') (:y frame')))
+                    position (get-in threads-position-map [(:id comment-thread) :position])
+                    new-x (+ (:x position) (:x moved))
+                    new-y (+ (:y position) (:y moved))]
+                (update-comment-thread-position comment-thread [new-x new-y] (:id frame))))]
+
+        (->> (:comment-threads state)
+             (vals)
+             (mapv #(assoc % :position (get-in threads-position-map [(:id %) :position])))
```
niwinz: don't use mapv if you really don't need the output to be vector, and in this case, filter is used later so vector is clearly not needed.

frontend/src/app/main/ui/comments.cljs
```diff
@@ -313,33 +317,126 @@
...
+        {:keys [on-pointer-down
+                on-pointer-up
+                on-mouse-move
+                state
+                on-lost-pointer-capture]} (use-buble zoom thread)
+
+        pos-x (* (or (:new-position-x @state) (:x pos)) zoom)
+        pos-y (* (or (:new-position-y @state) (:y pos)) zoom)
+
+        on-pointer-down* (fn [event]
```
niwinz: Missing the `use-fn/use-callback` hooks I think

frontend/src/app/main/ui/viewer/comments.cljs
```diff
@@ -96,7 +98,16 @@
...
 
+        update-thread-position (fn update-thread-position [thread]
+                                 (if (contains? threads-position-map (:id thread))
+                                   (-> thread
+                                       (assoc :position (get-in threads-position-map [(:id thread) :position]))
+                                       (assoc :frame-id (get-in threads-position-map [(:id thread) :frame-id])))
+                                   thread))
+
         threads     (->> (vals threads-map)
+                         (mapv update-thread-position)
```
niwinz: Don't use mapv if you don't need explicitly the vector type as return value.

frontend/src/app/main/data/workspace/comments.cljs
superalex: Great!

PR #2107, opened 2022-07-26
:sparkles: 2645 Hovering layers bounding box

andrewzhurov:
This PR introduces outlining of shapes on hover upon their representation in Design hierarchy and Layer selection menu.

record

Closes https://tree.taiga.io/project/penpot/us/2645

### DIFF
--- a/frontend/src/app/main/data/workspace.cljs
+++ b/frontend/src/app/main/data/workspace.cljs
@@ -45,6 +45,7 @@
    [app.main.data.workspace.path.shapes-to-path :as dwps]
    [app.main.data.workspace.persistence :as dwp]
    [app.main.data.workspace.selection :as dws]
+   [app.main.data.workspace.highlight :as dwh]
    [app.main.data.workspace.shape-layout :as dwsl]
    [app.main.data.workspace.shapes :as dwsh]
    [app.main.data.workspace.state-helpers :as wsh]
@@ -1731,6 +1732,10 @@
 (dm/export dws/select-inside-group)
 (dm/export dws/select-shape)
 (dm/export dws/shift-select-shapes)
+
+;; Highlight
+(dm/export dwh/highlight-shape)
+(dm/export dwh/dehighlight-shape)
 
 ;; Groups
 (dm/export dwg/mask-group)
--- a/frontend/src/app/main/data/workspace/highlight.cljs
+++ b/frontend/src/app/main/data/workspace/highlight.cljs
@@ -1,0 +1,28 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) UXBOX Labs SL
+
+(ns app.main.data.workspace.highlight
+  (:require
+   [app.common.spec :as us]
+   [potok.core :as ptk]))
+
+;; --- Manage shape's highlight status
+
+(defn highlight-shape
+  [id]
+  (us/verify ::us/uuid id)
+  (ptk/reify ::highlight-shape
+    ptk/UpdateEvent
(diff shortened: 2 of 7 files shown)


niwinz:
Looks very nice @andrewzhurov 
Thank you very much for the contribution.

Probably today I will not have enough quality time for it, but tomorrow I will checkout and test it. Also, with your permission I will make some cosmetic changes on your code. I will do it on separated commits that explains the changes.

andrewzhurov:
Thank you for your kind words.

> Also, with your permission I will make some cosmetic changes on your code.

Absolutely! I would love to learn how it can be tuned up.

andrewzhurov:
I have two concerns about the solution:
1. It doesn't introduce a uniform 'outline shape on hover over its representation' behaviour, as there is also Design hierarchy in View mode which isn't covered.
    Perhaps we'd like to cover it as well?
2. Managing 'what is outlined' by mouse-enter and mouse-leave events is an imperative style prone to errors, it would be way more robust to have a pointer position tracked and current hover subject derived out of it, much like it's done for [shape hover within workspace viewport](https://github.com/penpot/penpot/blob/f448d18d5a2232082128920a9b3542a348cfdbbc/frontend/src/app/main/ui/workspace/viewport/hooks.cljs#L146). Although I have a concern that it may not be performant and seems to be out-of-design.

Would love to hear your take on it.

Let me know if there is anything I would be able to tune up.

andrewzhurov:
Also I'm not sure when to
```
          (dom/prevent-default event)
          (dom/stop-propagation event)
```
Do we need it on on-mouse-leave and on-mouse-enter? (if managing hover state this way is any good, that is :D)

niwinz:
> I have two concerns about the solution:
> 
> 1. It doesn't introduce a uniform 'outline shape on hover over its representation' behaviour, as there is also Design hierarchy in View mode which isn't covered.
>    Perhaps we'd like to cover it as well?

Hmm there are no related US nor specification for view mode, so I don't know. We have a lot of devs on holidays so I can't check it right now but I will raise it when they come back. For now it is ok  as-is.

> 2. Managing 'what is outlined' by mouse-enter and mouse-leave events is an imperative style prone to errors, it would be way more robust to have a pointer position tracked and current hover subject derived out of it, much like it's done for [shape hover within workspace viewport](https://github.com/penpot/penpot/blob/f448d18d5a2232082128920a9b3542a348cfdbbc/frontend/src/app/main/ui/workspace/viewport/hooks.cljs#L146). Although I have a concern that it may not be performant and seems to be out-of-design.

Nah, the viewport approach is very specific becuase it uses an index of all objects for the position. Here the position is clear, is a simple neested list. The mouse-enter/leave looks like a good approach for this case.

> 
> Would love to hear your take on it.
> 
> Let me know if there is anything I would be able to tune up.

Unfortunately today I haven't had time to review this. I have begun to review the oldest PR's and I have found several problems and several conflicts that are difficult to solve and it has taken me longer than expected. Tomorrow I will continue reviewing and merging PR's.

niwinz:
this PR is superseded by #2114 that already includes your commits without changes; as you can observe it also comes with many additional commits, I have made some changes to your code in the posterior commits and I have taken the opportunity to fix other related things. You can look commit by commit to understand the changes.

Thanks, I close this PR because the other already includes your commits

niwinz:
I have completely forgotten, the commits to enter have to have a kind of signature. With your permission, I have added it to your commits on the related MR:

image\

You can more about it on https://github.com/penpot/penpot/blob/develop/CONTRIBUTING.md#developers-certificate-of-origin-dco.

andrewzhurov:
Sounds good. I'll take a look at your PR, thanks for corrections.

Thanks for taking care of signature, it is with my permission indeed.
One thing to note is that it's Andrei Zhurau, I hope it's no biggie.
Sorry for the confusion, it tends to happen with people a lot, a wisely chosen nickname indeed, haha.

PR #2108, opened 2022-07-26
Components v2 (first PR)

hirunatan:
This PR contains:
 * All tasks marked as ready for test in **[COMPONENTS]** stories of Sprint 34.
 * Many helper functions from `comon/pages` moved to new `common/types` module, and add many more.
 * Add scaffolding for writing complex unit tests for `common` subsystem.

By default, components 2.0 behavior is disabled except in development environment. To enable it, open console and type

`features.components_v2()`

(or edit `frontend/src/app/main/features.cljs` and add a `toggle-feature` call).

When a file has been edited with 2.0 features enabled, if you try to open it later with the features disabled, you will get an error.


backend/src/app/rpc/mutations/files.clj
```diff
@@ -110,16 +112,29 @@
...
 (s/def ::set-file-shared
   (s/keys :req-un [::profile-id ::id ::is-shared]))
 
 (sv/defmethod ::set-file-shared
-  [{:keys [pool] :as cfg} {:keys [id profile-id] :as params}]
+  [{:keys [pool] :as cfg} {:keys [id profile-id is-shared] :as params}]
   (db/with-atomic [conn pool]
     (files/check-edition-permissions! conn profile-id id)
+    (when-not is-shared
+        (absorb-library conn params)
```
niwinz: Indentation :fire:

backend/src/app/rpc/mutations/files.clj
```diff
@@ -110,16 +112,29 @@
...
+        (unlink-files conn params))
     (set-file-shared conn params)))
 
+(def sql:unlink-files
+  "delete from file_library_rel
+    where library_file_id = ?")
+
+(defn- unlink-files
+  [conn {:keys [id] :as params}]
+  (db/exec-one! conn [sql:unlink-files id]))
```
niwinz: Can be done with simplier `db/delete!` helper instead of ad-hoc sql in this case.

backend/src/app/rpc/mutations/files.clj
```diff
@@ -146,6 +162,35 @@
...
+                    updated-data (ctf/absorb-assets (:data file) (:data library))]
+
+                (db/update! conn :file
+                            {:revn (inc (:revn file))
+                             :data (blob/encode updated-data)
+                             :modified-at ts}
+                            {:id (:id file)})))]
+
+        (dorun (->> (db/exec! conn [sql:find-files id])
+                    (map process-file)))))))
```
niwinz: The find-files ad-hoc sql query  can be solved with `db/query` helper.

The `dorun` +`map` I think can be expressed using `run!` function if the return value is ignored.

common/src/app/common/geom/shapes.cljc
```diff
@@ -41,11 +41,21 @@
...
+  [shape]
+  ; TODO: perhaps we need to store this calculation in a shape attribute
+  (gpr/points->rect (:points shape)))
+
 (defn left-bound
+  "Returns the lowest x coord of the shape BEFORE applying transformations."
+  ; TODO: perhaps some day we want after transformations, but for the
+  ;       moment it's enough as is now.
   [shape]
   (get shape :x (:x (:selrect shape)))) ; Paths don't have :x attribute
```
niwinz: I think a better approach here would be: 

`(or (:x shape) (:x (:selrect shape)))`

It will avoid executing the second part if the first one success

common/src/app/common/pages/changes.cljc
```diff
@@ -28,15 +34,11 @@
...
-(defn vec-without-nils
-  [coll]
-  (into [] (remove nil?) coll))
-
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Page Transformation Changes
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-;; --- Changes Processing Impl
+;; === Changes Processing Impl
```
niwinz, line 41: The `---` is used in almost all files, there are a reason to change it? I mean it is just for curiosity.

common/src/app/common/pages/changes.cljc
```diff
@@ -494,3 +452,36 @@
...
+          need-sync? (fn [operation]
+                       ; We need to trigger a sync if the shape has changed any
+                       ; attribute that participates in components syncronization.
+                       (and (= (:type operation) :set)
+                            (component-sync-attrs (:attr operation))))
+          any-sync? (some need-sync? operations)]
+      (when any-sync?
+        (into #{} (->> shape-and-parents
+                       (filter #(:main-instance? %)) ; Select shapes that are main component instances
+                       (map :id)))))))
```
niwinz: Use transducers here.
Additionally: there are no need for anonymous function on `(filter #(:main-instance? %))` -> it can be directly `(filter :main-instance?)`

common/src/app/common/types/color.cljc
```diff
@@ -213,11 +236,87 @@
...
+  "Check if the shape uses the given library color."
+  [shape library-id color-id]
+  (let [all-colors (get-all-colors shape)]
+    (some #(and (= (:ref-id %) color-id)
+                (= (:ref-file %) library-id))
+          all-colors)))
+
+(defn- process-shape-colors
+  "Execute an update function on all colors of a shape."
+  [shape func]
```
niwinz: I think a better name can be somethig like `update-fn` instead of `func` that is not very common and not very clear name In my opinion

common/src/app/common/types/color.cljc
```diff
@@ -213,11 +236,87 @@
...
+      (reduce process-stroke $ (d/enumerate (:strokes $)))
+      (reduce process-shadow $ (d/enumerate (:shadow $)))
+      (reduce process-grid $ (d/enumerate (:grids $)))
+      (process-text $))))
+
+(defn remap-colors
+  "Change the shape so that any use of the given color now points to
+  the given library."
+  [shape library-id color]
+  (let [remap-color (fn [shape position shape-color _ attach-fn _]
```
niwinz: can be letfn in this case I guess

common/src/app/common/types/color.cljc
```diff
@@ -213,11 +236,87 @@
...
+                        shape))]
+
+    (process-shape-colors shape remap-color)))
+
 (defn sync-shape-colors
   "Look for usage of any color of the given library inside the shape,
   and, in this case, copy the library color into the shape."
   [shape library-id library-colors]
-  (let [sync-color (fn [shape position shape-color set-fn detach-fn]
+  (let [sync-color (fn [shape position shape-color set-fn _ detach-fn]
```
niwinz: The same here, if let only has function definitions, `letfn` is preferible

common/src/app/common/types/components_list.cljc
```diff
@@ -0,0 +1,38 @@
...
+    (cond-> file-data
+      :always
+      (assoc-in [:components id]
+                {:id id
+                 :name name
+                 :path path
+                 :objects (d/index-by :id shapes)})
+
+      components-v2
+      (update-in [:components id] #(assoc %
```
niwinz: no need for anonymous function here

frontend/src/app/main/data/workspace/libraries.cljs
```diff
@@ -716,6 +756,48 @@
...
+            (->> stream
+                 (rx/filter #(or (= :app.main.data.workspace/finalize-page (ptk/type %))
+                                 (= ::watch-component-changes (ptk/type %)))))
+
+            workspace-data-str
+            (->> (rx/concat
+                  (rx/of nil)
+                  (rx/from-atom refs/workspace-data {:emit-current-value? true})))
+
+            change-str
```
niwinz: I think use of str suffix is confusing, `str` can mean `string`, I think `-s` is a better and shorter one

frontend/src/app/main/data/workspace/libraries.cljs
```diff
@@ -716,6 +756,48 @@
...
+                 (rx/observe-on :async))
+
+            check-changes
+            (fn [[event data]]
+              (let [changes (-> event deref :changes)
+                    components-changed (reduce #(into %1 (ch/components-changed data %2))
+                                               #{}
+                                               changes)]
+                (when (d/not-empty? components-changed)
+                  (apply st/emit!
```
niwinz: I think `run!` is better here than `apply`. I mean

```clojure
(->> components-changed
     (map #(update-component-sync % (:id data)))
     (run! st/emit!))
```

backend/src/app/rpc/mutations/files.clj
hirunatan: done. When all comments are resolved, I'll make a commit with them.

backend/src/app/rpc/mutations/files.clj
hirunatan: done

backend/src/app/rpc/mutations/files.clj
hirunatan: both done

common/src/app/common/geom/shapes.cljc
hirunatan: done

common/src/app/common/pages/changes.cljc
hirunatan: Actually I don't remember well :smile:. I think it's something similar to markdown heading levels. `===` for *changes* and *operations* helpers, and `---` for *media*, *components* and *typographies*, that are kind of second level under *changes*.

But probably a big "fumada mental".

common/src/app/common/pages/changes.cljc

common/src/app/common/types/color.cljc
hirunatan: Changed. I prefer the name `process-fn`, because this is not the tipycal update function, that receives an object and returns it modified.

niwinz: :+1:

common/src/app/common/types/color.cljc
hirunatan: changed

common/src/app/common/types/color.cljc
hirunatan: Don't know much what is the advantage. For me it seems more or less the same.

hirunatan: Changed anyway.

common/src/app/common/types/components_list.cljc
hirunatan: done

frontend/src/app/main/data/workspace/libraries.cljs
hirunatan: Don't like very much the suffix -s, but you are true that -str is normally used for strings. Can't think of one better, so I've changed it to -s.

frontend/src/app/main/data/workspace/libraries.cljs
hirunatan: Perhaps is clearer and more appropriate, yes.

Althoug the pattern `(apply st/emit! (map xxxx))` is used in many other places.

niwinz: Is used in many sites yes, but this is not the optimal one.

PR #2114, opened 2022-07-28
Layer outlines hightlight on hovering

niwinz:
This PR supersedes the https://github.com/penpot/penpot/pull/2107
And contains the original commis by @andrewzhurov and many others, with cosmetic changes to the original code and many additional fixes and improvements to the existing code


frontend/src/app/main/ui/workspace/viewport/outline.cljs
```diff
@@ -90,18 +91,17 @@
...
-                      (filter #(not= edition %))
-                      (map #(get objects %))
-                      (filterv show-outline?)
-                      (filter some?))
-                 ;; outline highlighted shapes even if they are hidden or blocked
-                 (->> highlighted
-                      (filter #(not= edition %))
-                      (map #(get objects %))
-                      (filter some?))))]
+        shapes (d/concat-set
```
andrewzhurov: That's a nice helper! Surprised it's not called 'setcat' :D

Looks more declarative and semantical

frontend/src/app/main/ui/workspace/viewport/outline.cljs
```diff
@@ -76,32 +76,34 @@
...
+                     (->> outlines
+                          (filter #(not= edition %))
+                          (map #(get objects %))
+                          (filterv show-outline?)
+                          (filter some?))
+                     ;; outline highlighted shapes even if they are hidden or blocked
+                     (->> highlighted
+                          (filter #(not= edition %))
+                          (map #(get objects %))
+                          (filter some?)))]
```
andrewzhurov: Seems way easier to parse visually

frontend/src/app/main/ui/workspace/viewport.cljs
```diff
@@ -53,6 +53,7 @@
                 panning
                 selrect
                 transform
+                highlighted
```
andrewzhurov, line 56: That's handy!

frontend/src/app/main/ui/workspace/viewport/outline.cljs
```diff
@@ -92,18 +92,17 @@
...
-                     (->> highlighted
-                          (filter #(not= edition %))
-                          (map #(get objects %))
-                          (filter some?)))]
+        lookup      (d/getf objects)
+        edition?    (fn [o] (= edition o))
+
+        shapes      (-> #{}
+                        (into (comp (remove edition?)
+                                    (keep lookup)
```
andrewzhurov, line 100: Wow, I didn't know this fn existed
I like how it spares the need to `(remove nil?)` after a `(map lookup)`, more concise

frontend/src/app/main/ui/workspace/viewport/outline.cljs
```diff
@@ -92,18 +92,17 @@
...
+                                    (filter show-outline?))
+                              (set/union selected hover))
+                        (into (comp (remove edition?)
+                                    (keep lookup))
+                              highlighted))]
```
andrewzhurov, line 105: Albeit a more imperative approach than d/concat-set, surprisingly looks pretty readable. Perhaps due to raising a level of abstraction by giving anonymous functions names

frontend/src/app/main/ui/workspace/sidebar/layers.cljs
```diff
@@ -84,101 +84,114 @@
...
         select-shape
-        (fn [event]
-          (dom/prevent-default event)
-          (reset! scroll-to-middle? false)
-          (let [id (:id item)]
-            (cond
-              (kbd/shift? event)
-              (st/emit! (dw/shift-select-shapes id))
+        (mf/use-fn
+         (mf/deps id)
```
andrewzhurov, line 134: Won't use-fn have the same result if we omit this expr?

frontend/src/app/main/ui/workspace/viewport/outline.cljs
niwinz: Hmm, I think you are pretty right on this! Now that you have mentioned the name, it looks more interesting that the current one hehe. I will propose/do the change the next week If I have "free" time hehe.

frontend/src/app/main/ui/workspace/sidebar/layers.cljs
```diff
@@ -84,101 +84,114 @@
...
 
         toggle-collapse
-        (fn [event]
-          (dom/stop-propagation event)
-          (if (and expanded? (kbd/shift? event))
-            (st/emit! (dwc/collapse-all))
-            (st/emit! (dwc/toggle-collapse id))))
+        (mf/use-fn
+         (mf/deps expanded?)
+         (fn [event]
```
andrewzhurov, line 107: Is it equivalent to
```
(mf/with-memo [expanded?]
  (fn [event]
    ...
```
?

frontend/src/app/main/ui/workspace/viewport/outline.cljs
niwinz: yep, this is probably the main reason, sometimes giving a good alias for a posible annonymous function makes thins more easy readable on first & quick look

frontend/src/app/main/ui/workspace/sidebar/layers.cljs
andrewzhurov: Since useCallback(fn, deps) is equivalent to useMemo(() => fn, deps)
[source](https://hyp.is/1RVjig8GEe2WoHvn4o74fg/reactjs.org/docs/hooks-reference.html)

frontend/src/app/main/ui/workspace/sidebar/layers.cljs
niwinz: `use-fn` in this case is the React.useCallback hook, it allows having stable reference to the function callback instance upon all renders while the id not changes. This makes that on VDOM diffing it does not need to remove the old callback from the real dom and attach a new one. On normal code it can be really irrelevant, but in code that renders a lot and is in the "hot path" every millisecond counts and this helps on this.

frontend/src/app/main/ui/workspace/sidebar/layers.cljs
niwinz: Yep, in fact the `use-fn` (the React.useCallback) is a specialization of React.useMemo (penpot's mf/use-memo or mf/with-memo macro)

frontend/src/app/main/ui/workspace/context_menu.cljs
```diff
@@ -70,9 +70,10 @@
              (when (and (some? dom) (some? submenu-node))
                (dom/set-css-property! submenu-node "top" (str (.-offsetTop dom) "px"))))))]
 
-    (when on-unmount
-      (mf/use-effect
-       (fn [] on-unmount)))
+
+    (mf/use-effect
+     (mf/deps on-unmount)
+     (constantly on-unmount))
```
andrewzhurov, line 77: Good catch! Thanks.:)

frontend/src/app/main/ui/workspace/context_menu.cljs
```diff
@@ -448,14 +458,11 @@
                     :on-click do-delete}]))
 
 (mf/defc shape-context-menu
+  {::mf/wrap [mf/memo]}
   [{:keys [mdata] :as props}]
   (let [{:keys [disable-booleans? disable-flatten?]} mdata
         shapes (mf/deref refs/selected-objects)
-        hover-ids (mf/deref refs/current-hover-ids)
-        hover-objs (mf/deref (refs/objects-by-id hover-ids))
```
andrewzhurov, line 442: It's been puzzling me why this line causes a rerender for a while now, it seems to return the same value on each render, yet it reruns again and again..

CHANGES.md
```diff
@@ -5,12 +5,15 @@
...
 
-- Add some cosmetic changes in viewer mode [Taiga #3688](https://tree.taiga.io/project/penpot/us/3688)
+- Add cosmetic changes in viewer mode [Taiga #3688](https://tree.taiga.io/project/penpot/us/3688)
+- Outline highlights on layer hovering [Taiga #2645](https://tree.taiga.io/project/penpot/us/2645) by @andrewzhurov
 
 ### :bug: Bugs fixed
 ### :arrow_up: Deps updates
 ### :heart: Community contributions by (Thank you!)
 
+- To @andrewzhurov for many code contributions on this release.
```
andrewzhurov, line 15: ^.^

frontend/src/app/main/ui/workspace/sidebar/layers.cljs
andrewzhurov: Thanks for giving details about performance, I see how memo is of use.:)
I was wondering, could we omit (mf/deps id), having it as
```
(mf/use-fn
  (fn [event]
    ...
```
to the same extent?

andrewzhurov: The rational behind it is that we don't list `id` in other memos, since it does not change, so if we are to stay uniform perhaps here we'd have empty deps or, rather, absent deps

frontend/src/app/main/ui/workspace/context_menu.cljs
niwinz: It caused rerender bucause refs/objects-by-id is always generating a new ref. Is a bit confusing, but we don't need to use hooked version of deref all the times. We don't need always need reactivity on values that we know that will not be changed.

frontend/src/app/main/ui/workspace/sidebar/layers.cljs
niwinz: The id dep can be omited if we ensure when layer-item component is always used with id as :key,  but this is a error prone asumption and in this case is better provide the hook with deps.

PR #2122, opened 2022-08-01
:bug: Fix layers get out of their group when moved

andrewzhurov:
https://user-images.githubusercontent.com/13056889/182096235-c6b77f9d-011e-4f73-bfb2-0cfcae225a81.mp4

Am I breaking something by removing that function?
I tried to get my head around of what can fall out and fell short, but there got to be something..

Closes https://tree.taiga.io/project/penpot/issue/3932

### DIFF
--- a/frontend/src/app/main/data/workspace/transforms.cljs
+++ b/frontend/src/app/main/data/workspace/transforms.cljs
@@ -736,14 +736,6 @@
         (rx/of (set-modifiers [id] {:displacement displ} false true)
                (apply-modifiers))))))
 
-(defn check-frame-move?
-  [target-frame-id objects position shape]
-
-  (let [current-frame (get objects (:frame-id shape))]
-    ;; If the current frame contains the point and it's a child of the target
-    (and (gsh/has-point? current-frame position)
-         (cph/is-child? objects target-frame-id (:id current-frame)))))
-
 (defn- calculate-frame-for-move
   [ids]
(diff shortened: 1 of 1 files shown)


niwinz:
Hmm I think the current approach is broken. I'm not clearly understand why is done in this manner.
After testing your commit, it almost fixes, but there are case where it still happens:

https://user-images.githubusercontent.com/843689/182167057-ce0efd16-4b56-4e1c-a209-fb8d284182d0.mp4

(when the grouped shape is moved outside the frame... i don't know if this is expected or not, so I need to check it with the design/ux team tomorrow).

In any case I'm not feel very sure about all this, because I don't understand enough why this is implemented in this manner. I tell you more tomorrow.

andrewzhurov:
That's a good case, I wouldn't be sure either.
I've taken a look at how Figma does it, as I assume we would like to have a similar behaviour, and it moves shape out of its group

https://user-images.githubusercontent.com/13056889/182191808-38a70ed8-608d-4799-9420-a753f15f461a.mp4

niwinz:
Yep, I have confirmed it with the designers, and your solution is correct.
So I will merge your PR with some additional changes to the surrounding code today.

niwinz:
This PR is superseded by https://github.com/penpot/penpot/pull/2131
Your commit is included there with other changes to the affected code.
Thansk  :heart:

PR #2123, opened 2022-08-01
:bug: Fix svg upload

andrewzhurov:
SVGs were failing to upload
image

Thus not appearing in the Assets
image
As reported in https://tree.taiga.io/project/penpot/issue/3885

After the fix it's back to normal
image

Here is the director's cut of the upload process comparison

https://user-images.githubusercontent.com/13056889/182109918-c08e9366-0809-4589-9b7f-597324c9c719.mp4


niwinz:
Works as expected.

Thanks for the fix :heart:

niwinz:
Also, backported to staging branch.

PR #2129, opened 2022-08-02
:sparkles: Select all inside a group select only the objects at this …

pabloalba:
Resolves Ctrl+A inside a group should select only the objects at this group level

### DIFF
--- a/frontend/src/app/main/data/workspace/selection.cljs
+++ b/frontend/src/app/main/data/workspace/selection.cljs
@@ -178,29 +178,56 @@
       (let [objects (wsh/lookup-page-objects state)]
         (rx/of (dwc/expand-all-parents ids objects))))))
 
+
+
+(defn- select-siblings
+  [state parent]
+  (let [children (wsh/lookup-shapes state (:shapes parent))
+        selected (into (d/ordered-set)
+                       (comp (remove :blocked) (map :id))
+                       children)]
+    (rx/of (select-shapes selected))))
+
+(defn- select-all-frame
+  [state]
+  (let [focus (:workspace-focus-selected state)
+        objects  (-> (wsh/lookup-page-objects state)
+                     (cp/focus-objects focus))
+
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/data/workspace/selection.cljs
```diff
@@ -178,29 +178,58 @@
...
-
-            selected (into (d/ordered-set)
-                           (comp (remove :blocked) (map :id))
-                           selected)]
-
-        (rx/of (select-shapes selected))))))
+           (let [current-selection-parents (->> (wsh/lookup-selected state)
+                                                (wsh/lookup-shapes state)
+                                                (map :parent-id)
+                                                distinct)
```
niwinz: I think this can be better expressed in this way:

```clojure
 (->> (wsh/lookup-selected state)
      (wsh/lookup-shapes state)
      (into #{} (map :parent-id)))
```

Because this will generate a set, and is countable, the map + disctinct generates a seq where the count scales O(n)

frontend/src/app/main/data/workspace/selection.cljs
```diff
@@ -178,29 +178,58 @@
...
+                 num-parents (count current-selection-parents)
+                 parent (when (= num-parents 1)
+                          (wsh/lookup-shape state (first current-selection-parents)))
+                 is-frame (= (:type parent) :frame)]
+             
+             (case num-parents
+               0 (select-all-frame state)
+               1 (if is-frame
```
niwinz: I think this can be replaced by directly call to `cph/is-frame-shape?` predicate

frontend/src/app/main/data/workspace/selection.cljs
pabloalba: Changed

frontend/src/app/main/data/workspace/selection.cljs
pabloalba: Changed

PR #2142, opened 2022-08-04
:recycle: Refactor session management

niwinz:
This PR refactors how session is managed. 

I have replaced the previous session updated mechanism that uses a "dedicated thread loop" for updating all sessions of received requests with a more efficient one that just does on demand on user requests.

The new approach consists on setting a default expiration of 7 days to all sessions. If you enter the application every day, the session will be transparently renewed on some random request from the user. If user stays for 7 days not connection, the session will expire without renovation and then deleted from the table by the GC.

This also solves the problem that causes [session removal after browser closing](https://tree.taiga.io/project/penpot/issue/3937) because previous cookies without expiration are identified by browser as "session" cookie and all session cookies are removed on browser close. Now with proper expiration and transparent token renovation, the user experience will improve considerably.

### DIFF
--- a/backend/src/app/http/session.clj
+++ b/backend/src/app/http/session.clj
@@ -7,33 +7,35 @@
 (ns app.http.session
   (:require
    [app.common.data :as d]
-   [app.common.exceptions :as ex]
    [app.common.logging :as l]
-   [app.config :as cfg]
+   [app.config :as cf]
    [app.db :as db]
    [app.db.sql :as sql]
-   [app.metrics :as mtx]
-   [app.util.async :as aa]
    [app.util.time :as dt]
    [app.worker :as wrk]
-   [clojure.core.async :as a]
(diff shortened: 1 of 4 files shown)


backend/src/app/http/session.clj
```diff
@@ -7,33 +7,35 @@
...
-(def authenticated-cookie-name "authenticated")
+;; A cookie that we can use to check from other sites of the same
+;; domain if a user is authenticated.
+(def default-authenticated-cookie-name "authenticated")
+
+;; Default value for cookie max-age
+(def default-cookie-max-age (dt/duration {:days 7}))
+
+;; Default age for automatic session reneval
+(def default-reneval-max-age (dt/duration {:hours 6}))
```
superalex: I think it should be renewal

niwinz: Fixed

PR #2145, opened 2022-08-05
:tada: Add Libraries & Templates carousel

pabloalba:
Related to https://tree.taiga.io/project/penpot/us/3860


backend/kk
```diff
@@ -0,0 +1,2 @@
+sleep 40
```
superalex: KK?

PR #2148, opened 2022-08-09
:tada: Add zoom style to share link url

EvaMarco:
This PR solves https://tree.taiga.io/project/penpot/us/3874


frontend/src/app/main/ui/share_link.cljs
```diff
@@ -37,20 +37,21 @@
    ::mf/register-as :share-link}
   [{:keys [file page]}]
   (let [current-page page
-        slinks   (mf/deref refs/share-links)
-        router   (mf/deref refs/router)
-        route    (mf/deref refs/route)
+        slinks    (mf/deref refs/share-links)
+        router    (mf/deref refs/router)
+        route     (mf/deref refs/route)
+        zoom-type (mf/deref refs/viewer-zoom-type)
```
niwinz, line 43: Accessing viewer zoom-type from generic share-link component also looks counterintuitive.

frontend/src/app/main/data/viewer.cljs
```diff
@@ -196,32 +204,39 @@
     ptk/UpdateEvent
     (update [_ state]
       (let [increase #(min (* % 1.3) 200)]
-        (update-in state [:viewer-local :zoom] (fnil increase 1))))))
+        (-> state
+            (update-in state [:viewer-local :zoom] (fnil increase 1))
+            (assoc-in  [:viewer-local :zoom-type] :default))))))
```
niwinz: I'm not clearly understand the need to assign the zoom-type here and in the following events.

niwinz: In fact, looking the code, the zoom-type prop is not used in any view, so I think we can just :fire: it

PR #2182, opened 2022-08-23
Viewer performance issues

niwinz:


frontend/src/app/main/ui.cljs
```diff
@@ -29,6 +29,7 @@
 (mf/defc on-main-error
   [{:keys [error] :as props}]
   (mf/with-effect
+    (js/console.log error)
```
hirunatan, line 32: perhaps console.error?

niwinz: Yep, but this is debug only so no importance here.

Issue #2167, opened 2022-08-16
Fixed objects jitter while scrolling


ghost:
**To Reproduce**

1.  Turn on "Fix when scrolling" for some objects
2.  Start the prototype
3.  Start scrolling

**Expected behavior**

The object stays still and doesn't move or jitter at all

**Actual behavior**

The object jitters while scrolling. I also noticed that when the developer tools are open on the right, objects no longer stutter
 
**Screenshots**

https://user-images.githubusercontent.com/56515875/184787978-6b070ab0-1bed-42b6-9281-1b98cc19f6b1.mp4
(the bar at the top and the blue rectangle are the fixed objects)

**Desktop:**
- OS: KDE neon 5.25
- Browser & version: Firefox 103.0 64-bit

**Environment:**
- Host: design.penpot.app

myfunnyandy:
Definitely a bug. I've added it to our issue tracker: https://tree.taiga.io/project/penpot/issue/3989. We will try to fix it soon. 

Thanks for reporting it!

superalex:
Hi @logicalwillow 

This one should be fixed with the last release, could you please check if it's still happening?

ghost:
I still see it happening, especially when using the scroll wheel instead of the scrollbar

superalex:
Thanks @logicalwillow, now in my case it's only happening in firefox, could you please check if it jitters too on other browsers?

ghost:
Sorry, I don't have access to any other browsers.

PR #2205, opened 2022-08-30
:bug: Fix jittering on firefox scroll fixed elements

superalex:
Related to:
- https://github.com/penpot/penpot/issues/2167
- https://tree.taiga.io/project/penpot/issue/3989


niwinz:
looks like the file with `normalize-wheel` namespace is missing..
also I think it should be under dom... I mean `app.util.dom.normalize-wheel`

superalex:
> looks like the file with `normalize-wheel` namespace is missing.. also I think it should be under dom... I mean `app.util.dom.normalize-wheel`

That's exactly where it is :) https://github.com/penpot/penpot/tree/develop/frontend/src/app/util/dom/normalize_wheel.js

PR #2210, opened 2022-08-30
:bug: Fix base uri in repo call

Windfarer:
When we use "import Penpot files", the calling api's base uri is still the original one, not the value we set in `penpotPublicURI`.

I'm not familiar with clojurescript, but I found that the private var `base-uri`'s operand maybe evaluate before the correct public uri to be set. So in this PR, I use the get-public-uri as the value of base-uri and call it wherever it is used to get the right uri

I'm not sure if it is a correct fix, but this can fix the import files issue. And I'm happy to know if you have a better solution.

### DIFF
--- a/frontend/src/app/main/repo.cljs
+++ b/frontend/src/app/main/repo.cljs
@@ -43,7 +43,7 @@
                :status status
                :data body})))
 
-(def ^:private base-uri cf/public-uri)
+(def ^:private base-uri cf/get-public-uri)
 
 (defn- send-query!
   "A simple helper for send and receive transit data on the penpot
@@ -56,7 +56,7 @@
                           http/conditional-error-decode-transit
                           http/conditional-decode-transit)]
      (->> (http/send! {:method :get
-                       :uri (u/join base-uri "api/rpc/query/" (name id))
+                       :uri (u/join (base-uri) "api/rpc/query/" (name id))
                        :credentials "include"
                        :query params})
           (rx/map decode-transit)
@@ -66,8 +66,9 @@
   "A simple helper for a common case of sending and receiving transit
   data to the penpot mutation api."
   [id params]
+  (log/debug :hint "send-mutation" (base-uri))
   (->> (http/send! {:method :post
-                    :uri (u/join base-uri "api/rpc/mutation/" (name id))
+                    :uri (u/join (base-uri) "api/rpc/mutation/" (name id))
                     :credentials "include"
                     :body (http/transit-data params)})
        (rx/map http/conditional-decode-transit)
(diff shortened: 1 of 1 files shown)


niwinz:
i will look on it tomorrow, thanks \o/

niwinz:
After looking on it, I think this change is not necessary... the configuration is loaded before all application code is loaded. So when application is loaded it should correctly use the `window.penpotPublicURI` variable defined in the `js/config.js` file... so we don't need to call the get-public-uri function each time we make a request.

Can you describe the problem you have? because I have tested it setting a custom value on the config and everything is working as expected (the request all goes to the new destination) without needing change any line of code.

Windfarer:
Let me describe the problem in detail.

I deploy the penpot stack using https://github.com/penpot/penpot/blob/develop/docker/images/docker-compose.yaml this docker-compose file, and set the image tag version to `1.15.0-beta`

Then I set `PENPOT_PUBLIC_URI=http://localhost/penpot/` in config.env  

Mount a config.js file into frontend container, the content is 

```
// Frontend configuration
//var penpotFlags = "";
var penpotPublicURI = "http://localhost/penpot/"
```

After above modifications, I start this compose. And use another nginx as a reverse proxy to serve penpot. The config is following:

```
map $http_upgrade $connection_upgrade {
    default upgrade;
    '' close;
}

server {
    listen       80;

    location /penpot {
    rewrite ^/penpot(/.*)$ $1 break;
    proxy_pass http://192.168.1.4:9001; # this is my host ip
        proxy_redirect                  off;
        proxy_buffers                   32 16k;
        proxy_busy_buffers_size         64k;
        proxy_cache                     off;
	proxy_http_version 1.1;
         proxy_set_header Connection $connection_upgrade;
        }
  }
```
To serve penpot frontend and backend on the /penpot/ path.

It is ok to use most of the functions in penpot. The frontend assets and api call will be successful under /penpot/ path. But there's a wired problem occurred when I try to upload a penpot file. The api call to `/api/rpc/mutation/create-temp-file` is not under the /penpot/ path.

There are some related screenshots:
When I entered the penpot UI, the console output the correct public-ui value: `http://localhost/penpot/` and most of the functions work well:
Screenshot from 2022-09-01 23-41-46


The function I got this problem:
Screenshot from 2022-09-01 23-42-18

When I selected the file and hit the continue button, the console printed the following error:
Screenshot from 2022-09-01 23-44-58

The network shows that it call that api to the wrong public-api to `/api/rpc/mutation/create-temp-file` directly, and got a 404.
Screenshot from 2022-09-01 23-45-26

So after that, I try to modify the source code and found that if call the get-public-api like this PR, this create-temp-file request will use the correct public-uri.

I have no idea why this happened, and guess that maybe a issue about code execution order.

Hope I've described this issue clearly, maybe you could try as mine process to reproduce this issue. If you have any further questions, please feel free to contact me.

niwinz:
Now I understand the problem. In general I should say penpot is not tested to work on sub-directory, so this bugs can happen.
I will look on this because you say that almost all works with the exception of the import.

The import process is done on worker, so this can be the cause, I will look on it...

niwinz:
I have identified the problem and fixed it on https://github.com/penpot/penpot/pull/2223
The PR includes many other changes for a general improvements on how public-uri is handled but the real fix consists on properly setting the public-uri on the worker.

niwinz:
thanks for digging into it, and letting us know about the problem

PR #2214, opened 2022-08-31
:bug: Fix undo on delete page does not preserve its order

pabloalba:
Fixes https://tree.taiga.io/project/penpot/issue/3375

### DIFF
--- a/common/src/app/common/types/pages_list.cljc
+++ b/common/src/app/common/types/pages_list.cljc
@@ -6,7 +6,8 @@
 
 (ns app.common.types.pages-list
   (:require
-   [app.common.data :as d]))
+   [app.common.data :as d]
+   [app.common.pages.helpers :as cph]))
 
 (defn get-page
   [file-data id]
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/data/workspace.cljs
```diff
@@ -326,7 +326,10 @@
   (ptk/reify ::delete-page
     ptk/WatchEvent
     (watch [it state _]
-      (let [page (get-in state [:workspace-data :pages-index id])
+      (let [pages (get-in state [:workspace-data :pages])
+            page-num (d/index-of pages id)
+            page (get-in state [:workspace-data :pages-index id])
+            page (assoc page :page-num page-num)
```
niwinz: I think is better call it `:index` on page data sturcture

common/src/app/common/types/pages_list.cljc
```diff
@@ -6,22 +6,27 @@
...
+        page (dissoc page :page-num)
+
+        ; It's legitimate to add a page that is already there,
         ; for example in an idempotent changes operation.
-        conj-if-not-exists (fn [pages id]
-                             (cond-> pages
-                               (not (d/seek #(= % id) pages))
-                               (conj id)))]
+        add-if-not-exists (fn [pages id]
+                            (when (not (d/seek #(= % id) pages))
```
niwinz: the `when` returns nil if condition evaluates to false, I think this is not expected because we should not assoc nil to pages on file-data

pabloalba: fixed

frontend/src/app/main/data/workspace.cljs
pabloalba: fixed

PR #2232, opened 2022-09-06
:sparkles: Ungroup frames

Alotor:
- Ungroup frames [Taiga #4012](https://tree.taiga.io/project/penpot/us/4012)

### DIFF
--- a/frontend/src/app/main/data/workspace/groups.cljs
+++ b/frontend/src/app/main/data/workspace/groups.cljs
@@ -94,7 +94,7 @@
 
     [group changes]))
 
-(defn prepare-remove-group
+(defn remove-group-changes
   [it page-id group objects]
   (let [children  (mapv #(get objects %) (:shapes group))
         parent-id (cph/get-parent-id objects (:id group))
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/data/workspace/groups.cljs
```diff
@@ -149,17 +162,22 @@
...
+
+            prepare
+            (fn [shape-id]
+              (let [shape (get objects shape-id)]
+                (cond
+                  (or (cph/group-shape? shape) (cph/bool-shape? shape))
+                  [(remove-group-changes it page-id shape objects)]
+
+                  (cph/frame-shape? shape)
+                  [(remove-frame-changes it page-id shape objects)])))
```
niwinz: This function returns always a vector of single element, and then... `mapcat`. I think it can return the value as-is and then, use `map` instead `mapcat`

frontend/src/app/main/ui/workspace/context_menu.cljs
```diff
@@ -213,6 +213,7 @@
         single?   (= (count shapes) 1)
         do-create-artboard-from-selection #(st/emit! (dw/create-artboard-from-selection))
 
+        has-frame? (->> shapes (d/seek #(= :frame (:type %))))
         has-group? (->> shapes (d/seek #(= :group (:type %))))
         has-bool? (->> shapes (d/seek #(= :bool (:type %))))
```
niwinz: All this anonymous callbacks I think should be replaced with the named predicate defined on common.page.helpers.

frontend/src/app/main/data/workspace/groups.cljs
Alotor: Not exactly, there is the omitted default `nil` when it won't match any conditions. When it's nil the mapcat will not modify the resulting list.

niwinz: yep, in this case instead of `map` you can use `keep` :P

Alotor: I'll remove the mapcat for a keep

PR #2275, opened 2022-09-16
:bug: Fix doubled quotes in frontend config

Lazalatin:
Docker parses environment variables literally, delivering quoted flags in the $PENPOT_FLAGS variable. This in turn leads to doubled quotes in the resulting config.js in front and after the flags, omitting them completely.

This PR fixes this behaviour.

### DIFF
--- a/docker/images/config.env
+++ b/docker/images/config.env
@@ -10,7 +10,7 @@ PENPOT_TENANT=pro
 
 ## Feature flags.
 
-PENPOT_FLAGS="enable-registration enable-login"
+PENPOT_FLAGS=enable-registration enable-login
 
 ## Temporal workaround because of bad builtin default
 


Lazalatin:
While my description says all about the change, it may be helpful to know how I came to it:

If using the `config.env` in its old state no unusual behaviour can be seen at first sight. This changes if one tries to disable self-registration or the username/password-based login in favor of i.e. gitlab social login as the only option. Options like `disable-login` do not change the view of the frontend. This is where my investigation started.

I used the `docker-compose.yml` and `config.env` from the repository.

Hope this helps with reconstructing the issue I had which lead to this PR.

niwinz:
Hello @Lazalatin 
Thanks for reaching us.

I'm still does not understand why this happens to somebody but I was unable to reproduce it.

This is the content of the js/config.js file starting the docker images with quoted flags (the current default):
image

Everything is working as expected, so I guess there are some differences in our environment that probably can explain this.
Are you running it under `docker-compose` or `docker compose` (i mean v2)? which versions are you using, and which base operating system are you using.

From my side is:

```bash
~❯ docker compose version                                                                                                                                                                                      
Docker Compose version v2.10.2
~❯ docker --version                                                                                                                                                                                                   
Docker version 20.10.17, build 100c701
~❯ lsb_release -a                                                                                                                                                                                                     14:03:00
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 11 (bullseye)
Release:	11
Codename:	bullseye
```

I don't want to change the default approach unless there are a good reason to do it. For now it works as expected on our default environment, but there are environments that does not handle correctly the environment variables.

niwinz:
I can confirm myself that if you are using the old v1 `docker-compose`, the environment variables are not handled correctly:
image

So from our side we recommend upgrade your docker-compose to use the `v2`, and the issue will be solved.

niwinz:
I closing this MR for inactivity, the problem as I said previously is because of use of the the legacy docker-compose.

Lazalatin:
@niwinz Please reopen this MR.

As legacy docker compose is still shipped with LTS distros as ubuntu 20.04 (Which has support until Q1 2025!) I would recommend to fix the issue using my MR. 

It does not hurt any compose v2+ users but hurts anyone relying on older ubuntu LTS and other slow progressing distros. Maybe you can use your telemetry data to evaluate if this is an issue which might affect a critical number of penpot users.

PS: I have many private issues on my plate right now so future answers may take until 4 weeks from now. In regards of "inactivity".

niwinz:
Agree, didn't considered this argument, so I will merge it. I will resolve the conflict myself. Thanks

Lazalatin:
Thank you for reconsidering! 👍🏻

niwinz:
I have added your commit on my enhancements branch #2346 thanks again

PR #2277, opened 2022-09-16
Fix WOFF fonts disabled in font file selection dialog

wojtekmaj:

Looks like `font/woff` mimetype is indeed recommended for WOFF fonts, but e.g. Microsoft Edge uses `application/font-woff`, making it impossible to select WOFF fonts  in file selection dialog.

> Screenshots/GIFs

Before:

Zrzut ekranu 2022-09-16 o 23 07 21

After:

Zrzut ekranu 2022-09-16 o 23 07 45

### DIFF
--- a/common/src/app/common/media.cljc
+++ b/common/src/app/common/media.cljc
@@ -9,7 +9,7 @@
    [clojure.spec.alpha :as s]
    [cuerdas.core :as str]))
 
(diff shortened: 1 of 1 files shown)


niwinz:
Nice catch, thank you :heart: 
I have rebased and merged it using the command line.

In order for us to merge directly from github, we need the commits to follow a formatting policy (which you can find in the CONTRIBUTING.md file)

Thanks

PR #2328, opened 2022-09-23
:bug: Fix inviting to non existing users can fail

pabloalba:
Fixes https://tree.taiga.io/project/penpot/issue/4108


niwinz:
I'm not clearly understand which fail case this PR fixes.
It also conflicts with my ongoing work related to profile blocking, email verification and accepting invitation in my branch; so I prefer firstly understand this and then port this changes yo my branch instead of merging and later have the merge of the death :P

PR #2368, opened 2022-09-28
Fix(frontend/comments): Change word-break on comments to break-word

lol768:
**Before**:

image

Problems: Hard to read, no dashes to indicate that words have been broken across lines. 

**After**:

image

Much easier to read, container overflows in the X direction for super long unbreakable words, which I think is preferable to the original behaviour.


superalex:
Merged here -> https://github.com/penpot/penpot/commit/4a3a18140318857436af016b0dbc4befe30d88e7

Thanks!

PR #2379, opened 2022-09-29
:sparkles: Add garbage collect of deleted components

hirunatan:
https://tree.taiga.io/project/penpot/task/4042

### DIFF
--- a/backend/src/app/tasks/file_gc.clj
+++ b/backend/src/app/tasks/file_gc.clj
@@ -13,6 +13,7 @@
    [app.common.data :as d]
    [app.common.logging :as l]
    [app.common.pages.migrations :as pmg]
+   [app.common.types.file :as ctf]
    [app.common.types.shape-tree :as ctt]
    [app.config :as cf]
    [app.db :as db]
@@ -45,7 +46,8 @@
   (fn [params]
     (db/with-atomic [conn pool]
       (let [min-age (or (:min-age params) (:min-age cfg))
-            cfg     (assoc cfg :min-age min-age :conn conn)]
+            id      (:id params)
+            cfg     (assoc cfg :min-age min-age :conn conn :id id)]
         (loop [total 0
                files (retrieve-candidates cfg)]
           (if-let [file (first files)]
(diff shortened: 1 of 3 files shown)


backend/src/app/tasks/file_gc.clj
```diff
@@ -162,7 +164,73 @@
...
+                 :kf first
+                 :initk (dt/now))))
+
+(defn- clean-deleted-components!
+  "Performs the garbage collection of unreferenced deleted components."
+  [conn library-id library-data]
+  (let [find-used-components-file
+        (fn [components file-data]
+          ; Find what of the components are used in the file.
+          (d/filterm #(ctf/used-in? file-data library-id (second %) :component)
```
niwinz: I think you don't need to use `d/filterm` here (see other related comments)

backend/src/app/tasks/file_gc.clj
```diff
@@ -162,7 +164,73 @@
...
+        find-used-components
+        (fn [components files-data]
+          ; Find what components are used in any of the files.
+          (loop [files-data      files-data
+                 components      components
+                 used-components {}]
+            (let [file-data (first files-data)]
+              (if (or (nil? file-data) (empty? components))
+                used-components
+                (let [used-components-file (find-used-components-file components file-data)]
```
niwinz, line 213: This probably need to be a set, because two lines below you are using `contains?` over it.

backend/src/app/tasks/file_gc.clj
```diff
@@ -162,7 +164,73 @@
...
+                  (recur (rest files-data)
+                         (d/filterm #(not (contains? used-components-file (:id %))) components)
```
niwinz: `contains?` over a vector does not works as you probably expect:
image

niwinz: And, don't use `d/filterm` prematurely, I mean, proceed all with lazy version and convert to the final data structure on before return.

backend/src/app/tasks/file_gc.clj
```diff
@@ -45,7 +46,8 @@
   (fn [params]
     (db/with-atomic [conn pool]
       (let [min-age (or (:min-age params) (:min-age cfg))
-            cfg     (assoc cfg :min-age min-age :conn conn)]
+            id      (:id params)
```
niwinz: can this id be destructured directly in params?

backend/src/app/tasks/file_gc.clj
```diff
@@ -162,7 +164,73 @@
...
+    limit 1")
+
+(defn- retrieve-client-files
+  "search al files that use the given library.
+   Returns a sequence of file-data (only reads database rows one by one)."
+  [conn library-id]
+  (let [get-chunk (fn [cursor]
+                    (let [rows (db/exec! conn [sql:retrieve-client-files library-id cursor])]
+                      [(some-> rows peek :modified-at)
+                       (map #(->> % :data blob/decode) rows)]))]
```
niwinz: this annonymous function probably can be expressed with `comp`:

`(comp blob/decode :data)`

backend/src/app/tasks/file_gc.clj
hirunatan: ok

backend/src/app/tasks/file_gc.clj
hirunatan: done

backend/src/app/tasks/file_gc.clj
hirunatan: done

backend/src/app/tasks/file_gc.clj
hirunatan: done

backend/src/app/tasks/file_gc.clj
hirunatan: done

PR #2386, opened 2022-09-30
:sparkles: Newsletter suscription

pabloalba:
Related with https://tree.taiga.io/project/penpot/us/3242


niwinz:
This MR has conflict.
Also: is backward compatibility here necessary? (I mean we probably already have some data stored with old attrs, this data will be lost, it is expected?)

PR #2387, opened 2022-09-30
:bug: Show new dashboard only in v2, and fix loading screen

hirunatan:
https://tree.taiga.io/project/penpot/task/4260

### DIFF
--- a/frontend/src/app/main/ui/dashboard/grid.cljs
+++ b/frontend/src/app/main/ui/dashboard/grid.cljs
@@ -355,8 +355,7 @@
                      (dnd/has-type? e "application/x-moz-file"))
              (dom/prevent-default e)
              (reset! dragging? false)
-             (import-files (.-files (.-dataTransfer e))))))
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/ui/dashboard/grid.cljs
```diff
@@ -34,7 +34,7 @@
    [cuerdas.core :as str]
    [rumext.v2 :as mf]))
 
-(log/set-level! :info)
+(log/set-level! :warn)
```
niwinz: `:info` was there for properly logging the thumbnails loading, I need it activated once deployed.

niwinz: we need a way to overwrite it externally but right now we don't have this option

hirunatan: wtf?

well, if so, we need to leave it in info.

PR #2413, opened 2022-10-06
:bug: Fix paste layer is not visible

superalex:
- Related to https://tree.taiga.io/project/penpot/issue/4283
- Related to https://tree.taiga.io/project/penpot/issue/4299

### DIFF
--- a/frontend/src/app/main/data/workspace.cljs
+++ b/frontend/src/app/main/data/workspace.cljs
@@ -1162,13 +1162,14 @@
                 res)))
 
           (maybe-translate [shape objects selected+children]
-            (if (and (not= (:type shape) :frame)
-                     (not (contains? selected+children (:frame-id shape))))
-              ;; When the parent frame is not selected we change to relative
-              ;; coordinates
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/data/workspace.cljs
```diff
@@ -1329,49 +1330,47 @@
...
-                      ;; Pasted objects mustn't exceed the selected frame y limit
-                      paste-y (if (> (+ (:height wrapper) (:y1 wrapper)) (:height frame-object))
-                                (+ (- (:y frame-object) (:y orig-pos)) (- (:height frame-object) (:height wrapper) margin-y))
-                                (:y frame-object))
-
-                      delta    (gpt/point paste-x paste-y)]
-
-                  [frame-id frame-id delta])
+                      origin-frame-object (get page-objects origin-frame-id)]
+                  (cond
```
Alotor: I don't like much this nested `cond`. Usually it's better to try to move the variables up and have only one `cond`

If you think that's more confusing change this `cond` to `if` because it has only one condition.

PR #2437, opened 2022-10-11
Storage: ObjectsMap

niwinz:


backend/src/app/rpc/queries/files.clj
```diff
@@ -227,29 +227,34 @@
...
 
 (defn retrieve-file
-  [{:keys [pool] :as cfg} id components-v2]
+  [{:keys [pool] :as cfg} id features]
   (let [file (->> (db/get-by-id pool :file id)
                   (decode-row)
                   (pmg/migrate-file))]
 
-    (if components-v2
+    (if (contains? features "components/v2")
```
hirunatan, line 236: Couldn't this be replaced by reading `ffeat/*current*`?

hirunatan: Ignore me... this function is used from other places, too.

PR #2443, opened 2022-10-13
:tada: Basic changes to use penpot as a library

superalex:


frontend/src/app/libs/file_builder.cljs
```diff
@@ -50,6 +99,13 @@
   (closeGroup [_]
     (set! file (fb/close-group file)))
 
+  (add-bool [_ data]
```
niwinz: why some functions uses snake-case and other camelCase?

frontend/src/app/main/ui/browser_hooks.cljs
```diff
@@ -0,0 +1,31 @@
+(ns app.main.ui.browser-hooks
```
niwinz: After the broadcast and storage changes, I think thare are no reason to have it separated, I'm wrong?

frontend/src/app/libs/file_builder.cljs
superalex: my failure :D

PR #2459, opened 2022-10-14
:recycle: Move internal.xxx namespaces to separated files

hirunatan:
To clarify the code for figma plugin contributors.

Hope I have put all the `:` and `::` correctly in their place...

### DIFF
--- a/common/src/app/common/types/file/media_object.cljc
+++ b/common/src/app/common/types/file/media_object.cljc
@@ -1,0 +1,21 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) UXBOX Labs SL
+
+(ns app.common.types.file.media-object
+  (:require
+   [app.common.spec :as us]
+   [clojure.spec.alpha :as s]))
+
+(s/def ::name string?)
+(s/def ::width ::us/safe-integer)
+(s/def ::height ::us/safe-integer)
+(s/def ::mtype string?)
+
+;; NOTE: This is marked as nilable for backward compatibility, but
+;; right now is just exists or not exists. We can thin in a gradual
+;; migration and then mark it as not nilable.
+(s/def ::path (s/nilable string?))
+
--- a/common/src/app/common/types/file/media_object.cljc
+++ b/common/src/app/common/types/file/media_object.cljc
@@ -9,6 +9,7 @@
    [app.common.spec :as us]
    [clojure.spec.alpha :as s]))
 
+(s/def ::id uuid?)
 (s/def ::name string?)
 (s/def ::width ::us/safe-integer)
 (s/def ::height ::us/safe-integer)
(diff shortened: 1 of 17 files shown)


common/src/app/common/types/file/media_object.cljc
```diff
@@ -0,0 +1,21 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) UXBOX Labs SL
```
niwinz: Should be `KALEIDOS INC`

common/src/app/common/types/page/flow.cljc
```diff
@@ -0,0 +1,24 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) UXBOX Labs SL
```
niwinz: Should be `KALEIDOS INC`

niwinz: and the rest of files, ensure correct copyright holder

common/src/app/common/types/page/grid.cljc
```diff
@@ -0,0 +1,44 @@
...
+;; Copyright (c) UXBOX Labs SL
+
+(ns app.common.types.page.grid
+  (:require
+   [app.common.spec :as us]
+   [clojure.spec.alpha :as s]))
+
+;; --- Board grids
+
+(s/def :grid/color string?)
```
niwinz: We should stop use single segment undeclared namespace.
I mean, if we have `color` and `internal/color`, we pobably need to create a specific ns for it with
`(:require [app.common.types.page.grid.internal :as-alias internal])`

And then use `(s/def ::internal/color string?)` for this case.
But having `color` meaning different things, is a bit confusing (I know that is not related to this PR, because it has been done in this way some time ago).

common/src/app/common/types/shape/text.cljc
```diff
@@ -0,0 +1,72 @@
...
+                   :position-data/font-family
+                   :position-data/font-size
+                   :position-data/font-style
+                   :position-data/font-weight
+                   :position-data/rtl
+                   :position-data/text
+                   :position-data/text-decoration
+                   :position-data/text-transform]))
+
+(s/def :position-data/x ::us/safe-number)
```
niwinz: Use the specific alias for this, for avoid single segment namespaces:

`(:require [app.common.types.shape.text.position-data :as-alias position-data])`

And then you can declare  spec with the alias `(s/def ::position-data/x ::us/safe-number)`

common/src/app/common/types/page/flow.cljc
hirunatan: Ok, but this should be in a different branch, because there are a lot of files with UXBOX Labs, unrelated tho this PR.

common/src/app/common/types/shape/text.cljc
hirunatan: I understand this implies moving those specs to a new file, to be able to `:require` it, don't it?

niwinz: No. It implies the use of `:as-alias` that just creates a reference without importing it.

common/src/app/common/types/page/flow.cljc
niwinz: yep, this is fixed on one of my pending to merge PR, but this is just for not introduce again incorrect string

hirunatan: ah wel... Done, then.

common/src/app/common/types/file/media_object.cljc
hirunatan: done

PR #2469, opened 2022-10-17
:tada: Add robust concurrency limiter

niwinz:


common/src/app/common/types/container.cljc
```diff
@@ -15,13 +15,12 @@
 
 (s/def ::type #{:page :component})
 (s/def ::id uuid?)
-(s/def ::name string?)
-(s/def ::path (s/nilable string?))
+(s/def ::name ::us/string)
+(s/def ::path (s/nilable ::us/string))
 
 (s/def ::container
-  ;; (s/keys :req-un [::id ::name ::ctst/objects]
```
hirunatan, line 22: I think this was commented out for performance, because validating spec of a tree of objects may be very slow.

And this validation is being triggered somewhere with real files in production.

I suppose it's good to activate it, but we need to be aware of this, and check when is this being triggered.

common/test/common_tests/test_types_shape_interactions.cljc
```diff
@@ -2,17 +2,17 @@
...
   (:require
-   [clojure.test :as t]
-   [clojure.pprint :refer [pprint]]
    [app.common.exceptions :as ex]
+   [app.common.geom.point :as gpt]
    [app.common.types.shape :as cts]
    [app.common.types.shape.interactions :as ctsi]
    [app.common.uuid :as uuid]
-   [app.common.geom.point :as gpt]))
+   [clojure.pprint :refer [pprint]]
```
hirunatan, line 14: delete?

PR #2484, opened 2022-10-19
:rewind: Backport Blob v5 from develop

niwinz:


hirunatan:
This seems OK to me, but prefer to wait until @niwinz is back, to check that everything is ok after merging.

niwinz:
This is just a backport, this prevents errors on opening files created on HOURLY on staging env

PR #2485, opened 2022-10-19
Add new langs and some fixes

niwinz:


superalex:
.po files shouldn't be included in this PR?

PR #2491, opened 2022-10-20
:recycle: Pointers Map and features negotiation related refactor

niwinz:


backend/src/app/util/pointer_map.clj
```diff
@@ -0,0 +1,224 @@
...
+
+(declare create)
+
+(defprotocol IPointerMap
+  (get-id [_])
+  (load! [_])
+  (modified? [_])
+  (clone [_]))
+
+(deftype PointerMap [id mdata
```
hirunatan, line 69: I miss some documentation here. For those kind of structures that are not obvious, it's important to have an introduction of what is, more or less what is used for, and a brief general description of how does it work. All of this is complicated to guess without reviewing deeply the code itself.

This doesn't need to be very long, just a couple of paragraphs, perhaps, but it's important.

niwinz: :+1:

niwinz: done

PR #2498, opened 2022-10-21
:bug: Fix bug about decoding :features PgArray

hirunatan:


niwinz:
I think a map is a wrong default, it should be a set

PR #2537, opened 2022-10-28
Autolayout & refactor transformations

Alotor:
https://tree.taiga.io/project/penpot/us/2262?milestone=328432


common/src/app/common/geom/shapes/flex_layout/bounds.cljc
```diff
@@ -0,0 +1,113 @@
...
+          (if (or (ctl/fill-height? child) (ctl/fill-height? child))
+            (child-layout-bound-points parent child)
+            points))]
+
+    (as-> children $
+      (mapcat child-bounds $)
+      (gco/transform-points $ (gco/center-shape parent) (:transform-inverse parent))
+      (gre/squared-points $)
+      (gpo/pad-points $ (- pad-top) (- pad-right) (- pad-bottom) (- pad-left))
+      (gre/points->rect $))))
```
niwinz: Looking on it, looks like the ony one step requires the subject in different position, in this case I guess a better alternative can be:
```clojure
(-> (mapcat child-bounds children)
    (gco/transform-points (gco/center-shape parent) (:transform-inverse parent))
    (gre/squared-points)
    (gpo/pad-points (- pad-top) (- pad-right) (- pad-bottom) (- pad-left))
    (gre/points->rect))
```

common/src/app/common/geom/shapes/flex_layout/drop_area.cljc
```diff
@@ -0,0 +1,169 @@
...
+
+                line-area (gsr/make-rect x y width height)
+
+                children (subvec children from-idx (+ from-idx num-children))
+
+                result (first (reduce redfn-child [result line-area] (d/with-next children)))]
+
+            [result (+ from-idx num-children) (+ x width) (+ y height)]))]
+
+    (first (reduce redfn-lines [[] 0] (d/with-next (:layout-lines layout-data))))))
```
niwinz: I suggest consider using `loop` here instead of `reduce`. I guess this will improve readability and performance

common/src/app/common/geom/shapes/flex_layout/lines.cljc
```diff
@@ -0,0 +1,312 @@
...
+
+              [{:line-min-width  next-min-width
+                :line-min-height next-min-height
+                :line-max-width  next-max-width
+                :line-max-height next-max-height
+                :num-children    1
+                :children-data   [child-data]}
+               (cond-> result (some? line-data) (conj line-data))])))
+
+        [line-data layout-lines] (reduce calculate-line-data [nil []] children)]
```
niwinz: The same here, if think `loop` will be better on performance (I don't really know if this is on hot path or not, so I just suggest the change)

common/src/app/common/geom/shapes/flex_layout/lines.cljc
```diff
@@ -0,0 +1,312 @@
...
+        [result remainder]
+        (let [cur-val (or (get current prop) (get current prop-min) 0)
+              max-val (get current prop-max)
+              cur-inc (if (> (+ cur-val per-line-target) max-val)
+                        (- max-val cur-val)
+                        per-line-target)
+              current (assoc current prop (+ cur-val cur-inc))
+              remainder (- remainder cur-inc)
+              result (conj result current)]
+          (recur (first items) (rest items) remainder result))))))
```
niwinz, line 133: This is just FYI, that there are a way to write loop without maintaining the both: first and rest on the bindings. Just maintaining the seq. Example:

```clojure
(loop [items     (seq items)
       remainder to-share
       result    []]
  (if-let [current (first items)]
    (let [cur-val (or (get current prop) (get current prop-min) 0)
          max-val (get current prop-max)
          cur-inc (if (> (+ cur-val per-line-target) max-val)
                    (- max-val cur-val)
                    per-line-target)
          current (assoc current prop (+ cur-val cur-inc))]
      (recur (rest items)
             (- remainder cur-inc)
             (conj result current)))
    [result remainder]))
```

I don't suggest to change, this is just a informative comment.

common/src/app/common/geom/shapes/flex_layout/lines.cljc
```diff
@@ -0,0 +1,312 @@
...
+  (let [layout-bounds (layout-bounds shape)
+        reverse?      (ctl/reverse? shape)
+        children      (cond->> children reverse? reverse)
+
+        ;; Creates the layout lines information
+        layout-lines
+        (->> (init-layout-lines shape children layout-bounds)
+             (add-lines-positions shape layout-bounds)
+             (mapv (partial add-line-spacing shape layout-bounds))
+             (mapv (partial add-children-resizes shape)))]
```
niwinz: prefer `map` + `map` then vec/into (or directly `into` with transducers) over `mapv` + `mapv`

common/src/app/common/geom/shapes/flex_layout/positions.cljc
```diff
@@ -0,0 +1,280 @@
...
+        col?           (ctl/col? parent)
+        space-between? (ctl/space-between? parent)
+        space-around?  (ctl/space-around? parent)
+        h-center?      (ctl/h-center? parent)
+        h-end?         (ctl/h-end? parent)
+        v-center?      (ctl/v-center? parent)
+        v-end?         (ctl/v-end? parent)
+        content-stretch? (ctl/content-stretch? parent)
+
+        hv   #(gpo/start-hv layout-bounds %)
```
niwinz: Use the same style :P 
image

common/src/app/common/geom/shapes/modifiers.cljc
```diff
@@ -7,284 +7,279 @@
...
-                          (:height raw-bounds)
-                          (max 1 (mth/round (:height raw-bounds))))
+(defn resolve-tree-sequence
+  "Given the ids that have changed search for layout roots to recalculate"
+  [ids objects]
 
-          target-p (cond-> (gpt/round (gpt/point raw-bounds))
-                     flip-x?
-                     (update :x + target-width)
+  (assert (or (nil? ids) (set? ids)) (dm/str "tree sequence from not set: " ids))
```
niwinz: Prefer using `us/assert!` macro for documented asserts:
```clojure
(us/assert!
 :expr (or (nil? ids) (set? ids))
 :hint (dm/str "tree sequence from not set: " ids))
 ```

File: common/src/app/common/geom/shapes/modifiers.cljc [hunk 8]
```diff
@@ -7,284 +7,279 @@
...
+
+                (cond-> result
+                  (not contains-parent?)
+                  (conj root)))))
+
+          (generate-tree ;; Generate a tree sequence from a given root id
+            [id]
+            (->> (tree-seq
+                  #(d/not-empty? (get-in objects [% :shapes]))
+                  #(get-in objects [% :shapes])
```
niwinz: `dm/get-in` instead of `get-in`

niwinz:
As general note, prefer use `dm/get-in` in performance sensitive code and when the vector traverse path is known at compile time.

The `dm/get-in` is 2x faster in general use cases, here a simple bench results:

- `get-in` with vector of 3 segments => `132ns`
- `dm/get-in` with vector of 3 segments => `72ns`

common/src/app/common/geom/shapes/rect.cljc
```diff
@@ -11,14 +11,21 @@
...
-      {:x x
-       :y y
-       :width width
-       :height height})))
+  ([p1 p2]
+   (let [x1 (min (:x p1) (:x p2))
+         y1 (min (:y p1) (:y p2))
+         x2 (max (:x p1) (:x p2))
+         y2 (max (:y p1) (:y p2))]
+     (make-rect x1 y1 (- x2 x1) (- y2 y1))))
```
niwinz, line 23: If the function is performance sensitive, consider extract common attrs first and then access directly to the binding. This will avoid hash-map lookup multiple times for same value.

common/src/app/common/geom/shapes/transforms.cljc
```diff
@@ -662,17 +433,48 @@
...
 (defn apply-group-modifiers
   "Apply the modifiers to the group children to calculate its selection rect"
-  [group objects modif-tree]
+  [parent objects modif-tree]
+
+  (let [children (->> (:shapes parent)
+                      (map (d/getf objects))
+                      (apply-children-modifiers objects modif-tree))]
+    (cond-> parent
+      (= :group (:type parent))
```
niwinz: Consider use a predicate defined in page helpers.

common/src/app/common/geom/shapes/flex_layout/bounds.cljc
Alotor: ok

common/src/app/common/geom/shapes/flex_layout/drop_area.cljc
Alotor: OK

common/src/app/common/geom/shapes/flex_layout/lines.cljc
Alotor: ok

common/src/app/common/geom/shapes/flex_layout/lines.cljc
Alotor: ok

common/src/app/common/geom/shapes/flex_layout/positions.cljc
Alotor: ok

common/src/app/common/geom/shapes/modifiers.cljc
Alotor: ok

common/src/app/common/geom/shapes/rect.cljc
Alotor: ok

common/src/app/common/geom/shapes/transforms.cljc
Alotor: ok

frontend/src/app/util/snap_data.cljs
```diff
@@ -79,33 +80,36 @@
...
-        (update-in [frame-id :x] (make-insert-tree-data (d/concat-vec frame-data grid-x-data) :x))
-        (update-in [frame-id :y] (make-insert-tree-data (d/concat-vec frame-data grid-y-data) :y)))))
+
+    (cond-> page-data
+      (not (ctl/layout-child? objects frame))
+
+      (-> ;; Update root frame information
+       (assoc-in [uuid/zero :objects-data frame-id] frame-data)
+       (update-in [parent-id :x] (make-insert-tree-data frame-data :x))
+       (update-in [parent-id :y] (make-insert-tree-data frame-data :y))
```
niwinz, line 90: Just a suggestion: prevent duplicated calls to `update-in` with common parent (in this case i'm refering to the `parent-id`), and replace it with `(update parent-id (fn [parent] (assoc ...)))`

Repeated calls to `update-in` creates the vector instance on each invocation.

frontend/src/app/util/snap_data.cljs
```diff
@@ -164,22 +168,22 @@
...
         root-data (get-in page-data [uuid/zero :objects-data frame-id])
         frame-data (get-in page-data [frame-id :objects-data frame-id])]
-    (-> page-data
-        (update-in [uuid/zero :x] (make-delete-tree-data root-data :x))
-        (update-in [uuid/zero :y] (make-delete-tree-data root-data :y))
-        (update-in [frame-id :x]  (make-delete-tree-data frame-data :x))
-        (update-in [frame-id :y]  (make-delete-tree-data frame-data :y))
-        (add-frame new-frame))))
+    (as-> page-data $
+      (update-in $ [uuid/zero :x] (make-delete-tree-data root-data :x))
```
niwinz, line 176: the same here

common/src/app/common/geom/shapes/common.cljc
```diff
@@ -31,6 +32,22 @@
     (gpt/point (/ (+ minx maxx) 2.0)
                (/ (+ miny maxy) 2.0))))
 
+(defn center-bounds [[a b c d]]
```
niwinz, line 35: This function receives a coll of fixed number of points, they can be directly 4 arguments?

common/src/app/common/types/modifiers.cljc
```diff
@@ -296,17 +354,20 @@
...
 (defn only-move?
   "Returns true if there are only move operations"
-  [modifier]
-  (or (and (= 1 (-> modifier :geometry-child count))
-           (= :move (-> modifier :geometry-child first :type)))
-      (and (= 1 (-> modifier :geometry-parent count))
-           (= :move (-> modifier :geometry-parent first :type)))))
+  [{:keys [geometry-child geometry-parent]}]
+  (and (every? #(= :move (:type %)) geometry-child)
+       (every? #(= :move (:type %)) geometry-parent)))
```
niwinz, line 359: I gues we can reuse the same predicate for both operations

common/src/app/common/geom/shapes/modifiers.cljc
```diff
@@ -173,119 +173,136 @@
...
+    (reduce apply-shape objects modif-tree)))
+
+(defn- apply-partial-objects-modifiers
+  [objects tree-seq modif-tree]
+
+  (letfn [(apply-shape [objects {:keys [id] :as shape}]
+            (if (cph/group-shape? shape)
+              (let [children (cph/get-children objects id)]
+                (assoc objects id
+                       (cond
```
niwinz, line 253: I suggest using simple `if` here

common/src/app/common/geom/shapes/common.cljc
Alotor: This will receive the `(:points shape)` value and it's a fixed value of 4 points. So we don't want to "spread" them.

common/src/app/common/types/modifiers.cljc
Alotor: ok

common/src/app/common/geom/shapes/modifiers.cljc
Alotor: yeah, I will simplify this

PR #2546, opened 2022-11-03
:sparkles: Better overlays interactions on boards inside boards

pabloalba:
Implements with https://tree.taiga.io/project/penpot/us/4386

### DIFF
--- a/common/src/app/common/types/shape/interactions.cljc
+++ b/common/src/app/common/types/shape/interactions.cljc
@@ -87,7 +87,7 @@
     :close-overlay
     :prev-screen
     :open-url})
-
+(s/def ::position-relative-to (s/nilable ::us/uuid))
 (s/def ::overlay-pos-type
   #{:manual
     :center
@@ -110,7 +110,7 @@
 (defmethod action-opts-spec :navigate [_]
   (s/keys :opt-un [::destination
                    ::preserve-scroll
-                     ::animation]))
+                   ::animation]))
 
 (defmethod action-opts-spec :open-overlay [_]
   (s/keys :req-un [::overlay-position
@@ -118,7 +118,8 @@
           :opt-un [::destination
                    ::close-click-outside
                    ::background-overlay
-                   ::animation]))
+                   ::animation
+                   ::position-relative-to]))
 
 (defmethod action-opts-spec :toggle-overlay [_]
   (s/keys :req-un [::overlay-position
@@ -126,11 +127,13 @@
           :opt-un [::destination
                    ::close-click-outside
                    ::background-overlay
-                   ::animation]))
+                   ::animation
+                   ::position-relative-to]))
 
 (defmethod action-opts-spec :close-overlay [_]
   (s/keys :opt-un [::destination
-                   ::animation]))
+                   ::animation
+                   ::position-relative-to]))
 
 (defmethod action-opts-spec :prev-screen [_]
   (s/keys :req-un []))
@@ -159,6 +162,7 @@
   {:event-type :click
    :action-type :navigate
    :destination nil
+   :position-relative-to nil
    :preserve-scroll false})
 
 (def default-delay 600)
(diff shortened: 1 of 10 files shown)


hirunatan:
This change works when the origin shape is a board. But when it's a shape inside a board (for example a button), the user expect that the position is relative to the board containing the shape, not the origin shape itself. This is how it works now in production.

The current problem is that it always use the bottom board, even if the shape is inside a nested board. It should use the board in that the shape is.

common/src/app/common/types/shape/interactions.cljc
```diff
@@ -262,6 +263,13 @@
     (assoc :overlay-pos-type :center
            :overlay-position (gpt/point 0 0))))
 
+(defn set-position-relative-to
+  [interaction position-relative-to]
+  (us/verify ::interaction interaction)
+  (us/verify ::position-relative-to position-relative-to)
+  (assert (has-destination interaction))
```
hirunatan: I understand that the new `:position-relative-to` parameter is related only to overlays, isn't it?

If so, this should be `(has-overlay-opts interaction)`

pabloalba: Mmm... I need it there to initialize the value to "self" on new interactions. Maybe I can initialize the value in another place, I need to check...

frontend/src/app/main/data/workspace/interactions.cljs
```diff
@@ -123,7 +123,8 @@
                     (fn [shape]
                       (let [new-interaction (ctsi/set-destination
                                              ctsi/default-interaction
-                                             destination)]
+                                             destination)
+                            new-interaction (ctsi/set-position-relative-to new-interaction (:id shape))]
```
hirunatan: I think more clean
```clojure
(let [new-interaction (-> ctsi/default-interaction
                          (ctsi/set-destination destination)
                          (ctsi/set-position-relative-to (:id shape))]
```

frontend/src/app/main/ui/workspace/sidebar/options/menus/interactions.cljs
```diff
@@ -184,6 +184,8 @@
         destination          (get objects (:destination interaction))
 
         frames               (mf/with-memo [objects] (ctt/get-viewer-frames objects {:all-frames? true}))
+        shape-parent-ids     (mf/with-memo [objects] (cph/get-parent-ids objects (:id shape)))
+        shape-parents        (filter (comp (set shape-parent-ids) :id) frames)
```
hirunatan: Think this one also needs to be wrapped in `(mf/with-memo [frames]`, to avoid repeating the filter each time the component is repainted.

common/src/app/common/types/shape/interactions.cljc
hirunatan: Maybe call to `set-position-relative-to` when you change the action type, and only if `(has-overlay-opts` is true.

common/src/app/common/types/shape/interactions.cljc
```diff
@@ -87,7 +87,7 @@
     :close-overlay
     :prev-screen
     :open-url})
-
+(s/def ::position-relative-to (s/nilable ::us/uuid))
```
hirunatan, line 90: You also need to add the new attribute to `action-opts-spec`, in the `:req-un` block of `:open-overlay` and `close-overlay`

pabloalba: Yes. Or more exactly it should be on opt-un for backwards compatibility

frontend/src/app/main/ui/workspace/sidebar/options/menus/interactions.cljs
pabloalba: Ok!

frontend/src/app/main/data/workspace/interactions.cljs
pabloalba: ok!

common/src/app/common/types/shape/interactions.cljc
pabloalba: Ok, I found a way :)

PR #2600, opened 2022-11-25
:tada: Add workspace read-only setting

pabloalba:
Implements https://tree.taiga.io/project/penpot/task/4514

### How to test
* Call in the console to debug.set_workspace_read_only(true) or debug.set_workspace_read_only(false) to activate/deactivate read only

### DIFF
--- a/frontend/src/app/main/data/workspace/state_helpers.cljs
+++ b/frontend/src/app/main/data/workspace/state_helpers.cljs
@@ -146,3 +146,4 @@
   (let [{:keys [x y width height]} (get-in state [:workspace-local :vbox])]
     (gpt/point (+ x (/ width 2)) (+ y (/ height 2)))))
 
+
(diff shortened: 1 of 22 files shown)


frontend/src/app/main/data/workspace/state_helpers.cljs
```diff
@@ -146,3 +146,9 @@
   (let [{:keys [x y width height]} (get-in state [:workspace-local :vbox])]
     (gpt/point (+ x (/ width 2)) (+ y (/ height 2)))))
 
+
+(defn lookup-workspace-read-only?
```
niwinz: Looks unused

frontend/src/app/main/ui/workspace/header.cljs
```diff
@@ -464,7 +468,8 @@
                 :project project
                 :file file
                 :team-id team-id
-                :page-id page-id}]]
+                :page-id page-id
+                :workspace-read-only? workspace-read-only?}]]
```
niwinz: You don't need to pass this all down having the react context set. Just use the use-ctx on the component if you need to know the value

pabloalba: Fixed!

frontend/src/app/main/data/workspace/state_helpers.cljs
pabloalba: Removed

PR #2619, opened 2022-12-01
:bug: Fix adding an extra blank page on import

hirunatan:
fixes https://tree.taiga.io/project/penpot/task/4543

### DIFF
--- a/backend/src/app/rpc/commands/files/create.clj
+++ b/backend/src/app/rpc/commands/files/create.clj
@@ -31,9 +31,9 @@
 
 (defn create-file
   [conn {:keys [id name project-id is-shared data revn
-                modified-at deleted-at
+                modified-at deleted-at create-page
                 ignore-sync-until features]
-         :or {is-shared false revn 0}
+         :or {is-shared false revn 0 create-page true}
          :as params}]
   (let [id       (or id (:id data) (uuid/next))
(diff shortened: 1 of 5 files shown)


backend/src/app/rpc/commands/files/create.clj
```diff
@@ -64,13 +65,16 @@
...
+
 (s/def ::create-file
   (s/keys :req-un [::files/profile-id
                    ::files/name
                    ::files/project-id]
           :opt-un [::files/id
                    ::files/is-shared
-                   ::files/features]))
+                   ::files/features
+                   ::nilable-page-id]))
```
niwinz: This will not going to validate never the `:page-id` attr.
Also you are touching `create-file` public API, when you really need to touch the create-temp-file.

hirunatan: I think it's useful to be able to create a file without automatically adding a blank page, both in create-file and in create-tem-file (although we only use it for the moment in the second one).

I have more doubts in using the nil value of page-id or adding another paramenter like `:avoid-auto-page` or similar. I think it validates well: you can omit the `:page-id` attribute, or send it with null value.

hirunatan: changed

PR #2640, opened 2022-12-07
Flex layout fixes

Alotor:


common/src/app/common/pages/changes_builder.cljc
```diff
@@ -276,6 +276,21 @@
          (update :undo-changes #(reduce mk-undo-change % shapes))
          (apply-changes-local)))))
 
+
+(defn changed-attrs
```
niwinz, line 280: Looks unused.

common/src/app/common/pages/changes_builder.cljc
```diff
@@ -276,6 +276,21 @@
...
+  [object update-fn {:keys [attrs]}]
+  (let [changed?
+        (fn [old new attr]
+          (let [old-val (get old attr)
+                new-val (get new attr)]
+            (not= old-val new-val)))
+        new-obj (update-fn object)]
+    (if (= object new-obj)
+      '()
```
niwinz: Can be simple when, returning `nil` in some cases, `nil` and `'()` are equivalent in seq processing

common/src/app/common/pages/changes_builder.cljc
Alotor: I know, I left it because I think could be used in the future

common/src/app/common/pages/changes_builder.cljc
Alotor: ok

PR #2693, opened 2022-12-27
:bug: Fix display type of component library not persistent

pabloalba:
Fixes https://tree.taiga.io/project/penpot/issue/4512

### DIFF
--- a/frontend/src/app/main/data/workspace.cljs
+++ b/frontend/src/app/main/data/workspace.cljs
@@ -1845,6 +1845,26 @@
       (assoc-in state [:workspace-local :inspect-expanded] expanded?))))
 
 
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; File Library persistent settings
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+
+(defn set-file-library-listing-thumbs
+  [listing-thumbs?]
+  (ptk/reify ::set-file-library-listing-thumbs
+    ptk/UpdateEvent
+    (update [_ state]
+      (assoc-in state [:workspace-global :file-library-listing-thumbs] listing-thumbs?))))
+
+(defn set-file-library-reverse-sort
+  [reverse-sort?]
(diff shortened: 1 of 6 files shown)


frontend/src/app/main/data/workspace.cljs
```diff
@@ -1842,6 +1842,26 @@
...
+;; File Library persistent settings
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+
+(defn set-file-library-listing-thumbs?
+  [listing-thumbs?]
+  (ptk/reify ::set-file-library-listing-thumbs?
+    ptk/UpdateEvent
+    (update [_ state]
+      (assoc-in state [:workspace-global :file-library-listing-thumbs?] listing-thumbs?))))
```
niwinz: I think the function name and the state keyword should not end with `?`
Especially when the function is not a predicate.

frontend/src/app/main/data/workspace.cljs
```diff
@@ -1842,6 +1842,26 @@
...
+
+(defn set-file-library-reverse-sort?
+  [reverse-sort?]
+  (ptk/reify ::set-file-library-reverse-sort?
+    ptk/UpdateEvent
+    (update [_ state]
+      (assoc-in state [:workspace-global :file-library-reverse-sort?] reverse-sort?))))
```
niwinz: The same here

frontend/src/app/main/refs.cljs
```diff
@@ -183,6 +183,12 @@
 (def context-menu
   (l/derived :context-menu workspace-local))
 
+(def file-library-listing-thumbs?
+  (l/derived :file-library-listing-thumbs? workspace-global))
```
niwinz: Here, the var name it is ok to be with `?`, but the state access should be changed accordingly to the previous change

frontend/src/app/main/data/workspace.cljs
pabloalba: Fixed

frontend/src/app/main/data/workspace.cljs
pabloalba: Fixed

frontend/src/app/main/refs.cljs
pabloalba: Fixed

PR #2697, opened 2022-12-28
:bug: Fix show outline with rounded corners on rects

pabloalba:
Fixes https://tree.taiga.io/project/penpot/issue/4053

### DIFF
--- a/frontend/src/app/main/ui/workspace/viewport/outline.cljs
+++ b/frontend/src/app/main/ui/workspace/viewport/outline.cljs
@@ -10,6 +10,7 @@
    [app.common.exceptions :as ex]
    [app.common.geom.shapes :as gsh]
    [app.main.ui.hooks :as hooks]
+   [app.main.ui.shapes.attrs :as attrs]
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/ui/workspace/viewport/outline.cljs
```diff
@@ -34,12 +35,15 @@
             (or (ex/ignoring (upf/format-path (:content shape)))
                 "")))
 
-        {:keys [x y width height selrect]} shape
+        {:keys [x y width height selrect rx ry]} shape
+
+        style-attrs (attrs/extract-style-attrs shape)
```
niwinz: `extract-style-attrs` is pretty expensive for just extract the `d` if it is found for the outline component.
We need to make the `d` calculation (with rounding, etc..) code more reusable and use it here instead of calling the `extract-style-attrs` for a single attribute

pabloalba: Fixed

PR #2745, opened 2023-01-09
:sparkles: Improves matrix serialization

AzazelN28:
Changes matrix serialization from `precision` to native JS method `toFixed` (faster)


common/src/app/common/geom/matrix.cljc
```diff
@@ -19,6 +19,22 @@
 (def precision 6)
 
 ;; --- Matrix Impl
+;
```
Alotor: remove the comment, please

AzazelN28: done

PR #2758, opened 2023-01-11
🐛 Fixes multiple React warnings about duplicate keys

AzazelN28:
Fixes multiple react warnings about duplicate keys on viewer mode.

### DIFF
--- a/frontend/src/app/main/ui/shapes/frame.cljs
+++ b/frontend/src/app/main/ui/shapes/frame.cljs
@@ -113,9 +113,10 @@
     {::mf/wrap-props false}
     [props]
 
-    (let [childs     (unchecked-get props "childs")]
+    (let [childs (unchecked-get props "childs")]
       [:> frame-container props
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/ui/shapes/frame.cljs
```diff
@@ -65,17 +65,17 @@
         render-id (mf/use-ctx muc/render-id)]
 
     [:*
-     [:g {:clip-path (when (not show-content) (frame-clip-url shape render-id))}
-      [:& frame-clip-def {:shape shape :render-id render-id}]
+     [:g {:key "clip-path" :clip-path (when (not show-content) (frame-clip-url shape render-id))}
```
Alotor: I don't think this are adding much value if they are not inside a loop

AzazelN28: true, I've removed this

PR #2792, opened 2023-01-17
Fix board component import

hirunatan:
Fixes https://tree.taiga.io/project/penpot/issue/4647

### DIFF
--- a/common/src/app/common/file_builder.cljc
+++ b/common/src/app/common/file_builder.cljc
@@ -567,9 +567,8 @@
 (defn start-component
   ([file data] (start-component file data :group))
   ([file data root-type]
-   (let [selrect (if (and (:x data) (:y data) (:width data) (:height data))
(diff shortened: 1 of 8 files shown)


common/src/app/common/file_builder.cljc
```diff
@@ -561,35 +565,38 @@
...
-          :main-instance-id main-instance-id
-          :main-instance-page main-instance-page
-          :shapes [obj]})
-
-        (assoc :last-id (:id obj))
-        (update :parent-stack conjv (:id obj))
-        (assoc :current-component-id (:id obj)))))
+  ([file data] (start-component file data :group))
+  ([file data root-type]
+   (let [selrect (if (and (:x data) (:y data) (:width data) (:height data))
```
Alotor: You could use `(or (gsh/make-selrect ...) cts/empty-selrect)` because `make-selrect` returns null if one of its arguments is not a number

hirunatan: done

PR #2827, opened 2023-01-24
:bug: Fix missing flex props on code generation

EvaMarco:
This PR solves this issue https://tree.taiga.io/project/penpot/issue/4681 (Check message on issue.)
Fixed css value properties as (nowrap, row-reverse, column-reverse, line height without px"
Fixed copy button on code tab

### DIFF
--- a/frontend/src/app/main/ui/formats.cljs
+++ b/frontend/src/app/main/ui/formats.cljs
@@ -55,7 +55,7 @@
       {:p1 p1 :p2 p2 :p3 p3 :p4 p4}
 
       (and (= p1 p3) (= p2 p4))
-      {:p1 p1 :p3 p3}
+      {:p1 p1 :p2 p2}
 
       (and (not= p1 p3) (= p2 p4))
       {:p1 p1 :p2 p2 :p3 p3}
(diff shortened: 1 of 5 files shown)


frontend/src/app/main/ui/formats.cljs
```diff
@@ -67,9 +67,11 @@
   (let [sizing (if (= type :width)
                  (:layout-item-h-sizing shape)
                  (:layout-item-v-sizing shape))]
-    (if (= sizing :fill)
-      "100%"
-      (str (format-pixels value)))))
+    (if (string? value)
```
Alotor, line 70: Two things, first it needs to consider the `sizing = auto` also I think the size type has priority over the value. 

My sugestion would be:

```clojure
(case sizing
  :fill ...
   :auto ....
   (if (number? value)
       (format-pixels value)
       value))    
```

frontend/src/app/util/code_gen.cljs
```diff
@@ -41,6 +41,33 @@
...
+(defn get-layout-orientation
+  [value]
+  (if (= :wrap value)
+    "wrap"
+    "nowrap"))
+
+(defn get-layout-direction
+  [value]
+  (case value
+    :row "row"
```
Alotor, line 65: Maybe we can change the keyword now that we still can... once we're in production this will be quite painful.

frontend/src/app/util/code_gen.cljs
```diff
@@ -98,9 +125,9 @@
              :text-transform]
    :to-prop {:fill-color "color"}
    :format  {:font-family #(str "'" % "'")
-             :font-style #(str "'" % "'")
+             :font-style #(str %)
```
Alotor, line 128: All these `str` should be changed to `dm/str` where dm is `app.common.data.macros` it's more efficient and should be used always for string concatenation.

PR #2851, opened 2023-01-27
:bug: Fix styling info at the libraries modal

superalex:
Related to https://tree.taiga.io/project/penpot/issue/4748


hirunatan:
The assets count is disordered. This should be 1 component, 2 graphics, 3 colors, 4 typographies

image

hirunatan:
The same file, when made a library and shown in the righ column,  has a different count.

image

But when it's added as external library, the count is ok now:

image

PR #2884, opened 2023-02-01
Post release fixes

Alotor:
:bug: Fix problems with strange file
:bug: Add limit to growth fill shapes to the bounds of the layout
:bug: Fix thumbnail not updating when changing pages
:bug: Fix problem with selection on nested boards
:sparkles: Change parent/children constraint for problematic configurations

This will solve the problems we have when there is a "hug content" parent with "fill" children. When we detect an incompatible configuration we change the parent or the children to fix it.

### DIFF
--- a/common/src/app/common/pages/helpers.cljc
+++ b/common/src/app/common/pages/helpers.cljc
@@ -109,6 +109,20 @@
         (recur (conj result parent-id) parent-id)
         result))))
 
+(defn get-parent-ids-with-index
+  "Returns a tuple with the list of parents and a map with the position within each parent"
+  [objects shape-id]
+  (loop [parent-list []
+         parent-indices {}
+         current shape-id]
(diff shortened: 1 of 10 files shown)


common/src/app/common/pages/helpers.cljc
```diff
@@ -109,6 +109,20 @@
         (recur (conj result parent-id) parent-id)
         result))))
 
+(defn get-parent-ids-with-index
+  "Returns a tuple with the list of parents and a map with the position in everyparent"
+  [objects shape-id]
+  (loop [result [[] {}]
```
niwinz: split the result in two different bindings, this will avoid create an additional vector instance on each iteration

Alotor: good idea. Thanks

PR #2891, opened 2023-02-01
:sparkles: Change parent/children constraint for problematic configur…

Alotor:
:sparkles: Change parent/children constraint for problematic configurations

This will solve the problems we have when there is a "hug content" parent with "fill" children. When we detect an incompatible configuration we change the parent or the children to fix it.


Alotor:
Will merge this MR with https://github.com/penpot/penpot/pull/2884

PR #2905, opened 2023-02-04
fix(docker): adds mailcatch to penpot network

mabasic:
Without this the backend complains that it cannot connect to the smtp host (when using mailcatcher). The reason is because the mailcatcher is not on the same network as the backend application.


niwinz:
Merged using command line, thanks

PR #2982, opened 2023-02-24
♻️ connect line-heigh values to variables and set scale

ondrejkonec:
This branch connect line-heights values with variables from scale, that was created.

**What was changed**:
- All line height values was redefined as unitless. This is best approach for browser to display everything correctly and there is low change for visual bugs. 
> References: 
https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#values
https://allthingssmitty.com/2017/01/30/nope-nope-nope-line-height-is-unitless/ (unofficial source)

Now there is scale that represent percentages - for e.g. `$lh-088` is `0.88%`, `$lh-192` is `192%` and so.

The main added value is that now we can see all the variables and can eventually unify them, name them better, etc. - but that is for the future, because it is necessary to go through each element separately, and I will focus on this when I rewrite the components.

**Note**:
I removed a couple of `line-heights` that were used inappropriately (they were not tied to the typography but to the height of the element) and solved (and improved) the centering of the iconic button on the slider. You can see results below:

**Original**:
Screenshot 2023-02-24 at 14 07 02

**Updated**:
Screenshot 2023-02-24 at 14 35 49

### DIFF
--- a/frontend/resources/styles/common/base.scss
+++ b/frontend/resources/styles/common/base.scss
@@ -71,19 +71,19 @@ button {
 p {
   font-size: $fs12;
   margin-bottom: 1rem;
-  line-height: $base-lh-sm;
+  line-height: $lh-133;
 
   @include bp(baby-bear) {
     font-size: $fs16;
-    line-height: $base-lh;
+    line-height: $lh-143;
   }
 }
 
 li {
-  line-height: $base-lh-sm;
+  line-height: $lh-133;
 
   @include bp(baby-bear) {
-    line-height: $base-lh;
+    line-height: $lh-143;
   }
 }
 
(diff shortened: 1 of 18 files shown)


ondrejkonec:
There you can see process. On the left, there are original values that mixes browser values, px, rem, percentage and unitless.

On the right, there is updated line-height (leading) scale, that is ready for improvements (merge, better naming etc.)

Screenshot 2023-02-24 at 14 45 45

niwinz:
hi @ondrejkonec , thanks for all the contributions.

Just a unrelated to the commit content comment:

Please, can you avoid use special/unicode characters on commits and branch names. I suggest to use the following commit message:

`:recycle: Connect line-heigh values to variables and set scale`

The commit message format is detailed in the [CONTRIBUTING.md](https://github.com/penpot/penpot/blob/develop/CONTRIBUTING.md#commit-guidelines)

ondrejkonec:
Hi @niwinz 

Thanks for letting me know, I did not know that there is difference and I will use recommended syntax for next time. Sorry about that.

Should I change the commit message or it's good for now and I can improve it next time?

niwinz:
Don't worry, we will fix it on merge :D
Next time is ok, thanks!

ondrejkonec:
You are welcome. 

Btw. same approach is for branches right? As I saw with other branches, I do not need to put "type" like `:recycle:` and so, right? I guess I misunderstood from the previous conversation.

niwinz:
the branch naming can be simplier, we internally use `<username>-<some-simple-descr>`, example: `ondrejkonec-connect-line-height-to-variables`, that makes very easy check who is the owner of the branch on at first look.

I will document it on CONTRIBUTING.md

ondrejkonec:
Thank you @niwinz, my next branch will be in this format ;)

niwinz:
merged using CLI, thanks

PR #2985, opened 2023-02-27
🐋 Add environment variables for specifying the export and backend url

### LINKED ISSUE #2984: How to set the backend url for frontend docker image?
Hi, My team has an internal app called penpot-backend, and as such we aren't able to use the compose one. Is there an environment variable to change this? 

Your reply is appreciated. Thanks in advance!

Supernova3339:
This PR is for the `frontend` docker image.
`penpotapp/frontend`

> Why was this change made?

When deploying Penpot, I use a weird docker combination that doesn't use compose files.
Penpot has the URLs hardcoded by default into the `nginx.conf` which is bad practice.

[Documentation PR](https://github.com/penpot/penpot-docs/pull/134)

closes #2984

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 ### :boom: Breaking changes & Deprecations
 
+- Adds environment variables for specifying the export and backend url for the frontend docker image. [Github #2983](https://github.com/penpot/penpot/pull/2983)
+
 ### :sparkles: New features
 - Adds more accessibility improvements in dashboard [Taiga #4577](https://tree.taiga.io/project/penpot/us/4577)
 - Adds paddings and gaps prediction on layout creation [Taiga #4838](https://tree.taiga.io/project/penpot/task/4838)
--- a/docker/images/files/nginx.conf
+++ b/docker/images/files/nginx.conf
@@ -82,7 +82,7 @@ http {
         }
 
         location /assets {
-            proxy_pass http://penpot-backend:6060/assets;
+            proxy_pass $BACKEND_URL/assets;
             recursive_error_pages on;
             proxy_intercept_errors on;
             error_page 301 302 307 = @handle_redirect;
@@ -95,11 +95,11 @@ http {
         }
 
         location /api/export {
-            proxy_pass http://penpot-exporter:6061;
+            proxy_pass $EXPORT_URL;
         }
 
         location /api {
-            proxy_pass http://penpot-backend:6060/api;
+            proxy_pass $BACKEND_URL/api;
         }
 
         # location /admin {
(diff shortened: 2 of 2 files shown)


niwinz:
Hi @Supernova3339 

Nice catch, looks a nice change, but I need to double check it and test it. I will go with this as fast as I finish with my current stuff I'm working on. 

Thanks

Supernova3339:
> Hi @Supernova3339
> 
> Nice catch, looks a nice change, but I need to double check it and test it. I will go with this as fast as I finish with my current stuff I'm working on.
> 
> Thanks

Thanks for the reply! Considering these values are hardcoded, not very good coding standards. Sent a PR to fix it. If you find anything that concerns you, please let me know. I did NOT touch anything that was commented out.

niwinz:
I'm pretty happy with your changes, I just need test it myself and then merge. I'll probably will do it tomorrow.
I'm honestly not 100% sure if you can use environment variables as-is in the nginx.conf file; that's why I really want to try it.

Supernova3339:
I can link you to an example repo if needed. That's how I found out about
this

abn0mad:
Hi there, I was also looking for a way to specify the backend and exporter URLs for a similar reason; running penpot as a series of rootless podman containers managed by systemd.

Won't this pull require something along the lines of [this](https://www.baeldung.com/linux/nginx-config-environment-variables) in order to work though?

i.e.

```
/bin/bash -c "envsubst '$${BACKEND_URL} $${EXPORTER_URL}' < /etc/nginx/nginx.conf.template > \ 
/etc/nginx/nginx.conf && nginx -g 'daemon off;'"

```

It would be great to see the option to specify the backend and exporter URLs for users of podman and users who prefer to stay away from docker-compose.

Supernova3339:
> Hi there, I was also looking for a way to specify the backend and exporter URLs for a similar reason; running penpot as a series of rootless podman pods managed by systemd.
> 
> Won't this pull require something along the lines of [this](https://www.baeldung.com/linux/nginx-config-environment-variables) in order to work though?
> 
> i.e.
> 
> ```
> /bin/bash -c "envsubst '$${BACKEND_URL} $${EXPORTER_URL}' < /etc/nginx/nginx.conf.template > \ 
> /etc/nginx/nginx.conf && nginx -g 'daemon off;'"
> 
> ```
> 
> It would be great to see the option to specify the backend and exporter URLs for users of podman and users who prefer to stay away from docker-compose. 

Not at all! I've tested it and my solution works perfectly. It's better to NOT overcomplicate things :)

abn0mad:
@Supernova3339 - Thanks for the quick reply :), well that's good news, I didn't know environment variables could be accepted by nginx in such a way, that opens up a lot of simplified deployments indeed. I look forward to having this pull request being accepted :) (and sorry for raising the question unnecessarily).

niwinz:
image

As I suspected, this does not work out of the box. Nginx does not support interpolation of environment variables out of the box. This is a screenshot of the nginx docker image readme:

image

It is posible to have this functionality but an additional step is necessary. I will look on how to approach this

niwinz:
https://github.com/penpot/penpot/pull/3010 replaces this PR,
thanks for the suggestion

PR #3092, opened 2023-03-31
:tada: Copy paste components, even to another page

pabloalba:
Implements https://tree.taiga.io/project/penpot/task/4578

### DIFF
--- a/frontend/src/app/main/data/workspace.cljs
+++ b/frontend/src/app/main/data/workspace.cljs
@@ -1600,9 +1600,10 @@
                                   (gpt/subtract (gpt/point (:selrect base)) orig-pos))]
                   [frame-id parent-id delta index]))))
 
-          ;; Change the indexes if the paste is done with an element selected
+          ;; Change the indexes of the pasted shapes
           (change-add-obj-index [paste-objects selected index change]
-            (let [set-index (fn [[result index] id]
+            (let [index (or index -1) ;; if there is no current element selected, we want the first (inc index) to be 0
+                  set-index (fn [[result index] id]
                               [(assoc result id index) (inc index)])
 
                   map-ids
@@ -1626,7 +1627,8 @@
 
           ;; Proceed with the standard shape paste process.
           (do-paste [it state mouse-pos media]
-            (let [file-id      (:current-file-id state)
+            (let [libraries    (wsh/get-libraries state)
+                  file-id      (:current-file-id state)
                   page         (wsh/lookup-page state)
                   media-idx    (d/index-by :prev-id media)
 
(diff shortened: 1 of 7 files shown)


frontend/src/app/main/data/workspace.cljs
```diff
@@ -1651,7 +1653,7 @@
 
                   all-objects (merge (:objects page) paste-objects)
 
-                  changes  (-> (dws/prepare-duplicate-changes all-objects page selected delta it nil)
+                  changes  (-> (dws/prepare-duplicate-changes all-objects page selected delta it libraries state)
```
hirunatan: Passing the whole application state to a function introduces a huge implicit dependency. I think it would be better to pass only the part that is needed.

frontend/src/app/main/data/workspace/libraries_helpers.cljs
```diff
@@ -146,24 +146,30 @@
...
+         [new-shape new-shapes]
+         (ctn/make-component-instance page
+                                      component
+                                      (:data library)
+                                      position
+                                      components-v2)
 
-    [new-shape changes]))
+         changes (cond-> (pcb/add-object changes (first new-shapes) {:ignore-touched true})
+                     (some? old-id) (pcb/amend-last-change #(assoc % :old-id old-id)))
```
hirunatan: I don't understand where this :old-id value is used :-?

frontend/src/app/main/data/workspace/selection.cljs
```diff
@@ -332,54 +333,81 @@
...
+     (-> changes
+         (prepare-duplicate-flows shapes page ids-map)
+         (prepare-duplicate-guides shapes page ids-map delta)))))
+
+(defn- prepare-duplicate-component-change
+  ([changes page component delta libraries state it]
+   (let [component-id (:component-id component)
+         file-id (:component-file component)
+         main-component    (ctf/get-component libraries file-id component-id)
+         moved-component   (gsh/move component delta)
```
hirunatan, line 376: I don't get this. `gsh/move` is for moving one shape, but here you pass it a component, why? Components do not have position.

frontend/src/app/main/data/workspace/selection.cljs
```diff
@@ -332,54 +333,81 @@
...
+                                                       state
+                                                       it)
+                      init-changes))]
+
+     (-> changes
+         (prepare-duplicate-flows shapes page ids-map)
+         (prepare-duplicate-guides shapes page ids-map delta)))))
+
+(defn- prepare-duplicate-component-change
+  ([changes page component delta libraries state it]
```
hirunatan: The attribute `component` should be renamed to `component-root`, because it's a shape, not a component object.

PR #3108, opened 2023-04-04
🔨 Fix default port for penpot-exporter

leonekmi:
By default, penpot-frontend will try to reach penpot-export on port 80, this is wrong, the exporter listens by default to port 6061.

An alternative solution is to add the following environment variable to the penpot-frontend container spec in [`docker-compose.yaml`](https://github.com/penpot/penpot/blob/develop/docker/images/docker-compose.yaml#L97):
`      - PENPOT_EXPORTER_URI=http://penpot-exporter:6061`


niwinz:
Thanks! It is better put a good default.

niwinz:
Merged using CLI

Issue #2520, opened 2022-10-24
feature: Click to select full values at the design sidebar


myfunnyandy:
### Is your feature request related to a problem? Please describe.

Click on all text fields at the design sidebar should select the whole value.

### Describe the solution you'd like.

For the sake of inputs consistency, when clicking in a numeric box on the right side panel (design properties), the whole value should get selected so those values can be easily overwritten.

![click-value](https://user-images.githubusercontent.com/708713/197551395-82944605-f8cf-40f6-bf32-0cc3193ca44f.gif)

### Additional context

[Taiga reference](https://tree.taiga.io/project/penpot/us/3043?no-milestone=1)

PR #3179, opened 2023-04-30
:tada: Add feature to select full values on click at the design sidebar

akshay-gupta7:
:tada: Add feature to select full values on click at the design sidebar #2520 

* On clicking a numeric box on the right sidebar (design properties), the whole value will get selected so those values can be easily overwritten
* Feature added to Typography fields(Font size, line height and letter spacing)
* Feature added to Blur value as well


superalex:
Merged manually, thanks! :)

PR #3180, opened 2023-05-01
:bug: rect filter bounds math fix

ryanbreen:
get-rect-filter-bounds was incorrectly applying delta-blur to x1 twice and to y1 never

Signed-off-by: Ryan Breen


superalex:
Thank you very much @ryanbreen 

I've merged manually to include your changes in the CHANGES.md :)

PR #3184, opened 2023-05-03
:sparkles: Removed sizing variables from radius

ondrejkonec:
### What
Remove sizing variables from border radius definition

### Why
Since this is not critical, I'm here to help with improvement. By this change, we are preparing for scalable system ([it's part of this iniciative](https://community.penpot.app/t/code-contributions-welcome/1870/6)). I think it's better **not to mix** spacing and other type of design decisions.

This approach can block us creating more meaningful structure. As an example, let's say we want to to update/change our spacing system and we have this spacing variables connected to radius. Everything has to be redesigned.

Another point is, that it not add any value using this properties at the moment. I'll update this radiuses when new UI is ready so we can be more complex.

It's part of my previous commits, where I created scale for border radii.
Screenshot 2023-05-03 at 17 42 27


superalex:
Thank you very much @ondrejkonec 

Merged manually to include you in the CHANGES.md ;)

Issue #2518, opened 2022-10-24
feature: Line height "auto"


myfunnyandy:
### Is your feature request related to a problem? Please describe.

Cannot set the value "auto" to line height text property.

### Describe the solution you'd like.

Deleting the value at the line height should leave it with the value "1.2".

The value that is applied and will be shown at the handoff is "line-height: 1.2".

### Additional context

[Taiga reference](https://tree.taiga.io/project/penpot/us/3048?no-milestone=1)

PR #3185, opened 2023-05-03
:tada: Add feature to set line-height to auto as 1.2

akshay-gupta7:
:tada: Add feature to set line-height to auto as 1.2 #2518 

* With this feature, deleting the value at the line height will leave it with the value "1.2"
* Hence, the value that will be applied and shown at the handoff is "line-height: 1.2"

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/options/menus/typography.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/menus/typography.cljs
@@ -378,10 +378,7 @@
 
         handle-change
         (fn [value attr]
-          (let [new-values (case attr
-                            :line-height (assoc values :line-height (if (or (str/empty? value) (nil? value)) 1.2 value))
-                            :spacing-options (assoc values :letter-spacing value))]
-            (on-change new-values)))]
+          (on-change {attr (str value)}))]
 
     [:div.spacing-options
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/ui/workspace/sidebar/options/menus/typography.cljs
```diff
@@ -374,9 +374,14 @@
         line-height (or line-height "1.2")
         letter-spacing (or letter-spacing "0")
 
+        line-height-nillable (if (= (str line-height) "1.2") false true)
+
         handle-change
         (fn [value attr]
-          (on-change {attr (str value)}))]
+          (let [new-values (case attr
+                            :line-height (assoc values :line-height (if (or (str/empty? value) (nil? value)) 1.2 value))
```
superalex: I think you can let the handle-change function as the original one if just add
```
:default "1.2"
```
To 
```
[:> numeric-input" ...
```

akshay-gupta7: Thank you so much for the suggestion. This worked like a charm without tinkering the handle-change function. I have made the changes

superalex:
Thank you very much!, merged manuall :)

PR #3189, opened 2023-05-04
:sparkles: Notify library updates when really needed

hirunatan:
https://tree.taiga.io/project/penpot/us/1397?milestone=345366

### DIFF
--- a/frontend/src/app/main/data/workspace.cljs
+++ b/frontend/src/app/main/data/workspace.cljs
@@ -260,12 +260,11 @@
 
     ptk/WatchEvent
     (watch [_ state _]
-      (let [ignore-until  (-> state :workspace-file :ignore-sync-until)
-            file-id       (-> state :workspace-file :id)
-            needs-update? (some #(and (> (:modified-at %) (:synced-at %))
-                                      (or (not ignore-until)
-                                          (> (:modified-at %) ignore-until)))
-                                libraries)]
+      (let [file-data     (:workspace-data state)
+            ignore-until  (dm/get-in state [:workspace-file :ignore-sync-until])
+            file-id       (dm/get-in state [:workspace-file :id])
+            needs-update? (seq (filter #(dwl/assets-need-sync % file-data ignore-until)
+                                       libraries))]
         (when needs-update?
           (rx/of (dwl/notify-sync-file file-id)))))))
 
@@ -1633,7 +1632,7 @@
           ;; Check if the shape is an instance whose master is defined in a
           ;; library that is not linked to the current file
           (foreign-instance? [shape paste-objects state]
-            (let [root         (cph/get-root-shape paste-objects shape)
+            (let [root         (ctn/get-component-shape paste-objects shape)
                   root-file-id (:component-file root)]
               (and (some? root)
                    (not= root-file-id (:current-file-id state))
--- a/frontend/src/app/main/data/workspace/libraries.cljs
+++ b/frontend/src/app/main/data/workspace/libraries.cljs
@@ -605,7 +605,7 @@
              container   (cph/get-container local-file :page page-id)
              shape       (ctn/get-shape container id)]
 
-         (when (ctk/in-component-instance? shape)
+         (when (ctk/instance-head? shape)
            (let [libraries (wsh/get-libraries state)
 
                  changes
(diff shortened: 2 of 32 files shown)


frontend/src/app/main/data/workspace/libraries.cljs
```diff
@@ -812,13 +814,26 @@
                {:file-id (get-in state [:workspace-file :id])
                 :date (dt/now)}))))
 
+(defn assets-need-sync
+  "Get a lazy sequence of all the assets of each type in the library that have
+  been modified after the last sync of the library. The sync date may be
+  overriden by providing a ingnore-until parameter.
```
pabloalba: typo ignore-until

frontend/src/app/main/data/workspace.cljs
```diff
@@ -260,12 +260,11 @@
 
     ptk/WatchEvent
     (watch [_ state _]
-      (let [ignore-until  (-> state :workspace-file :ignore-sync-until)
+      (let [file-data     (-> state :workspace-data)
+            ignore-until  (-> state :workspace-file :ignore-sync-until)
```
pabloalba: Is this better than 
file-data (:workspace-data state)
ignore-until (get-in state [:workspace-file :ignore-sync-until])
? Because I think we use a lot more the get-in form...

frontend/src/app/main/data/workspace/libraries.cljs
```diff
@@ -828,14 +843,15 @@
...
-                (tr "workspace.updates.there-are-updates")
-                :inline-actions
-                [{:label (tr "workspace.updates.update")
-                  :callback do-update}
-                 {:label (tr "workspace.updates.dismiss")
-                  :callback do-dismiss}]
-                :sync-dialog))))))
+        (when (seq libraries-need-sync)
+          (rx/of (dm/info-dialog
+                  (tr "workspace.updates.there-are-updates")
```
pabloalba, line 848: We don't want to show the updated components yet?

frontend/src/app/main/data/workspace/libraries.cljs
hirunatan: fixed

frontend/src/app/main/data/workspace.cljs
hirunatan: In the first case I'm not really sure.

In the second, actually get-in is slower because it receives a dynamic vector of arguments, and the compiler optimizer cannot convert it to static accessors. But we have a `(d/get-in` macro that solves this. So I'm not either sure if this construct is better to d/get-in.

frontend/src/app/main/data/workspace/libraries.cljs
hirunatan: We have not design of how to show the list of updated assets. This is left for a future phase.

frontend/src/app/main/data/workspace.cljs
niwinz: The `->` is faster than using `get-in`, if you are going to change `->` to get-in, please, use `dm/get-in` (faster, macro based version)

niwinz: and, don't use `->` when you have a single operation:

`(-> state :workspace-data)` should be `(:workspace-data state)`

niwinz: > In the first case I'm not really sure.
> 
> In the second, actually get-in is slower because it receives a dynamic vector of arguments, and the compiler optimizer cannot convert it to static accessors. But we have a `(d/get-in` macro that solves this. So I'm not either sure if this construct is better to d/get-in.

dm/get-in is analogous to -> performance

frontend/src/app/main/data/workspace.cljs
```diff
@@ -260,12 +260,11 @@
...
-      (let [ignore-until  (-> state :workspace-file :ignore-sync-until)
-            file-id       (-> state :workspace-file :id)
-            needs-update? (some #(and (> (:modified-at %) (:synced-at %))
-                                      (or (not ignore-until)
-                                          (> (:modified-at %) ignore-until)))
-                                libraries)]
+      (let [file-data     (:workspace-data state)
+            ignore-until  (dm/get-in state [:workspace-file :ignore-sync-until])
+            file-id       (dm/get-in state [:workspace-file :id])
+            needs-update? (seq (filter #(dwl/assets-need-sync % file-data ignore-until)
```
niwinz, line 266: The `seq` idiom should be used on `if`

frontend/src/app/main/data/workspace.cljs
hirunatan: changed

Issue #2519, opened 2022-10-24
feature: Focus input when searching a file at the projects dashboard


myfunnyandy:
### Describe the solution you'd like.

At the projects dashboard, when using the shortcut to search file (Ctrl/Cmd + F) and the dashboard search is activated the input field should get the browser focus so the user can start searching inmediatly.

### Additional context

[Taiga reference](https://tree.taiga.io/project/penpot/us/3049?no-milestone=1)

PR #3196, opened 2023-05-07
:tada: Add feature to focus input on search when searching a file at …

akshay-gupta7:
🎉  Add feature to focus input on search when searching a file at the projects dashboard #2519 

* At the projects dashboard, when using the shortcut to search file (Ctrl/Cmd + F) and the dashboard search is activated, the search input field will get the browser focus
* Hence, with this feature, the user can immediately start typing the search phrase


superalex:
Thank you very much! :)

PR #3209, opened 2023-05-12
:recycle: Refactor thumbnail rendering on workspace

AzazelN28:
## Changes

This PR introduces three main changes:

1. Changes how thumbnails are stored in the state and in the backend. Instead of using Data URIs, which can slow down the application, I've replaced them with Blob URIs.
2. Changes how thumbnails are rendered in the React component and in the Worker (now it uses `createImageBitmap` that it is much more performant than `<canvas>`). 
3. Better file structure. This should make the code easier to read and maintain.

## Thanks to

@niwinz for it's help in the state/backend part.
@Alotor for the initial rework.

### DIFF
--- a/frontend/src/app/main/ui/workspace/viewport/pixel_overlay.cljs
+++ b/frontend/src/app/main/ui/workspace/viewport/pixel_overlay.cljs
@@ -7,6 +7,7 @@
 (ns app.main.ui.workspace.viewport.pixel-overlay
   (:require
    [app.common.data :as d]
+   [app.common.data.macros :as dm]
    [app.common.pages.helpers :as cph]
    [app.common.uuid :as uuid]
    [app.main.data.modal :as modal]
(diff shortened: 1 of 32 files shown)


frontend/src/app/main/ui/workspace/viewport/pixel_overlay.cljs
```diff
@@ -92,8 +136,8 @@
                    x (- (.-clientX event) brx)
                    y (- (.-clientY event) bry)
 
-                   zoom-context (.getContext zoom-view-node "2d")
-                   canvas-context (.getContext canvas-node "2d")
+                   zoom-context (.getContext zoom-view-node "2d" {:willReadFrequently true})
+                   canvas-context (.getContext canvas-node "2d" {:willReadFrequently true})
```
niwinz: you probably miss the `#js` before the `{:willRead...}`

app/common/data/macros.cljs
Alotor, line 1: I think you should remove this

niwinz: fixed, and rebased

PR #3215, opened 2023-05-16
Superalex bugfixing 11

superalex:
This MR fixes different situations related to svg imports:

Also makes the import process more resilient and the position-data in texts safer

### DIFF
--- a/frontend/src/app/util/text_svg_position.cljs
+++ b/frontend/src/app/util/text_svg_position.cljs
@@ -90,27 +90,26 @@
   [shape-id]
   (when (some? shape-id)
     (p/let [text-data (calc-text-node-positions shape-id)]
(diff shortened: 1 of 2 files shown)


frontend/src/app/util/text_svg_position.cljs
```diff
@@ -90,7 +90,7 @@
   [shape-id]
   (when (some? shape-id)
     (p/let [text-data (calc-text-node-positions shape-id)]
-      (when (d/not-empty? text-data)
+      (if (d/not-empty? text-data)
```
Alotor: I think we don't need the `if` if `text-data` is empty or nil the `mapv` function will return `[]`

superalex: Thanks! :)

Issue #2512, opened 2022-10-24
feature: Ability to change shadows order


myfunnyandy:
### Is your feature request related to a problem? Please describe.

Shadows order can not be change, which making difficult to manage their visibility.

### Describe the solution you'd like.

Shadows order should be able to be changed using drag and drop (the same way as with strokes).

Upper shadows in the shadows list are placed at the top.

New shadows are positioned at the top by default. 

### Additional context

[Taiga reference](https://tree.taiga.io/project/penpot/us/3465?backlog-owner=426680&no-milestone=1)

PR #3221, opened 2023-05-17
:tada: Add ability to change shadows' order and place new shadows at t…

akshay-gupta7:
:tada: Add ability to change shadows' order and place new shadows at the top by default #2512 

* With this feature, user can now drag and change the order of shadows just like they did for strokes
* Also, the new shadow added will be at the top by default

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/options/menus/shadow.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/menus/shadow.cljs
@@ -10,9 +10,11 @@
    [app.common.data :as d]
    [app.common.uuid :as uuid]
    [app.main.data.workspace.changes :as dch]
+   [app.main.data.workspace.colors :as dc]
    [app.main.data.workspace.undo :as dwu]
    [app.main.store :as st]
    [app.main.ui.components.numeric-input :refer [numeric-input]]
+   [app.main.ui.hooks :as h]
    [app.main.ui.icons :as i]
    [app.main.ui.workspace.sidebar.options.common :refer [advanced-options]]
    [app.main.ui.workspace.sidebar.options.rows.color-row :refer [color-row]]
@@ -37,7 +39,7 @@
   (or (number? value) (not (js/isNaN (js/parseInt value)))))
 
 (mf/defc shadow-entry
-  [{:keys [ids index value]}]
+  [{:keys [ids index value on-reorder select-all disable-drag on-blur]}]
   (let [open-shadow (mf/use-state false)
 
         basic-offset-x-ref (mf/use-ref nil)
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/ui/workspace/sidebar/options/menus/shadow.cljs
```diff
@@ -64,6 +66,21 @@
         select-text
         (fn [ref] (fn [_] (dom/select-text! (mf/ref-val ref))))
 
+        on-drop
+        (fn [_ data]
+          (on-reorder (:index data)))
+
+        [dprops dref] (if (some? on-reorder)
```
niwinz, line 73: You really can't use hooks on this manner, the hooks can't be on a conditional. They should be always defined as static definitions and always in the same order. I just comment it for you information, I will change it my self.

niwinz:
The same here. I have created a new PR with your commit and some improvements.

akshay-gupta7:
> The same here. I have created a new PR with your commit and some improvements.

Thank you @niwinz for reviewing. 
Also, I saw your comment on the design limitations of the numeric input component. I want to work on improving the numeric component. I would like your high level inputs on how to go about achieving the same.

Regards

frontend/src/app/main/ui/workspace/sidebar/options/menus/shadow.cljs
akshay-gupta7: Thank you for highlighting this. Will keep this in mind going forward

niwinz:
The main idea on creating a wrapper component to the native element is that should behave very similar to the native
Right now, the event callbacks are called with value, but with no event, so I can't access the dom node. If I need dom need to use refs, that is ok for some cases but other they are completely unnecessary. 

Also, there are no way avoid re-render the component, because of the callbacks handling limitation (you need create a new function on each render).

I guess this explanation is not very clear, because the issues are not evident at simple look.
In any case this is just FYI, i will go to address it myself once I finish my current ongoing task.

akshay-gupta7:
> The main idea on creating a wrapper component to the native element is that should behave very similar to the native Right now, the event callbacks are called with value, but with no event, so I can't access the dom node. If I need dom need to use refs, that is ok for some cases but other they are completely unnecessary.
> 
> Also, there are no way avoid re-render the component, because of the callbacks handling limitation (you need create a new function on each render).
> 
> I guess this explanation is not very clear, because the issues are not evident at simple look. In any case this is just FYI, i will go to address it myself once I finish my current ongoing task.

Yes I get this to some extent. I faced a similar issue while trying to apply on-focus: select-all on these numeric-input components(#3179), so as a workaround I put these events on the wrapper of these elements, which is fine as a workaround, but going forward we will need to address this.
Thank you for taking the time to throw some light on this.
Regards

PR #3223, opened 2023-05-18
hiru bugfixing 

hirunatan:


common/src/app/common/pages/changes.cljc
```diff
@@ -677,6 +677,26 @@
...
+          xform (comp (filter :main-instance?)
+                      (map :id))]
+      (into #{} xform all-parents))))
+
+(defmethod components-changed :del-obj
+  [file-data {:keys [id page-id _component-id] :as change}]
+  (when page-id
+    (let [page (ctpl/get-page file-data page-id)
+          shape-and-parents (map #(ctn/get-shape page %)
+                                 (into [id] (cph/get-parent-ids (:objects page) id)))
```
niwinz: Is more efficient here use `cons` instead of `into`

common/src/app/common/pages/changes.cljc
```diff
@@ -677,6 +677,26 @@
                           (map :id))]
           (into #{} xform shape-and-parents))))))
 
+(defmethod components-changed :mov-objects
+  [file-data {:keys [page-id _component-id parent-id] :as change}]
+  (when page-id
+    (let [page (ctpl/get-page file-data page-id)
+          all-parents (map #(ctn/get-shape page %)
+                           (into [parent-id] (cph/get-parent-ids (:objects page) parent-id)))
```
niwinz: the same here, `cons` is better than `into` for this case

common/src/app/common/pages/changes.cljc

common/src/app/common/pages/changes.cljc

common/src/app/common/pages/changes.cljc
```diff
@@ -677,6 +678,34 @@
...
+  [file-data {:keys [page-id _component-id parent-id shapes] :as change}]
+  (when page-id
+    (let [page (ctpl/get-page file-data page-id)
+
+          xform (comp (filter :main-instance?)
+                      (map :id))
+
+          check-shape
+          (fn [shape-id others]
+            (let [all-parents (map #(ctn/get-shape page %)
```
niwinz: prefer partial here

common/src/app/common/pages/changes.cljc
```diff
@@ -677,6 +678,34 @@
...
+
+      (reduce #(set/union %1 (check-shape %2 []))
+              (check-shape parent-id [parent-id])
+              shapes))))
+
+(defmethod components-changed :del-obj
+  [file-data {:keys [id page-id _component-id] :as change}]
+  (when page-id
+    (let [page (ctpl/get-page file-data page-id)
+          shape-and-parents (map #(ctn/get-shape page %)
```
niwinz: the same here, prefer partial,
partial and anonymous functions are analogous, but internally, at least in JVM, it will generate an ad-hoc class for this function, if you use partial you avoid generating a class and reuse the one used in partial impl

common/src/app/common/pages/changes.cljc

common/src/app/common/pages/changes.cljc

PR #3211, opened 2023-05-14
:tada: Implement functionality to preview layer blend modes

akshay-gupta7:
:tada: Implement functionality to preview layer blend modes #2513 

* With this feature, users can preview a blend mode before selecting it
* Currently, this implementation excludes the case when 2 or more selected shapes have different blend modes(multiple). However, if the multiple shapes have same blend-mode, then preview mode will apply on all the shapes.
* If my implementation of this is fine and efficient, I would like to then open an enhancement to preview blend modes even when the  multiple selected shapes have different blend-modes

PR #3235, opened 2023-05-22
:tada: Implement functionality to preview layer blend modes

niwinz:
Replaces the MR: https://github.com/penpot/penpot/pull/3211

### DIFF
--- a/frontend/src/app/main/data/workspace.cljs
+++ b/frontend/src/app/main/data/workspace.cljs
@@ -2160,6 +2160,25 @@
 
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Preview blend modes
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn set-preview-blend-mode
+  [ids blend-mode]
+  (ptk/reify ::set-preview-blend-mode
+    ptk/UpdateEvent
+    (update [_ state]
+      (reduce #(assoc-in %1 [:workspace-preview-blend %2] blend-mode) state ids))))
+
+(defn unset-preview-blend-mode
+  [ids]
+  (ptk/reify ::unset-preview-blend-mode
+    ptk/UpdateEvent
+    (update [_ state]
+      (reduce #(update %1 :workspace-preview-blend dissoc %2) state ids))))
+
+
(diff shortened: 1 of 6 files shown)


frontend/src/app/main/refs.cljs
```diff
@@ -550,3 +550,8 @@
 (def current-file-id
   (l/derived :current-file-id st/state))
 
+(def preview-blend-modes
+  (l/derived :preview-blend-modes workspace-local))
+
+(defn preview-blend-modes-by-id [id]
+  (l/derived #(get % id) preview-blend-modes =))
```
niwinz: instead of `#(get % id)`, it should be `(l/key id)`

frontend/src/app/main/data/workspace.cljs
```diff
@@ -2159,6 +2159,24 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Preview blend modes
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn set-preview-blend-mode
+  [ids blend-mode]
+  (ptk/reify ::set-preview-blend-mode
+    ptk/UpdateEvent
+    (update [_ state]
+      (reduce #(assoc-in %1 [:workspace-local :preview-blend-modes %2] blend-mode) state ids))))
```
niwinz: I guess they don't need to be on `:workspace-local`, prefer using a single level... some thing like `:workspace-prevew-blend` as a root key on state

frontend/src/app/main/ui/shapes/shape.cljs
```diff
@@ -54,25 +55,30 @@
         pointer-events   (unchecked-get props "pointer-events")
         disable-shadows? (unchecked-get props "disable-shadows?")
 
+        preview-blend-mode-ref
+        (mf/use-memo (mf/deps (:id shape)) #(refs/preview-blend-modes-by-id (:id shape)))
```
niwinz: - prefer use `with-memo` instead use-memo,
- extract `shape-id` as separate binding on let and then use it (for avoid calling `(:id shape)`  twice
- replace all usage of `(:id shape)` on the rest of the shape, the `(:id shape)` is a Trie Lookup operation and for performance reasons is better lookup once and then reuse

superalex:
Hello @akshay-gupta7 

After your commit we have added two more:
 - @niwinz did a refactor related to performance
 - I added a commit to fix some problems with selection multiple and to avoid savings while just doing the preview of the blend mode

PR #3238, opened 2023-05-23
:tada: Change main shape name along with component name

pabloalba:


frontend/src/app/main/data/workspace/libraries.cljs
```diff
@@ -366,6 +366,11 @@
 
           (rx/of (dch/commit-changes changes)))))))
 
+(defn rename-component-and-main-instance
```
hirunatan, line 369: What do you think about putting this inside `rename-component` function itself?

I think that, since we want to always do both things, there is no need to have two separete functions.

PR #3245, opened 2023-05-24
:bug: Avoid infinite loop nesting copies inside components

hirunatan:
Fixes https://tree.taiga.io/project/penpot/issue/5314

### DIFF
--- a/common/src/app/common/pages/changes.cljc
+++ b/common/src/app/common/pages/changes.cljc
@@ -74,8 +74,8 @@
       [:frame-id {:optional true} ::sm/uuid]
       [:parent-id {:optional true} ::sm/uuid]
       [:index {:optional true} [:maybe :int]]
-      [:ignore-touched {:optional true} :boolean]
-      ]]
+      [:ignore-touched {:optional true} :boolean]]]
+      
 
     [:mod-obj
      [:map {:title "ModObjChange"}
@@ -214,9 +214,9 @@
     [:del-typography
      [:map {:title "DelTypogrphyChange"}
       [:type [:= :del-typography]]
-      [:id ::sm/uuid]]]
+      [:id ::sm/uuid]]]]])
 
-    ]])
+    
 
 (def change?
   (sm/pred-fn ::change))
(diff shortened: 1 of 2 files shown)


common/src/app/common/pages/changes.cljc
```diff
@@ -385,15 +385,27 @@
 
 (defmethod process-change :mov-objects
   [data {:keys [parent-id shapes index page-id component-id ignore-touched after-shape]}]
-  (letfn [(calculate-invalid-targets [objects shape-id]
+  (letfn [(nested-components? [objects shape-id]
+             (let [children          (cph/get-children-with-self objects shape-id)
+                   child-components  (into #{} (keep :component-id children))
```
niwinz: `(into #{} (keep :component-id children))` can be replaced with: `(into #{} (keep :component-id) children)`

The `(keep :component-id)` can be also defined at namespace level for reuse like:

`(def xf-keep-component-id (keep :component-id))`

And then re-used:

`(into #{} xf-keep-component-id children)`

common/src/app/common/pages/changes.cljc
```diff
@@ -385,15 +385,27 @@
...
+                   parents           (cph/get-parents-with-self objects parent-id)
+                   parent-components (into #{} (->> parents
+                                                    (filter :main-instance?)
+                                                    (map :component-id)))]
```
niwinz: here you also can replace

```
(into #{} (->> parents
               (filter :main-instance?)
               (map :component-id)))
```
with something like this

```
(let [xform (comp (filter :main-instance?) 
                  xf-keep-component-id)]
  (into #{} xform parents))
```

Remember, that transducers is a form of reusable and composable way to define transformations independently of the underlying datastructure.

common/src/app/common/pages/changes.cljc

common/src/app/common/pages/changes.cljc
hirunatan: transducers are my nemesis...

perhaps some day i will always remember to use them

Issue #2568, opened 2022-11-16
feature: Replace the "Drafts /" prefix in the project name with an `<a>` element


niekvb:
### Is your feature request related to a problem? Please describe.

I did not find a way to quickly open your *Drafts* dashboard in a new tab to open a second project.

In Figma you can do this by clicking the project name's prefix (eg *Drafts /*) to open this page in a new tab.

### Describe the solution you'd like.

Replace the *Drafts /* prefix of the project name (see screenshot) with an HTML `<a>` element.

This allows to open the drafts page in a new tab by either clicking it with your middle mouse button (scroll wheel click) or by clicking it with the right mouse button.

Alternatively, you could use the same implementation as Figma where it will open the drafts page in a new tab by default.

![Untitled4](https://user-images.githubusercontent.com/37668320/202301654-081db239-0214-43fd-9b58-ba02cb9bf722.png)

EvaMarco:
Hello again @niekvb:
You have been busy :), just yoking. 
This is a very nice idea, and I have created another issue in Taiga, you can check it in [this link](https://tree.taiga.io/project/penpot/us/4504).

PR #3246, opened 2023-05-24
:tada: Add feature to open project name in new tab from workspace

akshay-gupta7:
:tada: Add feature to open project name in new tab from workspace #2568 

* With this feature, when a user clicks on Project Name in the workspace header, the list of files in that particular project and team will now be opened in a new tab


superalex:
Merged manually, thanks!

PR #3260, opened 2023-05-30
:bug: Detect correctly color of outlines and controls of components

hirunatan:
The way of knowing if a shape is inside a component is different, for copies and main.

### DIFF
--- a/frontend/src/app/main/ui/workspace/viewport/outline.cljs
+++ b/frontend/src/app/main/ui/workspace/viewport/outline.cljs
@@ -9,6 +9,8 @@
    [app.common.data :as d]
    [app.common.exceptions :as ex]
    [app.common.geom.shapes :as gsh]
+   [app.common.types.container :as ctn]
+   [app.main.refs :as refs]
    [app.main.ui.hooks :as hooks]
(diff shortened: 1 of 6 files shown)


frontend/src/app/main/ui/workspace/viewport/outline.cljs
```diff
@@ -78,19 +86,16 @@
    ::mf/wrap [#(mf/memo' % (mf/check-props ["shapes" "zoom" "modifiers"]))]}
   [props]
 
-  (let [shapes (obj/get props "shapes")
-        zoom   (obj/get props "zoom")
-        modifiers   (obj/get props "modifiers")
-        color  (if (or (> (count shapes) 1) (nil? (:shape-ref (first shapes))))
```
pabloalba, line 84: Do we no longer need the count of the shapes? Only if it is in a component?

hirunatan: The count was used only to select the color: blue if only 1 shape selected and it's a component; green elsewise.

But now the color is selected individually for each shape. So the count is not necessary.

PR #3264, opened 2023-05-31
:bug: Fix copy paste can produce nested components

pabloalba:


common/src/app/common/pages/helpers.cljc
```diff
@@ -302,6 +303,17 @@
         (ctkl/get-component file id))
       (assoc :type type)))
 
+(defn nested-components? [objects shape-id parent-id]
```
hirunatan: Module `common.pages.helpers` is destined to disappear (it's still there because lack of time to move its contents elsewhere). I recommend putting this in `common.types.container`.

And by the way you can change the name. Looking at it isolated, i think it will be clearer `components-nesting-loop?`. And you could add a description like "Check if a nesting loop would be created if the given shape is moved below the given parent".

pabloalba: Renamed. About moving it to common.types.container, it has dependencies that will generate changes in over 20 files... It's better to address it separately

hirunatan: Ok. I still think that it's good to add a description to the function (the purpose is not obvious by looking at the code)

pabloalba: Done!

PR #3266, opened 2023-05-31
:bug: Distribute fix enabled when two elements were selected

dfelinto:
The distribute operations only make sense when there are at least 3 selected elements. Unlike the alignment ones which indeed should be enabled if there are at least two elements selected.

Current behavior:

image

Expected (patch) behavior:

image

### DIFF
--- a/frontend/src/app/main/data/workspace.cljs
+++ b/frontend/src/app/main/data/workspace.cljs
@@ -1014,7 +1014,7 @@
 (defn can-distribute? [selected]
   (cond
     (empty? selected) false
-    (< (count selected) 2) false
+    (< (count selected) 3) false
     :else true))
 
 (defn distribute-objects


superalex:
Hi @dfelinto 

Thank you very much for your PR, just a few comments:

- Could you please include DCO in your commit https://help.penpot.app/contributing-guide/code-contributions/#code-contributions-dco?
- It would be great if you could fix your commit message is formatted this way: https://help.penpot.app/contributing-guide/code-contributions/#code-contributions-commits

dfelinto:
Hi, I will do it, no problems.

---

By the way when I created the Pull Request, the default template actually says to use:
 `2. PR name follows [convention](http://karma-runner.github.io/4.0/dev/git-commit-msg.html)`
 
 Which is why I went with fix(`<type>`). This is what I see at the moment:

image

dfelinto:
Done, also updated #3267 :)

superalex:
Great!, thank you very much! :)

I've added your contribution here -> https://github.com/penpot/penpot/blob/develop/CHANGES.md?plain=1#L31

PR #3282, opened 2023-06-06
:bug: Fix showing touched mark on a shape belonging a main component

pabloalba:


hirunatan:
Mmmm.. I think the problem here is not hiding the touched flag in a main component. Is that shapes inside a main component never should be touched (because there is no remote shape to be touched respect to).

If we see such a shape with touched, we should detect how this could occur and avoid it. I don't think hiding it is useful.

PR #3284, opened 2023-06-07
:bug: Flix blend mode validation when importing svg

hirunatan:


frontend/src/app/main/data/workspace/svg_upload.cljs
```diff
@@ -67,10 +66,10 @@
                         str/trim
                         str/lower
                         keyword)]
-    (when-not (s/valid? ::cts/blend-mode clean-value)
+    (when-not (cts/blend-mode clean-value)
```
niwinz: Prefer `(contains? cts/blend-mode clean-value)`

Also, consider renaming `cts/blend-mode` to `cts/blend-modes`

hirunatan: Done. It's clearer with `contains`. yes.

And the `blend-modes`, i was afraid that it was used in many places, but not. So i've also changed it.

PR #3290, opened 2023-06-12
:bug: Solve error in notification of library changes

hirunatan:
Solves 3 issues:
https://tree.taiga.io/project/penpot/issue/5400
https://tree.taiga.io/project/penpot/issue/5370
https://tree.taiga.io/project/penpot/issue/5336

### DIFF
--- a/common/src/app/common/pages/changes.cljc
+++ b/common/src/app/common/pages/changes.cljc
@@ -216,8 +216,9 @@
       [:type [:= :del-typography]]
       [:id ::sm/uuid]]]]])
 
-
-
+(sm/def! ::changes
+  [:sequential {:gen/max 2} ::change])
+ 
 (def change?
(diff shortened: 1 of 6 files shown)


common/src/app/common/pages/changes.cljc
```diff
@@ -216,8 +216,9 @@
       [:type [:= :del-typography]]
       [:id ::sm/uuid]]]]])
 
-
-
+(sm/def! ::changes
```
hirunatan, line 219: In main.data.workspace.notifications/schema:handle-file-change there is a reference of `cpc/changes`, that is not defined. If I define it this way, it works, but I'm not sure that this is the best way.

niwinz: The fix is OK, is clearly a bug, it was defined probably initially and accidentally deleted.

PR #3308, opened 2023-06-15
:bug: Fix ungroup component

pabloalba:


frontend/src/app/main/data/workspace/groups.cljs
```diff
@@ -150,6 +153,43 @@
...
+        (ctst/clone-object shape
+                           (:parent-id shape)
+                           objects)
+
+        new-shape-id (:id new-shape)
+        new-shapes (->> new-shapes
+                        (filter #(not= (:id %) new-shape-id))
+                        (map #(if (= new-shape-id (:parent-id %))
+                                (assoc % :parent-id (:parent-id shape))
+                                %)))]
```
hirunatan: I think you may do this in `update-new-object` callback of  `clone-object`: this callback is for doing this kind of things. And so, you avoid a map. You will need only the filter.

Additionally, I think it's cleaner if you replace the `if` by a `cond->`.

frontend/src/app/main/data/workspace/groups.cljs
```diff
@@ -150,6 +153,43 @@
...
+     (fn [changes shape]
+       (pcb/add-object changes shape))
+     changes
+     new-shapes)))
+
+(defn remove-component-changes
+  [it page-id shape objects state]
+  (let [data (get state :workspace-data)
+        page (wsh/lookup-page state)
+        file (wsh/get-local-file state)
```
hirunatan: It's not mandatory, but I prefer that helper functions don't receive the whole app state, but only the elements they really need. I think it's better to leave state management only to event functions.

For example, I'll define this function to receive file-data and file. And extract the page from the file with common.types helpers.

frontend/src/app/main/data/workspace/groups.cljs
pabloalba: Changed

frontend/src/app/main/data/workspace/groups.cljs
pabloalba: Changed

Issue #3310, opened 2023-06-15
feature: Configurable Resolver in Docker Container


DaDummy:
### Is your feature request related to a problem? Please describe.

Currently the penpot docker containers do not properly work when run with kubernetes.

This stems from the issue that the nginx resolver is hardcoded to be 127.0.0.11: https://github.com/penpot/penpot/blob/1c1397a5d85d41f1588e93cb16aa26c023257635/docker/images/files/nginx.conf#LL41C14-L41C24

But for kubernetes you need to use `kube-dns.kube-system.svc` as resolver instead.

### Describe the solution you'd like.

Adjust nginx.conf and the Dockerfiles to allow overriding the default resolver through an environment variable as has been done for e.g. `PENPOT_BACKEND_URI`.

niwinz:
Should be fixed in 1.18.6, you should use the `PENPOT_INTERNAL_RESOLVER` to specify a custom resolver

DaDummy:
Thanks!

PR #3319, opened 2023-06-19
:sparkles: Allow override the default nginx resolver

niwinz:
using the PENPOT_INTERNAL_RESOLVER environment variable


niwinz:
Fixes https://github.com/penpot/penpot/issues/3310

Issue #3305, opened 2023-06-14
bug: Independent corner radius input tooltips only say "Radius" instead of which corner it is for


astudentinearth:
### Steps To Reproduce

1. Select a board
2. Choose "Independent radius" for corner radius
3. Hover on any of the inputs and read the tooltip

### Expected behavior

The input should tell which corner it will apply the radius to. This behavior is seen in "Independent padding", and margins in flex layout.

### Actual behavior

The tooltip doesn't say which corner the input affects. All 4 of the tooltips read "Radius"

### Screenshots or video

![image](https://github.com/penpot/penpot/assets/95227070/4ff8d041-dee8-47db-bda4-5e50b9114e03)

### Desktop (please complete the following information)

- OS: Arch Linux
- Browser: Firefox 114.0.1

### Environment (please complete the following information)

- Host: design.penpot.app

superalex:
Thank you very much @astudentinearth 

I've just created this Taiga issue to track it -> https://tree.taiga.io/project/penpot/issue/5486

PR #3332, opened 2023-06-21
:bug: Change independent corner radius input tooltips

astudentinearth:
Make the inputs show a tooltip for the relevant corner(e.g. "Top left") instead of "Radius".
Completes #3305 ([Taiga #5486](https://tree.taiga.io/project/penpot/issue/5486))
This pull request does not introduce breaking changes.

After the change the tooltips are shown like this:
image

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/options/menus/measures.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/menus/measures.cljs
@@ -412,7 +412,7 @@
 
             (= radius-mode :radius-4)
             [:*
-             [:div.input-element.mini {:title (tr "workspace.options.radius")}
+             [:div.input-element.mini {:title (tr "workspace.options.interaction-pos-top-left")}
               [:> numeric-input
                {:placeholder "--"
                 :min 0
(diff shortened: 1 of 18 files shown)


superalex:
Hey!, thank you very much @astudentinearth 

As radius and interactions are two different concepts I think it would be a good idea using new translation keys. Even if right now they have the same translation,  if in the future they change it will affect this too.

You could create the new keys for every language and copy the translations from each language.

astudentinearth:
@superalex I understand the concerns about translation keys. I will look into it once I'm available.

superalex:
Great!, thank you very much :)

astudentinearth:
I have created the new language keys, however I synced my fork branch with `origin/develop` on the web interface, so there is a merge commit in between. Would that cause a problem?

superalex:
Thank you very much!, I've just merged manually this :)

astudentinearth:
@superalex Thanks a lot. This was my first ever pull request❤️

superalex:
> @superalex Thanks a lot. This was my first ever pull requestheart

It is a great honour for us! <3 <3 <3

PR #3333, opened 2023-06-21
:bug: Fix Internal server error occurred after clicking on '3 dots' m…

pabloalba:
…enu of copy component on Design tab

Fixes https://tree.taiga.io/project/penpot/issue/5495

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
@@ -245,9 +245,9 @@
 
                                   [[(tr "workspace.shape.menu.detach-instance") do-detach-component]
                                    (when can-update-main?
-                                     [:*
-                                      [(tr "workspace.shape.menu.reset-overrides") do-reset-component]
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
```diff
@@ -245,9 +245,8 @@
 
                                   [[(tr "workspace.shape.menu.detach-instance") do-detach-component]
                                    (when can-update-main?
-                                     [:*
-                                      [(tr "workspace.shape.menu.reset-overrides") do-reset-component]
-                                      [(tr "workspace.shape.menu.update-main") do-update-component]])
+                                     [(tr "workspace.shape.menu.reset-overrides") do-reset-component]
```
niwinz: This line is ignored on `when`, this is why the `[:*` is used previously, to return both items.

pabloalba: You are right. But I can't use [:* because it will be detected as a submenu. I've made an alternative version...

PR #3395, opened 2023-07-10
Fix problem with thumbnails alpha channel

Alotor:

> Screenshots/GIFs


frontend/src/app/main/data/workspace/thumbnails.cljs
```diff
@@ -31,16 +34,38 @@
        (rx/filter #(= % id))
        (rx/take 1)))
 
+(defn node->ibmp
+  [node]
+  (p/resolved
```
niwinz, line 39: Use `rx/from` directly instead promesa here.

PR #3409, opened 2023-07-12
:bug: Fix rotate several elements in bulk

superalex:
Related to:
- https://tree.taiga.io/project/penpot/issue/5165

### DIFF
--- a/frontend/src/app/main/data/workspace/modifiers.cljs
+++ b/frontend/src/app/main/data/workspace/modifiers.cljs
@@ -371,6 +371,32 @@
 
          (assoc state :workspace-modifiers modif-tree))))))
 
+;; This function is similar to set-rotation-modifiers but:
+;; - It consideres the center for everyshape instead of the center of the total selrect
+;; - The angle param is the desired final value, not a delta
+(defn set-delta-rotation-modifiers
+  ([angle shapes]
+   (ptk/reify ::set-delta-rotation-modifiers
+     ptk/UpdateEvent
+     (update [_ state]
+       (let [objects     (wsh/lookup-page-objects state)
+             ids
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/data/workspace/modifiers.cljs
```diff
@@ -348,9 +348,12 @@
 ;; Rotation use different algorithm to calculate children modifiers (and do not use child constraints).
 (defn set-rotation-modifiers
   ([angle shapes]
-   (set-rotation-modifiers angle shapes (-> shapes gsh/selection-rect gsh/center-selrect)))
+   (set-rotation-modifiers angle shapes {}))
 
-  ([angle shapes center]
+  ([angle shapes {:keys [as-delta?] :or {as-delta? false}}]
```
hirunatan: I don't understand this. If someone was calling this function with the 3-arity variant `(set-rotation-modifiers angle shapes center)`, the center will now be interpreted as an invalid options argument.

frontend/src/app/main/data/workspace/modifiers.cljs
```diff
@@ -348,9 +348,12 @@
...
+   (set-rotation-modifiers angle shapes (-> shapes gsh/selection-rect gsh/center-selrect) {:as-delta? as-delta?}))
```
hirunatan: In case of changing rotation of several shapes, ¿do we need to do it relative to the selection center? ¿Or should each shape be rotated from its own center?

superalex: I think in this case it makes more sense to rotate each shape from it's own center but I'll speak about it with UX/UI. Thanks!

Issue #3377, opened 2023-07-04
UX Issue: Layer list should not scroll to the selected item


iamnbutler:
### Is your feature request related to a problem? Please describe.

Apologies if there is a pre existing issue for this, I did some searching but couldn't find one.

Selecting things on the canvas should not run a scroll animation to the selected item in the layer list – While this is visually interesting the first time, it doesn't bring value when navigating, is visually distracting, and creates unnecessary performance overhead.

When moving around the canvas a lot, and when you have many layers, often the animation won't even finish before you go to interact with the layer list. This created even more performance overhead, as all the hover animations start firing as the scroll continues.

I don't think there is enough positive value to justify the tradeoffs of scrolling the layer list like this.

### Describe the solution you'd like.

Selecting an item on the canvas should simply jump the layer list to where that item is, rather than animating it.

elhombretecla:
Hi @iamnbutler thank you so much for your feedback :)

We are going to review this enhancement as soon as we can. Here's the Taiga's link to track it https://tree.taiga.io/project/penpot/issue/5591

KristijanZic:
I think it's quite nice to track your way around the project. I do have my elements in groups but I presume the more elements you have the more the returns form animating are diminished. Maybe a better ux would be to just jump if the element distance is more than n tiles away from the current element. But animation is still quite nice for most projects I'd say.

PR #3414, opened 2023-07-13
✨  Improve layers autoscroll

AzazelN28:
## Solution

When scroll distance ratio is less than 1 it smoothly scrolls between items, otherwise it scrolls instantly.

[Taiga #5591](https://tree.taiga.io/project/penpot/issue/5591)
[GitHub #3377](https://github.com/penpot/penpot/issues/3377)


iamnbutler:
Great change, thanks ❤️

PR #3435, opened 2023-07-24
:bug: Fix color palette not being updated

AzazelN28:
[Taiga Issue #5552](https://tree.taiga.io/project/penpot/issue/5552)


niwinz:
I'm not clearly understand the fix, I mean, the use-equal-memo hook on the worst case should be noop, because if the value is different it will always return a different reference.

PR #3450, opened 2023-07-26
:bug: Fix CTRL+Z in workspace select

AzazelN28:
## Steps to reproduce

1. Create new file
2. Add rectangle
3. Add stroke to rectangle
4. Change stroke position from outside to center
5. Press Ctrl+z

## Expected behavior

It should undo the last change

## Behavior

Does nothing

## Fix

Mousetrap has a function called stopCallback that stops mousetrap internal event propagation if the returned value is true. I've added a new condition that checks if the element that triggers the event has the `data` `mousetrap-dont-stop`.

[Taiga Issue #5677](https://tree.taiga.io/project/penpot/issue/5677)

### DIFF
--- a/frontend/src/app/main/data/shortcuts_impl.js
+++ b/frontend/src/app/main/data/shortcuts_impl.js
@@ -17,17 +17,20 @@ if (Mousetrap.addKeycodes) {
 
 const target = Mousetrap.prototype || Mousetrap;
 target.stopCallback = function(e, element, combo) {
-  // if the element has the class "mousetrap" then no need to stop
-  if ((' ' + element.className + ' ').indexOf(' mousetrap ') > -1) {
-      return false;
+  // if the element has the data attribute "mousetrap-dont-stop" then no need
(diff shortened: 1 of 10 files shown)


frontend/src/app/main/data/shortcuts_impl.js
```diff
@@ -17,17 +17,21 @@ if (Mousetrap.addKeycodes) {
...
-         element.tagName == 'SELECT' || 
-         element.tagName == 'TEXTAREA' || 
-         (element.tagName == 'BUTTON' && combo.includes("tab")) ||
-         (element.contentEditable && element.contentEditable == 'true');
+  const shouldStop = element.tagName == "INPUT" ||
+    element.tagName == "SELECT" ||
+    element.tagName == "TEXTAREA" ||
+    (element.tagName == "BUTTON" && combo.includes("tab")) ||
+    (element.contentEditable && element.contentEditable == "true");
+  console.log("shouldStop", e, element, combo, shouldStop);
```
superalex: Should we remove this console.log?

PR #3451, opened 2023-07-26
:bug: Fix texts on deleteunpublish library

pabloalba:
Fixes https://tree.taiga.io/project/penpot/issue/5349


niwinz:
I think we need to delete the translation strings from all other locales.

pabloalba:
> I think we need to delete the translation strings from all other locales.

Done

PR #3487, opened 2023-08-03
:bug: Reset component is now against remote main

pabloalba:


common/src/app/common/types/file.cljc
```diff
@@ -178,6 +178,34 @@
         (cph/get-children-with-self (:objects instance-page) (:main-instance-id component)))
       (vals (:objects component)))))
 
+
+(defn get-top-instance
```
hirunatan: This one uses only the objects tree and is related to components, so It should be on common.types.container

common/src/app/common/types/file.cljc
```diff
@@ -178,6 +178,34 @@
...
+    (if (= parent-id uuid/zero)
+      current-top
+      (get-top-instance objects parent current-top))))
+
+
+(defn find-remote-shape
+  "Recursively go back by the :shape-ref of the shape until find the correct shape of the original component"
+  [container libraries shape]
+  (let [top-instance        (get-top-instance (:objects container) shape nil)
+        component           (find-component libraries (:component-id top-instance) {:included-delete? true})
```
hirunatan: Here you can use the faster get-component, since you have component-file

frontend/src/app/main/data/workspace/libraries_helpers.cljs
```diff
@@ -532,13 +532,26 @@
...
   (let [shape-inst     (ctn/get-shape container shape-id)]
     (if (ctk/in-component-copy? shape-inst)
-      (let [library        (dm/get-in libraries [(:component-file shape-inst) :data])
-            component      (or (ctkl/get-component library (:component-id shape-inst))
-                               (and reset?
-                                    (ctkl/get-deleted-component library (:component-id shape-inst))))
-
-            shape-main     (when component
-                             (ctf/get-ref-shape library component shape-inst))
+      (let [redirect-shaperef ;;Set the :shape-ref of a shape pointing to the :id of its remote-shape
```
hirunatan, line 535: I think this function doesn't add much value. Perhaps the code will be clearer if you just put the contents inline in the two places they are used. What do you think?

PR #3535, opened 2023-08-18
:sparkles: Show a confirmation dialog when an user tries to publish a…

pabloalba:
…n empty library

Implements to https://tree.taiga.io/project/penpot/us/5294


hirunatan:
Is the confirmation not needed when publishing from the dashboard or the main file menu?

PR #3538, opened 2023-08-21
:sparkles: Add function to validate shape referential integrity

hirunatan:


common/src/app/common/types/file.cljc
```diff
@@ -162,21 +169,23 @@
...
-      (vals (:objects component)))))
+(defn find-ref-shape
+  "Locate the near component in the local file or libraries, and retrieve the shape
+   referenced by the instance shape."
+  [file page libraries shape]
+  (let [root-shape     (ctn/get-component-shape (:objects page) shape)
+        component-file (if (= (:component-file root-shape) (:id file))
+                         file
+                         (get libraries (:component-file root-shape)))
+        component      (when component-file
```
pabloalba, line 185: Why not use the new resolve-component ?

common/src/app/common/types/file.cljc
```diff
@@ -837,3 +855,282 @@
...
+                          shape file page))))))
+
+(defn validate-component-not-main-not-head
+  "Validate that this shape is not main instance and not head."
+  [shape file page errors]
+  (when (some? (:main-instance shape))
+    (validate-error errors :component-main
+                    (str/format "Shape not expected to be main instance")
+                    shape file page))
+  (when (or (some? (:component-id shape))
```
pabloalba, line 948: Use helper ctk/instance-head?

common/src/app/common/types/file.cljc
hirunatan: It should be good. But we need component-file to get the shape, later. So i will need to calculate it, anyway. And if I have the file, then it's easier to directly call get-component.

Issue #3546, opened 2023-08-22
bug: cropped API response body when using a faux objectId value


roberto-red:
### Steps To Reproduce

1. Generate an access token
2. Create a Penpot project with a page and annonate their file ID and page ID
3. Make a HTTP POST call to the Penpot API for the `get-page` RPC specifying:
   - an `Authorization: Token YOUR_ACCESS_TOKEN` header
   - an `Accept: application/json` header
   - a `Content-Type: application/json` header
   - the correct `fileId` param in the body object
   - the correct `pageId` param in the body object
   - an `objectId` param in the body object with the `pageId` UUID value (any valid UUID that doesn't match with an object will do)
3. (alternative) Make a HTTP call like the above, but with:
   - a `pageId` param in the body object with the `fileId` UUID value (any valid UUID that doesn't match with a page will do)
   - the correct `objectId` param in the body object

### Expected behavior

Receive a response with `404` HTTP code and a properly formatted JSON body (with `type`, `code` and `hint` properties)

### Actual behavior

The response has a `200` HTTP code and the body is a cropped, invalid JSON (mind the white-space):
```
{
	"options": {
		"guides": {
		}
	},
	"objects": {
		
```

or

```
{
	"objects": {
		
```

### Environment (please complete the following information)

- Host: https://design.penpot.app
- Penpot version: 1.19

### Additional context

Example call using cURL with environment variables:

``` 
curl --request POST \
  --url https://design.penpot.app/api/rpc/command/get-page \
  --header 'Accept: application/json' \
  --header "Authorization: Token $ACCESS_TOKEN" \
  --header 'Content-Type: application/json' \
  --data "{
	\"fileId\": \"$FILE_ID\",
	\"pageId\": \"$PAGE_ID\",
	\"objectId\": \"$PAGE_ID\"
}"
}'

niwinz:
Hi @roberto-red 
I found the issue. Some context: this is mainly because internally, when you specify the objectId with an uuid pointing to an not existing object; in this case we generate an object with `null` as key y `null` as value (this is clearly a bug, and we will fix it), and this on rendering to JSON, raises an error saying that JSON does not supports object keys as `null`. 

A workaround, does not specify the objectId parameter or specify it with a correct ID.
Meanwhile we fix the underlying issue.

roberto-red:
Sure, it's not blocking for me right now, just nice for runtime error handling and for overall Penpot correctness. Thank you, @niwinz!

[...]

PR #3551, opened 2023-08-23
:bug: Fix unexpected output on get-page when invalid object-id is pro…

niwinz:


niwinz:
Fixes https://github.com/penpot/penpot/issues/3546

PR #3560, opened 2023-08-25
:zap: Add performance optimizations

niwinz:


common/src/app/common/geom/shapes/intersect.cljc
```diff
@@ -335,13 +337,32 @@
...
-(defn has-point?
-  "Check if the shape contains a point"
+(defn slow-has-point?
   [shape point]
-  (let [lines (points->lines (:points shape))]
-    ;; TODO: Will only work for simple shapes
+  (let [lines (points->lines (dm/get-prop shape :points))]
     (is-point-inside-evenodd? point lines)))
 
+(defn fast-has-point?
```
hirunatan, line 343: Will this give the correct result? This is checking if the point is inside the bounding rect of the shape. But for circles, for example, a point may be inside the rect but outside the shape (e.g. in a corner).

niwinz: Agree, and I will fix it. This change is because this function is mostly used for frames, which behave like rects.
Also, this function does not covers rotation, but this is how it worked until now so I don't go to change it.

niwinz: Added circle to the slow path

common/src/app/common/types/shape_tree.cljc
```diff
@@ -240,6 +240,22 @@
                -1))]
      (sort comp ids))))
 
+(defn sort-z-index-objets
```
hirunatan: If unstable-sort is faster than default sort, didn't it be useful to replace it in all calls of sort-z-index?

Or what tradeoffs does it have?

common/src/app/common/types/shape_tree.cljc
```diff
@@ -256,23 +256,27 @@
...
-     (or top-frame uuid/zero))))
-
-(defn frame-by-position
-  ([objects position] (frame-by-position objects position nil))
+   (dm/assert!
+    "expected a point"
+    (gpt/point? position))
+
+   (let [frames    (get-frames objects options)
+         frames    (sort-z-index-objets objects frames options)]
```
hirunatan: Typo: "objets"

common/src/app/common/types/shape_tree.cljc
niwinz: The stable vs unstable algorithms differs in how equal elements are treated. Stable algorithm preserves relative position of equal elements, unsable does not cares about that.
In our case, this is pretty irrelevant and using unsable algorithm removes a O(n) loop from sorting.
I have not changed it on sort-z-index just because I want to touch a small portion of code, but yes, I put a note here to replace it also on sort-z-indez impl

PR #3573, opened 2023-08-30
:bug: Correctly initialize geometry when creating a new group

hirunatan:
This was the cause of new components not visible in asset tabs in components-v1, until resized.

### DIFF
--- a/frontend/src/app/main/data/workspace/groups.cljs
+++ b/frontend/src/app/main/data/workspace/groups.cljs
@@ -91,6 +91,10 @@
                                     :name gname
                                     :shapes (mapv :id shapes)
                                     :selrect selrect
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/data/workspace/groups.cljs
```diff
@@ -90,7 +90,10 @@
         group     (cts/setup-shape {:type :group
                                     :name gname
                                     :shapes (mapv :id shapes)
-                                    :selrect selrect
+                                    :x (:x selrect)
```
niwinz, line 94: You also can leave selrect here, so it will be reused instead of creating a new one

hirunatan: done

PR #3575, opened 2023-08-30
:paperclip: Remove innecesary message on delete shared dialog

pabloalba:


niwinz:
I think the same messages should be deleted from all other translation files

PR #3589, opened 2023-09-01
:bug: Fix clone operaton of dbg handler

niwinz:


niwinz:
The linter fails because develop has  a linter issue

PR #3596, opened 2023-09-04
:sparkles: Assets groups review

pabloalba:
Implements https://tree.taiga.io/project/penpot/issue/5549 and also fixes https://tree.taiga.io/project/penpot/issue/5480

### DIFF
--- a/backend/src/app/srepl/fixes.clj
+++ b/backend/src/app/srepl/fixes.clj
@@ -281,3 +281,50 @@
                      {:id (:id file)})
 
          (files/persist-pointers! h/*conn* (:id file)))))))
+
+(defn fix-main-shape-name
+  ([file]
+   (prn (str "Updating " (:name file) " " (:id file)))
+   (if-not (contains? (:features file) "components/v2")
+     (do
+       (prn "   This file is not v2")
+       file)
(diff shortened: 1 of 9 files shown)


backend/src/app/srepl/fixes.clj
```diff
@@ -279,3 +279,48 @@
...
+           (fn [page]
+             (prn (str "Page " (:name page)))
+             (h/update-shapes page update-shape))]
+
+       (update file :data h/update-pages update-page))))
+
+  ([file save?]
+   (let [file (-> file
+                  (update :data blob/decode)
+                  (fix-main-shape-name))]
```
niwinz, line 321: I don't know how this fix function can be used but it has at least two flaws:
- The arity 1 of fix-main-shape-name can return nil if the file is not components-v2 that can just delete the file data
- The arity 2 handles pointers but does not setup the registry binding for delimit the pointers, so on large number of files, the last file will persist all pointers of all the other files processed.

niwinz: if this will be used with `h/update-file` then the pointer handling is already covered.
also, if you don't touch anything related to features maybe you just can skip encoding and decoding them

PR #3602, opened 2023-09-05
:zap: Allow passing overrides to frontend nginx config

m90:

This allows Docker users to mount an arbitrary number of additional nginx config files at `/etc/nginx/overrides.d` which will then be included in the configuration used. Passed config files take precedence over both the directives in the `http` block. The following `server` block could also be overwritten as the first match takes precedence.

Example in a compose setup:

```yml
  penpot_frontend:
    image: penpotapp/frontend:1.19.1
    networks:
      - penpot
      - penpot_public
    restart: unless-stopped
    volumes:
      - penpot_assets:/opt/data/assets
      - ./logging.conf:/etc/nginx/overrides.d/01logging.conf
    depends_on:
      - penpot_backend
      - penpot_exporter
    environment:
      PENPOT_FLAGS: >-
        disable-registration
        disable-login
        enable-login-with-password
      PENPOT_REDIS_URI: redis://penpot_redis/0
      PENPOT_BACKEND_URI: http://penpot_backend:6060
      PENPOT_EXPORTER_URI: http://penpot_exporter:6061
```

where `logging.conf` looks like

```
access_log off;
```

Users that don't use this feature will not be affected and do not need to change anything.

I would like to disable access logging for the `frontend` container as it is behind another reverse proxy anyways and logs can get quite noisy/heavy. `access_log /dev/stdout` is hard coded in the nginx config, so there is no way of configuring this currently.

----

I am not entirely sure if documentation lives in this repo as well and needs to be updated. In case yes, please lmk and I can do this still.

### DIFF
--- a/docker/images/files/nginx.conf
+++ b/docker/images/files/nginx.conf
@@ -50,6 +50,8 @@ http {
     proxy_cache_valid any 48h;
     proxy_cache_key "$host$request_uri";
 
+    include /etc/nginx/overrides.d/*.conf;
+
     server {
         listen 80 default_server;
         server_name _;


niwinz:
Looks nice. I will target it to staging, for make it available faster for docker users.

niwinz:
Merged directly to staging using command line, thanks.

niwinz:
Feel free to open an other PR on penpot-docs if you consider that documentation should be updated

PR #3605, opened 2023-09-07
:bug: Fix export file with components as basic objects

pabloalba:
Fixes https://tree.taiga.io/project/penpot/issue/5740

### DIFF
--- a/common/src/app/common/types/file.cljc
+++ b/common/src/app/common/types/file.cljc
@@ -12,8 +12,10 @@
    [app.common.files.features :as ffeat]
    [app.common.geom.point :as gpt]
    [app.common.geom.shapes :as gsh]
+   [app.common.logging :as l]
    [app.common.pages.helpers :as cph]
    [app.common.schema :as sm]
+   [app.common.text :as ct]
    [app.common.types.color :as ctc]
    [app.common.types.colors-list :as ctcl]
    [app.common.types.component :as ctk]
@@ -981,7 +983,7 @@
       (report-error :missing-component-root
                     (str/format "Referenced shape %s not found in near component" (:shape-ref shape))
                     shape file page))))
- 
+
 (defn validate-component-not-ref
   "Validate that this shape does not reference other one."
   [shape file page report-error]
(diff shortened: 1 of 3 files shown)


common/src/app/common/types/file.cljc
```diff
@@ -1149,3 +1151,65 @@
...
+
+        detach-objects
+        (fn [objects]
+          (->> objects
+               (d/mapm #(detach-shape objects %2))))
+
+        detach-pages
+        (fn [pages-index]
+          (->> pages-index
+               (d/mapm
```
niwinz: Consider using `update-vals` here instead of `d/mapm`
```clojure
(update-vals pages-index #(update % :objects detach-objects))
```

common/src/app/common/types/file.cljc
```diff
@@ -1149,3 +1151,65 @@
...
+            (not= file-id (get-component-ref-file objects shape))
+            (dissoc :component-id :component-file :shape-ref :component-root)
+
+            (= :text (:type shape))
+            (update :content detach-text)))
+
+        detach-objects
+        (fn [objects]
+          (->> objects
+               (d/mapm #(detach-shape objects %2))))
```
niwinz: The same here:
`(update-vals objects #(detach-shape objects %))`

common/src/app/common/types/file.cljc
```diff
@@ -1149,3 +1151,65 @@
...
+                   (dissoc :fill-color-ref-id :fill-color-ref-file)
+
+                   (not= file-id (:typography-ref-file %))
+                   (dissoc :typography-ref-id :typography-ref-file)))))
+
+        detach-shape
+        (fn [objects shape]
+          (l/info :msg "detach-shape")
+          (l/info :msg file-id)
+          (l/info :msg (get-component-ref-file objects shape))
```
niwinz: Please, use a single log line, in the same style as it is used in the backend binfile ns
The `info` logging level also looks incorrect, it should be probably `debug` or `trace` (check the binfile ns and use it in the same way)

common/src/app/common/types/file.cljc
pabloalba: Fixed

common/src/app/common/types/file.cljc
pabloalba: Fixed

PR #3612, opened 2023-09-08
:recycle: Refactor thumbnails

AzazelN28:
### Refactor

Instead of rendering all the frame thumbnails, now called imposters (in honor of the similar [rendering technique]( https://developer.nvidia.com/gpugems/gpugems3/part-iv-image-effects/chapter-21-true-impostors)) using DOM nodes, we use `react-dom/server` to render components directly as strings and then pass those strings to the `rasterizer` (`thumbnail_renderer` before).

:bulb: The initial idea was to use `react-dom/server` `renderToStaticMarkup`, but I think that using `renderToReadableStream` would make the app more responsive.

NOTE: The `renderToStaticMarkup` is working but right now the `renderToReadableStream` is showing the following warning and it is not working: 
image


frontend/src/app/main/imposters.cljs
```diff
@@ -0,0 +1,82 @@
...
+         x                  (dm/get-prop bounds :x)
+         y                  (dm/get-prop bounds :y)
+         width              (dm/get-prop bounds :width)
+         height             (dm/get-prop bounds :height)
+
+        ;;  bounds (:selrect shape)
+        ;;  x (:x bounds)
+        ;;  y (:y bounds)
+        ;;  width (:width bounds)
+        ;;  height (:height bounds)
```
niwinz: :fire:

frontend/src/app/main/imposters.cljs
```diff
@@ -0,0 +1,82 @@
...
+          (rx/catch rx/empty)
+          (rx/map (fn [styles] #js {:id object-id
+                                    :data data
+                                    :viewbox viewbox
+                                    :width fixed-width
+                                    :height fixed-height
+                                    :styles styles}))))))
+
+(defn render-by-id
+  "Render the shape by its id (IMPORTANT! id as uuid, not string)"
```
niwinz, line 66: Add an `(dm/asset! "expected uuid" (uuid? id))`

frontend/src/app/main/ui/workspace/viewport/pixel_overlay.cljs
```diff
@@ -110,29 +48,39 @@
...
-                   area-data (.getImageData canvas-context (- x 25) (- y 20) 50 40)]
-               (-> (js/createImageBitmap area-data)
-                   (p/then
-                    (fn [image]
-                      ;; Draw area
-                      (obj/set! zoom-context "imageSmoothingEnabled" false)
-                      (.drawImage zoom-context image 0 0 200 160))))
-               (st/emit! (dwc/pick-color [r g b a]))))))
+           (when-let [image-data (mf/ref-val canvas-image-data)]
+             (when-let [zoom-view-node (.getElementById js/document "picker-detail")]
```
niwinz: Use `dom/` helpers for this operation

PR #3623, opened 2023-09-12
:wrench: Add script to fix touched attributes

hirunatan:


backend/src/app/srepl/fixes.clj
```diff
@@ -328,3 +329,58 @@
...
+                       (println "Page " (:name page))
+                       (h/update-shapes page (partial update-shape page)))]
+
+     (println "Updating " (:name file) (:id file))
+     (update file :data h/update-pages update-page)))
+
+  ([file save?]
+   (let [file (-> file
+                  (update :data blob/decode)
+                  (fix-component-root))]
```
pabloalba: It should be fix-touched

backend/src/app/srepl/fixes.clj
```diff
@@ -328,3 +329,58 @@
...
+                                    (let [equal?
+                                          (if (= group :geometry-group)
+                                            (if (#{:width :height} attr)
+                                              (gsh/close-attrs? attr (get shape attr) (get ref-shape attr) 1)
+                                              true)
+                                            (gsh/close-attrs? attr (get shape attr) (get ref-shape attr)))]
+                                      (when (and (not equal?) (not (cph/touched-group? shape group)))
+                                        (println " -> set touched " (:name shape) (:id shape) attr group))
+                                      (cond-> shape
+                                        (not equal?)
```
pabloalba: I think it should be the same that for the println:

(and (not equal?) (not (cph/touched-group? shape group))

backend/src/app/srepl/fixes.clj
hirunatan: fixed

backend/src/app/srepl/fixes.clj
hirunatan: It's not relevant, because the update would not have any effect. But you are true, it's better so. Changed.

PR #3636, opened 2023-09-15
🎉 Warn about empty libraries on the share library dialog

pabloalba:
Implements https://tree.taiga.io/project/penpot/issue/5884


backend/src/app/rpc/commands/files.clj
```diff
@@ -749,6 +749,25 @@
...
+
+(sv/defmethod ::get-file-summary
+  "Retrieve a file summary by its ID. Only authenticated users."
+  {::doc/added "1.20"
+   ::sm/params schema:get-file
+   ::sm/result schema:file-with-permissions}
+  [{:keys [::db/pool] :as cfg} {:keys [::rpc/profile-id id features project-id] :as params}]
+  (db/with-atomic [conn pool]
+    (let [perms (get-permissions conn profile-id id)]
+      (check-read-permissions! perms)
```
niwinz: If you don't need to access to list of permis, you can execute the check directly:
`(check-read-permissioons! conn profile-id id)` so you can remove one level let

backend/src/app/rpc/commands/files.clj
```diff
@@ -749,6 +749,25 @@
...
     (get-team-recent-files conn team-id)))
 
+
+;; --- COMMAND QUERY: get-file-summary
+
+(sv/defmethod ::get-file-summary
+  "Retrieve a file summary by its ID. Only authenticated users."
+  {::doc/added "1.20"
+   ::sm/params schema:get-file
+   ::sm/result schema:file-with-permissions}
```
niwinz: this metadata is incorrect, I mean, ::sm/result is for adding schema for the output, and the output is clearly not `filewithpermissions`. It is an optional parameter, so just delete it.

frontend/src/app/main/ui/workspace/header.cljs
```diff
@@ -362,14 +363,7 @@
...
          (mf/deps file-name add-shared-fn)
-         #(modal/show! {:type :confirm
-                        :message ""
-                        :title (tr "modals.add-shared-confirm.message" file-name)
-                        :hint (tr "modals.add-shared-confirm.hint")
-                        :cancel-label :omit
-                        :accept-label (tr "modals.add-shared-confirm.accept")
-                        :accept-style :primary
-                        :on-accept add-shared-fn}))
+         #(st/emit! (show-shared-dialog (:id file) add-shared-fn)))
```
niwinz: You are using `(:id file)` inside clusure, but you are not declaring it on the deps, consider using `file-id` on the deps, and seems like `file-name` on deps is not used anymore necessary

frontend/src/app/main/ui/workspace/left_header.cljs
```diff
@@ -477,14 +478,7 @@
...
          (mf/deps file-name add-shared-fn)
-         #(modal/show! {:type :confirm
-                        :message ""
-                        :title (tr "modals.add-shared-confirm.message" file-name)
-                        :hint (tr "modals.add-shared-confirm.hint")
-                        :cancel-label :omit
-                        :accept-label (tr "modals.add-shared-confirm.accept")
-                        :accept-style :primary
-                        :on-accept add-shared-fn}))
+         #(st/emit! (show-shared-dialog (:id file) add-shared-fn)))
```
niwinz: The same here, file-id missing on deps, and file-name is not necessary dep

PR #3650, opened 2023-09-20
Grid bugfixing

Alotor:


common/src/app/common/pages/changes.cljc
```diff
@@ -367,7 +367,7 @@
                  (comp first first))
 
                 new-shapes
-                (into [] (sort-by id->idx < old-shapes))]
+                (into [] (sort-by #(or (id->idx %) -1) < old-shapes))]
```
niwinz: Use `d/nilv` here, the `or` generates a bunch of closures, and for this simple case is better to use `d/nilv` which is more efficient.

PR #3678, opened 2023-09-29
Codegen & grid fixes

Alotor:
Fix some issues with code generation
Fix problems with the grid layout


AzazelN28:
Instead of `Improved` it should be `Improve`.

PR #3698, opened 2023-10-09
:zap: Schedule thumbnail component updates

AzazelN28:


frontend/src/app/main/data/workspace/libraries.cljs
```diff
@@ -48,7 +48,7 @@
    [potok.core :as ptk]))
 
 ;; Change this to :info :debug or :trace to debug this module, or :warn to reset to default
-(log/set-level! :warn)
+(log/set-level! :trace)
```
niwinz: Return it to `:warn` or `:info`

PR #3700, opened 2023-10-09
:sparkles: Load current version on file import

Alotor:

> Screenshots/GIFs

### DIFF
--- a/backend/src/app/rpc/commands/binfile.clj
+++ b/backend/src/app/rpc/commands/binfile.clj
@@ -10,6 +10,7 @@
    [app.common.data :as d]
    [app.common.data.macros :as dm]
    [app.common.exceptions :as ex]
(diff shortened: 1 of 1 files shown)


backend/src/app/rpc/commands/binfile.clj
```diff
@@ -960,12 +961,13 @@
   "Import a penpot file in a binary format."
   {::doc/added "1.15"
    ::webhooks/event? true}
-  [{:keys [::db/pool] :as cfg} {:keys [::rpc/profile-id project-id file] :as params}]
+  [{:keys [::db/pool] :as cfg} {:keys [::rpc/profile-id project-id file version] :as params}]
```
niwinz: I don't think we need to pass version as parameter, you can access the constant directly on runtime on the JVM accessing to the app.common.files.defaults

PR #3717, opened 2023-10-19
:tada: Add tag to thumbnails

AzazelN28:

> Screenshots/GIFs


backend/src/app/rpc/commands/files_thumbnails.clj
```diff
@@ -271,7 +288,7 @@
 (s/def ::upsert-file-object-thumbnail
   (s/keys :req [::rpc/profile-id]
           :req-un [::file-id ::object-id]
-          :opt-un [::data]))
+          :opt-un [::data :tag]))
```
Alotor: this need to be `::tag` and add 

```clojure
(s/def ::tag ::us/string)
```

PR #3743, opened 2023-10-31
Hiru components bugfix

hirunatan:
Includes https://tree.taiga.io/project/penpot/issue/5905
and other things

### DIFF
--- a/common/src/app/common/files/validate.cljc
+++ b/common/src/app/common/files/validate.cljc
@@ -125,17 +125,29 @@
                           :child-id child-id)))))))
 
 (defn validate-frame
-  "Validate that the frame-id shape exists and is indeed a frame."
+  "Validate that the frame-id shape exists and is indeed a frame. Also it must point to the
+   parent shape (if this is a frame) or to the frame-id of the parent (if not)."
   [shape file page]
   (let [frame (ctst/get-shape page (:frame-id shape))]
     (if (nil? frame)
       (report-error :frame-not-found
                     (str/format "Frame %s not found" (:frame-id shape))
                     shape file page)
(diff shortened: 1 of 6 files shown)


common/src/app/common/files/validate.cljc
```diff
@@ -125,17 +125,29 @@
...
                     (str/format "Frame %s not found" (:frame-id shape))
                     shape file page)
-      (when (not= (:type frame) :frame)
+      (if (not= (:type frame) :frame)
         (report-error :invalid-frame
                       (str/format "Frame %s is not actually a frame" (:frame-id shape))
-                      shape file page)))))
+                      shape file page)
+        (let [parent (ctst/get-shape page (:parent-id shape))]
+          (when (some? parent)
```
pabloalba, line 141: Shouldn't it be an error if the parent isn't found?

common/src/app/common/types/container.cljc
```diff
@@ -192,6 +192,23 @@
...
+  [root objects file-id]
+  (let [new-id       (uuid/next)
+        new-root     (assoc root
+                            :component-id new-id
+                            :component-file file-id
+                            :component-root true
+                            :main-instance true)
+        new-children (->> (cph/get-children objects (:id root))
+                          (map #(dissoc % :component-root)))]
+    [(assoc new-root :id new-id)
```
pabloalba, line 208: this new-root has the new id, and its the same as the component-id

common/src/app/common/files/validate.cljc
hirunatan: Not necessary, because this situation is already tested in `validate-parent-children`. There is no need to report the problem again.

PR #3746, opened 2023-10-31
:bug: Fix automatic frame assignment in clone-object

hirunatan:


common/src/app/common/types/shape_tree.cljc
```diff
@@ -354,29 +354,35 @@
...
-         ;; Assign the correct frame-id for the given parent.
-         ;; It's the parent-id (if frame) or the parent's frame-id otherwise.
+         ;; Assign the correct frame-id for the given parent. It's the parent-id (if parent is frame)
+         ;; or the parent's frame-id otherwise. Only for the first cloned shapes. In recursive calls
+         ;; this is not needed.
          frame-id
-         (if (cph/frame-shape? objects parent-id)
-           parent-id
-           (dm/get-in objects [parent-id :frame-id]))]
+         (if calc-frame?
```
Alotor: You can use a single `cond` like...

```clojure
(cond 
  (and calc-frame? (cph/frame-shape? objects parent-id)) parent-id
  calc-frame?  (dm/get-in objects [parent-id :frame-id])
  :else (:frame-id object))
````

hirunatan: ok

PR #3756, opened 2023-11-02
✨ : Strip extension for uploaded image files

belen-albeza:
Fixes https://tree.taiga.io/project/penpot/issue/5485

With this patch newly uploaded images will have their extension removed. They'll look like this (notice the absence of `.png` to the filenames)

Screenshot 2023-11-02 at 5 27 39 PM

### DIFF
--- a/common/src/app/common/files/builder.cljc
+++ b/common/src/app/common/files/builder.cljc
@@ -746,3 +746,7 @@
           :page-id page-id
           :option :guides
           :value new-guides}))))
+
+(defn strip-image-extension [filename]
+  (let [image-extensions-re #"(\.png)|(\.jpg)|(\.jpeg)|(\.webp)|(\.gif)|(\.svg)$"]
+    (str/replace filename image-extensions-re "")))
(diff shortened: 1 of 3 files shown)


belen-albeza:
**Questions**:

- Is having this separate utiles `files.cljs` file the right way to do this? Is there other file more appropiate?
  - Does `common/src/app/common/files/builder.cljs` have anything to do with this?
- Does changing the name when uploading have other consequences?

Thanks!

niwinz:
The commits messages should follow the commit guideline defined in CONTRIBUTING.md
Try to have autocontaining commits, in this case the linter fix should be squashed with the first commit
Example: `:sparkles: Strip extension for shape name on asset upload`

niwinz:
> **Questions**:
> 
> * Is having this separate utiles `files.cljs` file the right way to do this? Is there other file more appropiate?
>   
>   * Does `common/src/app/common/files/builder.cljs` have anything to do with this?

The code looks perfect, but I think as you mention, the file `common/src/app/common/files/builder.cljs` looks like a better place for that helper than creating `app.util.files` namespace on frontend module. So, I suggest to move this code to the `common/src/app/common/files/builder.cljs` file.

> * Does changing the name when uploading have other consequences?

I guess no, it looks ok.
> 
> Thanks!

belen-albeza:
Thanks @niwinz ! Just moved the code to the builder file and fixed the git history

niwinz:
Nice,
Just FYI, i have capitalized the commit message

PR #3768, opened 2023-11-06
:lipstick: Add new UI at modals

EvaMarco:
This PR solves this [task](https://tree.taiga.io/project/penpot/task/6036)


.vscode/settings.json
```diff
@@ -5,5 +5,6 @@
         "**/.lsp": true,
         "**/.shadow-cljs": true,
         "**/node_modules": true
-    }
+    },
```
superalex: Why is this included ?

EvaMarco: My bad, it should be on gitingnore and I didn't realize

PR #3775, opened 2023-11-07
Bugfixes

Alotor:
- :lipstick: Show rounded corners when rulers hiden
- :bug: Fix crash when multiple typographies selected
- :bug: Fix problem resizing texts
- :bug: Fix problem with layout sizing

### DIFF
--- a/common/src/app/common/geom/shapes/min_size_layout.cljc
+++ b/common/src/app/common/geom/shapes/min_size_layout.cljc
@@ -17,53 +17,53 @@
 
 (defn child-min-width
   [child child-bounds bounds objects]
-  (let [min-width
-        (cond
-          (and (ctl/fill-width? child) (ctl/flex-layout? child))
-          (let [children (cph/get-immediate-children objects (dm/get-prop child :id) {:remove-hidden true})]
-            (max (ctl/child-min-width child)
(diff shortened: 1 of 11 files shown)


common/src/app/common/geom/shapes/min_size_layout.cljc
```diff
@@ -17,53 +17,53 @@
...
-                layout-data (gd/calc-layout-data child @(get bounds (:id child)) children bounds objects true)]
-            (max (ctl/child-min-width child)
-                 (gpo/width-points (gb/layout-content-bounds bounds child layout-data))))
-          
-          (ctl/fill-width? child)
-          (ctl/child-min-width child)
+  (cond
+    (and (ctl/fill-width? child) (ctl/flex-layout? child))
+    (ctl/child-min-width child)
+    ;; Uncomment this to activate "auto" as min size
```
pabloalba, line 23: Why? Is it a TODO? Maybe let it depend on a variable, instead of comment/uncomment

Alotor: It's just something I don't want to lose (for the moment) but I don't think it's necessary to parametrize. I think as it's right now is fine

PR #3794, opened 2023-11-13
Add dev deps.edn in the project root

PEZ:
Making clojure-lsp serve the whole repository when opened at the root in, say, Calva.

> Please provide enough information so that others can review your pull request:

In https://github.com/BetterThanTomorrow/calva/issues/2337 it came to my attention that Calva's clojure-lsp services are not in effect if the repository is opened at its root. The way to solve that is to add a deps.edn at the root to help clojure-lsp figure out the class paths and thus which files to analyze.

> Explain the **details** for making this change. What existing problem does the pull request solve?

Only adding a `deps.edn` at the root which references the “sub” projects.

> Screenshots/GIFs

This only effects development. I could add some screenshots, but don't know if it would bring so much info. 😄 

Please let me know if I can clarify this.

PR #3802, opened 2023-11-15
:wrench: Add file with calva config to affect the entire project

EvaMarco:
Replaces the PR the [PEZ](https://github.com/PEZ) to meet the commits standard.
 [This was the original PR ](https://github.com/penpot/penpot/pull/3794)

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,0 +1,6 @@
+;; This file exists to make clojure-lsp serve the whole repository
+{:deps
+ {dev/frontend {:local/root "frontend"}
+  dev/backend {:local/root "backend"}
+  dev/common {:local/root "common"}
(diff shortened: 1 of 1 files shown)


PEZ:
> [Add file with calva config to affect the entire project](https://github.com/penpot/penpot/pull/3802/commits/5c8cf54b4381f7ddfeb06468154dc8068e0306af)

It could be good to know that it is not Calva config, per se. It is Clojure CLI config, and generally will help any `clojure` repl started at the project root to figure out the classpath. And it also helps clojure-lsp which uses `clojure` to assemble the classpath which it needs to know in order to analyze the right files and be able to deliver its services. And Calva integrates clojure-lsp.

PR #3803, opened 2023-11-15
:bug: Several bugfixes

niwinz:


backend/src/app/features/components_v2.clj
```diff
@@ -591,30 +591,48 @@
...
+        (some-> *file-stats* (swap! assoc :processed/graphics total)))
+
+      (let [factory  (px/thread-factory :virtual true)
+            executor (px/fixed-executor :parallelism 10 :factory factory)
+            process  (fn [mobj position]
+                       (let [tp1 (dt/tpoint)]
+                         (try
+                           (process-media-object fdata page-id mobj position)
+                           (catch Throwable cause
+                             (l/wrn :hint "unable to process file media object (skiping)"
```
pabloalba, line 608: This only should hint "skiping" when *skip-on-error*

niwinz: Ok, I think I can fix this in another MR

PR #3806, opened 2023-11-15
:tada: Import/export thumbnails in .penpot files

AzazelN28:
### How to test

- Import file without thumbnails (an old file or a file without any thumbnails)
- Import file with thumbnails
- Export file with thumbnails and import it on a version of Penpot that doesn't read thumbnails (it should work)

### DIFF
--- a/backend/src/app/rpc/commands/binfile.clj
+++ b/backend/src/app/rpc/commands/binfile.clj
@@ -336,6 +336,14 @@
       (->> (db/exec! conn [sql ids])
            (mapv #(assoc % :file-id id))))))
 
+(defn- get-file-thumbnails
+  "Return all file thumbnails for a given file."
+  [{:keys [::db/pool]} id]
+  (pu/with-open [conn (db/open pool)]
+    (let [sql "SELECT * FROM file_tagged_object_thumbnail WHERE file_id = ?"]
+      (->> (db/exec! conn [sql id])
+           (mapv #(dissoc % :file-id))))))
+
 (def ^:private storage-object-id-xf
   (comp
    (mapcat (juxt :media-id :thumbnail-id))
@@ -511,14 +519,19 @@
 
   (doseq [file-id (-> *state* deref :files)]
     (let [detach? (and (not embed-assets?) (not include-libraries?))
-          file    (cond-> (get-file cfg file-id)
-                    detach?
-                    (-> (ctf/detach-external-references file-id)
-                        (dissoc :libraries))
-                    embed-assets?
-                    (update :data embed-file-assets cfg file-id))
+          thumbnails (get-file-thumbnails cfg file-id)
+          file  (cond-> (get-file cfg file-id)
+                  detach?
+                  (-> (ctf/detach-external-references file-id)
+                      (dissoc :libraries))
 
-          media   (get-file-media cfg file)]
+                  embed-assets?
+                  (update :data embed-file-assets cfg file-id)
+
+                  :always
+                  (assoc :thumbnails thumbnails))
+
+          media (get-file-media cfg file)]
 
       (l/dbg :hint "write penpot file"
              :id file-id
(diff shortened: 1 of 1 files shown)


backend/src/app/rpc/commands/binfile.clj
```diff
@@ -336,11 +336,23 @@
       (->> (db/exec! conn [sql ids])
            (mapv #(assoc % :file-id id))))))
 
+(defn- get-file-tagged-object-thumbnails
```
niwinz: The function name does not need reflect the table name, in this case, the intention is obtain all thumbnails for a file so the function name should indicate that: `get-file-thumbnails`

backend/src/app/rpc/commands/binfile.clj
```diff
@@ -336,11 +336,23 @@
...
+  [{:keys [::db/pool]} id]
+  (pu/with-open [conn (db/open pool)]
+    (let [sql (str "SELECT * FROM file_tagged_object_thumbnail WHERE file_id = ?")]
```
niwinz: You don't need call `str` here, if you only have one string segment and no need concat nothing

backend/src/app/rpc/commands/binfile.clj
```diff
@@ -336,11 +336,23 @@
...
+      (->> (db/exec! conn [sql id])
+           (mapv #(dissoc % :file-id))))))
+
 (def ^:private storage-object-id-xf
   (comp
    (mapcat (juxt :media-id :thumbnail-id))
    (filter uuid?)))
 
+(def ^:private storage-object-id-whatever
```
niwinz: You don't need to declare a new transducer, you can continue use the `storage-object-id-xf` for all purposes

backend/src/app/rpc/commands/binfile.clj
```diff
@@ -511,14 +523,19 @@
...
+          file  (cond-> (get-file cfg file-id)
+                  detach?
+                  (-> (ctf/detach-external-references file-id)
+                      (dissoc :libraries))
+
+                  embed-assets?
+                  (update :data embed-file-assets cfg file-id)
 
-          media   (get-file-media cfg file)]
+                  true
```
niwinz: The convention is use `:always` keyword for indicate that this branch should be executed always

backend/src/app/rpc/commands/binfile.clj
```diff
@@ -534,7 +551,8 @@
         (write-obj! file)
         (write-obj! media))
 
-      (vswap! *state* update :sids into storage-object-id-xf media))))
+      (vswap! *state* update :sids into storage-object-id-xf media)
+      (vswap! *state* update :sids into storage-object-id-whatever thumbnails))))
```
niwinz: Said previously, just use the `storage-object-id-xf` here

backend/src/app/rpc/commands/binfile.clj
```diff
@@ -705,6 +736,11 @@
                   :expected-id expected-file-id
                   :hint "the penpot file seems corrupt, found unexpected uuid (file-id)"))
 
+      (when (contains? file :thumbnails)
+        (l/dbg :hint "updating thumbnails" ::l/sync? true)
+        (vswap! *state* update :thumbnails into (update-file-thumbnails file file-id'))
+        (l/dbg :hint "updated thumbnails" :thumbnails (:thumbnails @*state*) ::l/sync? true))
```
niwinz: this log entry can be very verbose when a file has a huge amount of thumbnails, consider using a single log entry and maybe using a counter of thumbnails instead of two

backend/src/app/rpc/commands/binfile.clj
```diff
@@ -755,6 +792,8 @@
                          (update :features #(db/create-array conn "text" %))
                          (update :data blob/encode))]
 
+          (l/inf :hint "importing thumbnails" :thumbnails (:thumbnails file) ::l/sync? true)
```
niwinz: The same here, don't put all thumbnails in the log entry, too much verbose

backend/src/app/rpc/commands/binfile.clj
```diff
@@ -837,7 +876,17 @@
...
+                      {:on-conflict-do-nothing overwrite?}))))
+
+    (doseq [item (:thumbnails @*state*)]
+      (l/dbg :hint "inserting file tagged object thumbnail"
+             :file-id (:file-id item)
+             :object-id (:object-id item)
+             ::l/sync? true)
+      (db/insert! conn :file-tagged-object-thumbnail
+                  (-> item
+                      (d/update-when :media-id lookup-index))
```
niwinz: update-when here is not necessary, because the field is always present

PR #3827, opened 2023-11-21
Mini fixes in components

hirunatan:


common/src/app/common/files/validate.cljc
```diff
@@ -461,15 +461,16 @@
   "Validate full referential integrity and semantic coherence on file data.
 
   Raises a validation exception on first error found."
-  [{:keys [data] :as file} libraries]
+  [{:keys [data features] :as file} libraries]
+  (when (features "components/v2")
```
niwinz: use `contains?`, it is safer option

PR #3836, opened 2023-11-22
:zap: Use subdomain for rasterizer (OOPIF)

AzazelN28:
### How to test it (locally)

1. Add a subdomain to `/etc/hosts`

```
127.0.0.1	localhost
127.0.0.1	sub.localhost
```

2. Then add these settings to `frontend/resources/public/js/config.js`:

```javascript
var penpotPublicURI = "http://localhost:3449";
var penpotRasterizerURI = "http://sub.localhost:3449";
```

3. Done!

Now we can check that we have at least one extra thread for the rasterizer iframe in DevTools.

BEFORE (iframe runs in main thread)
Screenshot from 2023-11-27 15-08-06

AFTER (iframe now runs in separate thread http:/sub.localhost:3449)
Screenshot from 2023-11-27 15-09-06

### DIFF
--- a/frontend/src/app/rasterizer.cljs
+++ b/frontend/src/app/rasterizer.cljs
@@ -221,6 +221,10 @@
             scope   (unchecked-get evdata "scope")]
         (when (and (some? payload)
                    (= scope "penpot/rasterizer"))
+          (log/dbg :hint "received message"
+                   :id id
+                   :payload payload
+                   :scope scope)
           (->> (render payload)
                (rx/subs (partial send-success! id)
                         (partial send-failure! id))))))))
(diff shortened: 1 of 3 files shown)


frontend/src/app/rasterizer.cljs
```diff
@@ -160,7 +160,10 @@
 (defn- svg-prepare
   "Prepares an SVG for rendering (resolves images to Data URIs and adds intrinsic size)."
   [data styles width]
-  (let [svg (svg-parse data)]
+  (let [svg (svg-parse data)
+        ;; TODO: Necesitamos hacer este reemplazo porque las URLs que se pasan desde el
```
niwinz: This TODO is still valid?

PR #3845, opened 2023-11-23
:recycle: Refactor thumbnails queue to priority queue

AzazelN28:
If the requested thumbnail is a frame it is always added to the front of the queue otherwise is added to the back.


frontend/src/app/util/queue.cljs
```diff
@@ -9,7 +9,7 @@
             [app.common.math :as mth]
             [app.util.time :as t]))
 
-(l/set-level! :info)
+(l/set-level! :debug)
```
niwinz: should be :info ?

PR #3849, opened 2023-11-23
:bug: Bugfixing

pabloalba:
Fixes:
https://tree.taiga.io/project/penpot/issue/6238
https://tree.taiga.io/project/penpot/issue/6242
https://tree.taiga.io/project/penpot/issue/6257

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/assets/components.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/assets/components.cljs
@@ -36,7 +36,7 @@
    [potok.core :as ptk]
    [rumext.v2 :as mf]))
 
-(def drag-data* (atom false))
+(def drag-data* (atom {:local? false}))
 
 (defn set-drag-data! [data]
   (reset! drag-data* data))
@@ -102,10 +102,12 @@
 
         on-component-double-click
         (mf/use-fn
-         (mf/deps file-id component-id)
+         (mf/deps file-id component local)
          (fn [event]
            (dom/stop-propagation event)
-           (st/emit! (dw/go-to-main-instance file-id component-id))))
+           (if local
+             (st/emit! (dw/go-to-component component-id))
+             (st/emit! (dwl/nav-to-component-file file-id component)))))
 
         on-drop
         (mf/use-fn
@@ -519,7 +521,8 @@
          (fn [component-id event]
            (dom/prevent-default event)
            (let [pos (dom/get-client-position event)]
-             (when (and local? (not read-only?))
+
+             (when (not read-only?)
                (when-not (contains? selected component-id)
                  (on-clear-selection))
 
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/ui/workspace/sidebar/assets/components.cljs
```diff
@@ -580,7 +607,11 @@
          (mf/deps current-component-id file-id)
```
niwinz, line 607: Missing `local?` dep

frontend/src/app/main/ui/workspace/sidebar/assets/components.cljs
```diff
@@ -580,7 +607,11 @@
...
          (fn [event]
            (dom/stop-propagation event)
-           (st/emit! (dw/go-to-main-instance file-id current-component-id))))
+           (if local?
```
niwinz, line 610: Missing `local?` dep

frontend/src/app/main/ui/workspace/sidebar/assets/components.cljs
```diff
@@ -580,7 +607,11 @@
...
+             (st/emit! (dw/go-to-component current-component-id))
+             (let [component (-> (filter #(= (:id %) current-component-id) components)
```
niwinz: d/seek is better that filter + first

frontend/src/app/main/ui/workspace/sidebar/assets/components.cljs
```diff
@@ -571,6 +594,10 @@
         (mf/use-fn
          (mf/deps file-id)
          (fn [component event]
+           ;; dnd api only allow to acces to the dataTransfer data on on-drop (https://html.spec.whatwg.org/dev/dnd.html#concept-dnd-p)
+           ;; We need to know if the dragged element is from the local library on on-drag-enter, so we need to keep the info elsewhere
+           (set-drag-data! {:local? local?})
```
niwinz, line 599: I think we don't need wrap this in an object, can we use `local?` directly as value on the drag data?

niwinz: I mean, the `drag-data*` is initialized with `false` without object wrapping so I think we should be consistent on the data stored in this atom, it is ok to have plain boolean value or object containing a single prop.

pabloalba: Fixed in the initialization. I like it to be an object in case we need to keep more info there in the future.

frontend/src/app/main/ui/workspace/sidebar/assets/components.cljs
pabloalba: Cool! I didn't know that one

PR #3852, opened 2023-11-24
:tada: Add Storybook to the project

belen-albeza:
Fixes:

- https://tree.taiga.io/project/penpot/task/6192
- https://tree.taiga.io/project/penpot/task/6194

> TL;DR: This PR adds Storybook to the `frontend` project, using a dummy story example.

## To try out

Compile the design system exports:

```zsh
clojure -M:dev:shadow-cljs compile storybook
```

Run storybook:

```zsh
yarn run storybook
```

Open Storybook at http://localhost:6006

## Supported features

- Adding stories anywhere in the project in JSX format.
- Writing text docs anywhere in the project in MDX format (Markdown + JSX).
- Path aliases so we don't drown in `../../../` import paths within stories.

## Caveats & things to watch out for

- **Hot reloading does not work** reliably. In this case, touching or resaving the `public/resources/css/main.css` file should do the trick.
- Loading the stories is slow. My guess is that this is due to have to load all the JS files related to the CLJS runtime.
- We need to manually add the components we want to use within stories in the new `design_system.cljs` file.
- We need to manually re-export the icons as actual components in the `design_system` if we want to use those icons within the stories themselves (i.e. not within the components used in the stories)
- We need to manually decorate each story with the newly created `<StoryWrapper>` component.

## In upcoming PRs

- Use an actual Storybook decorator for the `<StoryWrapper>`
- Dockerize this so we don't need to run manually
- Run the build in CI and publish the storybook in a URL


frontend/.storybook/preview-head.html
```diff
@@ -0,0 +1 @@
+<link href="/css/main.css" rel="stylesheet" type="text/css" />
```
belen-albeza, line 1: I added this because

1) It does not interfere when hot reloading (when it's working); and
2) It prevents the components from not having any CSS at all in case of failure (sometimes when running storybook from scratch it would pick up the `main.css` styles at all if it failed previously to hot reloaded… must be some internal cache maybe?)

frontend/src/app/main/ui/components/design_system.cljs
```diff
@@ -0,0 +1,18 @@
...
+   [rumext.v2 :as mf]))
+
+(mf/defc story-wrapper
+  {::mf/wrap-props false}
+  [{:keys [children]}]
+    [:.default children])
+
+(def ^export default #js {
+  :icons #js {
+    :IconAddRefactor (mf/fnc icon-add-refactor [] icons/add-refactor)
```
belen-albeza: Just aliasing `IconAddRefactor` to `icons/add-refactor` didn't work. We need for the icon to be an actual React component instead of just raw HTML. Is there an easier way to do this?

frontend/vite.config.js
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from "vite";
+import { resolve } from "path";
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  resolve: {
+    alias: {
```
belen-albeza, line 7: For some reason, using the `viteFinal` config option in `.storybook/main.js` wasn't working, but this does…

frontend/src/app/main/ui/components/design_system.cljs
niwinz: It should be react component react element?

niwinz: Also, consider indenting this file using standard clojure formating

PR #3857, opened 2023-11-27
:tada: Add clojure fmt linter and fixer

niwinz:


.circleci/config.yml
```diff
@@ -28,38 +28,28 @@ jobs:
...
       - run: cd .clj-kondo && cat config.edn
+      - run: cat .cljfmt.edn
 
-      - run:
-          name: frontend styles prettier
-          working_directory: "./frontend"
-          command: |
-            yarn install
-            yarn run lint:scss
+      # - run:
```
superalex, line 33: why is this commented?

niwinz: Because we do not want CI failing until we format all the code. And we can't  format all the code on this PR, because it probably will cause too many conflicts with ongoing work.

PR #3868, opened 2023-11-29
:zap: Cache rasterizer resources

AzazelN28:
Rasterizer now caches all resources (mainly images and fonts) as blobs.

### DIFF
--- a/frontend/src/app/rasterizer.cljs
+++ b/frontend/src/app/rasterizer.cljs
@@ -20,10 +20,12 @@
    [beicon.core :as rx]
    [cuerdas.core :as str]))
 
-(log/set-level! :trace)
+(log/set-level! :info)
 
 (declare send-success!)
 (declare send-failure!)
+
+(defonce data-uri-cache (js/Map.))
 
 (defonce parent-origin
   (dm/str cf/public-uri))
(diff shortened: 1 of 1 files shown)


frontend/src/app/rasterizer.cljs
```diff
@@ -86,13 +88,18 @@
...
+      (rx/from (.text blob)))
+    (->> (http/send! {:uri uri
+                      :response-type :blob
+                      :method :get
+                      :mode :cors
+                      :omit-default-headers true})
+         (rx/map :body)
+         (rx/mapcat wapi/read-file-as-data-url)
+         (rx/tap (fn [data-uri]
+                   (.set data-uri-cache uri (wapi/create-blob data-uri "text/plain")))))))
```
niwinz, line 102: Do we need really create a blob here? Can we cache the text?
An other option maybe is reusing the `:body` blob and use it as cache value instead of creating a new plaintext blob.
I don't know which impl is more efficient.

AzazelN28: Yes, I have tried both versions and the problem with using the `:body` directly on the cache map is that we force that information to be kept in memory, while if we create a Blob we gain two things: 
1. Blobs are deleted when closing the `iframe`.
2. Blobs can be "promoted" to disk when memory pressure is very high.

PR #3871, opened 2023-11-30
:bug: Fix importing frame components with fills from V1 to V2

superalex:
Related to https://tree.taiga.io/project/penpot/issue/6221

### DIFF
--- a/backend/src/app/features/components_v2.clj
+++ b/backend/src/app/features/components_v2.clj
@@ -227,11 +227,11 @@
                     [shape]
                     (if (or (nil? (:parent-id shape)) (ctk/instance-head? shape))
                       (assoc shape
-                             :type :frame           ; Old groups must be converted
-                             :fills []              ; to frames and conform to spec
-                             :hide-in-viewer true
-                             :rx 0
-                             :ry 0)
+                        :type :frame                  ; Old groups must be converted
+                        :fills (or (:fills shape) []) ; to frames and conform to spec
+                        :hide-in-viewer (or (:hide-in-viewer shape) true)
+                        :rx (or (:rx shape) 0)
+                        :ry (or (:ry shape) 0))
                       shape))]
             (-> file-data
                 (update :pages-index update-vals fix-container)
@@ -376,12 +376,7 @@
                        :main-instance true
                        :component-root true
                        :component-file component-file
-                       :component-id component-id
-                       :type :frame           ; Old groups must be converted
-                       :fills []              ; to frames and conform to spec
-                       :hide-in-viewer true
-                       :rx 0
-                       :ry 0)
+                       :component-id component-id)
 
                 (nil? (:frame-id shape))
                 (assoc :frame-id frame-id)


backend/src/app/features/components_v2.clj
```diff
@@ -225,8 +225,8 @@
                     [shape]
                     (if (ctk/instance-head? shape)
                       (assoc shape
-                             :type :frame           ; Old groups must be converted
-                             :fills []              ; to frames and conform to spec
+                             :type :frame                 ; Old groups must be converted
+                             :fills (or (:fills shape)[]) ; to frames and conform to spec
                              :hide-in-viewer true
                              :rx 0
                              :ry 0)
```
hirunatan, line 232: When I wrote this, i forgot that in v1 component roots can be frames, too.

It would be better to put all these five attributes inside a `(cond->` so they are applied only if the type is not already a frame.

backend/src/app/features/components_v2.clj
```diff
@@ -225,8 +225,8 @@
                     [shape]
                     (if (ctk/instance-head? shape)
```
hirunatan: When I wrote this, i forgot that in v1 component roots can be frames, too.

It would be better to apply this assoc only if is an instance head and the type is not already a frame.

backend/src/app/features/components_v2.clj
```diff
@@ -323,8 +323,8 @@
                        :component-root true
                        :component-file component-file
                        :component-id component-id
-                       :type :frame           ; Old groups must be converted
-                       :fills []              ; to frames and conform to spec
+                       :type :frame                 ; Old groups must be converted
+                       :fills (or (:fills shape)[]) ; to frames and conform to spec
                        :hide-in-viewer true
                        :rx 0
                        :ry 0)
```
hirunatan: I'm thinking these five attributes are not necessary now, since they have already been added above. Please remove them (type, fills, hide-in-viewer, rx and ry) and test that it should still work ok.

common/src/app/common/files/libraries_helpers.cljc
```diff
@@ -68,7 +68,7 @@
                                                        (map :id shapes)
                                                        nil
                                                        root-name
-                                                       true))]
```
hirunatan: I'm not sure this may have undesired side effects.

Have you checked all places where `generate-add-component` is called from? And this parameter is ok being changed to false in all cases?

backend/src/app/features/components_v2.clj
```diff
@@ -227,8 +227,8 @@
                     [shape]
                     (if (or (nil? (:parent-id shape)) (ctk/instance-head? shape))
                       (assoc shape
-                             :type :frame           ; Old groups must be converted
-                             :fills []              ; to frames and conform to spec
+                             :type :frame                 ; Old groups must be converted
+                             :fills (or (:fills shape)[]) ; to frames and conform to spec
                              :hide-in-viewer true
                              :rx 0
                              :ry 0)
```
hirunatan: I'd rather use a `(cond->` to set each attribute only if the value is null in the original shape. Think that if the shape was already a frame in v1, it may have values in those attributes, that we need to preserve.

superalex: great!

PR #3875, opened 2023-11-30
:bug: Features related bugfixes (part2)

niwinz:


backend/dev/user.clj
```diff
@@ -22,6 +22,7 @@
    [app.common.types.file :as ctf]
    [app.common.uuid :as uuid]
    [app.config :as cf]
+   [app.db :as db]
```
hirunatan, line 25: Is this used?

niwinz: Yep, is a `user` ns, is just for make it easy to access db from repl

hirunatan: true!

PR #3894, opened 2023-12-05
:bug: Fix color picker not rendering Latin1 svgs

AzazelN28:
The use of `str->data-uri` was preventing the Color Picker to generate the thumbnail image because `js/btoa` doesn't accept characters outside the `Latin1` range.

```clojure
(defn str->data-uri
  [str type]
  (assert (string? str))
  (let [b64 (-> str
                js/btoa)]
    (dm/str "data:" type ";base64," b64)))

(defn svg-node->data-uri
  [svg-node]
  (let [xml  (-> (js/XMLSerializer.)
                 (.serializeToString svg-node))
        data-uri (str->data-uri xml "image/svg+xml")]
    data-uri))
```

[Taiga Issue #6251](https://tree.taiga.io/project/penpot/issue/6251)

### DIFF
--- a/frontend/src/app/main/ui/workspace/viewport/pixel_overlay.cljs
+++ b/frontend/src/app/main/ui/workspace/viewport/pixel_overlay.cljs
@@ -7,6 +7,7 @@
 (ns app.main.ui.workspace.viewport.pixel-overlay
   (:require
    [app.common.data :as d]
+   [app.common.data.macros :as dm]
    [app.common.pages.helpers :as cph]
    [app.common.uuid :as uuid]
    [app.main.data.modal :as modal]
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/ui/workspace/viewport/pixel_overlay.cljs
```diff
@@ -43,15 +44,21 @@
...
   its data equivalent."
   [svg]
   (let [svg-clone (.cloneNode svg true)]
     (->> (resolve-svg-images! svg-clone)
-         (rx/map (fn [_] (dom/svg-node->data-uri svg-clone))))))
+         (rx/mapcat (fn [svg-node]
+                      (let [xml    (js/XMLSerializer.)
+                            xmlstr (.serializeToString xml svg-node)
+                            blob   (wapi/create-blob #js [xmlstr] #js {:type "image/svg+xml"})]
+                        (->> (rx/from (.text blob))
```
niwinz: Is this blob creation is necessary? I mean, xmlstr is already a string, which you store on the blob and then read it again. I'm missing something?

AzazelN28: Oops, sorry, my bad.

PR #3902, opened 2023-12-11
:tada: Improve validation and repair

superalex:
Related to https://tree.taiga.io/project/penpot/issue/5898

### DIFF
--- a/common/src/app/common/files/repair.cljc
+++ b/common/src/app/common/files/repair.cljc
@@ -68,7 +68,19 @@
 
 (defmethod repair-error :child-not-found
   [_ {:keys [shape page-id args] :as error} file-data _]
-  (log/info :hint "Repairing shape :child-not-found" :id (:id shape) :name (:name shape) :page-id page-id)
+  (let [repair-shape
+        (fn [parent-shape]
+          (log/debug :hint "  -> Remove child" :child-id (:child-id args))
+          (update parent-shape :shapes (fn [shapes]
+                                         (d/removev #(= (:child-id args) %) shapes))))]
+    (log/info :hint "Repairing shape :child-not-found" :id (:id shape) :name (:name shape) :page-id page-id)
+    (-> (pcb/empty-changes nil page-id)
+        (pcb/with-file-data file-data)
+        (pcb/update-shapes [(:id shape)] repair-shape))))
+
+(defmethod repair-error :invalid-parent
+  [_ {:keys [shape page-id args] :as error} file-data _]
+  (log/info :hint "Repairing shape :invalid-parent" :id (:id shape) :name (:name shape) :page-id page-id)
   (-> (pcb/empty-changes nil page-id)
       (pcb/with-file-data file-data)
       (pcb/change-parent (:parent-id args) [shape] nil {:component-swap true})))
--- a/common/src/app/common/files/validate.cljc
+++ b/common/src/app/common/files/validate.cljc
@@ -6,7 +6,6 @@
 
 (ns app.common.files.validate
   (:require
-   [app.common.data.macros :as dm]
    [app.common.exceptions :as ex]
    [app.common.files.helpers :as cfh]
    [app.common.schema :as sm]
@@ -35,6 +34,7 @@
     :invalid-main-instance-id
     :invalid-main-instance-page
     :invalid-main-instance
+    :invalid-parent
     :component-main
     :should-be-component-root
     :should-not-be-component-root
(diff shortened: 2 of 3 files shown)


common/src/app/common/files/validate.cljc
```diff
@@ -123,11 +123,17 @@
...
               (report-error! :child-not-found
                 (str/ffmt "Child % not found" child-id)
-                child file page
-                :parent-id (:id shape)))))))))
+                shape file page
+                :parent-id (:id shape)
+                :child-id child-id)
+              (when (not= (:parent-id child) (:id shape))
+                (report-error! :invalid-parent
+                  (str/ffmt "Invalid parent %" child-id)
```
hirunatan: In this trace it would be useful to show both the child-id and the parent-id of the child

hirunatan: For example `(str/ffmt "Child %s has invalid parent %2 child-id parent-id)`

common/src/app/common/files/repair.cljc
```diff
@@ -68,7 +68,19 @@
 
 (defmethod repair-error :child-not-found
   [_ {:keys [shape page-id args] :as error} file-data _]
-  (log/info :hint "Repairing shape :child-not-found" :id (:id shape) :name (:name shape) :page-id page-id)
+  (let [repair-shape
+        (fn [parent-shape]
+          (log/debug :hint "  -> Remove children from" :shapes (:child-id args))
```
hirunatan: I think it would be clearer `(log/debug :hint "  -> Remove child" :child-id (:child-id args))`

common/src/app/common/files/repair.cljc
```diff
@@ -68,7 +68,19 @@
...
+          (update parent-shape :shapes (fn [shapes]
+                                         (filterv #(not= (:child-id args) %) shapes))))]
```
hirunatan: ¿better to use `(d/removev ..`?

common/src/app/common/files/repair.cljc
```diff
@@ -68,7 +68,19 @@
...
+    (log/info :hint "Repairing base shape :child-not-found" :id (:child-id args) :name "page" :page-id page-id)
```
hirunatan: I prefer to stick to standard message here, and give more info in log/debug above
`(log/info :hint "Repairing shape :child-not-found" :id (:id shape) :name (:name shape) :page-id page-id)`

common/src/app/common/files/validate.cljc
```diff
@@ -178,9 +184,21 @@
...
-                         (str/ffmt "Main instance page of component % is not valid" (:component-id shape))
-                         shape file page))))))
+          (let [component-page (ctf/get-component-page (:data file) component)
+                main-component (ctst/get-shape component-page (:main-instance-id component))]
+            ;; We must check if the same component has main instances in different pages. 
+            ;; In that case one of those instances shouldn't be main
+            (if (and
+                  (not= (:id page) (:id component-page))
+                  (= (:main-instance main-component) (:main-instance shape))
+                  (= (:component-id main-component) (:component-id shape)))
```
hirunatan: If we have come to here, we already know that the component page is different, and the instance-id is irrelevant. I think it will be enough to check if `main-component` exists. If so, we know that the current shape is not expected to be a main.

common/src/app/common/files/validate.cljc
```diff
@@ -178,9 +184,21 @@
...
-          (report-error! :invalid-main-instance-page
-                         (str/ffmt "Main instance page of component % is not valid" (:component-id shape))
-                         shape file page))))))
+          (let [component-page (ctf/get-component-page (:data file) component)
+                main-component (ctst/get-shape component-page (:main-instance-id component))]
+            ;; We must check if the same component has main instances in different pages. 
+            ;; In that case one of those instances shouldn't be main
+            (if (and
+                  (not= (:id page) (:id component-page))
+                  (= (:main-instance main-component) (:main-instance shape))
```
pabloalba: Maybe simply (:main-instance shape). No need to compare with the :main-instance of the main-component, that should be always true (if it's not true, that's another error)

frontend/src/app/main/data/workspace.cljs
```diff
@@ -122,8 +121,7 @@
     ptk/WatchEvent
     (watch [_ _ _]
       (rx/of (fbc/fix-bool-contents)
-             (fdf/fix-deleted-fonts)
-             (fbs/fix-broken-shapes)))))
```
pabloalba: Are we sure that we don't want to do it on file open anymore?

common/src/app/common/files/validate.cljc
hirunatan: I think you have made the change in the wrong line :smile:

hirunatan: And forgot the parent-id

superalex: oh!, sorry!

PR #3921, opened 2023-12-13
:sparkles:  Add anidation constraints for components

superalex:
The only allowed situations for anidated components is copies inside main, the rest is not allowed (copies inside copies, main inside mains, mains inside copies...)

- Create component shouldn't produce any invalid situations, test ctrl+k and right click with:
  - Selecting shapes
  - Selecting nested shapes
  - Selecting components (mains and instances)
  - Selecting children of components (mains and instances)
  - Selecting parents of components (mains and instances)

- Copy + paste must generate only valid situations. Test copying shapes/instances/mains to the different situations listed for creation
- Drag & drop in workspace must generate only valid situations. Test copying shapes/instances/mains to the different situations listed for creation
- Drag & drop in layers must generate only valid situations. Test copying shapes/instances/mains to the different situations listed for creation

This should solve the following issues
- https://tree.taiga.io/project/penpot/issue/6293 (this situation is not allowed now)
- https://tree.taiga.io/project/penpot/issue/6387
- https://tree.taiga.io/project/penpot/issue/6478

Document with test cases: https://docs.google.com/spreadsheets/d/1rRGw3x07SsY91_D4yuTfPsBtvW2rZan2cAHPlspNrpU/edit?usp=sharing

### DIFF
--- a/common/src/app/common/files/validate.cljc
+++ b/common/src/app/common/files/validate.cljc
@@ -393,7 +393,8 @@
                   (check-shape-copy-root-top shape file page libraries)))
 
               (if (ctk/main-instance? shape)
-                (if (= context :not-component)
+                ;; mains can't be nested into mains
+                (if (or (= context :not-component) (= context :main-top))
                   (report-error :nested-main-not-allowed
                                 "Nested main component only allowed inside other component"
                                 shape file page)
--- a/common/src/app/common/types/container.cljc
+++ b/common/src/app/common/types/container.cljc
@@ -387,3 +387,49 @@
       (if (ctk/in-component-copy? parent)
         true
         (has-any-copy-parent? objects (:parent-id shape))))))
+
+(defn has-any-main?
+  "Check if the shape has any children or parent that is a main component."
+  [objects shape]
+  (let [children (cfh/get-children-with-self objects (:id shape))
+        parents  (cfh/get-parents objects (:id shape))]
+    (or
+     (some ctk/main-instance? children)
+     (some ctk/main-instance? parents))))
+
+(defn valid-shape-for-component?
+  "Check if a main component can be generated from this shape in terms of nested components:
+  - A main can't be the ancestor of another main
+  - A main can't be nested in copies"
+  [objects shape]
+  (and
+   (not (has-any-main? objects shape))
+   (not (has-any-copy-parent? objects shape))))
+
+(defn- invalid-structure-for-component?
+  "Check if the structure generated nesting children in parent is invalid in terms of nested components"
+  [objects parent children]
+  (let [selected-main-instance? (some true? (map #(has-any-main? objects %) children))
+        parent-in-component?    (in-any-component? objects parent)
+        comps-nesting-loop?     (not (->> children
+                                          (map #(cfh/components-nesting-loop? objects (:id %) (:id parent)))
+                                          (every? nil?)))]
+    (or
+      ;;We don't want to change the structure of component copies
+     (ctk/in-component-copy? parent)
+      ;; If we are moving something containing a main instance the container can't be part of a component (neither main nor copy)
+     (and selected-main-instance? parent-in-component?)
+      ;; Avoid placing a shape as a direct or indirect child of itself,
+      ;; or inside its main component if it's in a copy.
+     comps-nesting-loop?)))
+
+(defn find-valid-parent-and-frame-ids
+  "Navigate trough the ancestors until find one that is valid"
+  [parent-id objects children]
+  (let [parent (get objects parent-id)]
+    (if (invalid-structure-for-component? objects parent children)
+      (find-valid-parent-and-frame-ids (:parent-id parent) objects children)
+      [parent-id
+       (if (= :frame (:type parent))
+         parent-id
+         (:frame-id parent))])))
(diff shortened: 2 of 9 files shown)


common/src/app/common/files/validate.cljc
```diff
@@ -393,7 +393,8 @@
                   (check-shape-copy-root-top shape file page libraries)))
 
               (if (ctk/main-instance? shape)
-                (if (= context :not-component)
+                ;; mains can't be nested into mains
+                (if (or (= context :not-component) (= context :main-top))
```
hirunatan, line 397: I think that this case should always report error. If you reach here, this is a main instance and it's not root, so it must be forbidden.

The `check-shape-main-root-nested` function is no longer relevant (better not to delete it, but now it won't be called).

superalex: I think it's more resilient this way, we can have generated this situation in some files on PRE environments, this way those files can be fixed automatically too

common/src/app/common/types/container.cljc
```diff
@@ -385,3 +385,48 @@
...
+(defn has-any-main?
+  "Check if the shape has any children or parent that is a main component."
+  [objects shape]
+  (let [instance-root (get-instance-root objects shape)]
+    (if (or (ctk/main-instance? shape) (ctk/main-instance? instance-root))
+      true
+      (some true? (map #(has-any-main? objects (get objects %)) (:shapes shape))))))
+
+(defn valid-shape-for-component?
+  "Check if a main component can be generated from this shape in terms of component anidation:
```
hirunatan: We prefer the term "nesting" over "anidation", to use a consistent vocabulary.

See https://design.penpot.dev/#/workspace/9143035f-1952-806e-8002-3ff852ca3cd5/9143035f-1952-806e-8002-3ff859163e5d?page-id=1317f2cb-e04a-806a-8002-45253d6cb753

superalex: thanks!

frontend/src/app/main/data/workspace.cljs
```diff
@@ -1845,18 +1845,9 @@
                   tree-root            (get-tree-root-shapes pobjects)
                   only-one-root-shape? (and
                                         (< 1 (count pobjects))
-                                        (= 1 (count tree-root)))
-                  all-objects           (merge page-objects pobjects)
-                  comps-nesting-loop?   (not (->> (keys pobjects)
```
hirunatan, line 1850: Why do you remove this? Aren't those checks necessary now?

common/src/app/common/files/validate.cljc
hirunatan: If we have a file has any incorrect situation because it was edited before this change, the validate function must report it as error, precisely to be able to get it repaired, doesn't it?

frontend/src/app/main/data/workspace/transforms.cljs
```diff
@@ -587,16 +588,12 @@
                               [(assoc move-vector :x 0) :x]
 
                               :else
-                              [move-vector nil])
+                              [move-vector nil])]
 
-                            nesting-loop? (some #(cfh/components-nesting-loop? objects (:id %) target-frame) shapes)
-                            is-component-copy? (ctk/in-component-copy? (get objects target-frame))]
```
hirunatan, line 593: Are these checks done somewhere else? I can't find it.

common/src/app/common/files/validate.cljc
superalex: That's what we are doing, reporting this situation as an error

frontend/src/app/main/data/workspace.cljs
superalex: They are included on the posterior call to 
`[parent-id frame-id]   (ctn/find-valid-parent-and-frame-ids candidate-parent-id page-objects (vals objects))`

frontend/src/app/main/data/workspace/transforms.cljs
superalex: yes, they are implicit in the call to:
`[target-frame _] (ctn/find-valid-parent-and-frame-ids target-frame objects shapes)`

common/src/app/common/types/container.cljc
```diff
@@ -385,3 +385,48 @@
...
         true
         (has-any-copy-parent? objects (:parent-id shape))))))
+
+(defn has-any-main?
+  "Check if the shape has any children or parent that is a main component."
+  [objects shape]
+  (let [instance-root (get-instance-root objects shape)]
+    (if (or (ctk/main-instance? shape) (ctk/main-instance? instance-root))
+      true
+      (some true? (map #(has-any-main? objects (get objects %)) (:shapes shape))))))
```
hirunatan: I think that this function may do more loops than needed. Perhaps it would be more efficient something like:
1) Get instance root.
2) Get all children with `cfh/get-children-with-self`
3) Iterate all children looking for a main

common/src/app/common/files/validate.cljc
hirunatan: But we need to talk about a strategy to repair files in PRE when this is deployed, to not break the work of the designers.

PR #3930, opened 2023-12-15
feature: Hide bounding-box when editing shape.

VasilevsVV:
closes #3232

Added behavior for hiding bounding-box of shapes. As described in: https://github.com/penpot/penpot/issues/3232 

Bounding box (`bb`) is being hidden for 1 sec, after any changes in sidebar->measures, made "by hand".

Realized via new flag for shape, indicating it being transformed.
Also: scheduling an updateEvent (`task`) to set shape's `bb` visible after 1 sec.
This `task` schedule resets after each trigger of event, making sure `bb` isn't flickering.

PR #3946, opened 2023-12-28
:sparkles: Add set of events for hiding and revealing bounding box for selected shape while transforming

### LINKED ISSUE #3232: feature: Hide bounding box when adjusting value

The bounding box of a selected object makes it harder to see the output when changing a value.

It would be great if the bounding box is hidden when adjusting a value in the right menu so we can see what exactly it looks like.

I like the approach of Figma where they hide the bounding box when making adjustments.  

### Additional context

https://github.com/penpot/penpot/assets/24385617/429712e4-9eaa-444e-85eb-9c6ca689fe45

https://github.com/penpot/penpot/assets/24385617/3178037c-dd66-4ee4-8e65-e93d05b447f2

niwinz:
Replaces: https://github.com/penpot/penpot/pull/3930
Closes #3232

Added behavior for hiding bounding-box of shapes. As described in: https://github.com/penpot/penpot/issues/3232

### DIFF
--- a/frontend/src/app/main/data/workspace.cljs
+++ b/frontend/src/app/main/data/workspace.cljs
@@ -2304,6 +2304,7 @@
 
 ;; Transform
 
+(dm/export dwt/trigger-bounding-box-cloaking)
 (dm/export dwt/start-resize)
 (dm/export dwt/update-dimensions)
 (dm/export dwt/change-orientation)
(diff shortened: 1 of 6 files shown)


VasilevsVV:
@niwinz , Thank you for your finishing touches with this PR.

PR #3956, opened 2024-01-03
:sparkles: Load debug CSS in local dev only

belen-albeza:
We had some debug CSS (ex: red background, yellow text). This PR sets this debugging CSS in a separate stylesheet that only gets loaded when the `NODE_ENV` variable is _not_ set to `production`.

It seems we are always setting this var to `production` when building the app (from taking a look at the scripts) but it'd be good if we could also double check.

If we don't like relying in `NODE_ENV` we could set a specific env var for this and modify the `Dockerfile` for the `devenv`.

### DIFF
--- a/frontend/gulpfile.js
+++ b/frontend/gulpfile.js
@@ -1,6 +1,6 @@
 import fs from "fs";
 import l from "lodash";
-import path from "path"
+import path from "path";
 
 import gulp from "gulp";
 import gulpConcat from "gulp-concat";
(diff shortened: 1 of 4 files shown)


belen-albeza:
I know this is not great, but other solutions I tried (append this code to the global stylesheet in gulp, etc.) looked even more convoluted. If someone has other ideas, pls let me know! (ideally these things should be done with feature flags)

PR #3977, opened 2024-01-08
:bug: Fix opening the swap panel on a copy of a deleted component fails

pabloalba:
Fixes https://tree.taiga.io/project/penpot/issue/6481


belen-albeza:
Thanks! 🙌

PR #3986, opened 2024-01-09
Misc bugfixes

hirunatan:
Solves .penpot part of https://tree.taiga.io/project/penpot/issue/6390
Fixes https://tree.taiga.io/project/penpot/issue/6071

### DIFF
--- a/backend/src/app/features/components_v2.clj
+++ b/backend/src/app/features/components_v2.clj
@@ -255,6 +255,7 @@
                       (assoc shape
                              :type :frame                  ; Old groups must be converted
                              :fills (or (:fills shape) []) ; to frames and conform to spec
+                             :shapes (or (:shapes shape) [])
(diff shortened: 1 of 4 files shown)


backend/src/app/features/components_v2.clj
```diff
@@ -255,6 +255,7 @@
                       (assoc shape
                              :type :frame                  ; Old groups must be converted
                              :fills (or (:fills shape) []) ; to frames and conform to spec
+                             :shapes (or (:shapes shape) [])
```
niwinz, line 258: this adds `:shapes` attrs to all shapes? or I'm wrong?

hirunatan: No. Only to shapes that are instance head (either they have a `:component-id` attribute or are the root shape inside a v1 component).

PR #4005, opened 2024-01-11
:bug: Bugfixes

AzazelN28:
1. [Taiga Issue #6455](https://tree.taiga.io/project/penpot/issue/6455) Card menu is hard to launch after search
2. [Taiga Issue #6477](https://tree.taiga.io/project/penpot/issue/6477) 400 error after reloading the page on the "Invitations" page
3. [Taiga Issue #6486](https://tree.taiga.io/project/penpot/issue/6486) Grey background on component assets preview

### DIFF
--- a/frontend/src/app/main/data/dashboard.cljs
+++ b/frontend/src/app/main/data/dashboard.cljs
@@ -366,14 +366,14 @@
 (defn hide-file-menu
   []
   (ptk/reify ::hide-file-menu
-     ptk/UpdateEvent
-     (update [_ state]
-       (update state :dashboard-local
-               assoc :menu-open false))))
+    ptk/UpdateEvent
+    (update [_ state]
+      (update state :dashboard-local
+              assoc :menu-open false))))
 
 (defn start-edit-file-name
   [file-id]
-  (ptk/reify ::start-edit-file-name
+  (ptk/reify ::start-edit-file-menu
     ptk/UpdateEvent
     (update [_ state]
       (update state :dashboard-local
--- a/frontend/src/app/main/data/dashboard.cljs
+++ b/frontend/src/app/main/data/dashboard.cljs
@@ -320,7 +320,9 @@
     (update [_ state]
       (update state :dashboard-local
               assoc :selected-files #{}
-              :selected-project nil))))
+              :selected-project nil
+              :menu-open false
+              :menu-pos nil))))
 
 (defn toggle-file-select
   [{:keys [id project-id] :as file}]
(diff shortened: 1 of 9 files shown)


frontend/src/app/main/data/dashboard.cljs
```diff
@@ -87,7 +87,7 @@
               ;; to currently active team
               (->> stream
                    (rx/filter (ptk/type? ::du/teams-fetched))
-                   (rx/observe-on :async)
+                   #_(rx/observe-on :async)
```
niwinz: why this code is commented?

frontend/src/app/main/data/dashboard.cljs
```diff
@@ -151,6 +151,7 @@
     ptk/WatchEvent
     (watch [_ state _]
       (let [team-id (:current-team-id state)]
+        (when (nil? team-id) (js-debugger))
```
niwinz: js-debugger

frontend/src/app/main/refs.cljs
```diff
@@ -82,17 +82,32 @@
...
-                     selected (get-in state [:dashboard-local :selected-files])
-                     xform    (comp (map get-file)
-                                    (map sim-file))]
-                 (->> (into #{} xform selected)
-                      (d/index-by :id))))
+               (dashboard-extract-selected (:dashboard-files state)
+                                           (dm/get-in state [:dashboard-local :selected-files])))
              st/state =))
 
+(def dashboard-file-menu-open?
```
niwinz: If this ref is only used from single namespace, it should be defined in the namespace where it is used.
Also, refs should not end with `?` in the name

frontend/src/app/main/refs.cljs
```diff
@@ -82,17 +82,32 @@
...
+        xform    (comp (map get-file)
+                       (map sim-file))]
+    (->> (into #{} xform selected)
+         (d/index-by :id))))
+
+(def dashboard-selected-search
+  (l/derived (fn [state]
+               ;; we need to this because :dashboard-search-result is a list
+               ;; of maps and we need a map of maps (using :id as key).
+               (let [files (reduce (fn [m v] (assoc m (:id v) v)) {} (:dashboard-search-result state))]
```
niwinz: For thus purpose I guess you have the `d/index-by` helper

frontend/src/app/main/data/dashboard.cljs
AzazelN28: ups, sorry, this rx/observe-on :async is one of the culprits for the race condition on team fetching

frontend/src/app/main/data/dashboard.cljs
AzazelN28: removed

frontend/src/app/main/refs.cljs
AzazelN28: Fixed, I've removed this ref

frontend/src/app/main/refs.cljs
AzazelN28: I don't understand this, just below the last line (in line 99) I'm using `dashboard-extract-selected`, a function that uses `d/index-by` to extract the search files.

niwinz: This reduce looks like doing the same as d/index-by, or I'm wrong?

frontend/src/app/main/data/dashboard.cljs
niwinz: Yep, but the fix can't be just removing this. The dashboard should load not depending on synchronity of some events. This events can or can not be async, and the dashboard should be load correctly independently of this.
This is why I suggested use the same logic as workspace is using (a boolean key that indicates that all is loaded correctly)

niwinz: Also, this async probably has a reason for being there, you probably fixed one issue and reopened other one that is not evident right now. The problem here is that there are no comment on code for proper explanation of the rationale for that async and that was our fault in the past (probably my fault not documenting it...)

frontend/src/app/main/refs.cljs
AzazelN28: Mmm... I don't think so, this reduce is transforming the :dashboard-search-result, a list of files, into a map of files using it's :id as key.

frontend/src/app/main/data/dashboard.cljs
niwinz: still here

frontend/src/app/main/ui/dashboard/grid.cljs
```diff
@@ -331,9 +332,11 @@
                (on-select event)) ;; TODO Fix this
              )))]
 
-    (mf/with-effect [selected? local]
-      (when (and (not selected?) (:menu-open @local))
-        (swap! local assoc :menu-open false)))
+    #_(mf/with-effect
```
niwinz: commented code, it should be here or should be deleted?

PR #4015, opened 2024-01-15
:tada: Add the ability to export import entire team

niwinz:


backend/src/app/binfile/v2.clj
```diff
@@ -0,0 +1,714 @@
...
+   [cuerdas.core :as str]
+   [datoteka.io :as io]
+   [promesa.util :as pu])
+  (:import
+   java.sql.DriverManager))
+
+(set! *warn-on-reflection* true)
+
+(def ^:dynamic *state* nil)
+(def ^:dynamic *options* nil)
```
hirunatan, line 46: Having a global state, could not cause race conditions if two imports or exports are executed simultaneously in the same server? Or this is not expected to happen?

backend/src/app/binfile/v2.clj
```diff
@@ -0,0 +1,714 @@
...
+   :storage-objects #{}})
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; PUBLIC API
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn export-team!
+  [cfg team-id]
+  (let [id  (uuid/next)
+        tp  (dt/tpoint)
```
hirunatan, line 607: I think excesive abbreviations do harm readability of the code. Some very common ones like `cfg` for config may be valid. But this one for example causes me to stop reading to check what is a `tpoint`.

I think it's very easy to write `time-point` or `log/info` and makes the code significantly easier to read. Taking into account that any code is read much more times than it's written, it pays the effort... :stuck_out_tongue:

backend/src/app/binfile/v2.clj
niwinz: This is not a global state. Is a dynamic scope variables, they are set per importation thread.

backend/src/app/binfile/v2.clj
niwinz: Agree

But concretly `tp` is used many times in the application, it is not new and not new to this ns.

backend/src/app/binfile/v2.clj
hirunatan: Ah, I didn't understand it well. So this is ok, then.

backend/src/app/binfile/v2.clj
hirunatan: Haha... of course not from this one. it was only a general rant :D

PR #4018, opened 2024-01-16
More bugfixes

hirunatan:
We have detected in some files `:shape-ref` that points to a shape that exists in the page, but does not belong to the main component. This is broken, but it was not detected by the validator. For example, this occurs in the file of this issue:

https://tree.taiga.io/project/penpot/issue/6879

This PR enhances the find ref shape function to not resolve the ref in this case, generating a validation error, and then makes the repair function smarter, so it may fix this error in many cases without information loss.

### DIFF
--- a/common/src/app/common/files/repair.cljc
+++ b/common/src/app/common/files/repair.cljc
@@ -269,16 +269,34 @@
 
 (defmethod repair-error :ref-shape-not-found
   [_ {:keys [shape page-id] :as error} file-data libraries]
-  (let [matching-shape (let [page           (ctpl/get-page file-data page-id)
-                             root-shape     (ctn/get-component-shape (:objects page) shape)
-                             component-file (if (= (:component-file root-shape) (:id file-data))
-                                              file-data
-                                              (-> (get libraries (:component-file root-shape)) :data))
-                             component      (when component-file
-                                              (ctkl/get-component (:data component-file) (:component-id root-shape) true))
-                             shapes         (ctf/get-component-shapes file-data component)]
-                         (d/seek #(= (:shape-ref %) (:shape-ref shape)) shapes))
+  (let [matching-shape (let [page              (ctpl/get-page file-data page-id)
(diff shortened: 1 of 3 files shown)


common/src/app/common/files/repair.cljc
```diff
@@ -269,16 +269,35 @@
...
+  (let [matching-shape (let [page              (ctpl/get-page file-data page-id)
+                             root-shape        (ctn/get-component-shape (:objects page) shape)
+                             component-file    (if (= (:component-file root-shape) (:id file-data))
+                                                 file-data
+                                                 (-> (get libraries (:component-file root-shape)) :data))
+                             component         (when component-file
+                                                 (ctkl/get-component component-file (:component-id root-shape) true))
+                             component-shapes  (ctf/get-component-shapes file-data component)]
+
+                         ; Check if the shape points to the remote main. If so, reassign to the near main.
```
niwinz: Comments should be always with `;;`

common/src/app/common/files/repair.cljc
```diff
@@ -269,16 +269,35 @@
...
+                         (if-let [near-shape-1 (d/seek #(= (:shape-ref %) (:shape-ref shape)) component-shapes)]
+                           near-shape-1
+                           ;; Check if it points to any random shape in the page. If so, try to find a matchng
+                           ;; shape in the near main component.
+                           (let [random-shape (ctn/get-shape page (:shape-ref shape))]
+                             (when random-shape
```
niwinz: can be collapsed to `when-let`

common/src/app/common/files/repair.cljc
hirunatan: Even if they are single line?

hirunatan: fixed

common/src/app/common/files/repair.cljc
hirunatan: fixed

PR #4053, opened 2024-01-25
:sparkles: Use new icons for feedback, pin and download + refactor the pin button

belen-albeza:
We no longer have two icons for the `pin` (filled and stroke-only), so some code changes needed to be done. Since the pin button appears in several places, I took the chance to refactor it.

I had to modify the `icon-xref` macro so it accepts an optional parameter of which class to use (before it would hardcode the name of the icon as the class). It seems we are not using those classes for anything, and this change only affects the pin button, but if they are needed for something, _do tell me_.

> 🙏 Alternatively, if someone could explain how to properly interpolate + concatenate these strings into the macro, it would be great. I tried and only got it to output `(stl/css :some-class-name) icon-pin-refactor`, not sure how to force the execution of the `(stl/css)` func.

Some screenshots of the pin button + icon:

Screenshot 2024-01-25 at 12 24 05 PM

Screenshot 2024-01-25 at 12 23 54 PM

Screenshot 2024-01-25 at 12 23 43 PM

Light theme looks like this:

Screenshot 2024-01-25 at 1 08 03 PM


frontend/resources/styles/common/refactor/common-dashboard.scss
```diff
@@ -33,13 +33,6 @@
     }
   }
 
-  svg {
-    fill: $db-secondary;
-    height: $s-12;
-    margin-right: $s-4;
-    width: $s-12;
-  }
-
```
belen-albeza, line 42: We can do this because the only `.dashboard-header` that was using SVGs was the projects' one

frontend/src/app/main/ui/icons.clj
```diff
@@ -11,9 +11,9 @@
    [rumext.v2]))
 
 (defmacro icon-xref
-  [id]
+  [id & class]
```
niwinz: This class is a seq of the optional arguments, so it is not that you are expecting

belen-albeza: I guess it kinda worked before because of the `~`? O_O (not sure what operator does) I think I fixed it now, thanks!

PR #4074, opened 2024-01-30
:bug: Fix weird positioning of component mixing undos and cut/paste

pabloalba:
Fixes https://tree.taiga.io/project/penpot/issue/6631 (see comments!)


common/src/app/common/files/changes.cljc
```diff
@@ -205,6 +205,7 @@
      [:map {:title "DelComponentChange"}
       [:type [:= :del-component]]
       [:id ::sm/uuid]
+      [:main-instance {:optional true} :any]
```
niwinz, line 208: This `:any` i think should be a proper type

PR #4086, opened 2024-01-31
Fix components without id

hirunatan:


niwinz:
merged with cherry-pick

PR #4095, opened 2024-02-01
Bug fixing

Alotor:
https://tree.taiga.io/project/penpot/issue/6836
https://tree.taiga.io/project/penpot/issue/6835
https://tree.taiga.io/project/penpot/issue/6830
https://tree.taiga.io/project/penpot/issue/6768
https://tree.taiga.io/project/penpot/issue/6760

### DIFF
--- a/frontend/src/app/main/ui/shapes/fills.cljs
+++ b/frontend/src/app/main/ui/shapes/fills.cljs
@@ -8,7 +8,10 @@
   (:require
    [app.common.data :as d]
    [app.common.data.macros :as dm]
+   [app.common.files.helpers :as cfh]
+   [app.common.geom.point :as gpt]
    [app.common.geom.shapes :as gsh]
+   [app.common.geom.shapes.text :as gst]
    [app.config :as cf]
    [app.main.ui.shapes.attrs :as attrs]
    [app.main.ui.shapes.gradients :as grad]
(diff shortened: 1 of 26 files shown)


frontend/src/app/main/ui/shapes/fills.cljs
```diff
@@ -62,32 +70,50 @@
...
+
                  props #js {:id (dm/str "fill-color-gradient-" render-id "-" fill-index)
                             :key (dm/str fill-index)
                             :gradient gradient
-                            :shape shape}]
-             (case (:type gradient)
-               :linear [:> grad/linear-gradient props]
-               :radial [:> grad/radial-gradient props]))))
+                            :shape obj}]
+             (case (d/name (:type gradient))
```
niwinz, line 97: I'm not clearly understand this change, or the need to to execute the d/name for this.
The type should be always keyword

Alotor: Yes! but there was a bug that sometimes that was a string. This MR solves that bug but I was worried of old saved files. This was more for retro-compatibility than for anything else.

niwinz: There is a migration for it. I mean right now the file does not passes validation if type is string and the migration is very recent so there are no files with this bug I think.

PR #4096, opened 2024-02-01
:bug: Fix shared library thumbnails

AzazelN28:
[Issue Taiga #6576](https://tree.taiga.io/project/penpot/issue/6576) Thumbnail in external library is not updated

### DIFF
--- a/frontend/src/app/main/data/workspace/libraries.cljs
+++ b/frontend/src/app/main/data/workspace/libraries.cljs
@@ -635,6 +635,21 @@
                                     :path-params path-params
                                     :query-params query-params}))))))
 
+(defn library-thumbnails-fetched
+  [thumbnails]
+  (ptk/reify ::library-thumbnails-fetched
+    ptk/UpdateEvent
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/data/workspace/libraries.cljs
```diff
@@ -635,6 +635,21 @@
                                     :path-params path-params
                                     :query-params query-params}))))))
 
+(defn library-thumbnails-fetched
+  [library-id thumbnails]
+  (ptk/reify ::library-thumbnails-fetched
+    ptk/UpdateEvent
+    (update [_ state]
+      (assoc-in state [:workspace-libraries library-id :thumbnails] thumbnails))))
```
niwinz: Each thumbnail has an unique ID. I'm not clearly understand why we have library thumbnails and file thumbnails stored in different locations inside state.
I think that having a single data structure with all thumbnails, makes the logic of choosing where to look for thumbnail more simple.

I'm missing something?

PR #4117, opened 2024-02-08
:bug: Fix display of mixed measurements +  border-radius input

belen-albeza:
- :bug: Fix mixed values displays for measurements in the design tab
- :bug: Fix multi radius input being reset to zero on blur

Fixes these tickets:
- https://tree.taiga.io/project/penpot/issue/6872
- https://tree.taiga.io/project/penpot/issue/6917

Screenshot 2024-02-08 at 1 13 41 PM

### DIFF
--- a/frontend/src/app/main/ui/components/numeric_input.cljs
+++ b/frontend/src/app/main/ui/components/numeric_input.cljs
@@ -53,7 +53,7 @@
 
         ;; This `value` represents the previous value and is used as
         ;; initil value for the simple math expression evaluation.
-        value       (d/parse-double value-str default)
+        value       (when (not= :multiple value-str) (d/parse-double value-str default))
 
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/ui/components/numeric_input.cljs
```diff
@@ -53,14 +53,13 @@
 
         ;; This `value` represents the previous value and is used as
         ;; initil value for the simple math expression evaluation.
-        value       (d/parse-double value-str default)
+        value       (if (not= :multiple value-str) (d/parse-double value-str default) nil)
```
Alotor: if you have a `(if <condition> <body> nil)` you can use: `(when <condition> <body>)` is the same thing because `when` returns `nil` if the condition is not fulfilled

PR #4131, opened 2024-02-12
:bug: Fix update main when there are swapped copies

hirunatan:
Fixes https://tree.taiga.io/project/penpot/issue/6909
And also the undo operation of update main component, that was broken.

### DIFF
--- a/common/src/app/common/types/container.cljc
+++ b/common/src/app/common/types/container.cljc
@@ -166,13 +166,6 @@
     :else
     (get-instance-root objects (get objects (:parent-id shape)))))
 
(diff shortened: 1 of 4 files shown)


common/src/app/common/types/container.cljc
```diff
@@ -166,13 +166,6 @@
     :else
     (get-instance-root objects (get objects (:parent-id shape)))))
 
-(defn get-copy-root
```
hirunatan, line 169: This helper was really redundant, it does the same as get-component-shape and was used only in one place.

PR #4149, opened 2024-02-14
:bug: Fix detach button position and color info overflow

Xaviju:
This PR fixes 

- https://tree.taiga.io/project/penpot/issue/6749
- https://tree.taiga.io/project/penpot/issue/6971

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/options/rows/color_row.scss
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/rows/color_row.scss
@@ -28,6 +28,7 @@
   height: $s-32;
   width: 100%;
   flex-grow: 1;
+  min-width: 0;
 
   &:hover {
     --detach-icon-foreground-color: var(--input-foreground-color-active);
@@ -55,6 +56,8 @@
     padding: 0 $s-2 0 $s-8;
     border-radius: $br-8 0 0 $br-8;
     background-color: transparent;
+    display: flex;
+    align-items: center;
     &:hover {
       background-color: transparent;
     }
--- a/frontend/src/app/main/ui/workspace/sidebar/options/rows/color_row.scss
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/rows/color_row.scss
@@ -10,14 +10,16 @@
   @include flexRow;
 
   &.dnd-over-top {
-    border-top: $s-1 solid var(--layer-row-foreground-color-drag);
+    border-block-start: $s-1 solid var(--layer-row-foreground-color-drag);
   }
   &.dnd-over-bot {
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/ui/workspace/sidebar/options/rows/color_row.scss
```diff
@@ -23,157 +23,161 @@
   gap: $s-2;
   border-radius: $s-8;
   background-color: var(--input-details-color);
-  height: $s-32;
-  width: 100%;
+  block-size: $s-32;
```
belen-albeza: fwiw we are using `height` and `width` in our code. I don't think we are rendering text in vertical script for any language)

frontend/src/app/main/ui/workspace/sidebar/options/rows/color_row.scss
```diff
@@ -23,157 +23,161 @@
...
-      padding: 0 $s-2 0 $s-8;
-      border-radius: $br-8 0 0 $br-8;
+  &:hover {
+    .detach-btn,
+    .select-btn {
       background-color: transparent;
-      &:hover {
-        background-color: transparent;
+      svg {
+        stroke: var(--input-foreground-color-active);
```
belen-albeza: I know we have this kind of code all over the place, but I try to avoid using raw `svg` selectors with the aid of both CSS custom properties + adding a class to the SVG itself.

If you are curious you might check out the `pin-button` component ([cljs](https://github.com/penpot/penpot/blob/b426db133ddee98b7c9977247f92c59eca5085c0/frontend/src/app/main/ui/dashboard/pin_button.cljs#L18-L19) and [scss](https://github.com/penpot/penpot/blob/b426db133ddee98b7c9977247f92c59eca5085c0/frontend/src/app/main/ui/dashboard/pin_button.scss#L24-L27))

EvaMarco (approved):

frontend/src/app/main/ui/workspace/sidebar/options/rows/color_row.scss
EvaMarco: we are not using block-size

Xaviju: I'll change it. 
Anyway, I suggest reconsidering this since it helps the whole team to get used to the terminology inline/block instead of width/height. It has no disadvantages that I can think of, and it can be updated gradually.

frontend/src/app/main/ui/workspace/sidebar/options/rows/color_row.scss
Xaviju: Fixed. This SVG already has a class. I've updated the selector.

EvaMarco (approved):

PR #4150, opened 2024-02-14
:recycle: Add minor refactor to file migrations

niwinz:
Relevant changes:

- Add the ability to create migration in both directions, defaulting
  to identity if not provided
- Move the version attribute to file table column for to make it more
  accessible (previously it was on data blob)
- Reduce db update operations on file-update rpc method

Also fixes this issue:
- https://tree.taiga.io/project/penpot/issue/6509

### DIFF
--- a/backend/src/app/binfile/common.clj
+++ b/backend/src/app/binfile/common.clj
@@ -12,7 +12,6 @@
    [app.common.data.macros :as dm]
    [app.common.exceptions :as ex]
    [app.common.features :as cfeat]
-   [app.common.files.defaults :as cfd]
    [app.common.files.migrations :as fmg]
    [app.common.files.validate :as fval]
    [app.common.logging :as l]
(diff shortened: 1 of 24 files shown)


backend/src/app/binfile/common.clj
```diff
@@ -409,19 +409,21 @@
 
 (defn- upsert-file!
   [conn file]
-  (let [sql (str "INSERT INTO file (id, project_id, name, revn, is_shared, data, created_at, modified_at) "
-                 "VALUES (?, ?, ?, ?, ?, ?, ?, ?) "
-                 "ON CONFLICT (id) DO UPDATE SET data=?")]
+  (let [sql (str "INSERT INTO file (id, project_id, name, revn, version is_shared, data, created_at, modified_at) "
```
superalex: "," after version?

PR #4152, opened 2024-02-14
:recycle: Update alerts design

EvaMarco:
This PR adds new alerts componentes
- inline-notification
- context-notification
- toast-notification

I've also changed the notification-hub whitch is in charge of dispatching notifications.


frontend/src/app/main/data/messages.cljs
```diff
@@ -93,18 +95,17 @@
           (rx/of hide))))))
 
 (defn error
-  ([content] (error content {}))
-  ([content {:keys [timeout] :or {timeout default-timeout}}]
-   (show {:content content
-          :type :error
-          :position :fixed
-          :timeout timeout})))
+  ([content] ((show {:content content
```
niwinz: `((show...` looks wrong
`([content...` should be `[content]` on a function with single arity

EvaMarco: Yes¡¡ I deleted one of the arities of that function, and I forgot to delete the (  ).

PR #4159, opened 2024-02-16
Alotor bugfix viewer

Alotor:
https://tree.taiga.io/project/penpot/issue/6984
https://tree.taiga.io/project/penpot/issue/6956
https://tree.taiga.io/project/penpot/issue/6963
https://tree.taiga.io/project/penpot/issue/6972

### DIFF
--- a/backend/src/app/rpc/commands/files.clj
+++ b/backend/src/app/rpc/commands/files.clj
@@ -351,6 +351,7 @@
 (sv/defmethod ::get-file-fragment
   "Retrieve a file fragment by its ID. Only authenticated users."
   {::doc/added "1.17"
+   ::rpc/auth false
    ::sm/params schema:get-file-fragment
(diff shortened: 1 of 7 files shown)


backend/src/app/rpc/commands/files.clj
```diff
@@ -351,6 +351,7 @@
 (sv/defmethod ::get-file-fragment
   "Retrieve a file fragment by its ID. Only authenticated users."
   {::doc/added "1.17"
+   ::rpc/auth false
```
AzazelN28, line 354: Why ::rpc/auth false? It could be a potential security issue, isn't it? 

If it does not the case and it doesn't require authentication, then the method description should be changed.

Alotor: As discussed, this is only to deactivate the "general" autentication and then the code makes a custom authentication for this case.

PR #4182, opened 2024-02-22
:bug: Fix switch inspect from html to svg

superalex:
- https://tree.taiga.io/project/penpot/issue/7011

### DIFF
--- a/frontend/src/app/main/ui/viewer/inspect/code.cljs
+++ b/frontend/src/app/main/ui/viewer/inspect/code.cljs
@@ -184,6 +184,7 @@
 
         set-markup
         (mf/use-callback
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/ui/viewer/inspect/code.cljs
```diff
@@ -185,6 +185,7 @@
         set-markup
         (mf/use-callback
          (fn [value]
+           (mf/deps markup-type*)
```
niwinz: I think mf/deps should be going one line upper

superalex: Ops!, you are right, thanks

PR #4198, opened 2024-02-27
:bug: Bug fixing

AzazelN28:
[Taiga Issue #6650](https://tree.taiga.io/project/penpot/issue/6650) Internal Error after creating the Path and clicking on the "Design" tab
[Taiga Issue #6983](https://tree.taiga.io/project/penpot/issue/6983) Selecting after the "edition mode" (enter) is not working properly
[Taiga Issue #6919](https://tree.taiga.io/project/penpot/issue/6919) Path dropdown overlaps main dropdown

### DIFF
--- a/frontend/src/app/main/refs.cljs
+++ b/frontend/src/app/main/refs.cljs
@@ -196,46 +196,6 @@
 (def context-menu
   (l/derived :context-menu workspace-local))
 
-(defn path-editing?
-  "Returns true if we're editing a path or creating a new one."
-  [state]
-  (let [selected    (dm/get-in state [:workspace-local :selected])
-        edition     (dm/get-in state [:workspace-local :edition])
-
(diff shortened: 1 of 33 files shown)


frontend/src/app/main/refs.cljs
```diff
@@ -196,8 +196,45 @@
...
+
+        path-edition? (or (and single? editing?
+                               (and (not (cph/text-shape? shape))
+                                    (not (cph/frame-shape? shape))))
+                          draw-path?
+                          edit-path?)]
+
+    path-edition?))
+
+(def toolbar-hidden
```
niwinz: This ref is only used on toolbar-top component, so I suggest move that logic there.
Also: the `path-editing?` impl has too many repeated lookups for `:workspace-local`. It think it can be implemented efficiently.

PR #4232, opened 2024-03-07
:bug: Fix copy and paste a main on another main doesn't work

pabloalba:
Fixes https://tree.taiga.io/project/penpot/issue/7096 (check both cases)

### DIFF
--- a/common/src/app/common/types/container.cljc
+++ b/common/src/app/common/types/container.cljc
@@ -416,12 +416,6 @@
      (some ctk/main-instance? children)
      (some ctk/main-instance? parents))))
 
-(defn has-any-main-children?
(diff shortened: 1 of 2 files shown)


common/src/app/common/types/container.cljc
```diff
@@ -413,15 +413,10 @@
   (let [children (cfh/get-children-with-self objects (:id shape))
         parents  (cfh/get-parents objects (:id shape))]
     (or
+     (ctk/main-instance? shape)
```
hirunatan: This seems redundant, isn't it? children includes self, so if shape is main it should be detected by the following line

pabloalba: Oh, that's correct!

PR #4255, opened 2024-03-12
:bug: Fix horizontal scroll bar in comments

belen-albeza:
Fixes https://tree.taiga.io/project/penpot/issue/7166

Also fixes the same issue appearing in the dashboard.

Screenshot 2024-03-12 at 3 09 53 PM

Screenshot 2024-03-12 at 3 10 35 PM

### DIFF
--- a/frontend/src/app/main/ui/comments.scss
+++ b/frontend/src/app/main/ui/comments.scss
@@ -98,8 +98,12 @@
 }
 
 .content {
+  position: relative;
   @include bodySmallTypography;
   color: var(--color-foreground-primary);
+  word-wrap: break-word;
+  overflow-wrap: break-word;
+  hyphens: auto;
 }
 
 .replies {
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/ui/comments.scss
```diff
@@ -98,8 +98,12 @@
 }
 
 .content {
+  position: relative;
   @include bodySmallTypography;
   color: var(--color-foreground-primary);
+  word-wrap: break-word;
+  overflow-wrap: break-word;
```
belen-albeza, line 105: We need to do this in addition to `hyphens` auto in the cases the browser-dependent algorithm does not break the word: comments in different languages than the one specified in our `<html lang>` attribute, made up words (for instance, a very long string with just `a`'s, etc.

Alotor: Didn't know the `hyphens` property. Very interesting.

PR #4266, opened 2024-03-14
:bug: Fix move copies into a component have geometry overriden

superalex:
https://tree.taiga.io/project/penpot/issue/7073


frontend/src/app/main/data/workspace/transforms.cljs
```diff
@@ -621,7 +621,7 @@
                       (let [undo-id (js/Symbol)]
                         (rx/of (dwu/start-undo-transaction undo-id)
                                (move-shapes-to-frame ids target-frame drop-index)
-                               (dwm/apply-modifiers {:undo-transation? false})
+                               (dwm/apply-modifiers {:undo-transation? false :force-ignore-ids ids})
```
superalex, line 624: may be just changing the order is enough

PR #4278, opened 2024-03-15
:bug: Fix move anidated structures withc component copies to other co…

superalex:
https://tree.taiga.io/project/penpot/issue/7203

### DIFF
--- a/frontend/src/app/main/data/workspace/transforms.cljs
+++ b/frontend/src/app/main/data/workspace/transforms.cljs
@@ -905,6 +905,11 @@
             moving-shapes-ids
             (map :id moving-shapes)
 
+            moving-shapes-children-ids
+            (->> moving-shapes
+                 (mapcat #(cfh/get-children-with-self objects (:id %)))
+                 (map :id))
+
             changes
             (-> (pcb/empty-changes it page-id)
                 (pcb/with-objects objects)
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/data/workspace/transforms.cljs
```diff
@@ -913,13 +918,13 @@
...
                 ;; Add component-root property when moving a component outside a component
                 (cond-> (not (ctn/get-instance-root objects frame))
-                  (pcb/update-shapes moving-shapes-ids (fn [shape]
-                                                         (if (ctk/instance-head? shape)
-                                                           (assoc shape :component-root true)
-                                                           shape))))
+                  (pcb/update-shapes moving-shapes-children-ids (fn [shape]
+                                                                  (if (ctk/instance-head? shape)
+                                                                    (assoc shape :component-root true)
+                                                                    shape))))
```
hirunatan: This should add component-root only to shapes that are in the top of a copy. I think that this way it will make root all heads. What do you think?

hirunatan: Confirmed. If you have a component copy with some other copies nested, and move it out of the main, it crashes because it tries to add component-root to all heads.

superalex: Ops, you are right! thanks!

PR #4297, opened 2024-03-19
Fix references when duplicating nested copies

hirunatan:
This PR fixes this issue https://tree.taiga.io/project/penpot/issue/7240

That was caused by the fix of this one https://tree.taiga.io/project/penpot/issue/7149 in PR https://github.com/penpot/penpot/pull/4254

To correctly fix the old one we need to advance shape-refs only the number of levels necessary in each case. Otherwise they end up attached to the remote main and the syncronization does not work.

### DIFF
--- a/common/src/app/common/types/container.cljc
+++ b/common/src/app/common/types/container.cljc
@@ -154,6 +154,17 @@
      :else
      (get-head-shape objects (get objects (:parent-id shape)) options))))
 
+(defn get-child-heads
+  "Get all recursive childs that are heads (when a head is found, do not
+   continue down looking for subsequent nested heads)."
+  [objects shape-id]
+  (let [shape (get objects shape-id)]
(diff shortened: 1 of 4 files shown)


common/src/app/common/types/container.cljc
```diff
@@ -170,6 +181,20 @@
...
        (reverse)))
 
+(defn get-nesting-level-delta
+  "Get how many levels a shape will 'go up' if moved under the new parent."
+  [objects shape new-parent]
+  (let [orig-heads (->> (get-parent-copy-heads objects shape)
+                        (remove #(= (:id %) (:id shape))))
+        dest-heads (get-parent-copy-heads objects new-parent)
+
+        ; Calculate how many parent heads share in common the original
```
niwinz: Comments are always with `;;`

PR #4306, opened 2024-03-21
:bug: Fix comments and history toggle

AzazelN28:
[Taiga Issue #7242](https://tree.taiga.io/project/penpot/issue/7242) History panel doesn't close when click on comments icon

### DIFF
--- a/frontend/src/app/main/data/workspace/drawing.cljs
+++ b/frontend/src/app/main/data/workspace/drawing.cljs
@@ -28,6 +28,10 @@
     ptk/UpdateEvent
     (update [_ state]
       (-> state
+          (update :workspace-layout (fn [workspace-layout]
+                                      (if (= tool :comments)
+                                        (disj workspace-layout :document-history)
+                                        workspace-layout)))
           (update :workspace-drawing assoc :tool tool)
           ;; When changing drawing tool disable "scale text" mode
           ;; automatically, to help users that ignore how this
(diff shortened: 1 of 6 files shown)


frontend/src/app/main/data/workspace/drawing.cljs
```diff
@@ -27,6 +27,8 @@
   (ptk/reify ::select-for-drawing
     ptk/UpdateEvent
     (update [_ state]
+      (when (= tool :comments)
+        (update state :workspace-layout disj :document-history))
```
niwinz: This code does nothing, I mean, it returns a new state but this state is ignored.

frontend/src/app/main/ui/workspace/right_header.cljs
```diff
@@ -169,33 +169,32 @@
...
-           (when (contains? layout :document-history)
-             (st/emit! (-> (dw/remove-layout-flag :document-history)
-                           (vary-meta assoc ::ev/origin "workspace-header"))))
-
            (if (= :comments selected-drawtool)
-             (st/emit! :interrupt)
+             (do
+               (st/emit! :interrupt)
+               (when (some? comments-local)
+                 (st/emit! :interrupt)))
```
niwinz: Do you need emit interrupt two times?

EvaMarco: yes, one to stop the path edition other to stop the drawing mode

niwinz: This looks like the problem is in other side. There should never be a need for two interrupts, a single interrupt should interrupt everything.

PR #4312, opened 2024-03-21
Remove swap slots

pabloalba:
Fixes https://tree.taiga.io/project/penpot/issue/7174

### DIFF
--- a/common/src/app/common/types/container.cljc
+++ b/common/src/app/common/types/container.cljc
@@ -211,20 +211,33 @@
     :else
     (get-instance-root objects (get objects (:parent-id shape)))))
 
+(defn find-component-main
+  "If the shape is a component main instance or is inside one, return that instance"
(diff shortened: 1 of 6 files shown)


common/src/app/common/types/container.cljc
```diff
@@ -211,20 +211,25 @@
...
   (cond
     (or (nil? shape) (cfh/root? shape))
-    false
+    nil
     (nil? (:parent-id shape))  ; This occurs in the root of components v1
-    true
+    shape
     (ctk/main-instance? shape)
-    true
-    (ctk/instance-head? shape)
```
hirunatan, line 224: This function is used from `in-any-component?`, below. And this one in turn is called from several places around the code. I think it's dangerous to change the behavior. Perhaps it's better to add a flag and leave the default behavior as it was before.

PR #4341, opened 2024-04-02
:bug: Fix comment reply menu

AzazelN28:
[Taiga Issue # 7335](https://tree.taiga.io/project/penpot/issue/7335) Comment reply menu is not autoclosing after user opens menu of another reply


frontend/src/app/main/data/comments.cljs
```diff
@@ -120,6 +121,7 @@
             (update :comment-threads assoc id (dissoc thread :comment))
             (update-in [:viewer :pages page-id :options :comment-threads-position] assoc id position)
             (update :comments-local assoc :open id)
+            (update :comments-local assoc :options nil)
```
niwinz, line 124: Maybe we can use `dissoc` instead of assoc with `nil` value

PR #4342, opened 2024-04-02
:bug: Fix comments not being visible on View mode

belen-albeza:
Fixes https://tree.taiga.io/project/penpot/issue/7338

- :bug: Fix comments padding
- :bug: Fix comments not being visible on view mode

Note that this doesn't change the comments positioning in the view mode, we did this in https://tree.taiga.io/project/penpot/issue/6799 but only for the workspace. I'll create [another ticket](https://tree.taiga.io/project/penpot/issue/7346) to handle this, since it's not trivial and fixing comments not appearing is kind of urgent.

Screenshot 2024-04-02 at 2 58 12 PM

### DIFF
--- a/frontend/src/app/main/ui/comments.cljs
+++ b/frontend/src/app/main/ui/comments.cljs
@@ -418,8 +418,7 @@
              :id (str "thread-" thread-id)
              :style {:left (str pos-x "px")
                      :top (str pos-y "px")
-                     :max-height max-height
-                     :overflow-y "scroll"}
+                     :max-height max-height}
              :on-click dom/stop-propagation}
 
        [:div {:class (stl/css :comments)}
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/ui/comments.cljs
```diff
@@ -381,7 +382,7 @@
                        (some? position-modifier)
                        (gpt/transform position-modifier))
 
-        max-height   (int (* (:height viewport) 0.75))
+        max-height   (when (some? viewport) (int (* (:height viewport) 0.75)))
```
niwinz, line 385: having the `or` on the line 359 for viewport, I guess this `some?` condition looks redundant

belen-albeza: Hi! Not exactly, without this change then `max-height` is set to `0` and not `nil`, and then it gets added to the element styles as `max-height: 0px` and the fix does not work (i.e. we need this property to disappear from the stylesheet altogether).

belen-albeza: Ah now I see what you mean. The problem is that GitHub is omitting some of the lines and it seems that the diff shows the same function, but they are two different functions, so the `viewport` var is different in each one.

PR #4439, opened 2024-04-15
Avoid non existent function warning

jordisala1991:
This is my first PR to penpot 🎉 . I am not sure to be targetting the correct branch. I guess you are using staging branch for bug fixing but I might be wrong. If I need to change base branch, please let me know 🙏 

When building penpot libraries (inside frontend directory)

```
clojure -M:dev:shadow-cljs release lib-penpot
```

You get the following warning:

```
------ WARNING #1 - :undeclared-var --------------------------------------------
 File: /home/penpot/penpot/frontend/src/app/libs/file_builder.cljs:256:11
--------------------------------------------------------------------------------
 253 |
 254 |   (export [_]
 255 |     (->> (export-file file)
 256 |          (rx/subs
-----------------^--------------------------------------------------------------
 Use of undeclared Var beicon.v2.core/subs
--------------------------------------------------------------------------------
 257 |           (fn [value]
 258 |             (when  (not (contains? value :type))
 259 |               (let [[file export-blob] value]
 260 |                 (dom/trigger-download (:name file) export-blob))))))))
--------------------------------------------------------------------------------
```

I guess this is because on Penpot 2.0, the https://github.com/funcool/beicon/tree/master was updated.

https://github.com/funcool/beicon/blob/master/CHANGES.md

This is the change that is mentioned in the docs, I think this is the only place where the function wasn't change when the library was updated.

After applying the change it no longer gives a warning:

```
[:lib-penpot] Compiling ...
[:lib-penpot] Build completed. (894 files, 1 compiled, 0 warnings, 40.23s)
```

> Screenshots/GIFs

### DIFF
--- a/frontend/src/app/libs/file_builder.cljs
+++ b/frontend/src/app/libs/file_builder.cljs
@@ -253,7 +253,7 @@
 
   (export [_]
     (->> (export-file file)
-         (rx/subs
+         (rx/subs!
           (fn [value]
             (when  (not (contains? value :type))
               (let [[file export-blob] value]


niwinz:
Thanks

PR #4486, opened 2024-04-22
:paperclip: Add provider to validate uri

AzazelN28:
Add provider to validate uri

[Task #7542](https://tree.taiga.io/project/penpot/task/7542)


backend/src/app/auth/oidc.clj
```diff
@@ -558,6 +559,7 @@
                                 :props props
                                 :exp (dt/in-future "4h")})
         uri   (build-auth-uri cfg state)]
+    (prn "auth-handler" uri state params)
```
superalex: Should we remove this prn?

PR #4535, opened 2024-04-29
Move frontend tests to common

hirunatan:


common/test/common_tests/logic/components_touched_test.cljc
```diff
@@ -0,0 +1,151 @@
...
+                 (thf/add-sample-shape :free-shape))
+
+        page (thf/current-page file)
+
+        ;; Action
+        ;; IMPORTANT: as modifying copies structure is now forbidden, this action
+        ;; will not have any effect, and so the parent shape won't also be touched.
+        changes (cflh/generate-relocate-shapes (pcb/empty-changes)
+                                               (:objects page)
+                                               #{(thi/id :copy-root)}   ; parents
```
pabloalba: This should be the parent of the shape that you are moving: `#{(:parent-id (thf/get-shape :free-shape))}` Or in this case `uuid/zero`

common/test/common_tests/logic/components_touched_test.cljc
```diff
@@ -0,0 +1,151 @@
...
+        page (thf/current-page file)
+        copy-root (thf/get-shape file :copy-root)
+        copy-child1 (thf/get-shape-by-id file (first (:shapes copy-root)))
+
+        ;; Action
+        ;; IMPORTANT: as modifying copies structure is now forbidden, this action
+        ;; will not have any effect, and so the parent shape won't also be touched.
+        changes (cflh/generate-relocate-shapes (pcb/empty-changes)
+                                               (:objects page)
+                                               #{(thi/id :copy-root)}   ; parents
```
pabloalba: This should be the parent of the shape that you are moving: `#{(:parent-id copy-child1)}`

PR #4537, opened 2024-04-29
:recycle: Components refactor: generator for relocate shapes (and tests)

pabloalba:
:recycle: Components refactor: generator for relocate shapes (and tests)

### DIFF
--- a/common/test/common_tests/helpers/debug.cljc
+++ b/common/test/common_tests/helpers/debug.cljc
@@ -1,0 +1,33 @@
+(ns common-tests.helpers.debug
+  (:require
+   [app.common.uuid :as uuid]
+   [common-tests.helpers.ids-map :as thi]))
+
+(defn dump-shape
+  "Dumps a shape, with each attribute in a line"
+  [shape]
+  (println "{")
+  (doseq [[k v] (sort shape)]
+    (when (some? v)
+      (println (str "    " k " : " v))))
+  (println "}"))
+
+(defn- stringify-keys [m keys]
+  (apply str (interpose ", " (map #(str % ": " (get m %)) keys))))
+
+(defn dump-page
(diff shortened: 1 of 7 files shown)


common/test/common_tests/helpers/files.cljc
```diff
@@ -131,6 +135,8 @@
         shape     (sample-shape label (dissoc params :parent-label))
         parent-id (when parent-label
                     (thi/id parent-label))
+        _         (when parent-label
+                    (t/is (some? parent-id)))
```
hirunatan: I wouldn't add checks here. We can assume that the test setup is well done. We are not testing the test itself.

common/test/common_tests/helpers/debug.cljc
```diff
@@ -0,0 +1,32 @@
...
+  (:require
+   [app.common.uuid :as uuid]
+   [common-tests.helpers.ids-map :as thi]))
+
+(defn dump-shape
+  "Dumps a shape, with each attribute in a line"
+  [shape]
+  (println "{")
+  (doseq [[k v] (sort shape)]
+    (println (str "    " k " : " v)))
```
hirunatan: I think that it would be good that if the value is nil, the key is not shown

common/test/common_tests/logic/comp_remove_swap_slots_test.cljc
```diff
@@ -0,0 +1,194 @@
...
+
+
+(t/deftest test-keep-swap-slot-relocating-yellow-to-b2
+  (let [;; ============================== Setup ===============================
+        file                   (setup-file)
+        page                   (thf/current-page file)
+        blue1                  (thf/get-shape file :blue1)
+        yellow                 (thf/get-shape file :frame-yellow)
+
+        ;; Move blue1 into yellow
```
hirunatan, line 161: I think Action should start here

PR #4555, opened 2024-05-06
 ✨ Add login page as Page Object Model

EvaMarco:

> Screenshots/GIFs


frontend/playwright/ui/pages/login-page.js
```diff
@@ -0,0 +1,27 @@
...
+    this.loginButton = page.getByRole("button", { name: "Login" });
+    this.password = page.getByLabel("Password");
+    this.userName = page.getByLabel("Email");
+    this.message = page.getByText("Email or password is incorrect");
+    this.badLoginMsg = page.getByText("Enter a valid email please");
+    this.initialHeading = page.getByRole("heading", { name: "Log into my account" });
+  }
+
+  async url() {
+    this.page.url()
```
AzazelN28: Is this missing a `return` statement?

BTW, in that case if you want to return the result of `this.page.url()`, the `async` keyword is unnecessary.

EvaMarco: ohh I forgot to delete that, it was an error.

EvaMarco: Thanks, I was not using that function, now it's fixed :)

PR #4559, opened 2024-05-07
:recycle: Components refactor: move generators for duplicate and add tests for remove swap slot on duplicate

pabloalba:


common/src/app/common/logic/libraries.cljc
```diff
@@ -1923,3 +1925,295 @@
     (cond-> changes
       (some? swap-slot)
       (generate-sync-head file-full libraries container id components-v2 true))))
+
+(defn generate-prepare-duplicate-flows
```
hirunatan: Perhaps "generate prepare" is quite redundant. I think "prepare" was the old naming convention for those helpers, and then we moved to "generate".

You could name this simply "generate-duplicate-flows", and same for the other ones.

pabloalba: Fixed

PR #4576, opened 2024-05-10
:sparkles: Add duplicated pages component tests

superalex:
- https://tree.taiga.io/project/penpot/task/7680
- https://tree.taiga.io/project/penpot/task/7681


common/src/app/common/test_helpers/compositions.cljc
```diff
@@ -166,4 +171,143 @@
                             :component2-params component2-params
                             :main2-root-params main2-root-params
                             :nested-head-params nested-head-params)
-      (thc/instantiate-component component2-label copy2-root-label copy2-root-params)))
\ No newline at end of file
+      (thc/instantiate-component component2-label copy2-root-label copy2-root-params)))
```
hirunatan, line 176: I think we should add section comments in the file, to classify
- file builder functions
- getter functions
- file modifier functions
(perhaps even in the future we can split the file, if it grows larger, but not for now)

common/src/app/common/test_helpers/compositions.cljc
```diff
@@ -166,4 +171,143 @@
...
+
+(defn- bottom-shape-by-id [file id & {:keys [page-label]}]
+  ;; Get the deepest descendant of a shape by id
```
hirunatan: Those comments are better put as docstrings, i think. They are different than the comments of the previous functions, that describe the file fragment they are building.

common/src/app/common/test_helpers/compositions.cljc
```diff
@@ -166,4 +171,143 @@
...
+                                     0
+                                     nil
+                                     {})
+
+        file' (thf/apply-changes file changes)]
+    (if propagate-fn
+      (propagate-fn file')
+      file')))
+
+(defn swap-component-in-shape [file shape-tag component-tag & {:keys [page-label propagate-fn]}]
```
hirunatan, line 247: I don't like very much having a function that does two different things (this is against Single Responsibility Principle). I'd prefer a function that simply gets the shape, and that can be chained with the other function that does the swap. 

But i haven't found any easy way of creating this chain in the test, that is readable enough.

common/src/app/common/test_helpers/compositions.cljc
superalex: Thanks!

PR #4583, opened 2024-05-14
Fix flakiness of workspace tests

belen-albeza:
Fixes https://tree.taiga.io/project/penpot/task/7730

- :bug: Fix workspace rect drawing test.
- :bug: Fix mocking websockets when running multiple tests.

A short explanation about the fixes:

- The drawing a shape test required some `await` for the `expect`s (it seems the SVG was created and _then_ its properties set).
- The issue with the WebSocket mocks is that we were not cleaning up our static map of sockets between test runs, so they worked when run on isolation, but not when we had multiple tests that required websockets. Clearing this map when we call `MockWebSocketHelper.init` fixes it.
- Restricted the workers in CI to `1`, [as recommended per Playwright](https://github.com/microsoft/playwright/issues/22666) to ensure more predictability (we will have to parallelize in CI with sharding).

### DIFF
--- a/frontend/playwright/helpers/MockWebSocketHelper.js
+++ b/frontend/playwright/helpers/MockWebSocketHelper.js
@@ -2,17 +2,19 @@ export class MockWebSocketHelper extends EventTarget {
   static #mocks = new Map();
 
   static async init(page) {
-    await page.exposeFunction("MockWebSocket$$constructor", (url, protocols) => {
-      const webSocket = new MockWebSocketHelper(page, url, protocols);
+    this.#mocks = new Map();
+
+    await page.exposeFunction("onMockWebSocketConstructor", (url) => {
+      const webSocket = new MockWebSocketHelper(page, url);
       this.#mocks.set(url, webSocket);
     });
-    await page.exposeFunction("MockWebSocket$$spyMessage", (url, data) => {
+    await page.exposeFunction("onMockWebSocketSpyMessage", (url, data) => {
       if (!this.#mocks.has(url)) {
         throw new Error(`WebSocket with URL ${url} not found`);
       }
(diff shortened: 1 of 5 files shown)


frontend/playwright/ui/pages/WorkspacePage.js
```diff
@@ -1,5 +1,6 @@
 import { expect } from "@playwright/test";
 import { BaseWebSocketPage } from "./BaseWebSocketPage";
+import { MockWebSocketHelper } from "../../helpers/MockWebSocketHelper";
```
AzazelN28: Is this necessary?

frontend/playwright/helpers/MockWebSocketHelper.js
```diff
@@ -2,17 +2,19 @@ export class MockWebSocketHelper extends EventTarget {
   static #mocks = new Map();
 
   static async init(page) {
-    await page.exposeFunction("MockWebSocket$$constructor", (url, protocols) => {
-      const webSocket = new MockWebSocketHelper(page, url, protocols);
+    MockWebSocketHelper.#mocks = new Map();
```
AzazelN28: In this scope you can use `this` instead of `MockWebSocketHelper`.

frontend/playwright/ui/pages/WorkspacePage.js
belen-albeza: Leftover 🙈

frontend/playwright/helpers/MockWebSocketHelper.js
belen-albeza: You're right!

EvaMarco (approved):

PR #4594, opened 2024-05-17
🐛 [WIP] Fix components not being correctly imported using lib-penpot

Cenadros:

Right now we're unable of correctly importing a simple component into penpot using the compiled library `lib-penpot` because there are some missing properties or wrong ids in the .svgs generated by the file builder.

In this PR we're going to try and fix this.


hirunatan:
Finally, the problem was more complex, there were some complicated issues with formats and feature initialization.

I've solved them in this other PR, use it for reference: https://github.com/penpot/penpot/pull/4616

PR #4616, opened 2024-05-23
Prepare builder for figma exporter

hirunatan:
Enhance file-builder to work better with components-v2 when called from the lib-penpot javascript library (in particular from the figma exporter plugin).

### DIFF
--- a/common/src/app/common/schema.cljc
+++ b/common/src/app/common/schema.cljc
@@ -240,7 +240,6 @@
   (when-not ^boolean (-validate s value)
     (let [hint    (d/nilv dm/*assert-context* "check error")
           explain (-explain s value)]
-      (println (humanize-explain explain))
       (throw (ex-info hint {:type :assertion
                             :code :data-validation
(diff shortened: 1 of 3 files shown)


common/src/app/common/schema.cljc
```diff
@@ -240,6 +240,7 @@
   (when-not ^boolean (-validate s value)
     (let [hint    (d/nilv dm/*assert-context* "check error")
           explain (-explain s value)]
+      (println (humanize-explain explain))
```
superalex: is this println necessary?

hirunatan: Let's see. The lib-penpot packaging had a problem, that if a schema error occurred there, the exception trace was minified and illegible. So Andrey added this println to be able to see the cause.

I'l remove it in a separated commit for now. Andrey wants to review it to try to solve properly. But I've removed it in a separated commit, to be able to locate it easier if we need it again until then.

PR #4684, opened 2024-06-06
:sparkles: Speed up e2e testing

superalex:
image

### DIFF
--- a/frontend/scripts/e2e-server.js
+++ b/frontend/scripts/e2e-server.js
@@ -1,9 +1,13 @@
 import express from "express";
+import compression from "compression";
+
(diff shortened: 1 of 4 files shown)


frontend/scripts/e2e-server.js
```diff
@@ -1,10 +1,14 @@
 import express from "express";
+import compression from "compression";
```
belen-albeza, line 2: Q: Shouldn't `package.json` be updated with this as a dev dependency?

superalex: Yes!, thank you very much! :)

belen-albeza (approved):

PR #4687, opened 2024-06-06
:sparkles: Add e2e tests for fix color palette default library

superalex:
Tests for https://tree.taiga.io/project/penpot/issue/8029


belen-albeza:
Thanks for adding the test, looks great.

PR #4690, opened 2024-06-06
:sparkles: Return blob when exporting using the lib-penpot

jordisala1991:

It would be nice if we can have more control on how and when we want to download the generated zip file (blob).

With this PR we can allow to put a loading indicator to say that we are generating the file in zip format using the export function

I am not sure if this is the correct code to get the blob file, because I dont know Clojure

Here is the PR using this changes: https://github.com/penpot/penpot-exporter-figma-plugin/pull/144

With this PR we can hide the loading indicator when the file is already generated and sent to download.

> Screenshots/GIFs


frontend/src/app/libs/file_builder.cljs
```diff
@@ -262,12 +261,14 @@
...
-              (let [[file export-blob] value]
-                (dom/trigger-download (:name file) export-blob))))))))
+    (js/Promise.
+     (fn [resolve _reject]
+       (-> (export-file file)
+           (rx/subscribe
+            (fn [value]
+              (when (not (contains? value :type))
+                (let [[_ export-blob] value]
+                  (resolve export-blob))))))))))
```
niwinz, line 271: A small improvement:

```clojure
;; on requires
[promesa.core :as p]


(p/create
 (fn [resolve reject]
   (-> (export-file file)
       (rx/take 1)
       (rx/subscribe
        (fn [value]
          (when (not (contains? value :type))
            (let [[_ export-blob] value]
              (resolve export-blob))))
        reject))))
```

jordisala1991: I tried your suggestion and got this error message when exporting:

```
VM733:13 Uncaught (in promise) TypeError: h.subscribe is not a function
    at hn._subscribe (<anonymous>:13:45556)
    at hn._trySubscribe (<anonymous>:13:2676)
    at hn.subscribe (<anonymous>:13:2620)
    at Wpn (<anonymous>:238:86370)
    at <anonymous>:347:34481
    at Gpn (<anonymous>:238:87279)
    at k.export (<anonymous>:347:34456)
    at E (<anonymous>:347:43512)
```

The stack trace probably won't be enough to debug...

jordisala1991: @hirunatan , can you help me with this one? I dont know enough clojure to apply the provided suggestion

niwinz: Yeah, my example was buggy, I will look on it today

niwinz: Created a MR, with your commit and the change https://github.com/penpot/penpot/pull/4802, if it is ok for you, I will merge it ASAP

PR #4692, opened 2024-06-06
Plugins storage

Alotor:
- Adds support for storing plugin's data
- Fix rename layers in plugins


common/src/app/common/files/changes.cljc
```diff
@@ -133,6 +133,21 @@
       [:id ::sm/uuid]
       [:name :string]]]
 
+    [:mod-file-plugin-data
```
niwinz, line 136: Just suggestion, `:mod-file-plugin-data` -> `:mod-plugin-data`

PR #4709, opened 2024-06-10
:bug: Fix internal error when I set up a stroke for multiple objects

superalex:
https://tree.taiga.io/project/penpot/issue/7558


frontend/src/app/main/ui/workspace/sidebar/options/menus/stroke.cljs
```diff
@@ -169,9 +169,10 @@
                      :on-collapsed toggle-content
                      :title        label
                      :class        (stl/css-case :title-spacing-stroke (not has-strokes?))}
-
-       [:button {:class (stl/css :add-stroke)
-                 :on-click on-add-stroke} i/add]]]
+       (when (not (= :multiple strokes))
+         [:button {:class (stl/css :add-stroke)
+                   :data-testid "add-stroke"
```
niwinz, line 174: I think we should standarize to `data-test` because we already use it.

superalex: for the e2e testing this one is used in a lot of places, if we want to standardize it to another value we could merge this and do a small refactor in another PR for all the data-test-id

PR #4721, opened 2024-06-11
:bug: Fix problem moving layout to frame

Alotor:
https://tree.taiga.io/project/penpot/issue/7760

### DIFF
--- a/frontend/playwright/ui/specs/workspace.spec.js
+++ b/frontend/playwright/ui/specs/workspace.spec.js
@@ -67,7 +67,6 @@ test("User adds a library and its automatically selected in the color palette", 
   await expect(workspacePage.palette.getByText('There are no color styles in your library yet')).toBeVisible();
 });
 
-
 test("User makes a group", async ({ page }) => {
   const workspacePage = new WorkspacePage(page);
   await workspacePage.setupEmptyFile();
--- a/common/src/app/common/types/shape/layout.cljc
+++ b/common/src/app/common/types/shape/layout.cljc
@@ -604,18 +604,23 @@
 
 (defn remove-layout-item-data
   [shape]
-  (dissoc shape
-          :layout-item-margin
-          :layout-item-margin-type
-          :layout-item-h-sizing
-          :layout-item-v-sizing
-          :layout-item-max-h
-          :layout-item-min-h
-          :layout-item-max-w
-          :layout-item-min-w
-          :layout-item-align-self
-          :layout-item-absolute
-          :layout-item-z-index))
+  (-> shape
+      (dissoc :layout-item-margin
+              :layout-item-margin-type
+              :layout-item-max-h
+              :layout-item-min-h
+              :layout-item-max-w
+              :layout-item-min-w
+              :layout-item-align-self
+              :layout-item-absolute
+              :layout-item-z-index)
+      (cond-> (or (not (any-layout? shape))
+                  (= :fill (:layout-item-h-sizing shape)))
+        (dissoc :layout-item-h-sizing)
+
+        (or (not (any-layout? shape))
+            (= :fill (:layout-item-v-sizing shape)))
+        (dissoc :layout-item-v-sizing))))
 
 (defn update-flex-scale
   [shape scale]
--- a/frontend/playwright/data/workspace/get-file-7760.json
+++ b/frontend/playwright/data/workspace/get-file-7760.json
@@ -1,0 +1,49 @@
+{
+  "~:features": {
+    "~#set": [
+      "layout/grid",
+      "styles/v2",
+      "fdata/pointer-map",
+      "fdata/objects-map",
+      "components/v2",
+      "fdata/shape-data-type"
+    ]
+  },
+  "~:permissions": {
+    "~:type": "~:membership",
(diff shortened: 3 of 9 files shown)


frontend/package.json
```diff
@@ -33,6 +33,7 @@
     "watch": "node ./scripts/watch.js",
     "e2e:server": "node ./scripts/e2e-server.js",
     "e2e:test": "playwright test --project default",
+    "e2e:ui": "playwright test --ui --project default",
```
belen-albeza: I don't think we should have commands in `package.json` we cannot run inside our devenv

frontend/playwright/ui/specs/workspace.spec.js
```diff
@@ -67,6 +67,31 @@ test("User adds a library and its automatically selected in the color palette",
...
+  await workspacePage.mockRPC("update-file?id=*", "workspace/update-file-create-rect.json");
+
+  await workspacePage.goToWorkspace({fileId: "cd90e028-326a-80b4-8004-7cdec16ffad5", pageId: "cd90e028-326a-80b4-8004-7cdec16ffad6"});
+
+  // Select the flex board and drag it into the other container board
+  await workspacePage.clickLeafLayer("Flex Board");
+
+  // Move the first board into the second
+  const hAuto = await workspacePage.page.locator(`[for='behaviour-h-auto']`);
+  const vAuto = await workspacePage.page.locator(`[for='behaviour-v-auto']`);
```
belen-albeza: We should avoid these types of selectors (by attribute). If we cannot get around [following our criteria](https://help.penpot.app/technical-guide/developer/frontend/#query-priority-for-locators), then we should default to a `data-testid`.

But in this case, we can actually do a semantic query, which is the preferred one:

Screenshot 2024-06-11 at 1 48 55 PM

We could access it with `getByTitle("Fit content")`. Since this title is repeated in both vertical and horizontal, we could change the `title` for `Fit content (vertical)`.

Also: I don't know why these strings are not translated 🤔

frontend/playwright/ui/specs/workspace.spec.js
```diff
@@ -67,6 +67,31 @@ test("User adds a library and its automatically selected in the color palette",
...
+
+  await expect(hAuto).toHaveClass(/checked/);
+  await expect(vAuto).toHaveClass(/checked/);
```
belen-albeza: We should avoid checking by class whenever possible, since it's very tied to the implementation.

Here we have an alternative, which is to check whether the checkbox is checked.

Ideally we should be able to use `getByRole("checkbox")`, but since it's hidden with `display: none` (it shouldn't be hidden that way, we should eventually change this) it does not appear in the accesibility tree and we cannot use role queries.

In the meantime, we can do:

```js
 await expect(vAuto.locator("input")).toBeChecked();
```

(maybe let's put a `TODO` comment so we don't forget to change this once we have these radio buttons coded properly in the DS)

frontend/playwright/ui/specs/workspace.spec.js
```diff
@@ -67,6 +67,31 @@ test("User adds a library and its automatically selected in the color palette",
   await expect(workspacePage.palette.getByText('There are no color styles in your library yet')).toBeVisible();
 });
 
+test("BUG 7760 - Layout losing properties when changing parents", async ({ page }) => {
```
belen-albeza: This test should belong in the `design-tab.spec` no?

frontend/playwright/ui/specs/workspace.spec.js
```diff
@@ -67,6 +67,31 @@ test("User adds a library and its automatically selected in the color palette",
...
+  const vAuto = await workspacePage.page.locator(`[for='behaviour-v-auto']`);
+
+  await expect(hAuto).toHaveClass(/checked/);
+  await expect(vAuto).toHaveClass(/checked/);
+
+  await workspacePage.moveSelectionToShape("Container Board");
+
+  // The first board properties should still be auto width/height
+  await expect(hAuto).toHaveClass(/checked/);
+  await expect(vAuto).toHaveClass(/checked/);
```
belen-albeza: Same as above

belen-albeza (changes requested):

frontend/package.json
Alotor: ok, it's more for convenience than anything

frontend/playwright/ui/specs/workspace.spec.js
Alotor: I can't use "Fit content" because both horizontal and vertical have the same title... I can change the title so it has the description of vertical/horizontal and then use this.

It's not translated because we said that the terms that are relative to CSS we won't translate.

frontend/playwright/ui/specs/workspace.spec.js
Alotor: This won't work. I've checked and the radio buttons inside won't update the checked property.

frontend/playwright/ui/specs/workspace.spec.js
Alotor: ok

belen-albeza (approved):

PR #4726, opened 2024-06-12
:sparkles: Improve email verify threshold

superalex:


backend/src/app/config.clj
```diff
@@ -213,6 +216,7 @@
 (s/def ::telemetry-uri ::us/string)
 (s/def ::telemetry-with-taiga ::us/boolean)
 (s/def ::tenant ::us/string)
+(s/def ::email-verify-threshold ::us/string)
```
niwinz: Should be `::dt/duration` i guess

PR #4756, opened 2024-06-18
The value "version" is obsolete

ghost:
As of 2024, the value called "version" in the Docker Compose file is now obsolete. It is no longer required or needed in the file.


niwinz:
Will be merged with commit message adapted to the commiting guideline in other PR. Thanks :heart:

PR #4760, opened 2024-06-18
:bug: Fix json encoding output truncation

niwinz:


backend/src/app/http/middleware.clj
```diff
@@ -34,11 +31,12 @@
   {:name ::params
    :compile (constantly ymw/wrap-params)})
 
-(def ^:private json-mapper
-  (json/mapper
-   {:encode-key-fn str/camel
-    :decode-key-fn (comp keyword str/kebab)
```
sneko, line 40: @niwinz I think the keys transformation must be kept. Otherwise it results in `is-owner` which is harder to go through in JS (`data['is-owner']` instead of `data.isOwner`)

sneko (changes requested):
See my comment about casing for the ease with JS but also to match the OpenAI schema.

Otherwise, I tested locally and it works great, no more truncation.

backend/src/app/http/middleware.clj
niwinz: oh, you are right, i will fix it

niwinz:
Casing issue fixed

ymolists:
@niwinz great work. Thanks for the heads up. is this already deployed ? i keep trying and keep getting the same cropped json

PR #4787, opened 2024-06-21
:bug: Fix thumbnails related issues

niwinz:
## How to test it

1. Create a two frames
2. Add a nested frame into the first one with a color
3. Move or change the color of the nested frame

## Current behaviour

The thumbnail of the frame parent doesn't updates on changes of its nested frames.

## Expected behavior

The thumbnail is properly updated.

https://github.com/penpot/penpot/assets/478699/f5253c1c-7360-45cc-9168-ea43d34e1acd


AzazelN28:
I don't know why but this is throwing the following Error everytime I try to create a new Frame (not necessarily nested):
image

niwinz:
Fixed

PR #4789, opened 2024-06-21
Fix export deleted components

hirunatan:
Fixes https://tree.taiga.io/project/penpot/issue/8073

This is not really a bug, but that this feature was not implemented. All existing code was working on components v1 or early versions of v2. I've changed greatly the way deleted components are exported / imported.

~~WARNING: there is still a bug. The shapes inside the deleted component are imported with wrong coordinates, relative to the component they belong. This causes that when the component is restored, the shapes are positioned outside. When importing, they must be moved to be absolute coordinates when added to the page before deleting the component.~~

### DIFF
--- a/common/src/app/common/files/builder.cljc
+++ b/common/src/app/common/files/builder.cljc
@@ -41,16 +41,19 @@
    (let [components-v2 (dm/get-in file [:data :options :components-v2])
          component-id  (:current-component-id file)
          change        (cond-> change
-                         (and add-container? (some? component-id))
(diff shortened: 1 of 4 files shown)


superalex:
I've added some code to solve the @hirunatan warning, so the PR should be complete

jordisala1991:
This PR misses the necessary bindings for the penpot library, I tried to add them, but I don't understand all the code here, so I think it is better if you add them to allow the library to generate zip files with deleted components.

PR #4813, opened 2024-06-27
:sparkles: Add extra events info to track teams and invitations

superalex:
https://tree.taiga.io/project/penpot/us/7713?milestone=396395

### DIFF
--- a/backend/src/app/rpc/commands/teams.clj
+++ b/backend/src/app/rpc/commands/teams.clj
@@ -867,7 +867,6 @@
                     :invitations invitations}
           {::audit/props {:invitations (count invitations)}})))))
 
-
 ;; --- Mutation: Create Team & Invite Members
 
 (def ^:private schema:create-team-with-invitations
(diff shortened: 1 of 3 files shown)


backend/src/app/rpc/commands/teams.clj
```diff
@@ -813,14 +814,15 @@
...
+   [:resend? :boolean]
    [:emails ::sm/set-of-emails]])
 
 (sv/defmethod ::create-team-invitations
   "A rpc call that allow to send a single or multiple invitations to
   join the team."
   {::doc/added "1.17"
    ::sm/params schema:create-team-invitations}
-  [{:keys [::db/pool] :as cfg} {:keys [::rpc/profile-id team-id emails role] :as params}]
+  [{:keys [::db/pool] :as cfg} {:keys [::rpc/profile-id team-id emails role resend?] :as params}]
```
niwinz: We should not use `?` on params on RPC methods, this makes them less usable from JS or other laguanges

PR #4837, opened 2024-07-03
✨ Add typography ds tokens

EvaMarco:
This PR solves [this task](https://tree.taiga.io/project/penpot/task/7867)


belen-albeza:
Note that we need to remove the web font assets we no longer use. We can do this once we verify this works with no issues.

PR #4843, opened 2024-07-04
♻️  Remove unused locale fn

EvaMarco:
This PR removes not necessary functions tochange the base font family in body. This PR has no task associated.


belen-albeza:
To elaborate a bit if we need to query this PR in the future:

Switching between the regular (Work Sans) font and a special one for Arabic characters is no longer done using a custom CSS property and changing it from CLJS depending on the browser locale, but using the `unicode-range` feature of Web Fonts.

PR #4844, opened 2024-07-04
🐛  Fix workspace visual test for assets modal

EvaMarco:
This  fixes a broken visual test by adding a new RPC and some checks


belen-albeza:
Added some extra checks to ensure the screenshots are taken in the same state. That should help to mitigate the flakiness.

PR #4870, opened 2024-07-11
✨  Add loader component to ds

EvaMarco:
This PR closes [this task](https://tree.taiga.io/project/penpot/task/8168)

### DIFF
--- a/frontend/src/app/main/ui/ds.cljs
+++ b/frontend/src/app/main/ui/ds.cljs
@@ -11,12 +11,14 @@
    [app.main.ui.ds.foundations.typography :refer [typography-list]]
    [app.main.ui.ds.foundations.typography.heading :refer [heading*]]
    [app.main.ui.ds.foundations.typography.text :refer [text*]]
+   [app.main.ui.ds.product.loader :refer [loader*]]
    [app.main.ui.ds.storybook :as sb]))
 
 (def default
   "A export used for storybook"
   #js {:Heading heading*
        :Icon icon*
+       :Loader loader*
        :RawSvg raw-svg*
        :Text text*
        ;; meta / misc
--- a/frontend/src/app/main/ui/ds/product/loader.cljs
+++ b/frontend/src/app/main/ui/ds/product/loader.cljs
@@ -1,0 +1,38 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC
+
+(ns app.main.ui.ds.product.loader
+  (:require-macros
+   [app.common.data.macros :as dm]
+   [app.main.style :as stl])
+  (:require
+   [rumext.v2 :as mf]))
+
+(mf/defc loader*
+  {::mf/props :obj}
+  [{:keys [class width height title] :rest props}]
+  (let [class (dm/str (or class "") " " (stl/css :loader))
+        both-provided (and width height)
+        neither-provided (and (nil? width) (nil? height))
+        props (mf/spread-props props {:viewBox "0 0 677.34762 182.15429"
+                                      :role "status"
+                                      :width (or width "100px")
+                                      :height (or height "27px")
+                                      :class class})]
+    (assert (or both-provided neither-provided)
+            (dm/str "Invalid props: both 'width' and 'height' must be provided or neither. "
+                    "Received width: " width ", height: " height))
+    ;; TODO: Add a translated label insted of the title prop.    
+    (assert title
+            (dm/str "You must provide an accesible name for the component"))
+    [:> "svg" props
+     [:title title]
+     [:g
+      [:path {:d
+              "M128.273 0l-3.9 2.77L0 91.078l128.273 91.076 549.075-.006V.008L128.273 0zm20.852 30l498.223.006V152.15l-498.223.007V30zm-25 9.74v102.678l-49.033-34.813-.578-32.64 49.61-35.225z"}]
+      [:path {:class (stl/css :loader-line)
+              :d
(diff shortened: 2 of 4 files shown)


frontend/src/app/main/ui/ds/app/loader.cljs
```diff
@@ -0,0 +1,28 @@
...
+  (:require
+   [rumext.v2 :as mf]))
+
+(mf/defc loader*
+  {::mf/props :obj}
+  [{:keys [class] :rest props}]
+  (let [class (dm/str (or class "") " " (stl/css :loader-pencil))
+        props (mf/spread-props props {:viewBox "0 0 677.34762 182.15429"
+                                      :height "182"
+                                      :width "667"
```
belen-albeza: We need to be able to let the user override the SVGs `width` and `height` attributes. Let's use default values that makes sense (I don't think we're using `182`x`667` anywhere in the app), but ultimately let the user override them with delegated props.

frontend/src/app/main/ui/ds/app/loader.mdx
belen-albeza, line 1: I think we can skip docs for this component:

- We don't have any indications / documentations from Design.
- The technical notes are not relevant (we can do what we say with every component of our design system)

frontend/src/app/main/ui/ds/app/loader.cljs
```diff
@@ -0,0 +1,28 @@
...
+                                      :class class})]
+    [:> "svg" props
+     [:g
```
belen-albeza: Unlike icons, this component would be use "as it is", and we need to think about accessibility. We should add a `<title>` element to the SVG, and probably either make this prop mandatory or set a reasonable default if we have it (like "Loading").

Since this is not an attribute that will be passed to the SVG, but a child element, it should be a prop.

An alternative to `<title>` would be to use `arialabeledby` but I _think_ I have seen this component being used without associated text. If that's not the case, we should enforce this prop.

frontend/src/app/main/ui/ds/app/loader.scss
```diff
@@ -0,0 +1,27 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) KALEIDOS INC
+
+//Component tokens
```
belen-albeza: nit: redundant, we can skip this comment

frontend/src/app/main/ui/ds/app/loader.scss
```diff
@@ -0,0 +1,27 @@
...
+$color-loader-foreground: var(--color-foreground-secondary);
```
belen-albeza: It's best to use CSS variables for component-level tokens, then we can leverage them to create variants, etc. https://octuweb.com/css-custom-properties/

Another advantage is that we will see them (with the proper component name) while debugging in the browser inspector.

frontend/src/app/main/ui/ds/app/loader.stories.jsx
```diff
@@ -0,0 +1,18 @@
...
+
+export default {
+  title: "App/Loader",
+  component: Components.Loader,
+};
+
+export const Default = {
+  render: () => (
+    <StoryWrapper theme="default">
+      <Loader style={{ width: "6.25rem", height: "6.25rem" }} />
```
belen-albeza: I wouldn't override the default width and height if the component is already visible. It's the default, after all :)

frontend/src/app/main/ui/ds/app/loader.stories.jsx
```diff
@@ -0,0 +1,18 @@
+import * as React from "react";
+import Components from "@target/components";
+
+const { Loader } = Components;
+const { StoryWrapper } = Components.storybook;
+
+export default {
+  title: "App/Loader",
```
belen-albeza: Ask Andrey (or folks from Design) to see what he suggests there (also update the namespace for this component)

frontend/src/app/main/ui/ds/app/loader.scss
```diff
@@ -0,0 +1,27 @@
...
+    transform: translateY(-150px);
+  }
+}
+
+.loader-line {
+  fill: $color-loader-foreground;
+  animation: line-pencil 0.8s infinite linear;
+}
+
+.loader-pencil {
```
belen-albeza: It's more clear if we name this `container` or something similar (or just `loader`, the same as the component). So we know these styles correspond to the outer-most element (also important because here is where we should define component-level tokens using custom properties)

belen-albeza (changes requested):

frontend/src/app/main/ui/ds/penpot/loader.cljs
```diff
@@ -0,0 +1,38 @@
...
+  [{:keys [class width height title] :rest props}]
+  (let [class (dm/str (or class "") " " (stl/css :loader))
+        both-provided (and width height)
+        neither-provided (and (nil? width) (nil? height))
+        props (mf/spread-props props {:viewBox "0 0 677.34762 182.15429"
+                                      :role "status"
+                                      :width (or width "100px")
+                                      :height (or height "27px")
+                                      :class class})]
+    (assert (or both-provided neither-provided)
```
belen-albeza, line 25: asserts need to happen before (there's also no reason we cannot have two lets in a function)

frontend/src/app/main/ui/ds/penpot/loader.stories.jsx
```diff
@@ -0,0 +1,26 @@
+import * as React from "react";
+import Components from "@target/components";
+
+const { Loader } = Components;
+const { StoryWrapper } = Components.storybook;
```
belen-albeza: This is going away with the new theme system, no need to use this wrapper

belen-albeza (approved):

PR #4876, opened 2024-07-11
:sparkles: Add spacing tokens to the design system

belen-albeza:
Closes https://tree.taiga.io/project/penpot/task/7914


frontend/src/app/main/ui/ds/storybook.scss
```diff
@@ -1,5 +1,3 @@
-@use "./colors.scss";
```
belen-albeza, line 1: This was causing `colors.scss` to be included twice. Since that stylesheet is considered a CSS module, is automatically added to the end CSS without it having to be included manually.

PR #4904, opened 2024-07-22
♻️  Replacing the old Pencil loader with the new Loader component

EvaMarco:
This PR replaces all old pencil loaders with the new design system component “loader”. 
In the application we can find 3 types of loaders
- Icon size ->  On import/export modals. We will use the loader* component
  - Import 
Screenshot from 2024-07-23 08-19-43
  - Export 
Screenshot from 2024-07-23 08-19-43

- Grid element -> The size is adjusted according to the number of elements.  We will use the loader* component
  - Grid libraries
Screenshot from 2024-07-23 08-15-05
  - Grid files
Screenshot from 2024-07-23 08-16-21

- Full page size -> Its rendered in the middle of the screen, and is 64px wide.  We will use the loader-overlay* component
  - Libraries
Screenshot from 2024-07-23 07-54-14
   - Files
Screenshot from 2024-07-23 07-55-49
  - Viewer,  verify token and workspace
  Screenshot from 2024-07-23 08-09-31

This PR closes[ this task](https://tree.taiga.io/project/penpot/task/8355)


frontend/src/app/main/ui/dashboard/placeholder.scss
```diff
@@ -90,3 +75,13 @@
...
+
+.placeholder-loader {
+  align-self: end;
+}
+
+.grid-empty-placeholder.loader {
+  height: 100%;
+  width: 100%;
+  justify-items: center;
+}
```
belen-albeza: Let's avoid this type of selector and use just a single class

frontend/src/app/main/ui/dashboard/placeholder.cljs
```diff
@@ -41,5 +42,8 @@
 (mf/defc loading-placeholder
   []
   [:div {:class (stl/css :grid-empty-placeholder :loader)}
```
belen-albeza: Is this wrapper necessary?

frontend/src/app/main/ui/viewer.cljs
```diff
@@ -634,5 +635,5 @@
       [:> viewer-content props])
 
     [:div {:class (stl/css :loader-content)}
```
belen-albeza: Check if this wrapper is needed at all

frontend/src/app/main/ui/workspace.cljs
```diff
@@ -125,7 +125,10 @@
 (mf/defc workspace-loader
   []
   [:div {:class (stl/css :workspace-loader)}
```
belen-albeza: let's also check this wrapper

belen-albeza (changes requested):
Thanks @EvaMarco !

There are situations in which we have the loader embedded in a `<div>` wrapper it might not be needed, since we can apply classes directly to the loader component for positioning (this was not possible before, so we have a lot of icons nested into wrappers because of this). Let's see if we can get rid off them.

Also: let's add screenshots or a list of the loaders subbed by this PR in Taiga so it can be tested properly, both by QA and myself :)

Thanks!

frontend/src/app/main/ui/viewer.cljs
EvaMarco: I have changed the base component to have the wrapper itself

frontend/src/app/main/ui/workspace.cljs
EvaMarco: Same as above

belen-albeza (changes requested):
Let's keep this to a single `<loader>` component that renders one variant or the other depending on a prop.

frontend/src/app/main/ui/dashboard/placeholder.scss
EvaMarco: Done

EvaMarco:
> Let's keep this to a single `<loader>` component that renders one variant or the other depending on a prop.

Good idea¡¡

belen-albeza:
Added another commit with the following:

- Remove accepting two different CSS classes props in `loader*`.
- The inline variant now aligns vertically to the center the icon and children.
- Used the overlay variant in grid thumbnails
- Not required both or neither for `width` and `height` (if one is missing, is calculated respecting the default aspect ratio)
- Added controls to Storybook stories.
- Used the storybook layout `"fullscreen"` for the overlay story so we don't need to artificially set the component's height to `100%`.

As for the first point, there was one instance in which the icon size was put via CSS and the use of a `--th-width` custom property set way up in the DOM. This is not the correct way to do it (we need the grid to provide the dimensions via a Context), but it's out of the scope of this PR. As a workaround, `--loader-icon-width` can be set by the user.

@EvaMarco pls see if this looks alright and if so, let's merge. Thanks!

belen-albeza (approved):

PR #4928, opened 2024-07-26
Alotor plugins 10

Alotor:
- ✨ Add export method into API
- ✨ Add undo block api to plugins
- ⬆️ Update plugins runtime
- ✨ Change local storage access

### DIFF
--- a/frontend/src/app/plugins/file.cljs
+++ b/frontend/src/app/plugins/file.cljs
@@ -5,17 +5,23 @@
 ;; Copyright (c) KALEIDOS INC
 
 (ns app.plugins.file
-  "RPC for plugins runtime."
   (:require
    [app.common.data.macros :as dm]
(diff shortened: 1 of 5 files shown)


frontend/src/app/plugins/file.cljs
```diff
@@ -105,7 +111,47 @@
...
+                    :team-id team-id
+                    :features features
+                    :export-type export-type
+                    :files [file]})
+                  (rx/mapcat #(->> (rx/of %) (rx/delay 1000)))
+                  (rx/mapcat
+                   (fn [msg]
+                     (case (:type msg)
+                       :error
+                       (rx/throw "Cannot export file")
```
niwinz: Consider throwing an instance of js/Error or ex-info which is already an instance of js/Error instead of text. Throwing text is not recommended

PR #4941, opened 2024-07-30
✨  Add tabs component to DS

EvaMarco:
This PR resolves [this task](https://tree.taiga.io/project/penpot/task/7891)


belen-albeza:
Thanks @EvaMarco  Added the following to the PR:

- Made the tab panel focusable.
- Fixed the controls in the `tab-switcher*` stories.
- There was a CSS bug triggered if you set `action-button-position` without passing an `action-button`. I added an `assert` for this.

PR #4945, opened 2024-07-30
♻️  move entire image build process into Dockerfiles

### LINKED ISSUE #4900: feature: move entire container image build process into Dockerfiles

Currently the process for building production container images requires bundles to be build using a separate step involving `docker run`. While this works on a regular dev machine or in a VM, this makes it impossible to build penpot images in an unprivileged container like one often has in a CI. \
Also it's generally nice if a single Dockerfile allows for building the entire container from scratch as that integrates nicely with CI tooling as well.

I'd like to have the bundle creation be done in the Dockerfile and utilize a multi-stage build process to then have the bundle copied into the production image. \
I already refactored the code [in a fork](https://github.com/june128/penpot/tree/refactor-image-builds) implementing this feature and I would love to submit this refactor in a PR, so that penpot containers can easily be build in various CI tooling. 

_No response_

### Additional context

_No response_

june128:
With this change the entire image build process happens inside the Dockerfile. To not deviate too much from the current build process, I simply moved the bundle creation process from the `manage.sh` script into the Dockerfile for each image (backend, frontend and exporter). \
Then utilizing multi-stage building the bundle gets created in a separate step running the devenv image and then gets copied over to the production image.

With this change, all images can simply be built with just one invocation of `docker build` for each image. With that and especially with the removal of the `docker run` step in the `manage.sh` script, images can now be built in unprivileged containers and therefore in a wider variety of CI solutions.

Closes #4900

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+/bundles/
--- a/docker/images/Dockerfile.backend
+++ b/docker/images/Dockerfile.backend
@@ -1,3 +1,19 @@
+FROM penpotapp/devenv:latest AS build
+
+COPY --chown=penpot:users . /home/penpot/penpot
+WORKDIR /home/penpot/penpot
+USER penpot:users
+
+RUN ./manage.sh version > build_version.txt
+
+WORKDIR /home/penpot/penpot/backend
+RUN ./scripts/build $(cat ../build_version.txt)
+
+WORKDIR /home/penpot/penpot/bundles
+RUN cp -r ../backend/target/dist ./backend
+RUN cp -r ../build_version.txt ./backend/version.txt
+RUN ../manage.sh put-license-file ./backend
+
 FROM ubuntu:22.04
 
 LABEL maintainer="Andrey Antukh <niwi@niwi.nz>"
@@ -59,7 +75,7 @@ RUN set -eux; \
     tar -xf /tmp/openjdk.tar.gz --strip-components=1; \
     rm -rf /tmp/openjdk.tar.gz;
 
-COPY --chown=penpot:penpot ./bundle-backend/ /opt/penpot/backend/
+COPY --from=build --chown=penpot:penpot /home/penpot/penpot/bundles/backend/ /opt/penpot/backend/
 
 USER penpot:penpot
 WORKDIR /opt/penpot/backend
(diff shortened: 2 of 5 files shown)


june128:
I also would love to make the build process even cleaner by only copying the source without any prior build artifacts to the Dockerfile. To achieve that I wanted to simply symlink the `.dockerignore` to the `.gitignore`, however the `.gitignore` currently ignores a bunch of stuff, including files, which are relevant to the build, so that's why simply doing that wasn't possible.

I guess the build scripts delete all build artifacts on invocation, so it doesn't matter, but it would be nice to have this all cleaned up as well - maybe in a future PR.

niwinz:
Hello,

I've been looking at your proposal, and at first glance my opinion is that it is pretty nice. I think it brings simplicity to the current solution.

The problem right now is that some parts of your proposal cannot be merged as is (specifically the part of removing functions that is in the manage.sh script). This is because this script is also used by our deployment system in our infrastructure (we don't use docker images to deploy, we deplore them directly in a provisioned VM, so we use the same scripts for building the bundle).

There is also the problem of duplication since we would have two different sites defining how a bundle of each sub application is built.

Before making any decision and being able to merge it, I need to be able to have quality time to review and test it well and see what other implications it could have beyond what I see directly. And I will be able to do that in a few days, when I finish with the priority tasks that I have in hand.

Thanks.

june128:
Yeah, that makes sense. Thanks for considering this change tho and please take your time!
I'll also happily chat about the design of the build system at any time.

niwinz:
Back to this topic,

The main problem with this change is that we generate bundles and then use them for various purposes. We don't use docker for the deployment of our SAAS, so we do a build and then generate docker images and deploy the generated build to our SAAS.

Eliminating the build generation from manage.sh is not an option because it breaks our deployment. Or at least that would mean that we would have to take that to the deploy scripts repository; thus maintaining the same process in two places.
And regardless of whether there is duplication of code that does the same thing, we would have the fact that the build would also be done redundantly (2 times, one for deploy and another for the docker images).

And finally, any failure in the build, to debug it is much simpler if it were a Dockerfile. Or at least that is the perception we have.

So I am not very clear if we really gain anything with these proposed changes, over the current approach

PR #4952, opened 2024-08-01
Storybook build

belen-albeza:
Closes https://tree.taiga.io/project/penpot/task/8439

- :wrench: Enable storybook to be served from a subdir
- :sparkles: Compile storybook target in release and a separate DS stylesheet
- :sparkles: Watch css assets specifically for storybook
- :wrench: Update build script to include storybook

Ultimately, this PR:

- Allows the Storybook to be accessed from `localhost:3449/storybook` locally.
- Makes the `build-frontend-bundle` to also prepare a storybook build.
- Compiles the `storybook` target in release mode.
- Compiles a separate CSS stylesheet , `ds.css`, which only has the base, common CSS to all Penpot plus the styles for the design system (without all the other styles from the app).
- Adds a new watcher script to only watch and compile storybook-related CSS changes (useful when developing the design system)

To **test** this:

1. Rebuild the devenv locally with `./manage.sh build-devenv-local`
2. Start the local dev env.
3. Stop any running shadow-cljs process, such as the one at tab 1 in tmux.
4. Run `yarn run build:storybook` 
5. Go to http://localhost:3449/storybook


frontend/scripts/build
```diff
@@ -26,3 +26,7 @@ rsync -avr resources/public/ target/dist/
 
 sed -i -re "s/\%version\%/$CURRENT_VERSION/g" ./target/dist/index.html;
 sed -i -re "s/\%buildDate\%/$BUILD_DATE/g" ./target/dist/index.html;
+
+# build storybook
+yarn run storybook:build || exit 1;
+rsync -avr storybook-static/ target/dist/storybook-static
```
belen-albeza: Not sure if `target/dist/storybook-static` is the right destination. Maybe `target/dist-storybook` so it's separate from the rest?

PR #4958, opened 2024-08-02
:tada: Add file-data offload mechanism

niwinz:
Relevant changes:

1. Objects Storage

    Internally rename assets- storage to objects storage for make it more generic. It preserves backward compatibility with old settings; but the new settings should be preferred.

    `PENPOT_ASSETS_STORAGE_BACKEND` becomes `PENPOT_OBJECTS_STORAGE_BACKEND`
and the values: `assets-fs` -> `fs` and `assets-s3` -> `s3`

   `PENPOT_ASSETS_FS_DIRECTORY` becomes `PENPOT_OBJECTS_STORAGE_FS_DIRECTORY`
   `PENPOT_ASSETS_S3_BUCKET` becomes `PENPOT_OBJECTS_STORAGE_S3_BUCKET`
  and so on…

    Pending, not included in this MR: update the documentation.

2. Improve scalability of file-gc task.

    The file-gc task is separated into two tasks: file-gc and file-gc-scheduler. The file-gc from now is responsible for performing the gc on a single file (specified on params/props), and the file-gc-scheduler is responsible for looking for all eligible files for GC and properly scheduling a file-gc task for each eligible file.

    This change allows file-gc to execute on many workers instead of executing this on a single worker and single thread.

3. Performance enhancements on storage-gc-touched task.

    Performing the gc operation in small chunks and without transaction locking. So this operation will not occasionally block other operations.
    Use SELECT EXISTS instead of count() for reference checking allowing to perform much less db operations and making the ref checking faster.

4. Performance enhancements on telemetry related queries.

    Use better queries that makes better use of indexes. Reduces the load on big on-premise instances that have telemetry enabled.

5. Update FILLFACTOR on file table.

   This makes better use of HOT updates inside postgresql for heavy update tables. More about this here: https://www.crunchydata.com/blog/postgres-performance-boost-hot-updates-and-fill-factor

6. File Data Offload

    Allows store file data on objects storage when the file becomes inactive.

    Right now we have the file-gc task, which is run after some days of file inactivity. If the tiered-file-data-storage flag is activated, after the file-gc task the offload-file-data task will be scheduled. This task will move the data from the database row to the objects storage.

    The code is adapted for transparent access to file-data independently of the final location, so when you access it in a read-only manner to the file, it will work as expected.

    When you perform an update to a file with offloaded data, the file-data will be returned to the file table and the storage object will become eligible for GC.

How to test it:

1. No specific tests
2. Create some files; Run file-gc-schedule task on CLI with min-age=0 and observe how it schedules many file-gc tasks (has unit tests)
3. No specific tests, should work as before (has unit tests)
4. No specific tests
5. No specific tests
6. Enable feature, create file, run file-gc task with min-age=0 for the file-id previously created, observe that offload-file-data is scheduled and executed, look on file and file_data_fragment table that data becomes nil and data_backend and data_ref_id is filled.

### DIFF
--- a/backend/src/app/config.clj
+++ b/backend/src/app/config.clj
@@ -52,8 +52,8 @@
 
    :redis-uri "redis://redis/0"
 
-   :assets-storage-backend :assets-fs
-   :storage-assets-fs-directory "assets"
+   :objects-storage-backend "fs"
+   :objects-storage-fs-directory "assets"
 
    :assets-path "/internal/assets/"
    :smtp-default-reply-to "Penpot <no-reply@example.com>"
@@ -207,16 +207,24 @@
     [:prepl-host {:optional true} :string]
     [:prepl-port {:optional true} :int]
 
-    [:assets-storage-backend {:optional true} :keyword]
     [:media-directory {:optional true} :string] ;; REVIEW
     [:media-uri {:optional true} :string]
     [:assets-path {:optional true} :string]
 
+    ;; Legacy, will be removed in 2.5
+    [:assets-storage-backend {:optional true} :keyword]
     [:storage-assets-fs-directory {:optional true} :string]
     [:storage-assets-s3-bucket {:optional true} :string]
     [:storage-assets-s3-region {:optional true} :keyword]
     [:storage-assets-s3-endpoint {:optional true} :string]
-    [:storage-assets-s3-io-threads {:optional true} :int]]))
+    [:storage-assets-s3-io-threads {:optional true} :int]
+
+    [:objects-storage-backend {:optional true} :keyword]
+    [:objects-storage-fs-directory {:optional true} :string]
+    [:objects-storage-s3-bucket {:optional true} :string]
+    [:objects-storage-s3-region {:optional true} :keyword]
+    [:objects-storage-s3-endpoint {:optional true} :string]
+    [:objects-storage-s3-io-threads {:optional true} :int]]))
 
 (def default-flags
   [:enable-backend-api-doc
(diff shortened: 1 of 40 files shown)


backend/src/app/config.clj
```diff
@@ -207,23 +207,32 @@
...
+    [:objects-storage-s3-endpoint {:optional true} :string]
+    [:objects-storage-s3-io-threads {:optional true} :int]]))
 
 (def default-flags
   [:enable-backend-api-doc
    :enable-backend-openapi-doc
    :enable-backend-worker
    :enable-secure-session-cookies
    :enable-email-verification
+   :enable-tiered-file-data-storage
```
superalex: disabled by default?

backend/src/app/features/fdata.clj
```diff
@@ -55,31 +56,47 @@
...
+    (when-not fragment
       (ex/raise :type :internal
                 :code :fragment-not-found
                 :hint "fragment not found"
                 :file-id file-id
                 :fragment-id id))
 
-    (blob/decode content)))
+    (let [data (get-file-data system fragment)]
+      ;; FIXME: conditional thread scheduling for decoding big objects
```
superalex, line 95: should we leave this FIXME here?

backend/src/app/config.clj
niwinz: Fixed

backend/src/app/features/fdata.clj
niwinz: Yep, we will leave it.

PR #4979, opened 2024-08-11
🐛 Update time.cljc

eltociear:
prefered -> preferred


niwinz:
Thanks

But before merge, ensure you are using `:bug:` on the commit message instead of :bug: as unicode character.
If you prefer, I can do the change myself before merge.

niwinz:
replaced with https://github.com/penpot/penpot/pull/4996

niwinz:
Thanks

PR #5005, opened 2024-08-18
🐛 Fix cell size changed when scroll asset list

### LINKED ISSUE #5004: bug: component list scroll abnormal when it has many cells (20 +)
### Steps To Reproduce

1. set many componets in a shared files 
2. use this shared files
3. scroll components list

### Expected behavior

it should easy to scroll to bottom

### Actual behavior

The DOM continuously jumps upward while scrolling.espaicially more than 100 components

### Screenshots or video

Uploading 录屏2024-08-18 19.31.21.mov…

_No response_

_No response_

_No response_

### Frontend Stack Trace

_No response_

### Backend Stack Trace

_No response_

### Additional context

_No response_

VVege:
Issue: [penpot/penpot#5004](https://github.com/penpot/penpot/issues/5004)

Reason: The component thumbnails were being rendered in every cell, and the cells’ CSS had an aspect-ratio property. This caused the browser to calculate the actual width and height dynamically, leading to temporary changes in DOM size. As a result, during scrolling, these size changes affected the scrolling behavior.

Change: Set specific width and height values to prevent the DOM size from fluctuating during scrolling.

This is a small change, and I hope it can be accepted. Thank you very much!

### DIFF
--- a/frontend/resources/styles/common/refactor/color-defs.scss
+++ b/frontend/resources/styles/common/refactor/color-defs.scss
@@ -41,9 +41,6 @@
 
   --status-color-info-500: #0e9be9; // used on pixel grid and status widget
 
-  //GENERIC
-  --color-canvas: #e8e9ea; // Not defined on DS
-
   // APP COLORS
   --app-white: #ffffff; // Used in several places
   --app-black: #000; // Used on interactions, measurements and editor files
--- a/frontend/src/app/main/ui/ds/colors.scss
+++ b/frontend/src/app/main/ui/ds/colors.scss
@@ -9,6 +9,8 @@
 $mint-150: #7efff5;
 $mint-250: #00d1b8;
 $mint-700: #426158;
+$mint-150-60: #7efff599;
+$mint-250-10: #00d1b81a;
 
 $green-200: #a7e8d9;
 $green-500: #2d9f8f;
@@ -29,6 +31,8 @@ $purple-200: #e1d2f5;
 $purple-400: #bb97d8;
 $purple-600: #8c33eb;
 $purple-700: #6911d4;
+$purple-600-10: #8c33eb1a;
+$purple-700-60: #6911d499;
 
 $blue-200: #bae3fd;
 $blue-500: #0e9be9;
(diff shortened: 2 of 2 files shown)


niwinz:
Hello, thanks for your contribution.
Can you please change the commit message to use the `:bug:` instead of 🐛 unicode character

VVege:
> Hello, thanks for your contribution. Can you please change the commit message to use the `:bug:` instead of 🐛 unicode character

done

EvaMarco:
Hello Wvege¡¡ 
First thanks for your contribution. 
Second Your solution does not cover all use cases in the components section. Let me see if I can explain it well.

That sidebar has an adjustable width, i.e. there is a handler to resize it in width. With the aspect ratio property we wanted to maintain the look of the grid with square elements but that these were adjusted to the available size. 

This is how it should work
[Screencast from 2024-08-23 12-21-46.webm](https://github.com/user-attachments/assets/72e39e2d-be18-4604-8afc-ddd6f411d2c0)

And here's how your example works by setting the size
[Screencast from 2024-08-23 12-21-46.webm](https://github.com/user-attachments/assets/99f16b31-672a-446d-b1ae-e6398f98a0b2)
As you can see in your solution there are strange jumps and the gap between elements is not respected.

But knowing that aspect-ratio is generating problems I am looking for a solution to take advantage of the grid layout properties to avoid its use.

Here is my idea so far. https://github.com/penpot/penpot/pull/5022

VVege:
> Hello Wvege¡¡ First thanks for your contribution. Second Your solution does not cover all use cases in the components section. Let me see if I can explain it well.
> 
> That sidebar has an adjustable width, i.e. there is a handler to resize it in width. With the aspect ratio property we wanted to maintain the look of the grid with square elements but that these were adjusted to the available size.
> 
> This is how it should work [Screencast from 2024-08-23 12-21-46.webm](https://github.com/user-attachments/assets/72e39e2d-be18-4604-8afc-ddd6f411d2c0)
> 
> And here's how your example works by setting the size [Screencast from 2024-08-23 12-21-46.webm](https://github.com/user-attachments/assets/99f16b31-672a-446d-b1ae-e6398f98a0b2) As you can see in your solution there are strange jumps and the gap between elements is not respected.
> 
> But knowing that aspect-ratio is generating problems I am looking for a solution to take advantage of the grid layout properties to avoid its use.
> 
> Here is my idea so far. #5022

ok you are right,and i didnt consider this fuction. and i update my code , but I accidentally closed this PR，and I submit anther pr, https://github.com/penpot/penpot/pull/5025 .

PR #5009, opened 2024-08-19
:tada: Add `:assign` operation as alternative to `:set`

niwinz:
This operation (part of the update-file DSL) is alternative to the already existing `:set` operation. It is intended to be used as transform friendly alternative when JSON encoding is used (instead of transit+json) for interact with update-file rpc method.

The commit adds several FIXME coments that will be addressed on other PR.

### DIFF
--- a/common/src/app/common/types/container.cljc
+++ b/common/src/app/common/types/container.cljc
@@ -540,38 +540,51 @@
 ;; --- SHAPE UPDATE
 
 (defn set-shape-attr
-  [shape attr val & {:keys [on-changed ignore-touched ignore-geometry]}]
-  (let [group           (get ctk/sync-attrs attr)
-        shape-val       (get shape attr)
-        ignore          (or ignore-touched (= attr :position-data)) ;; position-data is a derived attribute and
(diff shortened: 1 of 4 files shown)


common/src/app/common/types/container.cljc
```diff
@@ -540,38 +540,51 @@
...
-        ignore          (or ignore-touched (= attr :position-data)) ;; position-data is a derived attribute and
-        is-geometry?    (and (or (= group :geometry-group)          ;; never triggers touched by itself
-                                 (and (= group :content-group) (= (:type shape) :path)))
-                             (not (#{:width :height} attr))) ;; :content in paths are also considered geometric
-                        ;; TODO: the check of :width and :height probably may be removed
-                        ;;       after the check added in data/workspace/modifiers/check-delta
-                        ;;       function. Better check it and test toroughly when activating
-                        ;;       components-v2 mode.
-        in-copy?        (ctk/in-component-copy? shape)
+  "Assing attribute to shape with touched logic.
```
superalex: assing -> assign :)

PR #5015, opened 2024-08-22
Update manage.sh

yamila-moreno:
Delete `-x` option as it gives an annoying output when we're not debugging.


niwinz:
I'm not clearly convinced on disabling this, I personally always want to know what is exactly executing right now and without debugging evidently removes that.

niwinz:
I merge it for now, but i not discarding that in some future it appears again hehe

PR #5038, opened 2024-08-30
:tada: Add export event for telemetry

pabloalba:


frontend/src/app/main/ui/viewer/inspect/exports.cljs
```diff
@@ -62,8 +64,14 @@
...
-                (st/emit! (de/request-simple-export {:export (first exports)}))
-                (st/emit! (de/request-multiple-export {:exports exports :filename filename}))))))
+                (st/emit!
+                 (de/request-simple-export {:export (first exports)})
+                 (ptk/event
+                  ::ev/event {::ev/name "export" :exports (str exports)}))
+                (st/emit!
+                 (de/request-multiple-export {:exports exports})
+                 (ptk/event
+                  ::ev/event {::ev/name "export" :exports (str exports)}))))))
```
niwinz: I think we should track the origin of the event, and use a more specific event name

niwinz: `export` is very generic and can be confused with exporting .zip or .penpot

superalex: agree :)

pabloalba: fixed

niwinz:
For me it is ok for merge it

superalex:
If I export multiple shapes from right button the event origin is not ok:





Also I don't know if adding a str version of the params is really useful or we can just ignore it:

niwinz:
Oh, I overlooked the exports as string thing. I think it doesn't make sense, either we pass it as data or we don't pass it but passing it as a string doesn't add much.

pabloalba:
> If I export multiple shapes from right button the event origin is not ok:

Fixed

pabloalba:
> Oh, I overlooked the exports as string thing. I think it doesn't make sense, either we pass it as data or we don't pass it but passing it as a string doesn't add much.

Yes, the export info doesn't have much sense. I've changed it for the number of exported shapes, that can be useful

PR #5071, opened 2024-09-06
:bug: Fix concurent edition issues with properties stored on page options

niwinz:
## On this PR:

Deprecate/Remove the set-option change (because of concurrent edition issues on storing big data structures on with it,where it is intended to store simple key value properties)
- Add new changes for CRUD of guides, default-grids (saved-grids), flows and thread comment positions
- Add migration for the page data structure change that maintains the old values for transient backward compatibility.

## How to test:

Previous steps: create a file with guides, flows, saved grids (only square, you can’t save other type of grids because of other bug) and custom background. Then:

- Test if with new code all this is preserved (means migration is working correctly)
- Test if you go back to develop branch, you still have all the data (the original one, means transient backward compatibility is working correctly)

Returning back to the branch of this MR:

- Test if you can create, modify or delete guides
- Test if you can save default grids
- Test if you can save a custom page background
- Test if you can create, modify or delete flows
- Test if plugins data is properly saved

It is also reasonable to test all the related parts of the application where this objects are used (per example viewer, export/import, etc..)


common/src/app/common/files/changes_builder.cljc
```diff
@@ -246,42 +259,80 @@
...
+                    (update :redo-changes conj {:type :set-guide
+                                                :page-id page-id
+                                                :id id
+                                                :params guide})
+                    (update :undo-changes conj {:type :set-guide
+                                                :page-id page-id
+                                                :id id
+                                                :params old-val}))]
+    ;; FIXME: not sure if we need this
+    (apply-changes-local changes)))
```
Alotor: I don't think is necessary. `apply-changes-local` persists the shape changes into the local copy of the objects. So should be only necessary for changes to shapes. For these options no shape changes are necessary.

frontend/src/app/main/ui/workspace/viewport/guides.cljs
```diff
@@ -256,14 +258,15 @@
     (and (>= (:position guide) (:y frame))
          (<= (:position guide) (+ (:y frame) (:height frame))))))
 
-(mf/defc guide
-  {::mf/wrap [mf/memo]}
-  [{:keys [guide hover? on-guide-change get-hover-frame vbox zoom hover-frame disabled-guides? frame-modifier]}]
+(mf/defc guide*
+  {::mf/wrap [mf/memo]
+   ::mf/props :obj}
+  [{:keys [guide hover? on-guide-change get-hover-frame vbox zoom hover-frame disabled-guides frame-modifier]}]
```
Alotor: maye change `hover?` to `is-hover`?

PR #5077, opened 2024-09-09
Set up devenv for Rust

belen-albeza:
Closes https://tree.taiga.io/project/penpot/task/8638

- Sets up our local devenv to be able to use Rust compiler + Cargo (build/dependency manager tool)
- Sets up a build script (`build:renderer`) to compile a Rust lib to WebAssembly, using [wasm-pack](https://github.com/rustwasm/wasm-pack).
- Adds a dummy Rust project that provides a web assembly function (`print`) that makes use of the browser API (`console.log`)
- Loads the wasm module from our workspace, walled by a config flag (`"enable-renderer-v2"`). Then, calls a wasm function that prints `"Hello from wasm fn!"` in the console.

To try this PR:

1. Build the new local devenv with `./manage.sh build-devenv-local`
2. Start the devenv `./manage.sh start-devenv`
3. In a tmux tab, `cd penpot/frontend` and build the renderer with `yarn run build:renderer`.
4. Open Penpot and open a file in the workspace
5. Check that `"Hello from wasm fn!"` _is not_ printed in the browser console.
6. Update your `resources/public/js/config.js` file to add the flag `"enable-renderer-v2"`.
7. Reload the workspace.
8. Check that `"Hello from wasm fn!"` does get logged in the console.


docker/devenv/Dockerfile
```diff
@@ -242,6 +244,31 @@ RUN set -ex; \
...
+    # Same steps as in Rust official Docker image https://github.com/rust-lang/docker-rust/blob/9f287282d513a84cb7c7f38f197838f15d37b6a9/1.81.0/bookworm/Dockerfile
+    dpkgArch="$(dpkg --print-architecture)"; \
+    case "${dpkgArch##*-}" in \
+        amd64) rustArch='x86_64-unknown-linux-gnu'; rustupSha256='6aeece6993e902708983b209d04c0d1dbb14ebb405ddb87def578d41f920f56d' ;; \
+        armhf) rustArch='armv7-unknown-linux-gnueabihf'; rustupSha256='3c4114923305f1cd3b96ce3454e9e549ad4aa7c07c03aec73d1a785e98388bed' ;; \
+        arm64) rustArch='aarch64-unknown-linux-gnu'; rustupSha256='1cffbf51e63e634c746f741de50649bbbcbd9dbe1de363c9ecef64e278dba2b2' ;; \
+        i386) rustArch='i686-unknown-linux-gnu'; rustupSha256='0a6bed6e9f21192a51f83977716466895706059afb880500ff1d0e751ada5237' ;; \
+        ppc64el) rustArch='powerpc64le-unknown-linux-gnu'; rustupSha256='079430f58ad4da1d1f4f5f2f0bd321422373213246a93b3ddb53dad627f5aa38' ;; \
+        s390x) rustArch='s390x-unknown-linux-gnu'; rustupSha256='e7f89da453c8ce5771c28279d1a01d5e83541d420695c74ec81a7ec5d287c51c' ;; \
+    *) echo >&2 "unsupported architecture: ${dpkgArch}"; exit 1 ;; \
```
niwinz, line 258: we only support amd64 and arm64 for our devenv,so the rest are useless because they will never execute

docker/devenv/files/start-tmux.sh
```diff
@@ -26,6 +26,7 @@ tmux send-keys -t penpot 'yarn run watch' enter
 tmux new-window -t penpot:1 -n 'frontend shadow'
 tmux select-window -t penpot:1
 tmux send-keys -t penpot 'cd penpot/frontend' enter C-l
+tmux send-keys -t penpot 'yarn run build:renderer' enter C-l
```
niwinz: I think for now, we don't need to add new window to the tmux just for building process. This is still an experiment and it should be build on demand (In the same way this renderer is under configuration flag, so we activate it explicitly when we going to test it, we can open a specific window on demand for building it)

Adding new window for all users independently if they go or not to develop/test this I think is not necessary

frontend/scripts/build
```diff
@@ -20,6 +20,7 @@ yarn install || exit 1;
 rm -rf resources/public;
 rm -rf target/dist;
 
+yarn run build:renderer
```
niwinz: This is a experiment code that is not going to go to be deployed right now until we have something usable. 

We shouldn't penalize every frontend build with a module that is only going to be used locally for the moment.

frontend/shadow-cljs.edn
```diff
@@ -67,6 +67,11 @@
     :output-wrapper false
     :warnings {:fn-deprecated false}}
 
+   :js-options
+   {:entry-keys ["module" "browser" "main"]
+    :resolve {"renderer" {:target :file
+                          :file "resources/public/js/renderer/renderer.js"}}}
```
niwinz: If the module is going to be imported on the code, it should not be located on resources/public/, resource public is a directory from where the browser (example: from index.html) is importing.

But as you already import it from code inside src, on generating the bundle, it will be already be bundled inside public/js by shadow-cljs.

frontend/src/app/util/wasm.cljs
```diff
@@ -0,0 +1,23 @@
...
+
+(defn foo [] (renderer/print "Lorem ipsum"))
+
+(defn init
+  []
+  (ptk/reify ::init
+    ptk/WatchEvent
+    (watch [_ _ _] ;; TODO: mirar la docu de potok
+      (->> (rx/from (renderer-init))
+           (rx/tap foo)
```
niwinz: I know that this code is still an experiment, and it is ok to have this on app.util.wasm but I think on the end, it should not be under `app.util`. We probably need to organize it under `app` (I mean something like `app.renderer-v2`.
There is NO need to make changes now, but we do have to take this into account for the future to organize the code well.

frontend/shadow-cljs.edn
belen-albeza: You're right, this is a leftover from a previous version, shouldn't be here

PR #5115, opened 2024-09-24
Plugins update

Alotor:


frontend/src/app/main/data/plugins.cljs
```diff
@@ -0,0 +1,42 @@
...
+(ns app.main.data.plugins
+  (:require
+   [app.plugins.register :as pr]
+   [potok.v2.core :as ptk]))
+
+(defn open-plugin!
+  [{:keys [plugin-id name description host code icon permissions]}]
+  (try
+    (.ɵloadPlugin
+     js/window
```
niwinz: should be using app.util.global instead js/window

frontend/src/app/main/data/plugins.cljs
```diff
@@ -0,0 +1,42 @@
...
+      (assoc state ::open-plugin (:plugin-id plugin)))))
+
+(defn check-open-plugin
+  []
+  (ptk/reify ::check-open-plugin
+    ptk/WatchEvent
+    (watch [_ state _]
+      (when-let [pid (::open-plugin state)]
+        (open-plugin! (pr/get-plugin pid))
+        (fn [state]
```
niwinz: You probably want here is the (rx/of (fn [state] ...)) instead of simply return a funcition

frontend/src/app/main/ui/routes.cljs
```diff
@@ -94,10 +95,17 @@
 (defn on-navigate
   [router path]
   (let [location (.-location js/document)
+        [base-path qs] (str/split path "?")
+
+        qstring
```
niwinz: maybe we can use the app.common.uri/query-string->map helper that already does this

frontend/src/app/main/data/plugins.cljs
Alotor: ups! you're right. I moved the code and broke this.

frontend/src/app/main/data/plugins.cljs
Alotor: ok

PR #5118, opened 2024-09-25
[WIP] Target Emscripten

belen-albeza:
After install Emscripten with a package manager, this should work (it generates both a `.js` and a `.wasm` file):

```
rustc src/lib.rs --target=wasm32-unknown-emscripten
```

However we need to make it work with cargo (I'm getting a linking error, probably due to the way I installed emscripten). 

**UPDATE**

Compile with:

```
EMCC_CFLAGS="--no-entry -s MODULARIZE=1 -s EXPORT_NAME=createRustSkiaModule -s ENVIRONMENT=web -s MAX_WEBGL_VERSION=2 -s EXPORTED_RUNTIME_METHODS=GL" cargo build --target=wasm32-unknown-emscripten
```


belen-albeza:
Closing in favor of #5136

PR #5135, opened 2024-10-02
✨  Add viewer role

EvaMarco:
Implements :

- https://tree.taiga.io/project/penpot/us/1056?milestone=407454
- https://tree.taiga.io/project/penpot/us/6590?milestone=407454
- https://tree.taiga.io/project/penpot/us/6811?milestone=407454
- https://tree.taiga.io/project/penpot/us/8869?milestone=407454
- https://tree.taiga.io/project/penpot/us/8893?milestone=407454
- https://tree.taiga.io/project/penpot/task/8973


backend/src/app/rpc/commands/teams.clj
```diff
@@ -707,6 +709,13 @@
       (db/delete! conn :team-profile-rel {:profile-id member-id
                                           :team-id team-id})
 
+      (mbus/pub! msgbus
+                 :topic member-id
+                 :message {:type :removed-from-team
```
niwinz: I think we should approach this message from distinct point of view. My proposal is to rename it to `:team-membership-change` with the appropriate key for indicate the operation. Example:

```clojure
{:type :team-membership-change
 :operation :delete
 :subs-id member-id
 :member-id member-id
 :team-id team-id
 :team-name "team name"}
 ```

File: backend/src/app/rpc/commands/teams.clj [hunk 2]
```diff
@@ -655,6 +650,13 @@
       (ex/raise :type :validation
                 :code :cant-promote-to-owner))
 
+    (mbus/pub! msgbus
+               :topic member-id
+               :message {:type :team-permissions-change
```
niwinz: After thinking a bit, I think the message should be `:team-permission-change` (removing the `s` from permissions)

niwinz: Or better: `:team-role-change`

frontend/src/app/main/data/common.cljs
```diff
@@ -170,3 +173,65 @@
...
+        (rx/of (ntf/info msg))))
+
+    ptk/UpdateEvent
+    (update [_ state]
+      (let [route (if workspace?
+                    [:workspace-file :permissions]
+                    [:teams team-id :permissions])]
+        (update-in state route
+                   (fn [permissions]
+                     (cond
```
niwinz: I think we can extract this logic to `app.common.types.team`, the logic of changing a permission objects with a role keyword is also used on backend I think, so we can start reusing this code also.

frontend/src/app/main/data/workspace/text/shortcuts.cljs
```diff
@@ -189,7 +189,10 @@
 
 (defn- update-attrs-when-no-readonly [props]
   (let [undo-id (js/Symbol)
-        read-only?           (deref refs/workspace-read-only?)
+        file                 (deref refs/workspace-file)
+        user-viewer?         (not (get-in file [:permissions :can-edit]))
```
niwinz: use `dm/get-in` for performance

frontend/src/app/main/data/changes.cljs
```diff
@@ -178,19 +178,23 @@
...
-                   (assoc :file-id file-id)
-                   (assoc :file-revn (resolve-file-revn state file-id))
-                   (assoc :undo-changes uchg)
-                   (assoc :redo-changes rchg)
-                   (commit)))))))
+      (let [file-id      (or file-id (:current-file-id state))
+            uchg         (vec undo-changes)
+            rchg         (vec redo-changes)
+            features     (features/get-team-enabled-features state)
+            user-viewer? (not (get-in state [:workspace-file :permissions :can-edit]))]
```
niwinz: Prefer, dm/get-in over get-in when the vector of traversal path is known statically
Also:
- it is ok for a first approach, but I think we need to start to abstract the permission checking to named funcions

pabloalba: I'm not sure... The permissions are in different places for dashboard and workspace. I've tried  a few things, but it becomes a little convoluted. I propose don't make this messy, and address the abstraction when we need it

frontend/src/app/main/data/common.cljs
```diff
@@ -174,6 +174,12 @@
              (rx/tap on-success)
              (rx/catch on-error))))))
 
+(def change-role-msg
+  {:viewer "dashboard.permissions-change.viewer"
+   :editor "dashboard.permissions-change.editor"
+   :admin  "dashboard.permissions-change.admin"
+   :owner  "dashboard.permissions-change.owner"})
```
niwinz: I think instead of defining plain map, it should be defined as function, because you need to use the `tr` function statically visible for the `translations` script correctly collect used and unused translation strings.

With this approach, the resolution for `tr` some lines below is dynamic and can't be statically analized by external scripts

common/src/app/common/types/team.cljc
```diff
@@ -7,4 +7,11 @@
...
 (def valid-roles
-  #{:owner :admin :editor :viewer})
\ No newline at end of file
+  #{:owner :admin :editor :viewer})
+
+(def permissions-for-role
+  {:viewer [:can-edit false :is-admin false :is-owner false]
+   :editor [:can-edit true :is-admin false :is-owner false]
+   :admin  [:can-edit true :is-admin true :is-owner false]
+   :owner  [:can-edit true :is-admin true :is-owner true]})
```
niwinz: I guess this should be a map instead of vector?

frontend/src/app/main/data/common.cljs
pabloalba: fixed

common/src/app/common/types/team.cljc
pabloalba: ops! fixed

PR #5172, opened 2024-10-15
:tada: Add binfile-v3 export/import file format

niwinz:
For enable, you should use `enable-export-file-v3` flag on frontend (on `config.js`)
It maintains the backward compatibility with the previous format.

Missing from this PR:
- the ability to rename files before import them
- see import progress in the same way as legacy-zip format


backend/src/app/binfile/v3.clj
```diff
@@ -0,0 +1,952 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC
+
+(ns app.binfile.v3
+  "A sqlite3 based binary file exportation with support for exportation
```
hirunatan: This is not sqlite3 but zip based.

backend/src/app/binfile/v3.clj
```diff
@@ -0,0 +1,952 @@
...
+  [^ZipFile input]
+  (into #{} (iterator-seq (.entries input))))
+
+(defn- get-zip-entry*
+  [^ZipFile input ^String name]
+  (.getEntry input name))
+
+(defn- get-zip-entry
+  [input path]
+  (let [entry (get-zip-entry* input path)]
```
hirunatan, line 369: why is it called `path` here but `name` in the * function?

hirunatan (approved):

PR #5201, opened 2024-10-25
:sparkles: Shape with buffer

niwinz:


frontend/src/app/main/data/workspace/selection.cljs
```diff
@@ -68,13 +68,13 @@
...
-                                (update :x2 + (:x delta))
-                                (update :y2 + (:y delta)))
+              (let [selrect (-> (cr/clone selrect)
+                                (cr/update! :x2 + (:x delta))
+                                (cr/update! :y2 + (:y delta)))
                     selrect (if ^boolean space?
                               (-> selrect
-                                  (update :x1 + (:x delta))
-                                  (update :y1 + (:y delta)))
+                                  (cr/update! :x1 + (:x delta))
```
niwinz: if you dont use clone, you should never use cr/update! this will break unexpectly at some point because you are mutating object that is not expected to be mutated

niwinz: if you remove cr/clone, you should always replace cr/update! with `update`

PR #5235, opened 2024-10-30
:tada: Handle WebGL Context

AzazelN28:
When the context is lost, an event is triggered that updates the render state and shows a message to restore context.

### DIFF
--- a/frontend/src/app/main/ui/workspace/viewport_wasm.cljs
+++ b/frontend/src/app/main/ui/workspace/viewport_wasm.cljs
@@ -111,6 +111,8 @@
         modifiers         (mf/deref refs/workspace-modifiers)
         text-modifiers    (mf/deref refs/workspace-text-modifier)
 
+        render-context-lost? (mf/deref refs/render-context-lost?)
+
         objects-modified  (mf/with-memo [base-objects text-modifiers modifiers]
                             (apply-modifiers-to-selected selected base-objects text-modifiers modifiers))
 
@@ -174,6 +176,8 @@
         grid-editing?     (and edition (ctl/grid-layout? base-objects edition))
 
         mode-inspect?       (= options-mode :inspect)
+
+        on-render-restore-context #(.reload js/location)
 
         on-click          (actions/on-click hover selected edition drawing-path? drawing-tool space? selrect z?)
         on-context-menu   (actions/on-context-menu hover hover-ids read-only?)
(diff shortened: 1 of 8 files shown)


render-wasm/src/render.rs
```diff
@@ -31,8 +36,23 @@ impl State {
...
+// fn gl_context_lost(_event_type: i32, _reserved: *const std::os::raw::c_void, _user_data: *mut std::os::raw::c_void) -> bool {
+//     println!("Context Lost!");
+//     true
+// }
+
+// fn gl_context_restored(_event_type: i32, _reserved: *const std::os::raw::c_void, _user_data: *mut std::os::raw::c_void) -> bool {
+//     println!("Context Restored!");
+//     true
+// }
+
```
belen-albeza: nit: let's remove these leftover commented code

frontend/src/app/render_wasm.cljs
```diff
@@ -82,29 +86,67 @@
...
-        _       (.makeContextCurrent ^js gl handle)
+        _       (.makeContextCurrent ^js gl handle)]
 
-        ;; Initialize Skia
-        state   (init-fn (.-width ^js canvas)
-                         (.-height ^js canvas))]
+    (when (dbg/enabled? :gl-context)
+      (init-debug-webgl-context-state context))
+
+    (.addEventListener canvas "webglcontextlost" on-webgl-context)
```
belen-albeza, line 143: Question: when is this triggered, exactly? When a user returns to an inactive tab, for instance? Or when the tab loses focus?

frontend/src/app/main/ui/workspace/viewport_wasm.cljs
```diff
@@ -635,4 +641,12 @@
...
-          :bottom-padding (when palete-size (+ palete-size 8))}]]]]]))
+          :bottom-padding (when palete-size (+ palete-size 8))}]]]]
+
+     (when render-context-lost?
+       [:div {:id "context-lost" :class (stl/css :context-lost)}
+        [:h1 "MEECK!!! CONTEXT LOST"]
+        [:h2 "Design is my passion"]
+        [:button
+         {:on-click on-render-restore-context}
+         "Restore context"]])]))
```
belen-albeza: Question: Why the restore needs to be triggered by a user? Isn't `restore-canvas` called by the `webglcontextrestore` handler?

frontend/src/app/render_wasm.cljs
AzazelN28: This is usually triggered when GPU resources are exhausted, for example: when are multiple WebGL tabs open and browser needs to open a new one. It can also happen when you suspend your computer.

frontend/src/app/main/ui/workspace/viewport_wasm.cljs
AzazelN28: Sometimes the context is recoverable, in that case `webglcontextrestored` is called automatically and sometimes it isn't so we need to inform the user that she needs to take action.

PR #5261, opened 2024-11-07
:sparkles: Integrate viewer role with plugin menus and popup

pabloalba:
Implements https://tree.taiga.io/project/penpot/us/8928

### DIFF
--- a/frontend/src/app/main/data/plugins.cljs
+++ b/frontend/src/app/main/data/plugins.cljs
@@ -8,12 +8,22 @@
   (:require
    [app.common.data.macros :as dm]
    [app.main.data.modal :as modal]
+   [app.main.data.notifications :as ntf]
    [app.main.store :as st]
    [app.plugins.register :as preg]
    [app.util.globals :as ug]
    [app.util.http :as http]
+   [app.util.i18n :as i18n :refer [tr]]
+   [app.util.time :as dt]
    [beicon.v2.core :as rx]
    [potok.v2.core :as ptk]))
+
+(defn save-plugin-permissions-peek
+  [id permissions]
+  (ptk/reify ::save-plugin-permissions-peek
+    ptk/UpdateEvent
+    (update [_ state]
+      (assoc-in state [:plugins-permissions-peek :data id] permissions))))
 
 (defn fetch-manifest
   [plugin-url]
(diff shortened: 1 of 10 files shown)


frontend/src/app/main/ui/workspace/main_menu.cljs
```diff
@@ -653,24 +655,42 @@
...
+                                                       :host host}))
+                      (dp/open-plugin! manifest user-can-edit?))
+                    (dom/stop-propagation event))))
+               on-key-down
+               (mf/use-fn
+                (mf/deps can-open? name host manifest user-can-edit?)
+                (fn [event]
+                  (when can-open?
+                    (when (kbd/enter? event)
+                      (do
```
niwinz: this `do` is redundant

frontend/src/app/main/ui/workspace/plugins.cljs
```diff
@@ -204,6 +222,7 @@
              [:& plugin-entry {:key (dm/str "plugin-" idx)
                                :index idx
                                :manifest manifest
+                               :user-can-edit? user-can-edit?
```
niwinz: Don't use `?` on component props, props is a data structure

frontend/src/app/main/data/plugins.cljs
```diff
@@ -8,13 +8,23 @@
...
    [app.main.store :as st]
    [app.plugins.register :as preg]
    [app.util.globals :as ug]
    [app.util.http :as http]
+   [app.util.i18n :as i18n :refer [tr]]
+   [app.util.time :as dt]
    [beicon.v2.core :as rx]
    [potok.v2.core :as ptk]))
 
+(defn save-plugin-permissions-peek
```
niwinz, line 21: I'm not very sure if `-peek` is the right suffix here, I think you may want something like `-pick`
And probably you should declare this function as private as it is used only from the same namespace

frontend/src/app/main/data/plugins.cljs
```diff
@@ -8,13 +8,23 @@
...
+  [id permissions]
+  (ptk/reify ::save-plugin-permissions-peek
+    ptk/UpdateEvent
+    (update [_ state]
+      (assoc-in state [:profile :props :plugins-permissions-peek :data id] permissions))))
```
niwinz: I think is not a good idea to use profile props to store this, firstly because this property is easy to be overwritten on single fetch of profile; i mean is a backend synchronized property.
Also you don't need to creeate so nested data structure. I mean the all global state is profile oriented, so you don't need at all neest it under profile, you can create a new, specific to the plugins permisions key on the root of the state, making it more simple to access. Also, looks like `:data` only contains the `:id`, can this be stored directly under parent key without neesting it under `:data`?

pabloalba: I put this there because is where the plugins info is stored. This is kind of a plugins permissions cache, so it made sense to have it next to the real plugins info

frontend/src/app/main/data/plugins.cljs
pabloalba: This is like a "sneak peek" of the permissions. I think "-pick" doesn't make sense. I tough about "-cache" but that is misleading, it is not a cache...

frontend/src/app/main/ui/workspace/main_menu.cljs
pabloalba: fixed

frontend/src/app/main/ui/workspace/plugins.cljs
pabloalba: Fixed

PR #5290, opened 2024-11-12
:bug: Several bugfixes related to the request-access feature

niwinz:
Should be revised by commit. Each commit has its own small description.

TLDR:
- Fixes an issue on handle conflict on creation of team access request (probably due to race condition, no taiga issue, detected on our automatic error reports)
- Fixed incorrect file data loading (offloaded file data will fail) on creating team access request
- Improve query efficiency on team access request creation RPC method and remove unnecessary validation
- Fix input-multi component initialization and handling initial data
- Removes the hard coupling between input-multi component and invitations, now it uses the form initial data instead (the default mechanism)
- Refresh the team members after sending an invitation (for better UX when a user accepts a team access request)

### DIFF
--- a/backend/src/app/util/time.clj
+++ b/backend/src/app/util/time.clj
@@ -158,6 +158,7 @@
      :iso8601 (Instant/from (.parse DateTimeFormatter/ISO_INSTANT ^String s)))))
 
 (defn is-after?
+  "Analgous to: da > db"
   [da db]
   (.isAfter ^Instant da ^Instant db))
 
--- a/frontend/src/app/main/ui/components/forms.cljs
+++ b/frontend/src/app/main/ui/components/forms.cljs
@@ -420,14 +420,25 @@
   (into [] (distinct) (conj coll item)))
 
 (mf/defc multi-input
-  [{:keys [form label class name trim valid-item-fn caution-item-fn on-submit invite-email] :as props}]
+  [{:keys [form label class name trim valid-item-fn caution-item-fn on-submit] :as props}]
   (let [form       (or form (mf/use-ctx form-ctx))
         input-name (get props :name)
         touched?   (get-in @form [:touched input-name])
         error      (get-in @form [:errors input-name])
         focus?     (mf/use-state false)
 
-        items      (mf/use-state [])
+        items      (mf/use-state
+                    (fn []
+                      (let [initial (get-in @form [:data input-name])]
+                        (if (or (vector? initial)
+                                (set? initial))
+                          (mapv (fn [val]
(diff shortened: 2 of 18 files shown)


backend/src/app/util/time.clj
```diff
@@ -158,6 +158,7 @@
      :iso8601 (Instant/from (.parse DateTimeFormatter/ISO_INSTANT ^String s)))))
 
 (defn is-after?
+  "Analgous to: d1 > d2"
```
pabloalba: Very minor, but it's better to respect the name of the parameters: da > db

frontend/src/app/main/ui/components/forms.cljs
```diff
@@ -420,14 +420,24 @@
...
 
-        items      (mf/use-state [])
+        items      (mf/use-state
+                    (fn []
+                      (let [initial (get-in @form [:data input-name])]
+                        (if (or (vector? initial)
+                                (set? initial))
+                          (mapv (fn [val]
+                                  {:text val
+                                   :valid (valid-item-fn val)})
```
pabloalba: missing `:caution (caution-item-fn val)`

frontend/src/app/main/ui/dashboard/team.cljs
```diff
@@ -74,14 +77,14 @@
...
-         (mf/deps team)
+         (mf/deps team invite-email)
          (fn []
            (st/emit! (modal/show {:type :invite-members
                                   :team team
                                   :origin :team
                                   :invite-email invite-email}))))]
 
-    (mf/with-effect []
+    (mf/with-effect [team]
```
pabloalba: It should be  (mf/with-effect [invite-email]

frontend/translations/en.po
```diff
@@ -421,7 +421,7 @@ msgid "dashboard.download-binary-file"
 msgstr "Download Penpot file (.penpot)"
 
 msgid "dashboard.download-binary-file-v3"
```
pabloalba, line 423: Missing spanish translation

backend/src/app/util/time.clj
niwinz: true

frontend/src/app/main/ui/components/forms.cljs
niwinz: true

frontend/translations/en.po
niwinz: Yes, it already was missing. I have still pending tasks for it. This will be fixed in other PR

PR #5294, opened 2024-11-13
:tada: Add feature flag for wasm render

superalex:
- https://tree.taiga.io/project/penpot/task/9262

### DIFF
--- a/common/src/app/common/features.cljc
+++ b/common/src/app/common/features.cljc
@@ -51,7 +51,8 @@
     "layout/grid"
     "plugins/runtime"
     "design-tokens/v1"
-    "text-editor/v2"})
+    "text-editor/v2"
+    "render-wasm/v1"})
 
 ;; A set of features enabled by default
 (def default-features
@@ -67,7 +68,8 @@
 (def frontend-only-features
   #{"styles/v2"
     "plugins/runtime"
-    "text-editor/v2"})
+    "text-editor/v2"
+    "render-wasm/v1"})
 
 ;; Features that are mainly backend only or there are a proper
 ;; fallback when frontend reports no support for it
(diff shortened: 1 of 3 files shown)


common/src/app/common/features.cljc
```diff
@@ -86,7 +88,8 @@
         "fdata/shape-data-type"
         "plugins/runtime"
         "design-tokens/v1"
-        "text-editor/v2"}
+        "text-editor/v2"
+        "render-wasm/v1"}
```
niwinz: Yo don't need to add here the flag, if it is already on `frontend-only-features` because these two sets are merged.

superalex: But it's the same situation that we had with text-editor/v2 and we needed to add it too

frontend/src/app/render_wasm.cljs
```diff
@@ -10,12 +10,13 @@
...
    [app.common.uuid :as uuid]
-   [app.config :as cf]
+   [app.main.features :as features]
+   [app.main.store :as st]
    [app.util.object :as obj]
    [promesa.core :as p]))
 
 (def enabled?
-  (contains? cf/flags :render-wasm))
+  (features/active-feature? @st/state "render-wasm/v1"))
```
niwinz, line 19: This is no longer valid, because it can't be defined in this way. I mean having it as feature flag that can be altered by team, this variable will be set to true or false on the initialization, when team is still not loaded and will never change. 

This static var at namespace level is no longer the way to check for feature, the feature should be checked correctly where it is initialized and this var should not exist

niwinz: I have a separated task for handle workspace initialization + wasm renderer, so I guess we still can merge it as-is and address it in the separated task.

PR #5295, opened 2024-11-13
:tada: Refactor naming anidated shapes with children for wasm render

superalex:


frontend/src/app/render_wasm.cljs
```diff
@@ -81,7 +81,7 @@
           (set-shape-selrect selrect)
           (set-shape-rotation rotation)
           (set-shape-transform transform)
-          (set-shapes children)
+          (set-children children)
```
niwinz: Should it be `set-shape-children`?

superalex: Fixed!, thanks :)

PR #5296, opened 2024-11-13
Bugs viewer role

pabloalba:
Fixes:

https://tree.taiga.io/project/penpot/issue/9098
https://tree.taiga.io/project/penpot/issue/9097

### DIFF
--- a/frontend/src/app/main/data/dashboard.cljs
+++ b/frontend/src/app/main/data/dashboard.cljs
@@ -1191,10 +1191,12 @@
                             {:name name
                              :team-id team-id})
             action-name   (if in-project? :create-file :create-project)
-            action        (if in-project? file-created project-created)]
+            action        (if in-project? file-created project-created)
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/data/dashboard/shortcuts.cljs
```diff
@@ -31,7 +32,9 @@
    :create-new-project {:tooltip "+"
                         :command "+"
                         :subsections [:general-dashboard]
-                        :fn #(st/emit! (dd/create-element))}
+                        :fn #(let [can-edit?  (:can-edit (deref refs/permissions))]
+                               (when can-edit?
+                                 (st/emit! (dd/create-element))))}
```
niwinz: The permission check: can it be handled in the event impl instead of outside the event? From the event impl you already have direct access to the state and to the permissions.

pabloalba: Ok

PR #5303, opened 2024-11-14
:sparkles: Add versions support in plugins

Alotor:
- Add support for file history versions in plugins

### DIFF
--- a/frontend/src/app/main/data/workspace/versions.cljs
+++ b/frontend/src/app/main/data/workspace/versions.cljs
@@ -72,6 +72,35 @@
                   (fetch-versions file-id)))))
          (rx/of (ptk/event ::ev/event {::ev/name "create-version"})))))))
 
+(defn create-version-from-plugins
+  [file-id label resolve reject]
+  (dm/assert! (uuid? file-id))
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/data/workspace/versions.cljs
```diff
@@ -72,6 +72,34 @@
...
+                 (rx/mapcat #(rp/cmd! :create-file-snapshot {:file-id file-id :label label}))
+
+                 (rx/mapcat
+                  (fn [{:keys [id]}]
+                    (->> (rp/cmd! :get-file-snapshots {:file-id file-id})
+                         (rx/take 1)
+                         (rx/map (fn [versions] (d/seek #(= id (:id %)) versions))))))
+                 (rx/tap resolve)
+                 (rx/ignore))
+            (rx/of (ptk/event ::ev/event {::ev/name "create-version"})))
```
niwinz: I think we need to add the origin (`::ev/origin "plugins"`)
And probably rename this event to `create-version-from-plugins`

Alotor: OK

PR #5336, opened 2024-11-20
:sparkles: Minor improvements to wasm shape and intial draft for path encoding

niwinz:


frontend/src/app/render_wasm/shape.cljs
```diff
@@ -101,20 +118,57 @@
       :fills      (api/set-shape-fills v)
       :blend-mode (api/set-shape-blend-mode v)
       :shapes     (api/set-shape-children v)
+
```
superalex: Should we leave this comment here without a TODO or something similar?

niwinz: It is ok for now leave this, is a prototype and we will iterate on it

PR #5379, opened 2024-11-26
:books: Doc for viewer role

madalenapmelo-kp:
Add viewer role to the team roles; also made some tweaks to the descriptions of the other roles


docs/user-guide/teams/index.njk
madalenapmelo-kp, line 1: @myfunnyandy maybe we could add more detail to these descriptions of the role, and consider removing the table below it. Let me know what you think

myfunnyandy:
The table was redundant and not accessible so it was acting almost as decoration. I think it was a good decision to remove it.

The rest is clear while giving the essential info needed. Thanks!

PR #5381, opened 2024-11-27
:bug: Fix incorrect thumbnail rendering on dashboard

niwinz:


backend/src/app/rpc/commands/files.clj
```diff
@@ -698,11 +698,7 @@
 
 (defn get-team-recent-files
   [conn team-id]
-  (->> (db/exec! conn [sql:team-recent-files team-id])
-       (mapv (fn [row]
```
AzazelN28, line 702: Why is this mapv removed?

PR #5387, opened 2024-11-28
docs(create-a-plugin): Fix Vite script conflicts

Jemeni11:
> Please provide enough information so that others can review your pull request

This PR enhances the `Step 5. Build the plugin file` section by addressing potential issues with the dev script when using Vite. Specifically, it introduces solutions to handle conflicts caused by running `vite build --watch` and `vite preview` simultaneously.

When running `vite build --watch & vite preview`, no preview URL is provided, and accessing `localhost:port` in the browser fails. While removing `--watch` resolves the issue, it prevents automatic rebuilding when files change. This update offers solutions to handle both needs effectively.

- Added an explanation of the issue when `vite build --watch` and `vite preview` run together without coordination.
- Provided two solutions:
  - Sequential Commands: Ensuring the build completes before starting the preview server.
  - Concurrent Execution with Task Runner: Running both commands simultaneously using concurrently.

> Screenshots/GIFs

N/A

### DIFF
--- a/docs/plugins/create-a-plugin.md
+++ b/docs/plugins/create-a-plugin.md
@@ -176,6 +176,32 @@ And then add the following scripts to your <code class="language-js">package.jso
 }
 ```
 
+However, if you encounter issues with the dev script when accessing the preview URL, it's likely caused by <code class="language-js">vite build --watch</code> and <code class="language-js">vite preview</code> running simultaneously without coordination. Here are two solutions to address this:
+
+##### Sequential Commands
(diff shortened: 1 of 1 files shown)


belen-albeza:
Hi @Jemeni11 thanks for the pull request! You are right this is a bug in the docs, but we believe this issue is specific to Vite and thus want to avoid transforming our docs into a "Vite troubleshoot" manual. We'll fix the example (likely use the same module that is available in our ready-to-use vite template for plugins).

Closing this PR, but again, thanks for your contribution!

PR #5394, opened 2024-11-28
:sparkles: Add path to color bullet title

Epskampie:
So, i noticed that the html `title` for library colors in the color picker doesn't include the group name from the assets.

Assets:
image

Color picker title:
image

As you can see "blue" is missing from the title, rather confusing if you have a lot of color groups. 

This PR adds it:
image

Let me know if I you prefer something in this PR to be different!


niwinz:
:heart:  thanks

PR #5402, opened 2024-12-02
Create DS swatch utility component

Xaviju:
This PR creates the DS utility component for `swatch`
See [link to DS component](https://design.penpot.app/#/workspace/df60c394-e535-80b3-8004-3e0ef306d47a/502c5b43-61ea-81d3-8004-88ad2e9a68fa?page-id=9483f131-cf50-800b-8004-14be7966fd9d&layout=assets) for further documentation.

This PR includes the component and the storybook documentation and examples.

### DIFF
--- a/frontend/src/app/main/ui/ds/utilities/swatch.cljs
+++ b/frontend/src/app/main/ui/ds/utilities/swatch.cljs
@@ -1,0 +1,103 @@
+
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC
+
+(ns app.main.ui.ds.utilities.swatch
+  (:require-macros
+   [app.main.style :as stl])
+  (:require
+   [app.common.data.macros :as dm]
+   [cuerdas.core :as str]
+   [rumext.v2 :as mf]))
+
+(def ^:private schema:swatch
+  [:map
+   [:background :string]
+   [:class {:optional true} :string]
+   [:format {:optional true} [:enum "square" "rounded"]]
+   [:size {:optional true} [:enum "small" "medium"]]
+   [:active {:optional true} :boolean]
+   [:on-click {:optional true} fn?]])
+
+(def hex-regex #"^#(?:[0-9a-fA-F]{3}){1,2}$")
+(def rgb-regex #"^rgb\((\d{1,3}),\s*(\d{1,3}),\s*(\d{1,3})\)$")
+(def hsl-regex #"^hsl\((\d{1,3}),\s*(\d{1,3})%,\s*(\d{1,3})%\)$")
+(def hsla-regex #"^hsla\((\d{1,3}),\s*(\d{1,3})%,\s*(\d{1,3})%,\s*(0|1|0?\.\d+)\)$")
+(def rgba-regex #"^rgba\((\d{1,3}),\s*(\d{1,3}),\s*(\d{1,3}),\s*(0|1|0?\.\d+)\)$")
+
+(defn- gradient? [background]
+  (or
+   (str/starts-with? background "linear-gradient")
+   (str/starts-with? background "radial-gradient")))
+
+(defn- color-solid? [background]
+  (boolean
+   (or (re-matches hex-regex background)
+       (or (re-matches hsl-regex background)
+           (re-matches rgb-regex background)))))
+
+(defn- color-opacity? [background]
+  (boolean
+   (or (re-matches hsla-regex background)
+       (re-matches rgba-regex background))))
+
+(defn- extract-color-and-opacity [background]
+  (cond
+    (re-matches rgba-regex background)
+    (let [[_ r g b a] (re-matches rgba-regex background)]
+      {:color (dm/str "rgb(" r ", " g ", " b ")")
+       :opacity (js/parseFloat a)})
+
+    (re-matches hsla-regex background)
+    (let [[_ h s l a] (re-matches hsla-regex background)]
+      {:color (dm/str "hsl(" h ", " s "%, " l "%)")
+       :opacity (js/parseFloat a)})
+
+    :else
+    {:color background
+     :opacity 1.0}))
+
+(mf/defc swatch*
+  {::mf/props :obj
+   ::mf/schema schema:swatch}
+  [{:keys [background on-click format size active class]
+    :rest props}]
+  (let [element-type (if on-click "button" "div")
+        button-type (if on-click "button" nil)
+        format (or format "square")
+        size (or size "small")
+        active (or active false)
+        {:keys [color opacity]} (extract-color-and-opacity background)
+        class (dm/str class " " (stl/css-case
+                                 :swatch true
+                                 :small (= size "small")
+                                 :medium (= size "medium")
+                                 :square (= format "square")
+                                 :active (= active true)
+                                 :interactive (= element-type "button")
+                                 :rounded (= format "rounded")))
+        props (mf/spread-props props {:class class :on-click on-click :type button-type})]
+
+    [:> element-type props
+     (cond
+       (color-solid? background)
+       [:span {:class (stl/css :swatch-solid)
+               :style {:background background}}]
+
+       (color-opacity? background)
+       [:span {:class (stl/css :swatch-opacity)}
+        [:span {:class (stl/css :swatch-solid-side)
+                :style {:background color}}]
+        [:span {:class (stl/css :swatch-opacity-side)
+                :style {:background color :opacity opacity}}]]
+
+       (gradient? background)
+       [:span {:class (stl/css :swatch-gradient)
+               :style {:background-image (str background ", repeating-conic-gradient(lightgray 0% 25%, white 0% 50%)")}}]
+
+       :else
+       [:span {:class (stl/css :swatch-image)
+               :style {:background-image (str "url('" background "'), repeating-conic-gradient(lightgray 0% 25%, white 0% 50%)")}}])]))
--- a/frontend/src/app/main/ui/ds/utilities/swatch.scss
+++ b/frontend/src/app/main/ui/ds/utilities/swatch.scss
@@ -1,0 +1,104 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) KALEIDOS INC
+
+@use "../_borders.scss" as *;
+@use "../_sizes.scss" as *;
+@use "../colors.scss" as *;
+
+.swatch {
+  --border-color: var(--color-accent-primary-muted);
+  --border-radius: #{$br-4};
+  --border-color-active: var(--color-foreground-primary);
+  --border-color-active-inset: var(--color-background-primary);
+
+  --checkerboard-background: repeating-conic-gradient(lightgray 0% 25%, white 0% 50%);
+  --checkerboard-size: 0.5rem 0.5rem;
+
+  border: 1px solid var(--border-color);
+  border-radius: var(--border-radius);
+  overflow: hidden;
+
+  &:focus {
+    --border-color: var(--color-accent-primary);
+  }
+}
+
+.small {
+  inline-size: $sz-16;
+  block-size: $sz-16;
+}
+
+.medium {
+  --checkerboard-size: 1rem 1rem;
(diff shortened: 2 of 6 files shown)


frontend/src/app/main/ui/ds/utilities/swatch.cljs
```diff
@@ -0,0 +1,108 @@
...
+  (:require-macros
+   [app.main.style :as stl])
+  (:require
+   [cuerdas.core :as str]
+   [rumext.v2 :as mf]))
+
+(def ^:private schema:swatch
+  [:map
+   [:background :string]
+   [:ariaLabel {:optional true} :string]
```
EvaMarco: we usually use `[:aria-label ...]`

frontend/src/app/main/ui/ds/utilities/swatch.cljs
```diff
@@ -0,0 +1,108 @@
...
+    (str/starts-with? background "radial-gradient"))))
+
+(defn- is-color-opacity? [background]
+  (boolean
+   (or (re-matches hsla-regex background)
+       (re-matches rgba-regex background))))
+
+(defn- extract-color-and-opacity [background]
+  (cond
+
```
EvaMarco: Remove empty line

frontend/src/app/main/ui/ds/utilities/swatch.cljs
```diff
@@ -0,0 +1,108 @@
...
+    (re-matches rgba-regex background)
+    (let [[_ r g b a] (re-matches rgba-regex background)]
+      {:color (str "rgb(" r ", " g ", " b ")")
```
EvaMarco: Use dm/str for better performance

frontend/src/app/main/ui/ds/utilities/swatch.scss
```diff
@@ -0,0 +1,91 @@
...
+  --checkerboard-background: repeating-conic-gradient(lightgray 0% 25%, white 0% 50%);
+  --checkerboard-size: 0.5rem 0.5rem;
+
+  border: $b-1 solid var(--border-color);
+  border-radius: var(--border-radius);
+  overflow: hidden;
+
+  & > div {
+    block-size: 100%;
+  }
```
belen-albeza: Avoid this nested selector

frontend/src/app/main/ui/ds/utilities/swatch.scss
```diff
@@ -0,0 +1,91 @@
...
+  block-size: $sz-16;
+}
+
+.medium {
+  inline-size: $sz-24;
+  block-size: $sz-24;
+
+  .swatch-opacity {
+    --checkerboard-size: 1rem 1rem;
+  }
```
belen-albeza: Please do not nest CSS selectors, it creates unnecessary specificity. See [our UI Guidelines](https://help.penpot.app/technical-guide/developer/ui/#favor-lower-specificity).

frontend/src/app/main/ui/ds/utilities/swatch.cljs
```diff
@@ -0,0 +1,108 @@
...
+
+(mf/defc swatch*
+  {::mf/props :obj
+   ::mf/schema schema:swatch}
+  [{:keys [background on-activate format size ariaLabel active] :as props}]
+  (let [format (or format "square")
+        size (or size "small")
+        active (or active false)
+        {:keys [color opacity]} (extract-color-and-opacity background)
+        on-click (when on-activate (mf/use-fn (mf/deps on-activate)
```
belen-albeza: I think this `on-activate` is weird. This looks like `on-click` and `on-keydown` should actually be delegated properties of a `<button>` (like we're doing with the current buttons in the DS) and set `disabled` if the swatch cannot be interacted with.

frontend/src/app/main/ui/ds/utilities/swatch.cljs
Xaviju: Fixed 👍

frontend/src/app/main/ui/ds/utilities/swatch.cljs
Xaviju: Fixed 👍

frontend/src/app/main/ui/ds/utilities/swatch.cljs
Xaviju: Fixed 👍

frontend/src/app/main/ui/ds/utilities/swatch.scss
Xaviju: Fixed 👍

frontend/src/app/main/ui/ds/utilities/swatch.scss
Xaviju: Fixed 👍

frontend/src/app/main/ui/ds/utilities/swatch.cljs
Xaviju: Fix: this `on-activate` was a poor implementation for the requirement of a dynamic tag in the swatch component.

I've fixed it, setting the appropriate tag for the element depending on the component props. 
In case the author sets an on-click prop on the swatch component, it will become a button, otherwise will remain as presentational with a simple div element.

frontend/src/app/main/ui/ds/utilities/swatch.cljs
```diff
@@ -0,0 +1,101 @@
...
+   [:aria-label {:optional true} :string]
+   [:format {:optional true} [:enum "square" "rounded"]]
+   [:size {:optional true} [:enum "small" "medium"]]
+   [:active {:optional true} :boolean]
+   [:on-click {:optional true} fn?]])
+
+(def hsla-regex #"^hsla\((\d{1,3}),\s*(\d{1,3})%,\s*(\d{1,3})%,\s*(0|1|0?\.\d+)\)$")
+(def rgba-regex #"^rgba\((\d{1,3}),\s*(\d{1,3}),\s*(\d{1,3}),\s*(0|1|0?\.\d+)\)$")
+
+(defn- is-image? [background]
```
belen-albeza: Minor thing: For functions in clojure, we usually skip the `is-` prefix if we're using the `?` suffix at the end

frontend/src/app/main/ui/ds/utilities/swatch.cljs
```diff
@@ -0,0 +1,101 @@
...
+  (boolean
```
belen-albeza: Question: why do we need this `boolean` here?  `str/starts-with?` already returns one, right?

frontend/src/app/main/ui/ds/utilities/swatch.cljs
```diff
@@ -0,0 +1,101 @@
...
+                              :square (= format "square")
+                              :active (= active true)
+                              :interactive (= element-type "button")
+                              :rounded (= format "rounded"))
+                      :aria-label (when on-click aria-label)
+                      :on-click (when on-click on-click)}
+
+     (cond
+       (is-image? background)
+       [:div {:class (stl/css :swatch-image)
```
belen-albeza: shouldn't all these outer `<div>`s be `<button>` elements?

frontend/src/app/main/ui/ds/utilities/swatch.stories.jsx
```diff
@@ -0,0 +1,86 @@
...
+// Copyright (c) KALEIDOS INC
+
+import * as React from "react";
+import Components from "@target/components";
+import { action } from "@storybook/addon-actions";
+
+const { Swatch } = Components;
+
+export default {
+  title: "Utilities/Swatch",
```
belen-albeza: Let's put this Utilities under `Foundations` as well.

frontend/src/app/main/ui/ds/utilities/swatch.cljs
```diff
@@ -0,0 +1,101 @@
...
+  [{:keys [background on-click format size aria-label active] :as props}]
+  (let [element-type (if on-click "button" "div")
+        format (or format "square")
+        size (or size "small")
+        active (or active false)
+        {:keys [color opacity]} (extract-color-and-opacity background)
+        on-click (mf/use-fn (mf/deps on-click)
+                            (fn [event]
+                              (when (fn? on-click)
+                                (on-click event))))]
```
belen-albeza: we should just delegate the `on-click` like we do with the `button*` component:

```clojure
(mf/defc swatch*
  ;; ...
  [{:keys [background format size is-active class] :rest props}]
  (let [ ;; ...
        class (dm/str class " " (stl/css-case ...))
        props (mf/spread-props props {:class class})]
        ;; here `props` contains an `on-click` if the user provided one
        [:button props ]))
```

frontend/src/app/main/ui/ds/utilities/swatch.cljs
```diff
@@ -0,0 +1,101 @@
...
+       :opacity (js/parseFloat a)})
+
+    :else
+    {:color background
+     :opacity 1.0}))
+
+(mf/defc swatch*
+  {::mf/props :obj
+   ::mf/schema schema:swatch}
+  [{:keys [background on-click format size aria-label active] :as props}]
```
belen-albeza: careful: we should always allow to provide a `class` for components in our DS (so users can add styles to position it inside their container, pass CSS custom properties, etc)

belen-albeza (changes requested):
Thanks Xavi! Left some comments, but the TL;DR: let's use `<button>` for clickable things, and let's leverage delegated props (and thus allow for a custom `className` coming from the user, as well as `on-click`, `aria-label`, etc.)

frontend/src/app/main/ui/ds/utilities/swatch.cljs
Xaviju: Fixed

frontend/src/app/main/ui/ds/utilities/swatch.cljs
Xaviju: I didn't know, wanted to ensure that was a boolean instead of anything else

frontend/src/app/main/ui/ds/utilities/swatch.cljs
Xaviju: No, there is only a common button that wraps all these elements when the swatch is interactive. These are the specific implementation of the swatch background type.

frontend/src/app/main/ui/ds/utilities/swatch.stories.jsx
Xaviju: Done

frontend/src/app/main/ui/ds/utilities/swatch.cljs
Xaviju: I think I did it and looks more elegant :+1:

Xaviju:
Thanks for the review, could you please check it again? @belen-albeza

frontend/src/app/main/ui/ds/utilities/swatch.cljs
belen-albeza: But why have an external wrapper when we can do it here? I see two alternatives:

- Always use `<button>` and set `disable` when not interactive.
- Render either a `<button>` or a `<div>` depending if it's interactive or not.

Accessibility-wise, if something is interactive it should have the proper aria semantics. `<button>` give us those for free (and also sets activation via keyboard, etc. for free)

frontend/src/app/main/ui/ds/utilities/swatch.stories.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+export const LinearGradient = {
+  args: {
+    background: "linear-gradient(to right, transparent, mistyrose)",
+  },
+};
+
+export const Image = {
+  args: {
+    background: "http://www.libpng.org/pub/png/img_png/OwlAlpha-0.5.png",
```
belen-albeza: nit: can we use a local URL or a `data` one? So it works when we are offline (or if they take down that url)

Xaviju: Used a local image as example. 
In case it breaks, the value can be edited by the user in the Controls.

Xaviju:
@belen-albeza Should be ready to review again :face_with_head_bandage:

belen-albeza:
@Xaviju sorry for the back and forth! I found a small bug, but it's easily fixed and I can merge this when it's done. The swatch has a very thin gray border (see screenshot). It can be fixed by setting the `<button>` background to `none`.
image

Xaviju:
@belen-albeza fixed!

belen-albeza (approved):
Thanks @Xaviju ! Merging to unblock this. I spotted a few instances of hardcoded `rem` or `px` values, pls consider a follow up to use the tokens instead. Thanks!

PR #5404, opened 2024-12-03
♻️ Remove Rx and RY from shapes

EvaMarco:
This PR solves this PR -> https://tree.taiga.io/project/penpot/us/9404?milestone=418915
Shapes that allow `border radius` used to have `:rx` and `:ry` values, but only one (:rx) was used in the interface. 
If the "border radius" was changed to 4 entries, the shape was modified by adding `:r1, :r2, :r3, :r4` without removing `:rx` or `:ry`.
`:rx` and `:ry` are no longer needed.
To test:
- The `border radius` input will appear only with some shapes that accept this values (rect, boards, images, components).
- The initial value should be 0 in all corners
- The edge radius entry should always be collapsed when a shape is selected (only 1 entry).
      - If there is the same value in all four attributes, the entry should show the value.
      - If the values are different, the entry should show “Mixed”. 
- When only 1 entry is shown, the border radius will be applied to all 4 corners.
- When there are 4 entries, each entry must modify one corner. 
- When selecting more than one shape with different border-radius between them
           -  when 1 input ->  entry “Mixed” would appear
           - when 4 inputs  
                     -  “--” should appear when are different
                     -  If the value is common you must be able to see it.


common/src/app/common/files/shapes_helpers.cljc
```diff
@@ -164,10 +168,15 @@
   [changes frame-id parent-id objects index frame-name without-fill? target-cell-id]
 
   (let [base-parent (get objects parent-id)
+        _ (prn "prepare create empty")
```
EvaMarco: todo delete this

frontend/src/app/main/ui/workspace/tokens/changes.cljs
```diff
@@ -112,12 +112,10 @@
...
                       (fn [shape]
-                        (when (ctsr/has-radius? shape)
-                          (cond-> shape
-                            (:rx shape) (ctsr/switch-to-radius-4)
-                            :always (ctsr/set-radius-4 (first attributes) value))))
+                        (when (ctsr/can-get-border-radius? shape)
+                          (ctsr/set-radius-to-single-corner shape (first attributes) value)))
                       {:reg-objects? true
-                       :attrs [:rx :ry :r1 :r2 :r3 :r4]}))
+                       :attrs ctt/border-radius-keys}))
```
niwinz, line 118: :heart:

PR #5410, opened 2024-12-03
:tada: Support for radial gradients with wasm render

superalex:
https://tree.taiga.io/project/penpot/task/9267

### DIFF
--- a/render-wasm/src/shapes/fills.rs
+++ b/render-wasm/src/shapes/fills.rs
@@ -28,6 +28,7 @@ pub struct Gradient {
     opacity: f32,
     start: (f32, f32),
     end: (f32, f32),
+    width: f32,
 }
 
 impl Gradient {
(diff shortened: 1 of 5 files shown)


render-wasm/src/shapes/fills.rs
```diff
@@ -80,17 +116,39 @@ impl ImageFill {
...
     Image(ImageFill),
 }
 
 impl Fill {
-    pub fn new_linear_gradient(start: (f32, f32), end: (f32, f32), opacity: f32) -> Self {
+    pub fn new_linear_gradient(
+        start: (f32, f32),
+        end: (f32, f32),
+        opacity: f32,
+        width: f32,
```
belen-albeza: If we're not using `width` in the linear gradient, let's remove it from the params list and set it to zero in the new struct.

superalex: Thank you very much! fixed! :)

PR #5412, opened 2024-12-03
:sparkles: Add test AB renaming "Libraries" to "Add library"

pabloalba:
Implements https://tree.taiga.io/project/penpot/us/9341

### DIFF
--- a/frontend/translations/en.po
+++ b/frontend/translations/en.po
@@ -4058,6 +4058,9 @@ msgstr "Group name"
 msgid "workspace.assets.libraries"
 msgstr "Libraries"
 
+msgid "workspace.assets.add-library"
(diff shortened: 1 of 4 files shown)


frontend/translations/en.po
```diff
@@ -4058,6 +4058,9 @@ msgstr "Group name"
 msgid "workspace.assets.libraries"
 msgstr "Libraries"
 
+msgid "workspace.assets.libraries.add"
```
niwinz: I think the key should be something `workspace.assets.add-library` .
Translation keys should have a maximum of 4 segments but preferable 3.
This is a minor issue, of course.

pabloalba: Fixed

PR #5418, opened 2024-12-04
:sparkles: Add test AB for adding a few "Suggested" libraries

pabloalba:
Implements https://tree.taiga.io/project/penpot/us/9339?milestone=419643

### DIFF
--- a/frontend/src/app/main/ui/workspace/libraries.cljs
+++ b/frontend/src/app/main/ui/workspace/libraries.cljs
@@ -87,8 +87,10 @@
                  (conj (tr "workspace.libraries.typography" typography-count))))
      "\u00A0")))
 
-(mf/defc describe-library-blocks
-  [{:keys [components-count graphics-count colors-count typography-count] :as props}]
+(mf/defc describe-library-blocks*
+  {::mf/props :obj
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/ui/workspace/libraries.cljs
```diff
@@ -104,9 +107,45 @@
       (tr "workspace.libraries.typography" typography-count)])])
 
 
+(mf/defc sample-library-entry
+  [{:keys [library project-id team-id importing] :as props}]
+  (let [id         (:id library)
+        importing? (deref importing)
```
niwinz, line 117: All new react components should follow the new style (name ends in `*` and it should have `::mf/props` to `:obj` and referenced/used with `[:>`)

niwinz: I'm also thinking that passing the whole state `importing` is the way to implement this. I mean, in general is not a good practice, but there are always exceptions.

niwinz: On the end, I think it is ok for this case, so no change is required for my previous comment

PR #5419, opened 2024-12-04
:sparkles: Add reorder handler to fills, strokes and shadows

Alotor:
https://tree.taiga.io/project/penpot/us/9216?milestone=419643

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/options/menus/shadow.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/menus/shadow.cljs
@@ -17,6 +17,7 @@
    [app.main.data.workspace.undo :as dwu]
    [app.main.store :as st]
    [app.main.ui.components.numeric-input :refer [numeric-input*]]
+   [app.main.ui.components.reorder-handler :refer [reorder-handler]]
    [app.main.ui.components.select :refer [select]]
    [app.main.ui.components.title-bar :refer [title-bar]]
    [app.main.ui.ds.buttons.icon-button :refer [icon-button*]]
@@ -152,8 +153,10 @@
     [:div {:class (stl/css-case :global/shadow-option true
                                 :shadow-element true
                                 :dnd-over-top (= (:over dprops) :top)
-                                :dnd-over-bot (= (:over dprops) :bot))
-           :ref dref}
+                                :dnd-over-bot (= (:over dprops) :bot))}
+     (when (some? on-reorder)
+       [:& reorder-handler {:ref dref}])
+
      [:*
       [:div {:class (stl/css :basic-options)}
        [:div {:class (stl/css-case :shadow-info true
--- a/frontend/src/app/main/ui/workspace/sidebar/options/menus/shadow.scss
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/menus/shadow.scss
@@ -5,6 +5,7 @@
 // Copyright (c) KALEIDOS INC
 
 @import "refactor/common-refactor.scss";
+@import "../../../../ds/_sizes.scss";
 
 .element-set {
   margin: 0;
@@ -34,6 +35,21 @@
 
 .shadow-element {
   @include flexColumn;
+  position: relative;
+
+  --reorder-left-position: calc(-1 * var(--sp-m) - var(--sp-xxs));
+
+  &:hover {
+    --reorder-icon-visibility: visible;
+  }
+
+  &.dnd-over-top {
+    --reorder-top-display: block;
+  }
+
+  &.dnd-over-bot {
+    --reorder-bottom-display: block;
+  }
 }
 
 .basic-options {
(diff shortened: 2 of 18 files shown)


frontend/src/app/main/ui/workspace/sidebar/options/menus/shadow.scss
```diff
@@ -34,6 +35,20 @@
...
+    visibility: visible;
+  }
+
+  &.dnd-over-top .reorder-separator-top {
+    display: block;
+  }
+
+  &.dnd-over-bot .reorder-separator-bottom {
+    display: block;
+  }
```
belen-albeza: pls avoid CSS selectors with high specificity 🙏 Details in our [UI Guidelines](https://help.penpot.app/technical-guide/developer/ui/#favor-lower-specificity)

Alotor: But I need it to get the ":hover" (and the class applied to the parent). I can move the selector to the other one but the "compiled" specify is the same.

Alotor: I'll change it so it passes a property instead of using nested selectors.

frontend/resources/styles/common/refactor/reorder.scss
```diff
@@ -0,0 +1,36 @@
+.reorder {
+  cursor: grab;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  left: var(--reorder-left-position, calc(-1 * $s-16));
```
belen-albeza: All stylesheets in `common/refactor` are deprecated, let's not add more CSS there.

We need to avoid using global classes for styles and use CSS modules. Since this CSS is meant to be shared among multiple components, it should be a mixin (see examples in `typography.scss`) or a placeholder selector (see `_buttons.scss`, for instance), and then import them from the relevant components.

frontend/resources/styles/common/refactor/reorder.scss
```diff
@@ -0,0 +1,36 @@
...
+  z-index: $z-index-2;
+  position: absolute;
+  top: calc(-1 * $s-6);
```
belen-albeza: Please use the tokens defined in the design system (for instance in `spacing.scss`) and use the semantic ones (i.e. `--sp-l` when available).

frontend/src/app/main/ui/workspace/sidebar/options/menus/shadow.cljs
```diff
@@ -152,8 +153,17 @@
     [:div {:class (stl/css-case :global/shadow-option true
                                 :shadow-element true
                                 :dnd-over-top (= (:over dprops) :top)
-                                :dnd-over-bot (= (:over dprops) :bot))
-           :ref dref}
+                                :dnd-over-bot (= (:over dprops) :bot))}
+     (when (some? on-reorder)
+       [:*
+        [:div {:ref dref :class (stl/css :reorder)}
```
belen-albeza: Q: why the icon needs to be wrapped in a `<div>` ?

frontend/src/app/main/ui/workspace/sidebar/options/rows/stroke_row.cljs
```diff
@@ -140,8 +141,18 @@
...
+
+     (when (some? on-reorder)
+       [:*
+        [:div {:ref dref :class (stl/css :reorder)}
+         [:> icon*
+          {:id ic/reorder
+           :class (stl/css :reorder-icon)
+           :aria-hidden true}]]
+        [:hr {:class (stl/css :reorder-separator-top)}]
+        [:hr {:class (stl/css :reorder-separator-bottom)}]])
```
belen-albeza: I see this being repeated all over, maybe it should be a component itself?

belen-albeza (changes requested):
Thanks @Alotor ! Added some comments, the biggest thing to do is to move the CSS out of the `refactor` (they are global styles) and use CSS modules and the design system tokens.

frontend/src/app/main/ui/workspace/sidebar/options/rows/stroke_row.cljs
Alotor: I think so. This will also solve the issue with the "refactor" because then the styles will be owned by the component

frontend/src/app/main/ui/workspace/sidebar/options/menus/shadow.cljs
Alotor: Because the clickable area is bigger than the icon. This wrapper occupies the 100% of the height and then the icon is positioned in the center of the area.

belen-albeza (changes requested):
Thanks Alonso! Plesase, let's make sure we use the new spacing tokens (i.e. the ones in `ds/spacing.scss`:

- `$s-12` -> `--sp-m`
- `$s-16` -> `--sp-l`
- `$s-6` -> we don't have such a token, either choose `--sp-s` or `--sp-xs`.

belen-albeza (approved):

PR #5420, opened 2024-12-04
🐛  Fix admin can demote owner

EvaMarco:
This PR fixes this https://tree.taiga.io/project/penpot/issue/9443


superalex:
Shouldn't be this against staging?

PR #5429, opened 2024-12-05
:bug: Palba bugfixing 01

pabloalba:
Fixes:

https://tree.taiga.io/project/penpot/issue/9435
https://tree.taiga.io/project/penpot/issue/9350

### DIFF
--- a/frontend/src/app/main/ui/components/context_menu_a11y.scss
+++ b/frontend/src/app/main/ui/components/context_menu_a11y.scss
@@ -36,7 +36,7 @@
     background-color: var(--menu-background-color);
     overflow: auto;
     & .separator {
-      height: $s-12;
+      border-top: solid $s-1 var(--color-background-quaternary);
     }
 
     &.min-width {
--- a/frontend/src/app/main/ui/components/context_menu_a11y.scss
+++ b/frontend/src/app/main/ui/components/context_menu_a11y.scss
@@ -18,6 +18,7 @@
     opacity: $op-10;
     visibility: visible;
   }
+
   &.fixed {
     position: fixed;
   }
@@ -35,6 +36,7 @@
     border: $s-2 solid var(--panel-border-color);
     background-color: var(--menu-background-color);
     overflow: auto;
+
     & .separator {
       border-top: solid $s-1 var(--color-background-quaternary);
     }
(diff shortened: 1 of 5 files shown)


frontend/src/app/main/ui/components/context_menu_a11y.scss
```diff
@@ -35,8 +36,12 @@
     border: $s-2 solid var(--panel-border-color);
     background-color: var(--menu-background-color);
     overflow: auto;
+
     & .separator {
-      height: $s-12;
+      border-top: solid 1px var(--color-background-quaternary);
```
superalex: We don't use "px" units, I think this should be a "$s-1", you can check all the available values in frontend/resources/styles/common/refactor/spacing.scss

frontend/src/app/main/ui/components/context_menu_a11y.scss
```diff
@@ -35,8 +36,12 @@
...
+      border-right: 0;
```
superalex: Are those 0 values for right, left and bottom really needed?

frontend/src/app/main/ui/components/context_menu_a11y.scss
```diff
@@ -45,6 +50,7 @@
 
     .context-menu-item {
       display: flex;
+
```
superalex, line 50: Why adding those extra lines in the commit?

frontend/src/app/main/ui/dashboard/sidebar.scss
```diff
@@ -370,6 +378,10 @@
 
 .profile-separator {
   height: $s-6;
+  border-top: solid 1px var(--color-background-quaternary);
```
superalex: Same here

belen-albeza:
Hi, it looks like the history panel bug could benefit from a Playwright test for future regressions? I also saw this bug in Taiga https://tree.taiga.io/project/penpot/issue/9492 , which I don't know if it's also fixed by this (in this case, we could add that interaction to the test too)

pabloalba:
> Hi, it looks like the history panel bug could benefit from a Playwright test for future regressions?

ok


> I also saw this bug in Taiga https://tree.taiga.io/project/penpot/issue/9492 , which I don't know if it's also fixed by this (in this case, we could add that interaction to the test too)

No, that is a different issue

PR #5433, opened 2024-12-09
:tada: Render wasm ellipses support

superalex:


render-wasm/src/main.rs
```diff
@@ -96,10 +97,17 @@ pub extern "C" fn set_view_xy(x: f32, y: f32) {
...
     let id = uuid_from_u32_quartet(a, b, c, d);
-    state.use_shape(id);
+    state.use_shape(id, Kind::Rect);
+}
+
+#[no_mangle]
+pub extern "C" fn use_shape_circle(a: u32, b: u32, c: u32, d: u32) {
+    let state = unsafe { STATE.as_mut() }.expect("got an invalid state pointer");
+    let id = uuid_from_u32_quartet(a, b, c, d);
+    state.use_shape(id, Kind::Circle);
```
belen-albeza: Question: I guess we need to do this just in case we need to create the shape within Rust. But once we're already storing the right kind when creating the shape, why do we need to pass the kind around to every shape-related function?

PR #5445, opened 2024-12-10
:books: User guide import/export section update

myfunnyandy:
https://tree.taiga.io/project/penpot/task/9375


myfunnyandy:
@madalenapmelo-kp I've moved the section about the formats to the bottom. This info was promoted to the first position of the section because we wanted to anticipate questions about the differences between the .penpot and the .zip. From now on we only provide one format, reducing the need for additional explanations, so I decided to focus the section on how to import/export. Let me know what you think :)

madalenapmelo-kp (approved):

PR #5455, opened 2024-12-11
:books: File history versions

madalenapmelo-kp:
https://tree.taiga.io/project/penpot/task/9377


myfunnyandy:
@madalenapmelo-kp I've added the images and also made a couple of changes to the text:
* Ensure headings do not go beyond h4. This is to be consistent with the rest of the docs and the styles are not prepared for h5 and h6. 
* I've added a subsection about restoring versions, because this is not specific of the version type (manually, auto, pinned).

If you're ok with these changes I think we're ready to merge :)

myfunnyandy (approved):
@madalenapmelo-kp looking great already, merging!

Issue #5478, opened 2024-12-14
bug: nginx use wrong hostname for ssl handshake


backrunner:
### Steps To Reproduce

1. Access an asset via /assets/by-id/xxxxxxxxxxx with S3 assets mode
2. Get 502 error, and can see an output with "SSL_do_handshake() failed (SSL: error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure:SSL alert number 40) while SSL handshaking to upstream" in the frontend container's output
3. The asset itself existed and can be access via right bucket domain

### Expected behavior

No SSL errors show up

### Actual behavior

Cannot access the assets via nginx proxy

### Desktop (please complete the following information)

Windows, Latest Chrome

### Environment (please complete the following information)

Self-hosted in docker with default nginx config and using S3 (Cloudflare R2) as assets storage backend

### Additional context

I tried to add "proxy_ssl_server_name on;" to @handle_redirect, the issue is gone. I'm not sure that if this modification will affect other S3 compatible storage, if that's fine, I can create a PR to add it to the default nginx config.

BohdanK-W32:
+1, same issue

BohdanK-W32:
Checked this one today, I agree with @backrunner, that one seems to fix the issue.

Here's the temporary workaround:

Dockerfile content for frontend:
``` Dockerfile
FROM penpotapp/frontend:latest

RUN sed -i -e 's/location @handle_redirect {/location @handle_redirect {\nproxy_ssl_server_name on;/g' /etc/nginx/nginx.conf.template

CMD ["nginx", "-g", "daemon off;"]
```

So you can just create `Dockerfile.frontend` file or something like that and replace `image` to `build` in the `docker-compose.yml`'s `penpot-frontend` service

```yaml
  penpot-frontend:
    build:
      context: .
      dockerfile: Dockerfile.frontend
```

Hope it helps

niwinz:
The fix for this issue will be released with 2.4 docker images

PR #5480, opened 2024-12-15
🐛 Fix assets proxy SSL handshake error

backrunner:

Fix issue: https://github.com/penpot/penpot/issues/5478

When using S3 assets storage backend with Cloudflare R2, Nginx will send wrong server name for SSL handshake, it will cause all the asset proxy requests return 502.

To fix this issue, we can simply add `proxy_ssl_server_name on;` to the Nginx config.

### DIFF
--- a/docker/devenv/files/nginx.conf
+++ b/docker/devenv/files/nginx.conf
@@ -90,6 +90,7 @@ http {
             proxy_hide_header x-amz-meta-server-side-encryption;
             proxy_hide_header x-amz-server-side-encryption;
             proxy_pass $redirect_uri;
+            proxy_ssl_server_name on;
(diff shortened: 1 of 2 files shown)


BohdanK-W32:
Hi, @backrunner!, here's one more file to update: https://github.com/penpot/penpot/blob/main/docker/images/files/nginx.conf

backrunner:
> Hi, @backrunner!, here's one more file to update: https://github.com/penpot/penpot/blob/main/docker/images/files/nginx.conf

My bad, I missed that file 😂, already modified it

niwinz:
Merged using CLI.
Thanks :heart:

PR #5518, opened 2025-01-07
:bug: Fix wasm render strokes bugs

superalex:
- In open paths: inner, center and outer strokes should be rendered as center: https://tree.taiga.io/project/penpot/task/9671
- In dotted paths the width was not correctly calculated (it should be doubled for inner and outeR): https://tree.taiga.io/project/penpot/task/9697

### DIFF
--- a/render-wasm/src/shapes/renderable.rs
+++ b/render-wasm/src/shapes/renderable.rs
@@ -131,7 +131,7 @@ fn render_stroke(
             Kind::Rect(rect) => draw_stroke_on_rect(surface.canvas(), stroke, rect, &selrect),
             Kind::Circle(rect) => draw_stroke_on_circle(surface.canvas(), stroke, rect, &selrect),
             Kind::Path(path) => {
-                draw_stroke_on_path(surface.canvas(), stroke, path, &selrect, path_transform)
+                draw_stroke_on_path(surface.canvas(), stroke, path, &selrect, path_transform);
             }
         }
     }
@@ -165,9 +165,10 @@ fn draw_stroke_on_path(
     let mut skia_path = path.to_skia_path();
     skia_path.transform(path_transform.unwrap());
 
-    let paint_stroke = stroke.to_stroked_paint(selrect);
+    let kind = stroke.render_kind(path.is_open());
+    let paint_stroke = stroke.to_stroked_paint(kind.clone(), selrect);
     // Draw the different kind of strokes for a path requires different strategies:
-    match stroke.kind {
+    match kind {
         // For inner stroke we draw a center stroke (with double width) and clip to the original path (that way the extra outer stroke is removed)
         StrokeKind::InnerStroke => {
             canvas.clip_path(&skia_path, skia::ClipOp::Intersect, true);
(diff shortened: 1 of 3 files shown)


render-wasm/src/shapes/renderable.rs
```diff
@@ -317,13 +319,15 @@ pub fn draw_image_stroke_in_container(
...
     // Clear outer stroke for paths if necessary. When adding an outer stroke we need to empty the stroke added too in the inner area.
-    if let (Kind::Path(p), StrokeKind::OuterStroke) = (kind, &stroke.kind) {
-        let mut path = p.to_skia_path();
-        path.transform(path_transform.unwrap());
-        let mut clear_paint = skia::Paint::default();
-        clear_paint.set_blend_mode(skia::BlendMode::Clear);
-        clear_paint.set_anti_alias(true);
-        canvas.draw_path(&path, &clear_paint);
+    if let Kind::Path(p) = kind {
+        if stroke.kind(p.is_open()) == StrokeKind::OuterStroke {
```
belen-albeza: Not sure about this condition. Is there any other possibilities that `stroke.kind()` returns something different than `StrokeKind::CenterStroke` in the future?

If not, I would keep the original `let` guard as it was (`(Kind::Path(p), StrokeKind::OuterStroke)`) , and just check for `!p.is_open` here.

render-wasm/src/shapes/strokes.rs
```diff
@@ -46,10 +46,18 @@ pub struct Stroke {
...
+    kind: StrokeKind,
 }
 
 impl Stroke {
+    pub fn kind(&self, is_open: bool) -> StrokeKind {
+        if is_open {
+            StrokeKind::CenterStroke
+        } else {
+            self.kind.clone()
+        }
```
belen-albeza: As commented above, we might not need to check for `is_open` here

render-wasm/src/shapes/strokes.rs
```diff
@@ -46,10 +46,18 @@ pub struct Stroke {
...
-    pub kind: StrokeKind,
+    kind: StrokeKind,
 }
 
 impl Stroke {
+    pub fn kind(&self, is_open: bool) -> StrokeKind {
+        if is_open {
+            StrokeKind::CenterStroke
+        } else {
+            self.kind.clone()
```
belen-albeza: If we make `StrokeKind` derive `Copy` as well, we don't need to do a `.clone` here.

render-wasm/src/shapes/renderable.rs
superalex: Sorry but I don't get it..could you pase some example code?

stroke.kind() can return something different than StrokeKind::CenterStroke right now, it depends on the internal kind attribute and if the path is open or not. I don't know how to keep the original let comparing StrokeKind::OuterStroke with stroke.kind(p.is_open())

render-wasm/src/shapes/strokes.rs
superalex: I think we still need it, Inner and Outer strokes behave as Center ones if they are open, may be it would be more clear renaming this method to something like drawing_kind ?

render-wasm/src/shapes/strokes.rs
superalex: Thanks! :)

render-wasm/src/shapes/renderable.rs
superalex: I renamed the function to render_kind and added some comments <3

render-wasm/src/shapes/strokes.rs
superalex: I renamed the function to render_kind and added some comments :)

PR #5522, opened 2025-01-08
:bug: Fix race condition when rendering the background

belen-albeza:
We have a Wasm panic when loading a penpot file and trying to render the background before we have initialized the shape there.

Screenshot 2025-01-08 at 11 38 45 AM

(We are trying to fetch the root shape from the tree and do an `unwrap`, but the tree is empty)

### DIFF
--- a/frontend/src/app/main/ui/workspace/viewport_wasm.cljs
+++ b/frontend/src/app/main/ui/workspace/viewport_wasm.cljs
@@ -284,8 +284,7 @@
              (p/fmap (fn [ready?]
                        (when ready?
                          (reset! canvas-init? true)
-                         (wasm.api/assign-canvas canvas)
-                         (wasm.api/set-canvas-background background)))))
+                         (wasm.api/assign-canvas canvas)))))
         (fn []
           (wasm.api/clear-canvas))))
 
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/ui/workspace/viewport_wasm.cljs
```diff
@@ -284,24 +284,23 @@
              (p/fmap (fn [ready?]
                        (when ready?
                          (reset! canvas-init? true)
-                         (wasm.api/assign-canvas canvas)
-                         (wasm.api/set-canvas-background background)))))
+                         (wasm.api/assign-canvas canvas)))))
         (fn []
           (wasm.api/clear-canvas))))
 
     (mf/with-effect [vport]
```
AzazelN28, line 291: this should also use @canvas-init? as a dependency, right?

belen-albeza: I don't think so, this is run when `canvas-init?` is false (in pure React I guess we would set as a dependency the `setCanvasInit`, but here we're using `reset!` and we don't want this effect to be re-run when `@canvas-init?` changes.

PR #5535, opened 2025-01-09
Token studio develop

hirunatan:


frontend/src/app/main/data/workspace/colors.cljs
```diff
@@ -82,7 +82,7 @@
           (assoc-in [:workspace-global :picked-shift?] shift?)))))
 
 (defn transform-fill
-  [state ids color transform]
+  [state ids color transform & options]
```
niwinz: I'm not clearly understand this change, I gues it should be `& {:as options}`, but as it is used around the changed code it always implies using `apply` that not looks good.

frontend/src/app/main/refs.cljs
```diff
@@ -468,6 +468,14 @@
 (def workspace-active-theme-paths
   (l/derived (d/nilf ctob/get-active-theme-paths) tokens-lib))
 
+(defn token-sets-at-path-all-active?
```
niwinz: This is not a predicate, it returns a derived atom, so it should not be ending with `?`, because it confusing

frontend/src/app/main/ui/ds/buttons/button.cljs
```diff
@@ -17,20 +17,24 @@
...
     [:and :string [:fn #(contains? icon-list %)]]]
+   [:on-ref {:optional true} fn?]
    [:variant {:optional true}
     [:maybe [:enum "primary" "secondary" "ghost" "destructive"]]]])
 
 (mf/defc button*
   {::mf/props :obj
    ::mf/schema schema:button}
-  [{:keys [variant icon children class] :rest props}]
+  [{:keys [variant icon children class on-ref] :rest props}]
```
niwinz, line 27: Why this called `on-ref`, this should be called `ref` and it should be using `forward-ref` mechanism

common/src/app/common/logic/shapes.cljc
```diff
@@ -14,8 +14,33 @@
    [app.common.types.container :as ctn]
    [app.common.types.shape.interactions :as ctsi]
    [app.common.types.shape.layout :as ctl]
+   [app.common.types.token :as cto]
    [app.common.uuid :as uuid]))
 
+(defn- check-unapply-tokens
```
niwinz: -> unaply-tokens & DOCSTRING

hirunatan: done

frontend/src/app/main/refs.cljs
hirunatan: done

frontend/src/app/main/ui/ds/buttons/button.cljs
hirunatan: Eva says that this is a dynamic ref that cannot be passed there with a forward-ref, so they instead pass a function that sets the reference on the node after creation. Same construct as in `app.main.ui.ds.controls.shared.options-dropdown`.

frontend/src/app/main/data/workspace/colors.cljs
hirunatan: Changed. Used arities option because some involved functions are in the critical path, I think. And also the `update-shapes` uses it.

frontend/src/app/main/ui/ds/buttons/button.cljs
niwinz: This does not make sense, as it is a minor issue, i will pass it, but I'm sure that this should be solved with forward-ref

PR #5569, opened 2025-01-13
Include Design Tokens in .penpot exported files

hirunatan:


CosmicJesterX (commented):
who does not want an token library? =)

PR #5572, opened 2025-01-13
Token Sets Drag & Drop

### LINKED ISSUE #16: `Set groups UI` Drag & Drop
Similar to D&D in layers & groups

https://github.com/user-attachments/assets/3300b5e5-5acf-4477-bb83-7c1389857681

Restore drag & drop UI for set and set groups.

When dragging a set/group we should move it to the dropped location.

We can use the rename events to finalize dropping of items, although we probably need some kind of ordering step? https://github.com/tokens-studio/tokens-studio-for-penpot/blob/71f656cc5ee7e7b374b39f4606de94919a290aa9/frontend/src/app/main/data/tokens.cljs#L136-L144

For sets we already had an implementation for D&D using the provided `use-sortable` from the penpot lib

https://github.com/tokens-studio/tokens-studio-for-penpot/blob/9fe4919a2b4ab5b18b021a77b9a7ca79920ce7d3/frontend/src/app/main/ui/workspace/tokens/sets.cljs#L118

### LINKED ISSUE #50: bug: Activating a group set in a theme throws an internal error
### Steps To Reproduce

In an empty file:
1. Create a group of sets with two sets in it
2. Create a theme
3. Go to edit theme sets and click on the group checkox in order to activate the group
4. The app crashes with an error (report attached)
[report (7).txt](https://github.com/user-attachments/files/18420943/report.7.txt)

### Expected behavior

The group is activated

### Actual behavior

The app crashes

### Screenshots or video

image

_No response_

_No response_

_No response_

### Frontend Stack Trace

_No response_

### Backend Stack Trace

_No response_

### Additional context

_No response_

floscr:
This PR adds drag & drop for sets and set groups.

To not further introduce any implementation leaks of set/set-group prefixes with the drag & drop implementation and keeping the release date in mind, I've changed a couple of other things in the selection logic.

Now there are divided functions for sets and set groups, so the consumer side doesn't know about prefixed paths.

## Drag & Drop UX

Drag & Drop should behave UX wise simlar to D&D of layers and layer groups.

Two key differences from the Layers D&D are:
- Set groups are not visible the UI if they don't have sets inside
- Sets inside set groups must have a unique name. 
  So no two sets/set groups should be named equally at their level in the sets tree.
  But sets & set groups may have the same name at the same level.
  
### Drag & Drop

https://github.com/user-attachments/assets/04741d69-4786-4b99-94b1-a3a3a4b9c57f

### Errors

https://github.com/user-attachments/assets/272a9ae4-0545-4bf8-bc8e-8bd01ebee531

### DIFF
--- a/common/src/app/common/logic/tokens.cljc
+++ b/common/src/app/common/logic/tokens.cljc
@@ -29,14 +29,103 @@
   [changes tokens-lib set-name]
   (generate-update-active-sets changes tokens-lib #(ctob/toggle-set % set-name)))
 
-(defn generate-toggle-token-set-group
-  "Toggle a token set group at `prefixed-set-path-str` in `tokens-lib` without modifying a user theme."
-  [changes tokens-lib prefixed-set-path-str]
-  (let [deactivate? (contains? #{:all :partial} (ctob/sets-at-path-all-active? tokens-lib prefixed-set-path-str))
-        sets-names  (->> (ctob/get-sets-at-prefix-path tokens-lib prefixed-set-path-str)
+(defn toggle-token-set-group
+  "Toggle a token set group at `group-path` in `tokens-lib` for a `tokens-lib-theme`."
+  [group-path tokens-lib tokens-lib-theme]
+  (let [deactivate? (contains? #{:all :partial} (ctob/sets-at-path-all-active? tokens-lib group-path))
+        sets-names  (->> (ctob/get-sets-at-path tokens-lib group-path)
                          (map :name)
-                         (into #{}))
-        update-fn   (if deactivate?
-                      #(ctob/disable-sets % sets-names)
-                      #(ctob/enable-sets % sets-names))]
-    (generate-update-active-sets changes tokens-lib update-fn)))
+                         (into #{}))]
+    (if deactivate?
+      (ctob/disable-sets tokens-lib-theme sets-names)
+      (ctob/enable-sets tokens-lib-theme sets-names))))
+
+(defn generate-toggle-token-set-group
+  "Toggle a token set group at `group-path` in `tokens-lib` without modifying a user theme."
+  [changes tokens-lib group-path]
+  (generate-update-active-sets changes tokens-lib #(toggle-token-set-group group-path tokens-lib %)))
+
+(defn vec-starts-with? [v1 v2]
+  (= (subvec v1 0 (min (count v1) (count v2))) v2))
+
+(defn calculate-move-token-set-or-set-group
+  [tokens-lib {:keys [from-index to-index position collapsed-paths]
+               :or {collapsed-paths #{}}}]
+  (let [tree (-> (ctob/get-set-tree tokens-lib)
+                 (ctob/walk-sets-tree-seq :walk-children? #(contains? collapsed-paths %)))
+        from (nth tree from-index)
+        to (nth tree to-index)
+        before (case position
+                 :top to
+                 :bot (nth tree (inc to-index) nil)
+                 :center nil)
+        prev-before (if (:group? from)
+                      (->> (drop (inc from-index) tree)
+                           (filter (fn [element]
+                                     (<= (:depth element) (:depth from))))
+                           (first))
+                      (nth tree (inc from-index) nil))
+
+        drop-as-direct-group-child? (or
+                                     (= :center position)
+                                     (and
+                                      (= :bot position)
+                                      (:group? to)
+                                      (not (get collapsed-paths (:path to)))))
+        from-path (:path from)
+        to-parent-path (if drop-as-direct-group-child?
+                         (:path to)
+                         (into [] (butlast (:path to))))
+        to-path (conj to-parent-path (last from-path))
+
+        identical? (or (= from-index to-index)
+                       (and (= from-path to-path)
+                            (case position
+                              :top (= from-index (dec to-index))
+                              :bot (= from-index to-index)
+                              nil)))
+        to-exists? (and
+                    (not= (:parent-path from) to-parent-path)
+                    (if (:group? from)
+                      (ctob/set-group-path-exists? tokens-lib to-path)
+                      (ctob/set-path-exists? tokens-lib to-path)))
+        parent-to-child-drop? (and
+                               (not= (:parent-path from) to-parent-path)
+                               (:group? from)
+                               (vec-starts-with? to-path (:path from)))]
+    (cond
+      identical? nil
+      to-exists?
+      (throw (ex-info "move token set error: path exists"
+                      {:error :path-exists
+                       :path to-path}))
+      parent-to-child-drop?
+      (throw (ex-info "move token set error: parent-to-child"
+                      {:error :parent-to-child
+                       :from-path from-path
+                       :to-path to-path}))
+      :else
+      (cond-> {:from-path from-path
+               :to-path to-path
+               :before-path nil
+               :before-group? nil}
+        before (assoc :before-path (:path before)
+                      :before-group? (:group? before))
+        prev-before (assoc :prev-before-path (:path prev-before)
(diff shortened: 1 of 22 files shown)


common/src/app/common/logic/tokens.cljc
```diff
@@ -29,14 +29,91 @@
   [changes tokens-lib set-name]
   (generate-update-active-sets changes tokens-lib #(ctob/toggle-set % set-name)))
 
-(defn generate-toggle-token-set-group
-  "Toggle a token set group at `prefixed-set-path-str` in `tokens-lib` without modifying a user theme."
-  [changes tokens-lib prefixed-set-path-str]
-  (let [deactivate? (contains? #{:all :partial} (ctob/sets-at-path-all-active? tokens-lib prefixed-set-path-str))
-        sets-names  (->> (ctob/get-sets-at-prefix-path tokens-lib prefixed-set-path-str)
+(defn toggle-token-set-group
```
hirunatan, line 32: I think this can be now a method in `tokens-lib` module

common/src/app/common/logic/tokens.cljc
```diff
@@ -29,14 +29,91 @@
...
+
+(defn generate-toggle-token-set-group
+  "Toggle a token set group at `group-path` in `tokens-lib` without modifying a user theme."
+  [changes tokens-lib group-path]
+  (generate-update-active-sets changes tokens-lib #(toggle-token-set-group group-path tokens-lib %)))
+
+(defn vec-starts-with? [v1 v2]
+  (= (subvec v1 0 (min (count v1) (count v2))) v2))
+
+(defn generate-drop-token-set-or-set-group
```
hirunatan: - "drop" is a UI concept, so it should not be mentioned here.
- This is not a generate function, because it doesn't create changes.

I suggest the name `calculate-move-token-set-or-set-group`.

common/src/app/common/logic/tokens.cljc
```diff
@@ -29,14 +29,91 @@
...
+  [tokens-lib {:keys [from-index to-index position collapsed-paths] :as input
```
hirunatan: If you never use the whole map of options you can remove `:as input`.

common/src/app/common/logic/tokens.cljc
```diff
@@ -29,14 +29,91 @@
...
+  [changes tokens-lib drop-opts]
+  (when-let [drop-opts' (generate-drop-token-set-or-set-group tokens-lib drop-opts)]
+    (cond
+      (:error drop-opts') drop-opts'
+      :else (pcb/move-token-set-before changes drop-opts'))))
+
+(defn generate-drop-token-set-group
+  "Create changes for dropping a token set or token set group."
+  [changes tokens-lib drop-opts]
+  (generate-drop-token-set-or-set-group tokens-lib drop-opts)
```
hirunatan: This line is redundant, isn't it?

common/src/app/common/logic/tokens.cljc
```diff
@@ -29,14 +29,91 @@
...
+  (when-let [drop-opts' (generate-drop-token-set-or-set-group tokens-lib drop-opts)]
```
hirunatan: A `generate` function always receives a `changes` object, and return another `changes` that may be the same, or one with more changes added. It should not return `nil` or a different object.

I suggest changing the `when-let` by an `if-let` that return `changes` in the else part. And perhaps throwing an exception with the error if the move cannot be done.

common/src/app/common/types/tokens_lib.cljc
```diff
@@ -448,8 +464,8 @@
   (add-sets [_ token-set] "add a collection of sets to the library, at the end")
   (update-set [_ set-name f] "modify a set in the library")
   (delete-set-path [_ set-path] "delete a set in the library")
-  (move-set-before [_ set-name before-set-name] "move a set with `set-name` before a set with `before-set-name` in the library.
-When `before-set-name` is nil, move set to bottom")
+  (move-set [_ from-path to-path before-path before-group?] "TODO")
+  (move-set-group [_ from-path to-path before-path before-group?] "TODO")
```
hirunatan: Please fill in the docstrings

frontend/src/app/main/data/tokens.cljs
```diff
@@ -51,20 +54,17 @@
...
-(defn set-selected-token-set-path
-  [full-path]
-  (ptk/reify ::set-selected-token-set-path
-    ptk/UpdateEvent
-    (update [_ state]
-      (wtts/assoc-selected-token-set-path state full-path))))
+(defn get-token-data-from-token-id
+  [id]
+  (let [workspace-data (deref refs/workspace-data)]
+    (get (:tokens workspace-data) id)))
```
hirunatan: is there a `:tokens` key in workspace-data?

frontend/src/app/main/ui/workspace/tokens/sidebar.cljs
```diff
@@ -246,15 +246,15 @@
...
+        selected-token-set-name (mf/deref refs/workspace-selected-token-set-name)
 
         token-groups (mf/with-memo [tokens selected-token-set-tokens]
                        (-> (select-keys tokens (keys selected-token-set-tokens))
                            (sorted-token-groups)))]
     [:*
      [:& token-context-menu]
      [:& title-bar {:all-clickable true
-                    :title (dm/str "TOKENS - " (ctob/set-path->set-name selected-token-set-path))}]
+                    :title (dm/str "TOKENS - " selected-token-set-name)}]
```
hirunatan: This string should be i18n-ed

common/src/app/common/logic/tokens.cljc
floscr: This one is a bit clunky, as it needs to operate on a Theme but needs context from the the whole lib.
It's used for the theme modal and the the change generate functions:

https://github.com/tokens-studio/penpot/blob/0f5fda994e85f70b8032d78b04789ffb7ab48e82/common/src/app/common/logic/tokens.cljc#L15-L17

We could just put it as a function in the tokens lib or give the lib as an argument for a theme method, but both feels wrong. Any better idea?

We could move it to the Token Lib as well, but then the ony two consumers need to add their temporary theme to the lib just to extract it again.

floscr:
I've changed all the points from the feeback except the question from above.
I've also added some unit tests for the move-token-set-group changes genererator.
The changes generator now throws on unallowed moves.

common/src/app/common/logic/tokens.cljc
```diff
@@ -29,14 +29,101 @@
...
+        before (assoc :before-path (:path before)
+                      :before-group? (:group? before))
+        prev-before (assoc :prev-before-path (:path prev-before)
+                           :prev-before-group? (:group? prev-before))))))
+
+(defn generate-move-token-set
+  "Create changes for dropping a token set or token set.
+  Throws for impossible moves."
+  [changes tokens-lib drop-opts]
+  (when-let [drop-opts' (calculate-move-token-set-or-set-group tokens-lib drop-opts)]
```
hirunatan: All `generate-xx` functions should always return a changes object, that may be the same as the input one. The idea is that you can chain several generates together.

Please change the `when-let` by an `if-let` that returns changes in the else part (and update the test).

common/test/common_tests/logic/token_test.cljc
```diff
@@ -2,6 +2,7 @@
   (:require
    [app.common.files.changes-builder :as pcb]
    [app.common.logic.tokens :as clt]
+   [app.common.pprint :refer [pprint]]
```
hirunatan: :fire: :smile:

common/src/app/common/logic/tokens.cljc
hirunatan: Well, leave it as is for now. After MVP I'd like to review the working of the temporary themes.

PR #5584, opened 2025-01-14
:sparkles: Open comments from email with parameter

Alotor:
https://tree.taiga.io/project/penpot/task/9743
https://tree.taiga.io/project/penpot/task/9742

### DIFF
--- a/frontend/src/app/main/data/workspace.cljs
+++ b/frontend/src/app/main/data/workspace.cljs
@@ -48,6 +48,7 @@
    [app.main.data.workspace.bool :as dwb]
    [app.main.data.workspace.collapse :as dwco]
    [app.main.data.workspace.colors :as dwcl]
+   [app.main.data.workspace.comments :as dwcm]
    [app.main.data.workspace.drawing :as dwd]
    [app.main.data.workspace.edition :as dwe]
    [app.main.data.workspace.fix-broken-shapes :as fbs]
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/data/workspace.cljs
```diff
@@ -355,6 +356,13 @@
                      (rx/take 1)
                      (rx/map zoom-to-frame)))
 
+              (when (:comment-id rparams)
```
niwinz: Is a minor issue but I think we can follow the same stile as prevous code:
image

frontend/src/app/main/ui/workspace/comments.cljs
```diff
@@ -119,16 +115,7 @@
...
-             (st/emit! (dcm/go-to-workspace :page-id (:page-id thread)
-                                            ::rt/new-window true)))
-           (tm/schedule
-            (fn []
-              (st/emit! (when (not= page-id (:page-id thread))
-                          (dw/select-for-drawing :comments))
-                        (dwcm/center-to-comment-thread thread)
-                        (-> (dcmt/open-thread thread)
-                            (with-meta {::ev/origin "workspace"})))))))]
+           (st/emit! (dwcm/navigate-to-comment (:id thread)))))]
```
niwinz: I think, with that change you makes a RPC request on each click on the thread, independently if it is loaded or not.

frontend/src/app/main/data/workspace.cljs
Alotor: OK

frontend/src/app/main/ui/workspace/comments.cljs
Alotor: You're right, I'll fix it.

PR #5596, opened 2025-01-15
[TOKENS] :books: Add gradients to storybook swatch component

Xaviju:
This PR adds:

- Helpers to manage elements of the components in the storybook
- Fixes the gradient in the swatch* component
- Displays the gradient variants as well as the library color (rounded variant)

Task https://tree.taiga.io/project/penpot/task/9419

### DIFF
--- a/frontend/src/app/main/ui/ds/helpers.cljs
+++ b/frontend/src/app/main/ui/ds/helpers.cljs
@@ -7,8 +7,11 @@
 (ns app.main.ui.ds.helpers
   "A collection of helpers for exporting them to be used on storybook code."
   (:require
+   [app.common.uuid :refer [random]]
    [rumext.v2 :as mf]))
 
 (def default
   (mf/object
-   {:uuid parse-uuid}))
+   {:generate-uuid random
+    :keyword keyword}))
+
--- a/frontend/src/app/main/ui/ds/utilities/swatch.cljs
+++ b/frontend/src/app/main/ui/ds/utilities/swatch.cljs
@@ -80,7 +80,10 @@
         button-type (if (not read-only?) "button" nil)
         size (or size "small")
         active (or active false)
-        gradient (:gradient background)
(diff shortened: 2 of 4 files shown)


frontend/src/app/main/ui/ds/helpers.cljs
```diff
@@ -11,4 +11,12 @@
 
 (def default
   (mf/object
-   {:uuid parse-uuid}))
+   {:uuid parse-uuid
+    :linear :linear
```
niwinz: Is better expose a function that converts a string to keyword, instead of all possible keywords.
That helpers namespace/module is intended to expose helpers for build instances of objects and not sample values as-is.

frontend/src/app/main/ui/ds/utilities/swatch.cljs
```diff
@@ -80,7 +84,9 @@
         button-type (if (not read-only?) "button" nil)
         size (or size "small")
         active (or active false)
-        gradient (:gradient background)
+        gradient-type (-> background :gradient :type)
+        gradient-stops (-> background :gradient :stops)
+        gradient-data (uc/gradient->css (gradient-data gradient-type gradient-stops))
```
niwinz: You don't need to define a function that simply builds a map, just build map inline without adding the function

frontend/src/app/main/ui/ds/utilities/swatch.stories.jsx
```diff
@@ -47,52 +48,45 @@ export const WithOpacity = {
...
+    },
   },
 };
 
-export const Active = {
+export const Rounded = {
   args: {
-    active: true,
+    background: {
+      id: helpers.uuid(crypto.randomUUID()),
```
niwinz: I think we can expose directly our own uuid generation function instead invoking crypto here (a very minor issue)

PR #5624, opened 2025-01-20
:whale: improve docker documentation related to the updates

yamila-moreno:
Add a piece of advice related to updating the docker installation.


niwinz:
This PR i think we can target `main` for make it available immediatelly

PR #5630, opened 2025-01-21
:whale: Ensure that postgresql and redis are running before starting Penpot services

bameda:
<div align="center">

</div>

This PR defines new dependency conditions, via healhcheck, to ensure that posrtgresql and redis are ready to receive connections before starting the rest of the penpot services (backend and exporter).

This is intended to solve the concurrency problems that arise randomly when starting docker compose.

## How to test it

Try to start several times, even from a ‘clean’ environment, and check that you don't get connection errors with postgresql and redis from the backend and the exporter and make sure you can access penpot from the browser.

### DIFF
--- a/docker/images/docker-compose.yaml
+++ b/docker/images/docker-compose.yaml
@@ -113,8 +113,10 @@ services:
       - penpot_assets:/opt/data/assets
 
     depends_on:
-      - penpot-postgres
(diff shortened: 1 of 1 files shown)


yamila-moreno (approved):
Working as intended. 🎉

PR #5632, opened 2025-01-21
:sparkles: Add new components to storybook

Alotor:
Adds the following components to storybook:

- Utilities/Date
- Notifications/Actionable
- Product/Avatar
- Product/CTA
- Product/Milestone


frontend/resources/styles/common/base.scss
```diff
@@ -18,6 +18,11 @@ $lh-115: 1.15;
 $lh-133: 1.33;
 $size-4: 1rem;
 
+/* Background for storybook */
+body.sb-show-main {
+  background-color: var(--color-background-secondary);
+}
+
```
belen-albeza: Not the right way to do this... If we want to set background colors for stories, we have https://storybook.js.org/docs/essentials/backgrounds 

If it's the background color of storybook itself, there's themes https://storybook.js.org/docs/configure/user-interface/theming

frontend/src/app/main/ui/ds/colors.scss
```diff
@@ -15,6 +15,7 @@ $mint-250-10: #00d1b81a;
 $green-200: #a7e8d9;
 $green-500: #2d9f8f;
 $green-950: #0a2927;
+$green-1000: #0d1d1b;
```
belen-albeza: I can't find this token in the Foundations penpot file. Either let's get Design folks to add it, or use an existing green (like `green-950`).

frontend/src/app/main/ui/ds/colors.scss
```diff
@@ -27,6 +28,7 @@ $red-950: #500124;
 
 $pink-400: #ff6fe0;
 
+$purple-100: #f3ebff;
```
belen-albeza: Same for this purple

frontend/src/app/main/ui/ds/notifications/actionable.cljs
```diff
@@ -0,0 +1,34 @@
...
+   [:cancelLabel {:optional true} :string]])
+
+(mf/defc actionable*
+  {::mf/props :obj
+   ::mf/schema schema:actionable}
+  [{:keys [class variant actionLabel cancelLabel children] :rest props}]
+
+  (let [class (d/append-class class (stl/css :notification))
+        props (mf/spread-props props {:class class :data-testid "actionable"})]
+    [:> "div" props
```
belen-albeza: This is probably an `<aside>`

frontend/src/app/main/ui/ds/notifications/actionable.scss
```diff
@@ -0,0 +1,26 @@
...
+  background: var(--color-background-primary);
+  border-radius: $br-8;
+  border: $b-1 solid var(--color-background-quaternary);
+  display: flex;
+  justify-content: space-between;
+  padding: var(--sp-l);
+
+  :global(a) {
+    color: var(--color-accent-primary);
+  }
```
belen-albeza: Please let's not use global styles

frontend/src/app/main/ui/ds/notifications/actionable.scss
```diff
@@ -0,0 +1,26 @@
...
+// Copyright (c) KALEIDOS INC
+
+@use "../_borders.scss" as *;
+
+.notification {
+  align-items: center;
+  background: var(--color-background-primary);
+  border-radius: $br-8;
+  border: $b-1 solid var(--color-background-quaternary);
+  display: flex;
```
belen-albeza: It's not bad, but usually `grid` simplifies some things since it adds an implicit `width` to the element (for things like cropping a line with an ellipsis), and offers more options, so I tend to favor it for the sake of maintanability.

frontend/src/app/main/ui/ds/product/avatar.cljs
```diff
@@ -0,0 +1,46 @@
...
+     {:class (d/append-class
+              class
+              (stl/css-case :avatar true
+                            :is-selected selected
+                            :is-small (= variant "S")
+                            :is-medium (= variant "M")
+                            :is-large (= variant "L")))
+      :style {"--color" color}
+      :title name}
+     [:div {:class (stl/css :avatar-image)}
```
belen-albeza, line 45: what is this `<div>` needed for?

frontend/src/app/main/ui/ds/product/avatar.cljs
```diff
@@ -0,0 +1,46 @@
...
+                  (avatars/generate {:name name :color color}))]
+    [:> (or tag "div")
+     {:class (d/append-class
+              class
+              (stl/css-case :avatar true
+                            :is-selected selected
+                            :is-small (= variant "S")
+                            :is-medium (= variant "M")
+                            :is-large (= variant "L")))
+      :style {"--color" color}
```
belen-albeza: Let's prefix this var with the name of the component: `--avatar-color`

frontend/src/app/main/ui/ds/product/avatar.scss
```diff
@@ -0,0 +1,48 @@
...
+
+.avatar-image {
+  overflow: hidden;
+  border-radius: $br-circle;
+  background-color: var(--color);
+}
+
+.is-selected {
+  --border-color: var(--color-accent-primary);
+  padding: 3px;
```
belen-albeza: Let's use a token

frontend/src/app/main/ui/ds/product/avatar.scss
```diff
@@ -0,0 +1,48 @@
...
+}
+
+.avatar:hover {
+  --border-color: var(--color-accent-primary);
+}
+
+.avatar:focus {
+  --border-color: var(--color-accent-primary);
+}
```
belen-albeza: Pseudo-classes like these (and also pseudo-elements) can be added nested, since it would be just syntax sugar (i.e. does not add unneeded specificity).

```scss
.avatar {
  /* ... */
  &:hover {
    /* ... */
  }
}
```

frontend/src/app/main/ui/ds/product/avatar.scss
```diff
@@ -0,0 +1,48 @@
...
+
+.is-medium {
+  width: $sz-32;
+  height: $sz-32;
+}
+
+.is-large {
+  width: $sz-40;
+  height: $sz-40;
+}
```
belen-albeza: For variants or these kind of styling options, we follow [this pattern](https://help.penpot.app/technical-guide/developer/ui/#implementing-variants)

frontend/src/app/main/ui/ds/product/cta.scss
```diff
@@ -0,0 +1,36 @@
...
+
+:global(.light) {
+  --color-cta-background: #{$purple-100};
+  --color-cta-border: #{$purple-600-10};
+}
+
+:global(.default) {
+  --color-cta-border: #{$green-950};
+  --color-cta-background: #{$green-1000};
+}
```
belen-albeza: The right way of doing this is via semantic tokens.

```scss
// colors.scss
:global(.light) {
  --whatever-design-has-called-this-token: #{$purple-100};
}

:global(.default) {
  --whatever-design-has-called-this-token: #{$green-1000};
}
```

And then, at a component level:

```scss
// cta.scss
.cta {
  --color-cta-background: var(--whatever-design-has-called-this-token);
}
```

Design is responsible of naming semantic tokens, hence `--whatever-design-has-called-this-token`. They usually ask for our input, but it's their decision and it should be added to the Foundations file.

frontend/src/app/main/ui/ds/product/cta.scss
```diff
@@ -0,0 +1,36 @@
...
+
+.cta {
+  border-radius: $br-8;
+  border: 1px solid var(--color-cta-border);
```
belen-albeza: Let's create a $br-1 token

frontend/src/app/main/ui/ds/product/cta.scss
```diff
@@ -0,0 +1,36 @@
...
+  padding: var(--sp-m);
+}
+
+.cta-message {
+  @include t.use-typography("body-small");
+  color: var(--color-foreground-secondary);
+
+  :global(a) {
+    color: var(--color-accent-primary);
+  }
```
belen-albeza: No global selectors in components

frontend/src/app/main/ui/ds/product/milestone.cljs
```diff
@@ -0,0 +1,92 @@
...
+(def ^:private schema:milestone
+  [:map
+   [:class {:optional true} :string]
+   [:active {:optional true} :boolean]
+   [:editing {:optional true} :boolean]
+   [:autosaved {:optional true} :boolean]
+   [:versionToggled {:optional true} :boolean]
+   [:userName :string]
+   [:userAvatar {:optional true} [:maybe :string]]
+   [:userColor :string]
```
belen-albeza: It feels like this should be a `user` object with `name`, `avatar` and `color`, no?

frontend/src/app/main/ui/ds/product/milestone.cljs
```diff
@@ -0,0 +1,92 @@
...
+   [:label :string]
+   [:date [:fn valid-date?]]
+   [:autosavedMessage :string]
+   [:snapshots [:vector [:fn valid-date?]]]])
+
+(mf/defc milestone*
```
belen-albeza: I feel that this component is too complex and could benefit of splitting some parts into private components (i.e. `<EditMilestone>`, `<MilestoneVersion>`, etc.)

frontend/src/app/main/ui/ds/product/milestone.scss
```diff
@@ -0,0 +1,164 @@
...
+  align-items: center;
+
+  column-gap: var(--sp-s);
+
+  &.is-selected,
+  &:hover {
+    --border-color: var(--color-accent-primary);
+  }
+
+  .name-input {
```
belen-albeza: Let's unnest this

frontend/src/app/main/ui/ds/product/milestone.scss
```diff
@@ -0,0 +1,164 @@
...
+
+    input {
+      height: auto;
+      margin: 0;
+      font-family: "worksans", "vazirmatn", sans-serif;
+      font-optical-sizing: auto;
+      font-weight: 400;
+      line-height: 1.3;
+      font-size: 0.875rem;
+    }
```
belen-albeza: Unnest this as well

frontend/src/app/main/ui/ds/product/milestone.scss
```diff
@@ -0,0 +1,164 @@
...
+    border: none;
+
+    input {
+      height: auto;
+      margin: 0;
+      font-family: "worksans", "vazirmatn", sans-serif;
+      font-optical-sizing: auto;
+      font-weight: 400;
+      line-height: 1.3;
+      font-size: 0.875rem;
```
belen-albeza: If this typography does not exist, we need to ask design to define it + add it to the Design System . Then, we should add this to the typography mixins and use its given name.

frontend/src/app/main/ui/ds/product/milestone.scss
```diff
@@ -0,0 +1,164 @@
...
+  justify-items: baseline;
+  margin: 0;
+  margin-top: var(--sp-xxs);
+  padding: 0;
+  cursor: pointer;
+
+  &:hover {
+    color: var(--color-accent-primary);
+
+    .icon-arrow {
```
belen-albeza: unnest

frontend/src/app/main/ui/ds/product/milestone.scss
```diff
@@ -0,0 +1,164 @@
...
+.icon-clock {
+  --icon-stroke-color: var(--color-accent-warning);
+}
+
+.icon-arrow {
+  --icon-stroke-color: var(--color-foreground-secondary);
+
+  &.toggled {
+    transform: rotate(90deg);
+  }
```
belen-albeza: unnest

frontend/src/app/main/ui/ds/utilities/date.cljs
```diff
@@ -0,0 +1,41 @@
...
+  [{:keys [class as date selected] :rest props}]
+  (let [as (or as "span")
+        class (d/append-class class (stl/css-case :date true :is-selected selected))
+        date (cond-> date (not (dt/datetime? date)) dt/datetime)]
+    [:> text* {:as as :typography t/body-medium :class class}
+     (dm/str
+      (dt/format date :date-full)
+      " . "
+      (dt/format date :time-24-simple)
+      "h")]))
```
belen-albeza: There's a proper HTML element for this: [`<time>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) (and we should also set its `datetime` attribute)

frontend/resources/styles/common/base.scss
Alotor: No, it won't work because in the global configuration `preview.js` we have:

```
    },
    backgrounds: { disable: true },
``` 
This will pick the global styles. And this is why y added the custom here.

frontend/src/app/main/ui/ds/colors.scss
Alotor: I forgot to change the values (because were defined afterwards) but the values are defined in the US, even if the DS is not updated yet:

image

frontend/src/app/main/ui/ds/notifications/actionable.scss
Alotor: How can I style the text that is a children otherwise? the text that it's added as a child is from markdown so we cannot customize its class.

frontend/src/app/main/ui/ds/product/avatar.cljs
Alotor: If you check the design it has two strokes. I need a second div basically to add the second stroke (that is really an internal padding).

frontend/src/app/main/ui/ds/product/cta.scss
Alotor: The same comment as in the notification. The text inside will be took from a markdown, and returns a "generic" `<a>` i don't know how to style this if not with globals.

frontend/src/app/main/ui/ds/product/milestone.scss
Alotor: I can't :) the input component has some properties defined and I cannot override it without a "more specific" selector.

frontend/src/app/main/ui/ds/colors.scss
Alotor: Design told me they would add `$mint-900`

frontend/src/app/main/ui/ds/colors.scss
Alotor: Design will add `$purple-100`

frontend/src/app/main/ui/ds/colors.scss
```diff
@@ -106,6 +106,9 @@ $grayish-red: #bfbfbf;
   --color-token-border: #{$aqua-400};
   --color-token-accent: #{$aqua-600};
   --color-token-foreground: #{$aqua-800};
+
+  --color-cta-background: #{$purple-100};
+  --color-cta-border: #{$purple-600-10};
```
belen-albeza: Double checking: Do these names come from design? They look like component-level token names

frontend/src/app/main/ui/ds/product/avatar.cljs
belen-albeza: We could use `outline` instead https://developer.mozilla.org/en-US/docs/Web/CSS/outline

PR #5648, opened 2025-01-22
:bug: Fix token name conflict on duplication

### LINKED ISSUE #53: Fix duplicating a token twice
When duplicating a token twice the second token wont show up because of the naming clash.

https://github.com/user-attachments/assets/7eca8a96-892e-49fd-9ea0-06e3425de2e2

not-in-stock:
# Fix Duplicated Token Name Conflict

## Issue
When duplicating a token twice, the second token is not created due to a naming clash.

## Solution
This PR resolves the issue by introducing a unique name generation function to prevent name conflicts when duplicating tokens.

### Steps to Reproduce
1. Create a token.
2. Click **Duplicate Token** in the token's context menu.
3. Click **Duplicate Token** again in the same token's context menu (on the original token, not the copy).

### Expected Result
- A new token is successfully created with a unique name.

### Actual Result
- The token is not created due to a name conflict.

### DIFF
--- a/common/src/app/common/files/helpers.cljc
+++ b/common/src/app/common/files/helpers.cljc
@@ -400,31 +400,47 @@
                    elements)]
     (into #{} (keep :name) elements)))
 
-(defn- extract-numeric-suffix
-  [basename]
-  (if-let [[_ p1 p2] (re-find #"(.*) ([0-9]+)$" basename)]
-    [p1 (+ 1 (d/parse-integer p2))]
-    [basename 1]))
+(defn- name-seq
+  "Creates a lazy sequence of names in the pattern:
+   (str name),
+   (str name (suffix-fn 1)),
+   (str name (suffix-fn 2)),
+   (str name (suffix-fn 3))..."
+  [base-name suffix-fn]
+  (concat
+   [base-name]
+   (map #(str/concat base-name (suffix-fn %))
+        (iterate inc 1))))
+
+(defn ^:private get-suffix
+  "Default suffix impelemtation"
+  [copy-count]
+  (str/concat " " copy-count))
 
 (defn generate-unique-name
-  "A unique name generator"
-  [used basename]
+  "Generates unique name by
+   returning first unique name from the sequence that doesn't exist in existing-names
+  - `base-name` - a name used for further unique name geneations
+  - `existing-names` - exitsing entity names to check uniqness against
+  - options
+    - `:suffix-fn` - a function that generates unique suffix
+       - accepts `copy-number`, returns `string`
+    - `:immediate-suffix?` - boolean, if `true` suffix added immediately"
+  [base-name existing-names & {:keys [suffix-fn immediate-suffix?]
+                               :or {suffix-fn get-suffix}}]
   (dm/assert!
    "expected a set of strings"
-   (sm/check-set-of-strings! used))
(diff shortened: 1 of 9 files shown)


floscr (approved):
Nice :+1:

not-in-stock:
A similar issue is also observed with files. Copy names don't have a unique enumeration 

https://github.com/user-attachments/assets/ba584272-1140-4346-af08-e89b23078946

And also with Projects

https://github.com/user-attachments/assets/de8f963f-dfe8-4e5a-b8a6-c5c9dd50fb77

not-in-stock:
And also with Projects

https://github.com/user-attachments/assets/de8f963f-dfe8-4e5a-b8a6-c5c9dd50fb77

frontend/src/app/main/data/helpers.cljs
```diff
@@ -168,3 +169,24 @@
...
+        (iterate inc 1))))
+
+(defn get-unique-copy-name
+  "Returns first unique name from the sequence that doesn't exist in existing-names"
+  [source-name existing-names suffix-fn]
+  (let [existing-name-set (set existing-names)
+        names (name-seq source-name suffix-fn)]
+    (->> names
+         (remove #(contains? existing-name-set %))
+         first)))
```
niwinz: We already have a method for doing that, maybe it is not as complete as this one but I think we can improve the current approach instead of creating a new one just for this https://github.com/penpot/penpot/blob/develop/common/src/app/common/files/helpers.cljc#L408

not-in-stock: @niwinz yes I can definitely improve it by incorporating my changes to an existing one.

niwinz:
Mark it ready to review again when is ready. This is justo for not hitting on my radar until that. Thanks :heart:

common/src/app/common/files/helpers.cljc
```diff
@@ -400,31 +400,52 @@
...
-
-  (if-not (contains? used basename)
-    basename
-    (let [[prefix initial] (extract-numeric-suffix basename)]
-      (loop [counter initial]
-        (let [candidate (str prefix " " counter)]
-          (if (contains? used candidate)
-            (recur (inc counter))
-            candidate))))))
+   (string? base-name))
```
not-in-stock: @niwinz
I have a small question: How necessary are these assertions?
I've noticed that in some places, `Spec` is used for schema validation, while in others, it's `malli`, and sometimes assertions like these are used.
It’s a bit confusing to determine which approach to choose for argument validation.

Also, I think the function itself could handle converting `existing-names` into a set. This way, we could relax its requirement from `::set-of-strings` to a more generic collection of strings.

niwinz: The spec is deprecated, but we can't replace all in one step. So we migrating away from spec gradually.

But if you simple need assets, just use `(assert...)`

not-in-stock:
@niwinz 
alright, I think now PR is ready for review again

niwinz:
At first quick view, looks very nice, I have today a bit busy with unexpected tasks, but I will merge it Monday as one of my first tasks.

niwinz:
The CI is failing because this PR is not rebased with latest changes on develop, but don't worry, i will rebase it myself.

PR #5651, opened 2025-01-22
📚  Update create-a-plugin.md

BDVGitHub:
Ask the user to check if a plugin does not already exist before creating it.

> Screenshots/GIFs


niwinz:
After consulting with the team internally, we are going to take your idea (thank you very much) but we are going to use a slightly different text than the one you proposed: https://github.com/penpot/penpot/pull/5656

In any case, thank you very much.

PR #5652, opened 2025-01-22
:bug: Update media references after instantiation of a component

niwinz:
## The issue

When a user instantiates a component with media references (image-fills per example) from an non-local library on the current file. The instance shape conserves the same values for :id on the `:fill-image` attribute as the original shape. As we should known, this ID represents the row on the relation between the **file** and **storage_object* and this relation is stored in the `file_media_object` table.

When the original file is deleted or the original component is deleted, all the component instances that uses some kind of media assets will point to deleted file media object row, which is susceptible to be garbage collected and leave the shape broken.

## The fix

**Check for media refs on `:add-obj` change and fix that references and remap them on the file on update-file**

On all changes of type `:add-obj` each shape is checked if it has media references, if references are found then that references will be processed by a secondary step on update-file request and will create the missing references and will update the file shapes accordingly with new ids.

## **How to test**

- Create a library file with single component with fill image
- Create other file A and link it to the Library file
- Instantiate a component from Library on the A

What is happening without the PR:

- The shapes created has fill-image pointing to a row on `file_media_object` that is part of file Library, there are no rows associated with file A.

What is happening with the FIX:

- The shapes created has fill-image pointing to a row on `file_media_object` that is part of file Library, but you also has an entry for the file A with different ID.
- If you refresh, the shape is now pointing to other ID (the correct one, with the corresponding row on file_media_object pointing  to the file A)

## Known Problems

When you instantiate a component, we still use original media references that points to the Library File instead to a correct one. This is because we can't create new ids on frontend and we can't wait for a rpc request for explicit creation of the references before create instance and draw it on workspace, it will introduce a lot of latency (because of need to wait backend response before allowing user drag and drop the new instance)

There are a solution for this, but it is out of scope of this PR that is already very big for a bugfix.

### DIFF
--- a/backend/src/app/binfile/common.clj
+++ b/backend/src/app/binfile/common.clj
@@ -247,7 +247,7 @@
     WHERE fmo.id = ANY(?::uuid[])
       AND file_id != ?")
 
-(defn update-media-references
+(defn update-media-references!
   "Given a file and a coll of media-refs, check if all provided
   references are correct or fix them in-place"
   [{:keys [::db/conn] :as cfg} {file-id :id :as file} media-refs]
@@ -260,10 +260,10 @@
                              (dissoc :created-at)
                              (dissoc :deleted-at))))
                 {}
-                (db/plan conn [sql:get-missing-media-references
-                               (->> (into #{} xf-map-id media-refs)
-                                    (db/create-array conn "uuid"))
-                               file-id]))
+                (db/exec! conn [sql:get-missing-media-references
+                                (->> (into #{} xf-map-id media-refs)
+                                     (db/create-array conn "uuid"))
+                                file-id]))
 
         lookup-index
         (fn [id]
--- a/backend/src/app/binfile/common.clj
+++ b/backend/src/app/binfile/common.clj
@@ -12,6 +12,7 @@
    [app.common.data.macros :as dm]
    [app.common.exceptions :as ex]
    [app.common.features :as cfeat]
+   [app.common.files.helpers :as cfh]
    [app.common.files.migrations :as fmg]
    [app.common.files.validate :as fval]
    [app.common.logging :as l]
@@ -29,7 +30,6 @@
    [app.util.time :as dt]
    [app.worker :as-alias wrk]
    [clojure.set :as set]
-   [clojure.walk :as walk]
    [cuerdas.core :as str]
    [datoteka.fs :as fs]
    [datoteka.io :as io]))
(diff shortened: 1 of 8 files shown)


backend/src/app/rpc/commands/files_update.clj
```diff
@@ -415,19 +416,31 @@
...
-                 (update :revn inc)
-                 (update :data cpc/process-changes changes)
-                 (update :data d/without-nils))]
+        libs
+        (when (and (or (contains? cf/flags :file-validation)
+                       (contains? cf/flags :soft-file-validation))
+                   (not skip-validate))
+          (get-file-libraries cfg file))
+
+        state
```
Alotor, line 436: I think a comment to inform of what is storing would be appropriate.

common/src/app/common/files/changes.cljc
```diff
@@ -623,6 +628,27 @@
         (fn [container]
           (ctst/add-shape id obj container frame-id parent-id index ignore-touched))]
 
+    ;; The main purpose of this is ensure that all created shapes has
```
Alotor: I think all this block it's better to be in its own private function so the "main" purpose of `add-obj` is not ofuscated by this big wall of comment and the functions that execute.

I propose something like:

```clojure
(when *state* (collect-media! obj))
```

Alotor: Also if we need to add this to more events will be easier.

backend/test/backend_tests/rpc_file_test.clj
```diff
@@ -1658,3 +1658,174 @@
             components (get-in result [:data :components])]
         (t/is (not (contains? components c-id)))))))
 
+
```
Alotor, line 1661: you can remove all this blankspaces

backend/test/backend_tests/rpc_file_test.clj
```diff
@@ -1658,3 +1658,174 @@
...
+                :file-id file-id
+                :is-local true
+                :name "testfile"
+                :content mfile}
+        out    (th/command! params)]
+
+    ;; (th/print-result! out)
+    (t/is (nil? (:error out)))
+    (:result out)))
+
```
Alotor, line 1681: idem

backend/src/app/binfile/common.clj
```diff
@@ -241,40 +241,65 @@
...
+  (let [missing-index
+        (reduce (fn [result {:keys [id] :as fmo}]
+                  (assoc result id
+                         (-> fmo
+                             (assoc :id (uuid/next))
+                             (assoc :file-id file-id)
+                             (dissoc :created-at)
+                             (dissoc :deleted-at))))
+                {}
+                (db/plan conn [sql:get-missing-media-references
```
Alotor: Review if `plan` is necessary

backend/src/app/binfile/common.clj
```diff
@@ -241,40 +241,65 @@
             :data nil}
            {::sql/columns [:media-id :file-id :revn]}))
 
+(def ^:private sql:get-missing-media-references
+  "SELECT fmo.*
+     FROM file_media_object AS fmo
+    WHERE fmo.id = ANY(?::uuid[])
+      AND file_id != ?")
+
+(defn update-media-references
```
Alotor: I would consider to add a `!` to indicate that this will produce an effect.

backend/test/backend_tests/rpc_file_test.clj
niwinz: This blank lines are explicit, mainly because the test file is large and the test case is also large, so when you scroll on file, you have a better visual feedback for where a test case impl starts and ends.

PR #5594, opened 2025-01-15
:bug: Fix color wrapper choppiness

not-in-stock:
# Fix color picker choppiness
### Steps to reproduce:
1. Open the color token creation form.
2. Click on a color to expand the color ramp.
3. Place the picker circle in the bottom-left corner of the ramp.
4. Attempt to move the hue slider.

### Expected result:
- The hue slider can be moved.
- The picker circle moves smoothly across the color ramp.

### Actual result:
- The hue slider cannot be moved.
- The picker circle exhibits choppy movement when dragged across the color ramp.

**Before**

https://github.com/user-attachments/assets/6bd9a0c6-77c6-4205-800d-19fd2077991b

**After**

https://github.com/user-attachments/assets/6bf063fd-18ad-4a55-acd5-250db6287ca2

PR #5655, opened 2025-01-23
:bug: Fix color wrapper choppiness

niwinz:
Replaces https://github.com/penpot/penpot/pull/5594

### DIFF
--- a/frontend/src/app/main/ui/workspace/colorpicker.cljs
+++ b/frontend/src/app/main/ui/workspace/colorpicker.cljs
@@ -33,7 +33,7 @@
    [app.main.ui.workspace.colorpicker.harmony :refer [harmony-selector]]
    [app.main.ui.workspace.colorpicker.hsva :refer [hsva-selector]]
    [app.main.ui.workspace.colorpicker.libraries :refer [libraries]]
-   [app.main.ui.workspace.colorpicker.ramp :refer [ramp-selector]]
+   [app.main.ui.workspace.colorpicker.ramp :refer [ramp-selector*]]
    [app.main.ui.workspace.colorpicker.shortcuts :as sc]
    [app.util.dom :as dom]
    [app.util.i18n :as i18n :refer [tr]]
@@ -346,7 +346,7 @@
                                       [:div {:class (stl/css :picker-detail-wrapper)}
                                        [:div {:class (stl/css :center-circle)}]
                                        [:canvas#picker-detail {:class (stl/css :picker-detail) :width 256 :height 140}]]
-                                      [:& ramp-selector
+                                      [:> ramp-selector*
                                        {:color current-color
                                         :disable-opacity disable-opacity
                                         :on-change handle-change-color
--- a/frontend/src/app/main/ui/workspace/colorpicker/ramp.cljs
+++ b/frontend/src/app/main/ui/workspace/colorpicker/ramp.cljs
@@ -7,8 +7,8 @@
 (ns app.main.ui.workspace.colorpicker.ramp
   (:require-macros [app.main.style :as stl])
   (:require
-   [app.common.data :as d]
    [app.common.colors :as cc]
+   [app.common.data :as d]
    [app.common.math :as mth]
    [app.main.ui.components.color-bullet :as cb]
    [app.main.ui.workspace.colorpicker.slider-selector :refer [slider-selector]]
@@ -52,7 +52,8 @@
                     :top (str (* 100 (- 1 (/ value 255))) "%")}}]]))
 
 
-(defn enrich-color-map [{:keys [h s v] :as color}]
+(defn- enrich-color-map
+  [{:keys [h s v] :as color}]
   (let [h (d/nilv h 0)
         s (d/nilv s 0)
         v (d/nilv v 0)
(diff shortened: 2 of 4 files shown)


frontend/src/app/main/ui/workspace/colorpicker.cljs
```diff
@@ -346,7 +346,7 @@
                                       [:div {:class (stl/css :picker-detail-wrapper)}
                                        [:div {:class (stl/css :center-circle)}]
                                        [:canvas#picker-detail {:class (stl/css :picker-detail) :width 256 :height 140}]]
-                                      [:& ramp-selector
+                                      [:> ramp-selector*
```
hirunatan, line 349: All `defc` components should not have a `*` suffix, and be called with `:>`. The `:&` syntax is deprecated. This way we avoid an automatic conversion between clojure and js formats that had an important performance cost.

frontend/src/app/main/ui/workspace/colorpicker/ramp.cljs
```diff
@@ -63,38 +64,52 @@
...
-(mf/defc ramp-selector
+(mf/defc ramp-selector*
   [{:keys [color disable-opacity on-change on-start-drag on-finish-drag]}]
-  (let [internal-color
+  (let [internal-color*
         (mf/use-state #(enrich-color-map color))
 
-        {:keys [h s v hex alpha]} @internal-color
+        internal-color
+        (deref internal-color*)
```
hirunatan, line 73: Storing the state value in a variable avoids `deref`ing it each time it's used.

`deref` and `@` are equivalent, this is a matter of coherent style.

frontend/src/app/main/ui/workspace/colorpicker/ramp.cljs
```diff
@@ -63,38 +64,52 @@
...
+
+        h     (get internal-color :h)
+        s     (get internal-color :s)
+        v     (get internal-color :v)
+        hex   (get internal-color :hex)
+        alpha (get internal-color :alpha)
```
hirunatan, line 79: This is not very important. Destructuring has a little added cost, because it has to check if the parameter is map or seq. This is done here only as an example, but it's only necessary in critical-path code.

frontend/src/app/main/ui/workspace/colorpicker/ramp.cljs
```diff
@@ -63,38 +64,52 @@
...
+
+        bullet-color
+        (mf/with-memo [hex alpha]
```
hirunatan, line 82: Store the value in a memoed variable, to avoid creating the map in each rerender. So the gc has less work.

frontend/src/app/main/ui/workspace/colorpicker/ramp.cljs
```diff
@@ -63,38 +64,52 @@
...
         on-change-value-saturation
-        (fn [new-saturation new-value]
-          (let [new-color (-> @internal-color
-                              (assoc :s new-saturation)
-                              (assoc :v new-value)
-                              (enrich-color-map))]
-            (reset! internal-color new-color)
-            (on-change new-color)))
+        (mf/use-fn
+         (mf/deps internal-color on-change)
```
hirunatan, line 87: It's important to cache all internal functions in a `mf/use-fn` to avoid creating a new function each time.

frontend/src/app/main/ui/workspace/colorpicker/ramp.cljs
```diff
@@ -63,38 +64,52 @@
...
 
         on-change-opacity
-        (fn [new-opacity]
-          (let [new-color (-> @internal-color
-                              (assoc :alpha new-opacity))]
-            (reset! internal-color new-color)
-            (on-change new-color)))]
-
-    (mf/with-effect [color]
-      (reset! internal-color (enrich-color-map color)))
```
hirunatan: This effect isn't necessary because the state is initialized with this value.

frontend/src/app/main/ui/workspace/tokens/form.cljs
```diff
@@ -156,30 +156,43 @@
...
+(defn hex->value
+  [hex]
+  (when-let [tc (tinycolor/valid-color hex)]
+    (let [hex (str "#" (tinycolor/->hex tc))
+          [r g b] (c/hex->rgb hex)
+          [h s v] (c/hex->hsv hex)]
+      {:hex hex
+       :r r :g g :b b
+       :h h :s s :v v
+       :alpha 1})))
```
hirunatan, line 168: Pure functions that do not access internal state should be put outside the component. This way they are created statically and not need to be create on each component instantiation.

frontend/src/app/main/ui/workspace/tokens/form.cljs
```diff
@@ -156,30 +156,43 @@
...
-                             [r g b] (c/hex->rgb hex)
-                             [h s v] (c/hex->hsv hex)]
-                         {:hex hex
-                          :r r :g g :b b
-                          :h h :s s :v v
-                          :alpha 1})))
-        on-change' (fn [{:keys [hex]}]
-                     (when-not (and @dragging? hex)
-                       (on-change hex)))]
+        dragging-ref     (mf/use-ref false)
```
hirunatan, line 173: The `mf/use-state` triggers a component redraw each time the state is changed. In this case this is not necessary, because the value is only used in other internal function.

We can use `mf/use-ref` that serves the same purpose, but doesn't triger redraws.

frontend/src/app/main/ui/workspace/tokens/form.cljs
```diff
@@ -156,30 +156,43 @@
...
 
     (colorpicker/use-color-picker-css-variables! wrapper-node-ref (hex->value color))
     [:div {:ref wrapper-node-ref}
      [:> ramp-selector*
       {:color (hex->value color)
        :disable-opacity true
-       :on-start-drag #(reset! dragging? true)
-       :on-finish-drag #(reset! dragging? false)
+       :on-start-drag on-start-drag
+       :on-finish-drag on-finish-drag
```
hirunatan, line 195: It's important not to have inline functions, even tiny ones, because they are also created each time the component is redrawn. This has a little performance cost and adds more objects for the garbage collector.

PR #5658, opened 2025-01-23
:bug: Fix user cant submit existing token when editing name only

### LINKED ISSUE #18: Editing token name doesn't allow submit until switched to value input
https://github.com/user-attachments/assets/2c0aeeae-7041-4c64-827b-5af01405b641

When editing a token the user has to switch to the value field for the name change to be valid.

## Steps to reproduce

1. Edit an existing token
2. Update the name
3. Submit should be not `disabled` when name is valid

floscr:
- Fixes https://github.com/tokens-studio/penpot/issues/18
- Adds integration test for editing a token

### DIFF
--- a/frontend/playwright/ui/specs/tokens.spec.js
+++ b/frontend/playwright/ui/specs/tokens.spec.js
@@ -7,7 +7,7 @@ test.beforeEach(async ({ page }) => {
   await BaseWebSocketPage.mockRPC(page, "get-teams", "get-teams-tokens.json");
 });
 
-const setupFileWithTokens = async (page) => {
+const setupEmptyTokensFile = async (page) => {
   const workspacePage = new WorkspacePage(page);
   await workspacePage.setupEmptyFile();
   await workspacePage.mockRPC(
@@ -29,12 +29,44 @@ const setupFileWithTokens = async (page) => {
   };
 };
 
+const setupTokensFile = async (page) => {
+  const workspacePage = new WorkspacePage(page);
+  await workspacePage.setupEmptyFile();
+  await workspacePage.mockRPC(
+    "get-team?id=*",
+    "workspace/get-team-tokens.json",
+  );
+  await workspacePage.mockRPC(/get\-file\?/, "workspace/get-file-tokens.json");
+  await workspacePage.mockRPC(
+    /get\-file\-fragment\?/,
+    "workspace/get-file-fragment-tokens.json",
+  );
+
+  await workspacePage.goToWorkspace({
+    fileId: "51e13852-1a8e-8037-8005-9e9413a1f1f6",
+    pageId: "51e13852-1a8e-8037-8005-9e9413a1f1f7",
+  });
+
+  const tokensTabButton = page.getByRole("tab", { name: "Tokens" });
+  await tokensTabButton.click();
+
+  return {
+    workspacePage,
+    tokensUpdateCreateModal: workspacePage.tokensUpdateCreateModal,
+    tokenThemesSetsSidebar: workspacePage.tokenThemesSetsSidebar,
+    tokenSetItems: workspacePage.tokenSetItems,
+    tokenSetGroupItems: workspacePage.tokenSetGroupItems,
+    tokensSidebar: workspacePage.tokensSidebar,
+    tokenContextMenuForToken: workspacePage.tokenContextMenuForToken,
+  };
+};
+
 test.describe("Tokens: Tokens Tab", () => {
   test("Clicking tokens tab button opens tokens sidebar tab", async ({
     page,
   }) => {
     const { workspacePage, tokensUpdateCreateModal, tokenThemesSetsSidebar } =
-      await setupFileWithTokens(page);
+      await setupEmptyTokensFile(page);
 
     const tokensTabPanel = page.getByRole("tabpanel", { name: "tokens" });
 
@@ -46,7 +78,7 @@ test.describe("Tokens: Tokens Tab", () => {
     page,
   }) => {
     const { workspacePage, tokensUpdateCreateModal, tokenThemesSetsSidebar } =
-      await setupFileWithTokens(page);
+      await setupEmptyTokensFile(page);
 
     const tokensTabPanel = page.getByRole("tabpanel", { name: "tokens" });
     await tokensTabPanel.getByTitle("Add token: Color").click();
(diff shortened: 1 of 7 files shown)


not-in-stock (approved):
LGTM!

frontend/src/app/main/ui/workspace/tokens/form.cljs
```diff
@@ -389,6 +391,22 @@
...
+     (fn []
+       (reset! form-token-cache-atom nil)))
+
+    ;; Update the value when editing an existing token
+    ;; so the user doesn't have to interact with the form to validate the token
+    (mf/use-effect
+     (mf/deps create? resolved-tokens token token-resolve-result set-resolve-value)
+     (fn []
+       (when (and (not create?)
+                  (not @token-resolve-result)
```
EvaMarco: Make deref of this state on let, and not @ on every use.

frontend/src/app/main/ui/workspace/tokens/form.cljs
```diff
@@ -389,6 +391,22 @@
...
+                  resolved-tokens)
+         (-> (get resolved-tokens @name-ref)
```
EvaMarco, line 407: The same as above

frontend/playwright/ui/specs/tokens.spec.js
```diff
@@ -156,7 +235,7 @@ test.describe("Tokens: Sets Tab", () => {
   //     tokenThemesSetsSidebar,
   //     tokenSetItems,
   //     tokenSetGroupItems,
-  //   } = await setupFileWithTokens(page);
+  //   } = await setupEmptyTokensFile(page);
```
EvaMarco, line 238: This test is commented, why?

frontend/src/app/main/ui/workspace/tokens/form.cljs
```diff
@@ -215,7 +215,9 @@
 (mf/defc form
   {::mf/wrap-props false}
   [{:keys [token token-type action selected-token-set-name]}]
-  (let [token (or token {:type token-type})
+  (let [create? (not (instance? ctob/Token token))
+
```
EvaMarco: remove empty line

EvaMarco (changes requested):

frontend/playwright/ui/specs/tokens.spec.js
floscr: It was removed during one of the merge conflicts and I search and replaced for this file.
I've added a ticket here to restore it:
https://github.com/tokens-studio/penpot/issues/61

frontend/src/app/main/ui/workspace/tokens/form.cljs
floscr: This is actually on purpose. I dont want this hook to trigger when the refs update, I only need to trigger it once when editing a token and the tokens have resolved.
Since this should only trigger 
So I want to explicitly want to deps compare the other two refs by reference not by value.

Is there a guideline in penpot when this is the case? Maybe exclude the refs from the deps?

EvaMarco: Doing deref on let has nothing to do with deps on effect.
We can save some time doing only one deref, and the as you want, don't place it on deps, so the value change don`t trigger re-render.

niwinz: Having state references (state objects) on deps make no sense, deps should almost always have values and no references. We don't need a special guideline, the react use-ref, return a reference that points a value and returns a stable reference, so you never need to put the ref on the deps. The same applies for rumext use-state, because it tries to offer a similar API. 

We have finally applied the change that made the reference returned in rumext use-state always stable, but that broke a behavior in the tokens part that used it in a way that was not recommended. To facilitate progress, we have temporarily reverted this change. So right now the object returned from use-state always changes if the underlying value changes, so use effect will trigger in any way. In other words, having the state reference object on deps and derefing on use-effect has no real meaning and is an anti pattern in any way and we should avoid this because on future we will apply again the change of having stable reference.

TLDR: @EvaMarco suggests derefing once and use many times and is correct

floscr: Thank you both for the explanation, I was missing that context :pray: 
I'll update both instances.

floscr: @niwinz  Would be neat to add a linter for such things, in [uix](https://github.com/pitch-io/uix/) there's a linter for react hooks to highlight missing dependencies, so it would be possible to lint for deref inside hooks
Linter namespace: https://github.com/pitch-io/uix/blob/master/core/src/uix/linter.clj
Would be super helpful for outside contributors.

niwinz: @floscr agree, i will put that on my tasklist, thanks

frontend/src/app/main/ui/workspace/tokens/form.cljs
floscr: This one uses `use-var` for `name-ref` to not re-render on update of the token name input.
I haven't seen any direct deref pattern for `use-var`, is it still applicable?

EvaMarco: Yes, In that file you have use `@name-ref` 5 times as I can see. 
That is doing `(deref name-ref)` five times each time it renders. 
If we do on let
```clojure
        name-ref* (mf/use-var (:name token))
        name-ref (deref name-ref*)
```
The deref funtion is only executed once. 
Then we will use `name-ref` when we need the value, and name-ref* when we need to set a different value. 

This is appart of use-var, or use-callbacks or deps. We want to reduce "derefs"

floscr: Okay, thanks :raised_hands:

floscr: Hmm I'm running into issues with this :thinking: 
The derefed value in the let statement never updates with the suggested changes for the use-var case only.

https://github.com/user-attachments/assets/97c9ebe7-6889-47a1-8850-3f57f3133f16

It's confusing me now also a bit as `use-var` uses a simple wrapper around reacts `useRef` and I would not expect the suggested change to work, but I wasn't sure if I have all the context around rumext.

Am I doing something wrong?

I've pushed some debug commits here:
https://github.com/penpot/penpot/pull/5658/commits/973f58284e6537b88be39d72ba8e7b80c286cf84

EvaMarco: That’s an error on my part. 
>This one uses use-var for name-ref to not re-render on update of the token name input.
I haven't seen any direct deref pattern for use-var, is it still applicable?
You tell me you where talking about use-var I din't read that. 

You where talking about use-var all the time...... No is not applicable.

floscr: No problem :relaxed:

EvaMarco (approved):

PR #5663, opened 2025-01-23
:bug: Add migration fixing files with shape-ref cycles

pabloalba:


common/src/app/common/files/migrations.cljc
```diff
@@ -1145,6 +1146,36 @@
...
+                      (fn [objects]
+                        (reduce-kv (fn [acc k v]
+                                     (if (contains? to-detach k)
+                                       (assoc acc k (ctk/detach-shape v))
+                                       (assoc acc k v)))
+                                   {}
+                                   objects)))))]
+
+    (-> data
+        (update :components update-vals update-component))))
```
hirunatan: I'd add a filter to process only deleted components. For more speed, and also to avoid the possibility of the code to have unintended effects in a component without `:objects`.

PR #5668, opened 2025-01-24
:bug: Fix click on the project name goes to blank dashboard from work…

superalex:
https://tree.taiga.io/project/penpot/issue/9877

### DIFF
--- a/frontend/src/app/main/ui/workspace/left_header.cljs
+++ b/frontend/src/app/main/ui/workspace/left_header.cljs
@@ -32,6 +32,7 @@
   (let [profile     (mf/deref refs/profile)
         file-id     (:id file)
         file-name   (:name file)
+        project-id  (:id project)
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/ui/workspace/left_header.cljs
```diff
@@ -78,7 +79,7 @@
 
         nav-to-project
         (mf/use-fn
-         #(st/emit! (dcm/go-to-dashboard-files ::rt/new-window true)))]
+         #(st/emit! (dcm/go-to-dashboard-files ::rt/new-window true :project-id project-id)))]
```
niwinz, line 83: `project-id` should be on mf/deps of use-fn

superalex: Thanks!

PR #5677, opened 2025-01-24
[Tokens] :sparkles Add combobox to select group in token themes modal

juanfran:
https://tree.taiga.io/project/penpot/task/9753


frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
```diff
@@ -158,40 +159,35 @@
...
+        options (map (fn [group]
+                       {:label group
+                        :id group})
+                     theme-groups)
         on-update-group (partial on-change-field :group)
-        on-update-name (partial on-change-field :name)]
+        on-update-name
+        (mf/use-fn
+         (fn [event]
+           (let [value (.-value (.-currentTarget event))]
```
EvaMarco: There is a namespace called dom where we have fns to get -current-target or get-value
```clojure
target  (dom/get-current-target event)
name   (dom/get-value target)
```
we can do this in one line also.

frontend/translations/en.po
```diff
@@ -6292,6 +6292,9 @@ msgstr "Edit themes"
 msgid "workspace.token.edit-token"
 msgstr "Edit token"
 
+msgid "workspace.token.group"
```
EvaMarco: This could be a "label.group"
because its shot, we can use it in different context, is very clear.

frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
niwinz: Yep, that helps should be used instead of that interop because they provide the right compiler hints for the interop calls that avoid issues with advanced compilation modes where almost all functions are renamed.

juanfran: :ok_hand: done

frontend/translations/en.po
juanfran: done

EvaMarco (approved):

PR #5683, opened 2025-01-27
🎉 Add alpha to token color

### LINKED ISSUE #6: Use Alpha value for color tokens
It is possible to import tokens with alpha values like `rgba(0,0,0,0.3)` but will be converted to hex values automatically.

Image

Make sure to use the correct colorpicker with alpha values possible in the form as well.

floscr:
Allows changing/adding alpha to a token color value.
Closes https://github.com/tokens-studio/penpot/issues/6

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/form.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/form.cljs
@@ -161,12 +161,13 @@
   [hex]
   (when-let [tc (tinycolor/valid-color hex)]
     (let [hex (str "#" (tinycolor/->hex tc))
+          alpha (tinycolor/alpha tc)
           [r g b] (c/hex->rgb hex)
           [h s v] (c/hex->hsv hex)]
       {:hex hex
        :r r :g g :b b
        :h h :s s :v v
-       :alpha 1})))
+       :alpha alpha})))
 
 (mf/defc ramp*
   [{:keys [color on-change]}]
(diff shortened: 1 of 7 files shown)


frontend/src/app/main/ui/workspace/tokens/form.cljs
```diff
@@ -313,10 +313,15 @@
...
-                           (set! (.-value (mf/ref-val value-input-ref)) hex-value)
-                           (on-update-value-debounced hex-value)))
+                         (fn [hex-value alpha]
+                           (let [color-value (if (= 1 alpha)
+                                               hex-value
+                                               (-> (tinycolor/valid-color hex-value)
+                                                   (tinycolor/set-alpha alpha)
+                                                   (tinycolor/->rgba-string)))]
+                             (reset! value-ref color-value)
+                             (set! (.-value (mf/ref-val value-input-ref)) color-value)
```
EvaMarco: Is better to use the dom.cljs name space and change that line for something like this
`(dom/set-value! node new-val)`
This is because they provide the right compiler hints for the interop calls that avoid issues with advanced compilation modes where almost all functions are renamed.

EvaMarco:
Can you add some test with alpha colors?

floscr:
> Can you add some test with alpha colors?

:heavy_check_mark: https://github.com/penpot/penpot/pull/5683/files#diff-892950bf20ba73c0adfdcd5d17024edd0e06017531a2efce8db04903e2f40b78

EvaMarco (approved):

PR #5687, opened 2025-01-27
:sparkles: Apply token values to shapes in all pages

hirunatan:
Fixes https://tree.taiga.io/project/penpot/issue/9924

Still pending to see how to apply updates of shapes of type text in a page that is not current. This will be done in a new PR.


niwinz:
Replaced by #5738

PR #5688, opened 2025-01-27
:bug: Update padding only on shapes with layout

### LINKED ISSUE #59: bug: Internal error when applying not accepted tokens
### Steps To Reproduce

- Create a dimension token.
- Create a board.
- Apply the dimension token to the left padding of the board.
- An internal error occurs.

### Expected behavior

This happens because boards do not have a "padding" property (unless they are converted to a layout).

This issue will be addressed in the enhancement described here: [GitHub Issue](https://github.com/tokens-studio/penpot/issues/8). However, if it's not possible to resolve it for the MVP, we should at least avoid throwing an internal error and simply not apply the token instead.

### Actual behavior

- An internal error occurs.

[report (9).txt](https://github.com/user-attachments/files/18522051/report.9.txt)

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

floscr:
Fixes https://github.com/tokens-studio/penpot/issues/59 and adds test case for padding

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/changes.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/changes.cljs
@@ -6,6 +6,7 @@
 
 (ns app.main.ui.workspace.tokens.changes
   (:require
+   [app.common.types.shape.layout :as ctsl]
    [app.common.types.shape.radius :as ctsr]
    [app.common.types.token :as ctt]
    [app.common.types.tokens-lib :as ctob]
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/ui/workspace/tokens/changes.cljs
```diff
@@ -173,22 +174,30 @@
                        (zipmap (repeat value)))]
     {:layout-gap layout-gap}))
 
+(defn shape-ids-with-layout [state shape-ids]
```
niwinz: Should be marked as private as this function is only used here. It helps detect unused functions when they become unused thanks to linter.

PR #5695, opened 2025-01-28
:tada: Import legacy format tokens

### LINKED ISSUE #49: Importing legacy format tokens
## Issue

In Penpot the tokens are compatible with DTCG format:

` { "red100": {
        "$value": "#f00",
        "$type": "color"
}
}
`
There could be tokens which are still in the legacy format which can come from various sources including exported tokens from the Tokens Studio Figma plugin:

`{ "red100": {
        "value": "#f00",
         "type": "color"
}
}`

When importing just token files into Penpot, currently we do not get any error and no tokens are imported except for the set names which is confusing.

## Technical details

In the import logic find a way to detect if tokens are in a legacy format,

A way would be when recursing the json tree at the first found object that has no children (= token), has `value|type` without prefixes.

## Proposed Solution

When a token file is imported in the legacy format we should detect it and throw an error message and provide an easy way of converting from the legacy format to the DTCG format.

"The file you have uploaded is not compatible with the DTCG format. Upload your file below to convert your file into a compatible format."

To convert legacy tokens use this website: https://dtcg-convert.netlify.app/

Slack thread: https://hyma-team.slack.com/archives/C069R040GDR/p1736503496695119

not-in-stock:
# **Add support for legacy JSON import**

## Issue
It's not possible to import tokens from legacy JSON files

## Solution
This PR resolves the issue by introducing a function detects file format before import JSON files

**Test legacy JSON file:**
[Meraki-for-penpot_legacy.json](https://github.com/user-attachments/files/18572649/Meraki-for-penpot_legacy.json)
**Test dctg JSON file:**
[Meraki-for-penpot.json](https://github.com/user-attachments/files/18576570/Meraki-for-penpot.json)

### DIFF
--- a/common/src/app/common/types/tokens_lib.cljc
+++ b/common/src/app/common/types/tokens_lib.cljc
@@ -670,6 +670,43 @@ used for managing active sets without a user created theme.")
 
 ;; === Import / Export from DTCG format
 
+(defn has-legacy-format?
+  "Searches through parsed token file and returns:
+   - true when first node satisfies `legacy-node?` predicate
+   - false when first node satisfies `dtcg-node?` predicate
+   - nil if neither combination is found"
+  ([data]
+   (let [legacy-node? (fn [node]
+                        (and (map? node)
+                             (or (and (contains? node "value")
+                                      (contains? node "type"))
+                                 (and (contains? node "value")
+                                      (sequential? (get node "value"))))))
+         dtcg-node? (fn [node]
+                      (and (map? node)
+                           (or (and (contains? node "$value")
+                                    (contains? node "$type"))
+                               (and (contains? node "$value")
+                                    (sequential? (get node "$value"))))))]
+     (has-legacy-format? data legacy-node? dtcg-node?)))
+  ([data legacy-node? dtcg-node?]
+   (let [branch? map?
+         children (fn [node] (vals node))
+         check-node (fn [node]
+                      (cond
+                        (legacy-node? node) true
+                        (dtcg-node? node) false
+                        :else nil))
+         walk (fn walk [node]
+                (lazy-seq
+                 (cons
+                  (check-node node)
+                  (when (branch? node)
+                    (mapcat walk (children node))))))]
+     (->> (walk data)
+          (filter some?)
+          first))))
+
 (defn walk-sets-tree-seq
   [nodes & {:keys [walk-children?]
             :or {walk-children? (constantly true)}}]
@@ -1142,7 +1179,6 @@ Will return a value that matches this schema:
       (-> sets
           (assoc "$themes" themes)
           (assoc-in ["$metadata" "tokenSetOrder"] ordered-set-names))))
-
   (decode-dtcg-json [_ parsed-json]
     (let [metadata (get parsed-json "$metadata")
           sets-data (dissoc parsed-json "$themes" "$metadata")
--- a/common/src/app/common/types/tokens_lib.cljc
+++ b/common/src/app/common/types/tokens_lib.cljc
@@ -670,25 +670,39 @@ used for managing active sets without a user created theme.")
 
 ;; === Import / Export from DTCG format
 
+(def ^:private legacy-node?
+  (sm/validator
+   [:or
+    [:map
+     ["value" :string]
+     ["type" :string]]
+    [:map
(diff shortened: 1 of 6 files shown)


common/src/app/common/types/tokens_lib.cljc
```diff
@@ -1173,7 +1174,27 @@ Will return a value that matches this schema:
...
+          sets-data (dissoc parsed-legacy-json "$themes" "$metadata")
+          dtcg-sets-data (walk/postwalk
+                          (fn [node]
+                            (cond-> node
+                              (and (map? node)
+                                   (contains? node "value")
+                                   (sequential? (get node "value")))
+                              (update "value"
+                                      (fn [seq-value]
+                                        (map #(set/rename-keys % {"type" "$type"}) seq-value)))
```
not-in-stock, line 1240: This condition is added to handle tokens like this:
https://github.com/penpot/penpot/pull/5695/files#diff-4179ab5570deb92b9019518866b7d9c6419bad5f7fe83ff91c6b7e20e068f944R715-R737

frontend/src/app/main/ui/workspace/tokens/style_dictionary.cljs
```diff
@@ -188,7 +188,9 @@
                      (throw (wte/error-ex-info :error.import/json-parse-error data e))))))
        (rx/map (fn [json-data]
                  (try
-                   (ctob/decode-dtcg-json (ctob/ensure-tokens-lib nil) json-data)
+                   (if-not (ctob/has-legacy-format? json-data)
```
floscr, line 191: What about the nil case?
Is this still decodable with `ctob/decode-dtcg-json` ?
(I guess we wrap it in try catch, so it doesn't really matter)

frontend/src/app/main/ui/workspace/tokens/sidebar.cljs
```diff
@@ -310,7 +310,6 @@
         on-display-file-explorer
         (mf/use-fn
          #(.click (mf/ref-val input-ref)))
-
```
floscr: Would remove those changes with the newline removals

common/src/app/common/types/tokens_lib.cljc
```diff
@@ -1173,7 +1210,27 @@ Will return a value that matches this schema:
                                        (set sets))))
          lib' themes-data)
         lib')))
-
```
floscr, line 1176: The functions in this area should have newlines between them

common/src/app/common/types/tokens_lib.cljc
```diff
@@ -670,6 +670,43 @@ used for managing active sets without a user created theme.")
...
+(defn has-legacy-format?
+  "Searches through parsed token file and returns:
+   - true when first node satisfies `legacy-node?` predicate
+   - false when first node satisfies `dtcg-node?` predicate
+   - nil if neither combination is found"
+  ([data]
+   (let [legacy-node? (fn [node]
+                        (and (map? node)
+                             (or (and (contains? node "value")
+                                      (contains? node "type"))
```
floscr: What about the edge-case of two tokens called `value` and `type`, wouldn't they get falsely detected as a legacy-node ?

E.g.:

```clj
{"error.value" {:value "..."}
 "error.type" {:value "..."}
 ```

File: frontend/src/app/main/ui/workspace/tokens/style_dictionary.cljs (hunk 2 above)
Comment (not-in-stock) [reply frontend/src/app/main/ui/workspace/tokens/style_dictionary.cljs:191]: Yes it will use `ctob/decode-dtcg-json` since nil a falsy value. But it shouldn’t be a problem because if there're no tokens in the file it doesn’t matter which function we use to process it.

File: frontend/src/app/main/ui/workspace/tokens/sidebar.cljs (hunk 3 above)
Comment (not-in-stock) [reply frontend/src/app/main/ui/workspace/tokens/sidebar.cljs]: Yes, sure, that's more of a typo

File: frontend/translations/en.po [hunk 6]
```diff
@@ -1751,6 +1751,9 @@ msgstr "Export"
 msgid "labels.import"
 msgstr "Import"
 
+msgid "labels.import-legacy"
+msgstr "Import (legacy)"
+
```
floscr: Remove unused

floscr:
Looks good now 👍

PR #5697, opened 2025-01-28
:bug: Fix border-radius applied to all corners on token update

### LINKED ISSUE #56: bug: editing a token value modifies other applied border-radius tokens
### Steps To Reproduce

- Create a border-radius token and apply it to one of the corners of a rectangle
- Create a color token and apply it to the fill of the said rectangle
- Edit the color token and change its value to another one

### Expected behavior

The color of the rectangle gets updated.

### Actual behavior

The color of the rectange is updated but the border radius token is now applied to the four corners.
This also happens modifying other types of tokens like rotation, opacity, etc.

### Screenshots or video

https://github.com/user-attachments/assets/0f2ba7e9-e987-435c-b80f-6f2aff0904dc

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

floscr:
Fixes border radius being overwritten for all corners when updating a token.
Closes https://github.com/tokens-studio/penpot/issues/56

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/changes.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/changes.cljs
@@ -104,16 +104,12 @@
                        :attrs ctt/border-radius-keys}))
 
 (defn update-shape-radius-for-corners [value shape-ids attributes]
-  (ptk/reify ::update-shape-radius-for-corners
-    ptk/WatchEvent
-    (watch [_ _ _]
-      (rx/of
-       (dwsh/update-shapes shape-ids
-                           (fn [shape]
-                             (ctsr/set-radius-for-corners shape attributes value))
-                           {:reg-objects? true
-                            :ignore-touched true
-                            :attrs ctt/border-radius-keys})))))
+  (dwsh/update-shapes shape-ids
+                      (fn [shape]
+                        (ctsr/set-radius-for-corners shape attributes value))
+                      {:reg-objects? true
+                       :ignore-touched true
+                       :attrs ctt/border-radius-keys}))
 
 (defn update-opacity [value shape-ids]
   (when (<= 0 value 1)
--- a/frontend/src/app/main/ui/workspace/tokens/changes.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/changes.cljs
@@ -96,22 +96,24 @@
 (defn update-shape-radius-all [value shape-ids]
   (dwsh/update-shapes shape-ids
                       (fn [shape]
-                        (when (ctsr/can-get-border-radius? shape)
-                          (ctsr/set-radius-to-all-corners shape value)))
+                        (ctsr/set-radius-to-all-corners shape value))
                       {:reg-objects? true
                        :ignore-touched true
                        :attrs ctt/border-radius-keys}))
 
-(defn update-shape-radius-single-corner [value shape-ids attributes]
-  ;; NOTE: This key should be namespaced on data tokens, but these events are not there.
-  (st/emit! (ptk/data-event :expand-border-radius))
-  (dwsh/update-shapes shape-ids
-                      (fn [shape]
-                        (when (ctsr/can-get-border-radius? shape)
-                          (ctsr/set-radius-to-single-corner shape (first attributes) value)))
-                      {:reg-objects? true
-                       :ignore-touched true
-                       :attrs ctt/border-radius-keys}))
+(defn update-shape-radius-for-corners [value shape-ids attributes]
+  (ptk/reify ::update-shape-radius-for-corners
+    ptk/WatchEvent
+    (watch [_ _ _]
+      (rx/of
+       ;; NOTE: This key should be namespaced on data tokens, but these events are not there.
+       (ptk/data-event :expand-border-radius)
+       (dwsh/update-shapes shape-ids
+                           (fn [shape]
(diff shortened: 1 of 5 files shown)


not-in-stock (approved):
LGTM!

frontend/src/app/main/ui/workspace/tokens/changes.cljs
```diff
@@ -96,22 +96,24 @@
...
-                      {:reg-objects? true
-                       :ignore-touched true
-                       :attrs ctt/border-radius-keys}))
+(defn update-shape-radius-for-corners [value shape-ids attributes]
+  (ptk/reify ::update-shape-radius-for-corners
+    ptk/WatchEvent
+    (watch [_ _ _]
+      (rx/of
+       ;; NOTE: This key should be namespaced on data tokens, but these events are not there.
+       (ptk/data-event :expand-border-radius)
```
floscr: Is this still necessary?

hirunatan: This is necessary, but it shouldn't be there. Better put it at the calling point (in context-menu). This function is also called from `main.ui.workspace.tokens/update-tokens` and then it's not useful to emit this event, since its done on not selected shapes.

frontend/src/app/main/ui/workspace/tokens/changes.cljs
```diff
@@ -96,22 +96,24 @@
...
-                      (fn [shape]
-                        (when (ctsr/can-get-border-radius? shape)
-                          (ctsr/set-radius-to-single-corner shape (first attributes) value)))
-                      {:reg-objects? true
-                       :ignore-touched true
-                       :attrs ctt/border-radius-keys}))
+(defn update-shape-radius-for-corners [value shape-ids attributes]
+  (ptk/reify ::update-shape-radius-for-corners
+    ptk/WatchEvent
+    (watch [_ _ _]
```
hirunatan: Why are you changing this from a simple function into an event?

frontend/src/app/main/ui/workspace/tokens/changes.cljs
```diff
@@ -96,22 +96,24 @@
 (defn update-shape-radius-all [value shape-ids]
   (dwsh/update-shapes shape-ids
                       (fn [shape]
-                        (when (ctsr/can-get-border-radius? shape)
-                          (ctsr/set-radius-to-all-corners shape value)))
+                        (ctsr/set-radius-to-all-corners shape value))
```
hirunatan, line 99: Why do you remove the check? Are you sure this can only be called for shapes that may have border radius?

floscr: Because these functions are already checking this:
https://github.com/tokens-studio/penpot/blob/430c5cd60206930e22742f6c342baa52bb721f34/common/src/app/common/types/shape/radius.cljc#L40-L58

frontend/src/app/main/ui/workspace/tokens/changes.cljs
floscr: No real reason, I thought it would be better everything is uniform in this ns

frontend/src/app/main/ui/workspace/tokens/changes.cljs
floscr: Yeah I'll move it in this PR then :+1:

frontend/src/app/main/ui/workspace/tokens/changes.cljs
hirunatan: ok

frontend/src/app/main/ui/workspace/tokens/changes.cljs
hirunatan: I don't understand this. All `update-xxx` functions are normal functions that return an event, not events theirselves.  I think this should be rolled back.

frontend/src/app/main/ui/workspace/tokens/context_menu.cljs
```diff
@@ -179,14 +180,24 @@
                               :on-update-shape wtch/update-layout-sizing-limits}
                              context-data)))
 
+(defn update-shape-radius-all [value shape-ids]
+  (st/emit!
+   (ptk/data-event ::expand-border-radius)
```
hirunatan: I think there is no need to expand here. This causes event that the border radius input expands to have four values. But if we set the same value of all borders it should be kept collapsed.

frontend/src/app/main/ui/workspace/tokens/changes.cljs
floscr: If you look at the functions below, it's a mixture of events and update functions:
https://github.com/tokens-studio/penpot/blob/6e641b2dd5c1c438d2713970684ed8ab47d50e06/frontend/src/app/main/ui/workspace/tokens/changes.cljs#L149-L202

Anyway I'm fine with either way

frontend/src/app/main/ui/workspace/tokens/context_menu.cljs
floscr: Ok I misunderstood what this does, I thought it would be some kind of data collection event.
In that case these should also not be `::`

hirunatan (approved):

PR #5722, opened 2025-01-30
:sparkles: Add has-files prop to create-file event

cocotime:
https://tree.taiga.io/project/penpot/us/9802?milestone=424203


frontend/src/app/main/ui/dashboard/files.cljs
```diff
@@ -158,7 +158,7 @@
          (mf/deps project)
          (fn [origin]
            (let [mdata  {:on-success on-file-created}
-                 params {:project-id (:id project)}]
+                 params {:project-id (:id project) :has-files (> file-count 0)}]
```
niwinz: This should not be a parameter, because it is not necessary as parameter. This should go directly on metadata, in the same way as ::ev/origin is passed some lines below

PR #5723, opened 2025-01-30
:bug: Fix combobox bg color and dropdown only with options (#9753)

juanfran:
https://tree.taiga.io/project/penpot/task/9753


niwinz:
The commit message should be like:

`Show dropdown and bg color only when options are not empty`

or

`Fix combobox dropdown visibility when no options are provided`

As described in CONTRIBUTING.md

Xaviju (approved):

PR #5726, opened 2025-01-30
:tada: Non blocking render wasm

superalex:


render-wasm/src/render.rs
```diff
@@ -241,65 +281,69 @@ impl RenderState {
...
+        let cancel_script = format!("cancelAnimationFrame({})", frame_id);
+        let c_cancel_script = std::ffi::CString::new(cancel_script).unwrap();
+        unsafe {
+            emscripten_run_script(c_cancel_script.as_ptr());
         }
+    }
 
-        if self.options.is_debug_visible() {
-            self.render_debug();
+    pub fn process_animation_frame(&mut self, tree: &mut HashMap<Uuid, Shape>, timestamp: i32) {
```
belen-albeza: Let's make this function return a `Result` instead of handling `.expect`s here

render-wasm/src/render.rs
```diff
@@ -337,58 +381,72 @@ impl RenderState {
...
-                    is_complete = self.render_shape_tree(&id, tree) && is_complete;
+                    if !node_id.is_nil() {
+                        self.render_shape(&mut element.clone(), element.clip());
+                    } else {
+                        self.apply_drawing_to_render_canvas();
+                    }
                     self.drawing_surface.canvas().restore();
+
+                    // Set the node as visited before processing children
+                    self.pending_stack.push((node_id, true));
```
belen-albeza: Let's rename `pending stack` to `pending_nodes`. When we were commenting the code earlier, I though this `pending_stack` contained the IDs of requestAnimationFrames (probably because we were talking about it, but still…)

render-wasm/src/render.rs
```diff
@@ -337,58 +381,72 @@ impl RenderState {
...
 
-            let mut paint = skia::Paint::default();
-            paint.set_blend_mode(element.blend_mode().into());
-            paint.set_alpha_f(element.opacity());
-            let filter = element.image_filter(self.viewbox.zoom * self.options.dpr());
-            if let Some(image_filter) = filter {
-                paint.set_image_filter(image_filter);
-            }
+        while let Some((node_id, visited_children)) = self.pending_stack.pop() {
+            if let Some(element) = tree.get(&node_id) {
```
belen-albeza: This function has lots of nesting and it can be hard to read. We can refactor it later, no prob, but as a suggestion for this PR, we could handle the pattern of 

```rust
if let Some(foo) = some_option {
  // do stuff
}
else {
  return Err("Lorem ipsum")
 }
 ```
 
 In a more concise manner, with [the `?` operator](https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html), assuming the function returns a `Result<_, String>` (like this one):
 
 ```rust
let foo = some_option.ok_or("Lorem ipsum".to_string())?;
 ``` 
 
 The `?` does an early return if it finds an `Err` or `None`.

File: render-wasm/src/render.rs [hunk 4]
```diff
@@ -337,58 +381,72 @@ impl RenderState {
...
+                );
             }
 
-            self.final_surface.canvas().restore();
-
-            return is_complete;
-        } else {
-            eprintln!("Error: Element with root_id {root_id} not found in the tree.");
-            return false;
+            if get_time() - timestamp > MAX_BLOCKING_TIME_MS {
```
belen-albeza: I worry the `performance.now()` call we do here might be too expensive to do it after each node. Usually these calls are reserved to just once per frame, whenever posible. A quick workaround could be to batch nodes and only do this call after rendering a specific amount

belen-albeza:
Hi @superalex thanks for the PR! Some questions I have:

 - I don't quite get the difference between `final_surface` and `render_surface`. Why keep both? Is it to pace the `flush`? Something else?
 - I found this glitch that does not happen in `hourly`: the rendering shapes get clipped when doing a pan
 
Screen Recording 2025-02-06 at 2 41 34 PM

- When opening the Lucide Icons file, if I do some pans, zoom ins & outs there are some times where the board just disappears. Probably related to the above, but there's also this situation:

Screenshot 2025-02-06 at 2 50 59 PM

render-wasm/src/render.rs
superalex: Great, thanks!

render-wasm/src/render.rs
superalex: Very useful :)

render-wasm/src/render.rs
superalex: Updated!

superalex:
Thank you very much!

> * I don't quite get the difference between `final_surface` and `render_surface`. Why keep both? Is it to pace the `flush`? Something else?

We need a surface where we can draw the cached images without losing the the shapes that need to be drawn. Without it, at the point we render the cached image to the final surface we had to redraw everything

> * I found this glitch that does not happen in `hourly`: the rendering shapes get clipped when doing a pan

I think this one makes sense, in hourly really the pan doesn't happen until the image is rendered, now the pan happens immediately. I think this should be improved when we have the tile system developed 

> * When opening the Lucide Icons file, if I do some pans, zoom ins & outs there are some times where the board just disappears. Probably related to the above, but there's also this situation:

I have to check this one, I couldn't reproduce it with my tests

render-wasm/src/render.rs
superalex: Thanks!, I think it makes sense adding a threshold now, an improvement could be doing a fast test on the user browser to check how fast it can render some shapes and set that number on the initialization

PR #5731, opened 2025-01-30
:bug: Fix color id on color bullet component

EvaMarco:
This PR fixes this issue https://tree.taiga.io/project/penpot/issue/10015

### DIFF
--- a/frontend/src/app/main/ui/components/color_bullet.cljs
+++ b/frontend/src/app/main/ui/components/color_bullet.cljs
@@ -68,7 +68,7 @@
              :title (color-title color)}]
       ;; No multiple selection
       (let [color    (if (string? color) {:color color :opacity 1} color)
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/ui/components/color_bullet.cljs
```diff
@@ -68,7 +68,7 @@
              :title (color-title color)}]
       ;; No multiple selection
       (let [color    (if (string? color) {:color color :opacity 1} color)
-            id       (:id color)
+            id       (:ref-id color)
```
niwinz: I think it should be `(or (:ref-id color) (:id color))` ...
because reasons ... (colors are broken on penpot heheh)

EvaMarco: good point

PR #5748, opened 2025-02-03
:recycle: Create shared notification utility component

Xaviju:
This PR:

- Creates a new `notification-pill` component that will be shared across notification elements
- Refactors `toast*` component to use this shared component amongst other minor fixes
- Switches the current toast component with the new `toast*` component for global notifications.

See story https://tree.taiga.io/project/penpot/us/10050?milestone=427620


frontend/src/app/main/ui/notifications.cljs
```diff
@@ -26,16 +27,16 @@
...
         toast?       (or (= :toast (:type notification))
-                         (some? (:timeout notification)))]
+                         (some? (:timeout notification)))
+        content     (or (:content notification) "")]
 
     (when notification
       (cond
         toast?
-        [:& toast-notification
+        [:& toast*
```
niwinz: Components with `*` in the name should always use `[:>` handler instead of `[:&`

frontend/src/app/main/ui/notifications.cljs
```diff
@@ -51,8 +52,7 @@
           :content (:content notification)}]
 
         :else
-        [:& toast-notification
+        [:& toast*
```
niwinz: The same here

frontend/src/app/main/ui/ds/notifications/shared/notification_pill.cljs
```diff
@@ -0,0 +1,39 @@
...
+;; Copyright (c) KALEIDOS INC
+
+(ns app.main.ui.ds.notifications.shared.notification-pill
+  (:require-macros
+   [app.main.style :as stl])
+  (:require
+   [app.main.ui.ds.foundations.assets.icon :as i]
+   [rumext.v2 :as mf]))
+
+(def ^:private icons-by-level
```
niwinz: This should be defined as a function, using `case`

frontend/src/app/main/ui/notifications.cljs
```diff
@@ -8,9 +8,10 @@
   (:require
    [app.main.data.notifications :as ntf]
    [app.main.store :as st]
+   [app.main.ui.ds.notifications.toast :refer [toast*]]
    [app.main.ui.notifications.context-notification :refer [context-notification]]
    [app.main.ui.notifications.inline-notification :refer [inline-notification]]
-   [app.main.ui.notifications.toast-notification :refer [toast-notification]]
+  ;;  [app.main.ui.notifications.toast-notification :refer [toast-notification]]
```
niwinz: :fire:
Also, if toast-notification component is no longer used, it also should be removed.

frontend/src/app/main/ui/ds/notifications/toast.cljs
```diff
@@ -10,36 +10,27 @@
...
-    [:> "aside" props
-     [:*
-      [:> i/icon* {:icon-id icon-id :class (stl/css :icon)}]
-      children
+  [{:keys [class level type children on-close] :rest props}]
+  (let [class (dm/str class " " (stl/css-case :toast true))
+        level (or level :info)
+        type (or type :context)
+        props (mf/spread-props props {:class class :role "alert"})]
+    (js/console.log "toast*" level)
```
niwinz: :fire:

frontend/src/app/main/ui/ds/notifications/shared/notification_pill.cljs
```diff
@@ -0,0 +1,42 @@
...
+   [rumext.v2 :as mf]))
+
+(defn icons-by-level
+  [level]
+  (case level
+    :info i/info
+    :warning i/msg-neutral
+    :error i/delete-text
+    :success i/status-tick
+    level))
```
niwinz: I think this should be:

```clojure
(defn icons-by-level
  [level]
  (case level
    :info i/info
    :warning i/msg-neutral
    :error i/delete-text
    :success i/status-tick
    i/msg-neutral))
```

The default value should be of the same type, I mean it should be `nil` or `icon` but not a keyword which is receiving by parameters.

frontend/src/app/main/ui/ds/notifications/shared/notification_pill.cljs
```diff
@@ -0,0 +1,42 @@
...
+   ::mf/schema schema:notification-pill}
+  [{:keys [level type children]}]
+  (let [class (stl/css-case :notification-pill true
+                            :type-toast (= type :toast)
+                            :type-context (= type :context)
+                            :level-warning  (= level :warning)
+                            :level-error    (= level :error)
+                            :level-success  (= level :success)
+                            :level-info     (= level :info))
+        icon-id (or (get icons-by-level level) i/msg-neutral)]
```
niwinz: I expect here call the `icon-by-level` function directly. Using `get` here is not correct.
```clojure
(icon-by-level level)
```

The default value should be handled already by the function, so you don't to handle it here.

frontend/src/app/main/ui/ds/notifications/toast.cljs
```diff
@@ -10,36 +10,26 @@
...
+  (let [class (dm/str class " " (stl/css-case :toast true))
+        level (or level :info)
+        type (or type :context)
+        props (mf/spread-props props {:class class :role "alert"})]
+    [:> "div" props
+     [:> notification-pill* {:level (keyword level) :type type} children]
       ;; TODO: this should be a buttom from the DS, but this variant is not designed yet.
       ;; https://tree.taiga.io/project/penpot/task/8492
-      [:> "button" {:on-click on-close :aria-label "Close" :class (stl/css :close-button)} [:> i/icon* {:icon-id i/close}]]]]))
+     [:> "button" {:on-click on-close :aria-label "Close" :class (stl/css :close-button)} [:> i/icon* {:icon-id i/close}]]]))
```
niwinz: Add some line breaks

frontend/src/app/main/ui/ds/notifications/toast.cljs
```diff
@@ -10,36 +10,26 @@
...
-     [:*
-      [:> i/icon* {:icon-id icon-id :class (stl/css :icon)}]
-      children
+  [{:keys [class level type children on-close] :rest props}]
+  (let [class (dm/str class " " (stl/css-case :toast true))
+        level (or level :info)
+        type (or type :context)
+        props (mf/spread-props props {:class class :role "alert"})]
+    [:> "div" props
+     [:> notification-pill* {:level (keyword level) :type type} children]
```
niwinz: The `(keyword level)` is executed always, independently if the level is a string or keyword.
I think you should check if coming level is string on let

```clojure
(let [...
      level (if (string? level)
              (keyword level)
              (d/nilv level :info))
```

frontend/src/app/main/ui/notifications.cljs
```diff
@@ -8,9 +8,9 @@
   (:require
    [app.main.data.notifications :as ntf]
    [app.main.store :as st]
+   [app.main.ui.ds.notifications.toast :refer [toast*]]
    [app.main.ui.notifications.context-notification :refer [context-notification]]
    [app.main.ui.notifications.inline-notification :refer [inline-notification]]
-   [app.main.ui.notifications.toast-notification :refer [toast-notification]]
```
niwinz, line 13: If toast notification is no longer used in penpot, the namespace defining the component also should be removed.

niwinz (changes requested):

frontend/src/app/main/ui/ds/notifications/shared/notification_pill.cljs
Xaviju: You are right, fixing it

frontend/src/app/main/ui/ds/notifications/toast.cljs
Xaviju: But according to the schema, it must always be a keyword.

`[:level {:optional true} [:maybe [:enum :info :warning :error :success]]]`

You mean to set this validation because of this component can be used in JavaScript?

frontend/src/app/main/ui/notifications.cljs
Xaviju: You mean deleting the full component definition? I'll check if it is not used anywhere

frontend/src/app/main/ui/ds/notifications/toast.cljs
Xaviju: I've added this validation for strings, but storybook (for instance) it works well using keywords as well.

niwinz: I mean, if you conde is always suppose to receive keyword, then  you don't need the `(keyword level)` because you are converting keyword to keyword... But as far as I understand this is done for make it more easy use from storybook, where it can receive strings there, in this case we need that check. It all depends if storybook is passing correct data or it passing strings for this.

frontend/src/app/main/ui/notifications.cljs
niwinz: Yes, on the first grep it looks unused.

niwinz (approved):

frontend/src/app/main/ui/ds/notifications/shared/notification_pill.scss
```diff
@@ -0,0 +1,68 @@
...
+  padding: var(--notification-padding);
+
+  display: flex;
+  gap: var(--sp-s);
+
+  color: var(--notification-fg-color);
+
+  &.type-toast {
+    padding-inline-end: var(--sp-xxxl);
+  }
```
belen-albeza: Let's avoid nested selectors

frontend/src/app/main/ui/ds/notifications/toast.cljs
```diff
@@ -9,37 +9,33 @@
...
-      [:> i/icon* {:icon-id icon-id :class (stl/css :icon)}]
-      children
+  [{:keys [class level type children on-close] :rest props}]
+  (let [class (dm/str class " " (stl/css-case :toast true))
+        level (if (string? level)
+                (keyword level)
+                (d/nilv level :info))
+        type (or type :context)
+        props (mf/spread-props props {:class class :role "alert"})]
+    [:> "div" props
```
belen-albeza: `<aside>`

frontend/src/app/main/ui/ds/notifications/toast.cljs
```diff
@@ -9,37 +9,33 @@
...
+     [:> notification-pill* {:level level :type type} children]
       ;; TODO: this should be a buttom from the DS, but this variant is not designed yet.
       ;; https://tree.taiga.io/project/penpot/task/8492
-      [:> "button" {:on-click on-close :aria-label "Close" :class (stl/css :close-button)} [:> i/icon* {:icon-id i/close}]]]]))
+     [:> "button" {:on-click on-close
+                   :aria-label "Close"
+                   :class (stl/css :close-button)}
```
belen-albeza, line 42: Not for this PR, but we should check with Design if/when they will be normalizing these kinds of buttons, so we can add the variant to the DS

frontend/src/app/main/ui/ds/notifications/toast.cljs
Xaviju: I changed it because in the [WAI ARIA patterns it recommends using a div element](https://www.w3.org/WAI/ARIA/apg/patterns/alert/examples/alert/)

Are you sure you want me to change it to `aside`?

frontend/src/app/main/ui/ds/notifications/toast.cljs
Xaviju: Its already commented and needs some design work to update the DS, so we kept it to avoid blocking this task.

frontend/src/app/main/ui/ds/notifications/toast.cljs
belen-albeza: Yes, from the WAI ARIA patterns website:

> The code in this example is not intended for production environments.
> [...]
> - Robust accessibility can be further optimized by choosing implementation patterns that [maximize use of semantic HTML](https://www.w3.org/TR/using-aria/#rule1) and heeding the warning that [No ARIA is better than Bad ARIA](https://www.w3.org/WAI/ARIA/apg/practices/read-me-first/#no_aria_better_bad_aria).

The `<aside>` element is used for content that is separated from the main content. Here the separation it's clear (the toasts live independently and are ephemeral). 

I also have doubts that we should have the `alert` aria role for non-error toasts, but I'll leave that to your own criteria.

Xaviju: The role alert is required, otherwise users with assistive technologies won't access to the information on the toast, while others do.

I'll make the changes

belen-albeza: I mentioned the role because of [this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role):

Screenshot 2025-02-11 at 9 32 58 AM

Not all toasts are "urgent" (that's why we have different levels), and there are [other roles for live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#roles_with_implicit_live_region_attributes) as well. 

That said, if you truly think `alert` is the most appropriate for all our toasts, that's OK. But if so, I would make the`aria-live` setting explicit and make non-errors be less intrusive than error toasts.

Xaviju:
Added a new commit with the changes required, please @belen-albeza could you review it again?

belen-albeza (approved):
Thanks @Xaviju !

PR #5760, opened 2025-02-04
:tada: Add variations POC

pabloalba:
Implements https://tree.taiga.io/project/penpot/task/10086

NOTE: To check this, you need to add `enable-feature-variants` to the PENPOT_FLAGS on backend and frontend, and then create a new team (new teams will have the feature activated)


frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -0,0 +1,229 @@
...
+(defn remove-property
+  "Remove the variant property on the position pos
+   in all the components with this variant-id"
+  [variant-id pos]
+  (ptk/reify ::remove-property
+    ptk/WatchEvent
+    (watch [it state _]
+      (let [page-id (:current-page-id state)
+            data    (dsh/lookup-file-data state)
+            objects    (dsh/lookup-page-objects state page-id)
```
niwinz: Avoid duplicate state lookups with the data already available on the context:

```clojure
(let [data    (dsh/lookup-file-data state)
      objects (-> (dsh/get-page data page-id)
                  (get :objects)
```

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -0,0 +1,229 @@
...
+
+(defn update-property-value
+  "Updates the variant property value on the position pos in a component"
+  [component-id pos value]
+  (ptk/reify ::update-property-value
+    ptk/WatchEvent
+    (watch [it state _]
+      (let [page-id   (:current-page-id state)
+            data      (dsh/lookup-file-data state)
+            objects   (dsh/lookup-page-objects state page-id)
```
niwinz: The same here

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -0,0 +1,229 @@
...
+            component (dm/get-in data [:components component-id])
```
niwinz: I think `app.common.types.component-list/get-component` helper should be used here

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -0,0 +1,229 @@
...
+            main-id   (:main-instance-id component)
+            properties (-> (:variant-properties component)
+                           (assoc-in [pos :value] value))
```
niwinz: Prefer `(update pos assoc :value value)`,

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -0,0 +1,229 @@
...
+
+(defn add-new-property
+  "Add a new variant property to all the components with this variant-id"
+  [variant-id]
+  (ptk/reify ::add-new-property
+    ptk/WatchEvent
+    (watch [it state _]
+      (let [page-id (:current-page-id state)
+            data    (dsh/lookup-file-data state)
+            objects    (dsh/lookup-page-objects state page-id)
```
niwinz: The same here

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -0,0 +1,229 @@
...
+
+            related-components (->> (:components data)
+                                    vals
+                                    (filter #(= (:variant-id %) variant-id))
+                                    reverse)
```
niwinz: Maybe this can be extracted to a separated, private helper on this namespace, looks it is repeated around

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -0,0 +1,229 @@
...
+   that component"
+  [id]
+  (ptk/reify ::transform-in-variant
+    ptk/WatchEvent
+    (watch [_ state _]
+      (let [variant-id (uuid/next)
+            new-component-id (uuid/next)
+            file-id      (:current-file-id state)
+            page-id      (:current-page-id state)
+            objects      (dsh/lookup-page-objects state page-id)
```
niwinz: Pass `file-id` here as second argument, for avoid a duplicate lookup of the current-file-id

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -0,0 +1,229 @@
...
+            main-id      (:id main)
+            undo-id  (js/Symbol)]
+
+
+        (rx/of
+         (dwu/start-undo-transaction undo-id)
+         (dwsh/create-artboard-from-selection variant-id)
+         (cl/remove-all-fills [variant-id] {:color clr/black :opacity 1})
+         (dwsl/create-layout-from-id variant-id :flex)
+         (dwsh/update-shapes [variant-id] #(assoc % :layout-item-h-sizing :auto
```
niwinz: `[variant-id]` is used many times, avoid creating vector instance many times defining it on let and reusing it

frontend/src/app/main/ui/workspace/sidebar/layer_name.cljs
```diff
@@ -38,6 +38,8 @@
 
         shape-for-rename (mf/deref lens:shape-for-rename)
 
+        shape-name       (or variant-name shape-name)
```
niwinz: `d/nilv`

frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
```diff
@@ -225,6 +230,81 @@
...
+(mf/defc component-variant*
+  {::mf/props :obj}
+  [{:keys [component shape libraries]}]
+  (let [id-component (:id component)
+        properties   (:variant-properties component)
+        data         (get-in libraries [(:component-file shape) :data])
+        variant-id   (:variant-id component)
+
+        related-components (->> data
+                                :components
```
niwinz: Repeated logic, can be reused from app.main.data.workspace.variants

frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
```diff
@@ -225,6 +230,81 @@
         (when (or editing? creating?)
           [:div {:class (stl/css  :counter)} (str size "/300")])]])))
 
+
+(mf/defc component-variant*
+  {::mf/props :obj}
+  [{:keys [component shape libraries]}]
+  (let [id-component (:id component)
+        properties   (:variant-properties component)
+        data         (get-in libraries [(:component-file shape) :data])
```
niwinz: Not an issue, just a note: we probably need to revisit this when feature is ready, because passing all around the whole libraries map has very bad implications on performance.

niwinz: So i'm not going to comment  more performance issues, because the whole component menu design is pretty bad with performance, so this should be addressed later, once the feature is complete or at least the sidebar menu part ready.

frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
```diff
@@ -643,5 +730,123 @@
...
             [:& component-annotation {:id id :shape shape :component component :rerender-fn rerender-fn}])
+
+          (when (and is-variant? (not swap-opened?) (not multi))
+            [:> component-variant* {:component component :shape shape :libraries libraries}])
+
           (when (dbg/enabled? :display-touched)
             [:div ":touched " (str (:touched shape))])])])))
+
+
+(mf/defc variant-menu
```
niwinz: New components should follow the new naming and call convention: name ends with `*` and instantiate with `:>` instead of `:&`  (also, no metada is required with that components)

frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
```diff
@@ -225,6 +230,81 @@
         (when (or editing? creating?)
           [:div {:class (stl/css  :counter)} (str size "/300")])]])))
 
+
+(mf/defc component-variant*
+  {::mf/props :obj}
```
niwinz: Minor: :fire:, not necessary when component follows the `*` naming convention, it is default for that components.

niwinz (changes requested):

common/src/app/common/logic/variants.cljc
```diff
@@ -0,0 +1,62 @@
...
+
+(defn generate-update-property-value
+  [changes component-id main-id pos value name]
+  (-> changes
+      (pcb/update-component component-id #(assoc-in % [:variant-properties pos :value] value))
+      (pcb/update-shapes [main-id] #(assoc % :variant-name name))))
+
+(defn generate-add-new-property
+  [changes related-components property-name]
+  (let [[_ changes]
```
niwinz, line 50: This can be implemented more efficiently using `loop`

common/src/app/common/logic/variants.cljc
```diff
@@ -0,0 +1,62 @@
...
+        (reduce (fn [[num changes] component]
```
niwinz, line 51: I think `loop` here can be more efficient than reduce

niwinz (approved):

common/src/app/common/logic/variants.cljc
niwinz: this does not blocks merge

common/src/app/common/logic/variants.cljc
```diff
@@ -0,0 +1,62 @@
...
+             #(assoc-in % [:variant-properties pos :name] new-name)))
+          changes
+          related-components))
+
+
+(defn generate-remove-property
+  [changes related-components pos]
+  (reduce (fn [changes component]
+            (let [props   (:variant-properties component)
+                  props   (vec (concat (subvec props 0 pos) (subvec props (inc pos))))
```
hirunatan, line 32: We have `d/remove-at-index` function, it would be clearer to use it.

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -0,0 +1,197 @@
...
+            undo-id  (js/Symbol)]
+        (rx/of
+         (dwu/start-undo-transaction undo-id)
+         (dch/commit-changes changes)
+         (dwu/commit-undo-transaction undo-id))))))
+
+(defn transform-in-variant
+  "Given the id of a main shape of a component, creates a variant structure for
+   that component"
+  [id]
```
hirunatan, line 158: I'd call this directly `main-id` (or better `main-instance-id`). The operation `transform-in-variant` does not have one id that is obviously the chosen one. In this case you are giving the main instance id of the component, but for example you could have used the component id. So the name `id` in short is ambiguous, better to be more explicit.

PR #5765, opened 2025-02-04
:recycle: Simplifies RPC pattern for tokens

### LINKED ISSUE #75: Tokens

floscr:
Closes https://github.com/tokens-studio/penpot/issues/75 of parent issue: https://github.com/tokens-studio/penpot/issues/66

### DIFF
--- a/common/src/app/common/files/changes.cljc
+++ b/common/src/app/common/files/changes.cljc
@@ -463,7 +463,7 @@
       [:type [:= :set-token]]
       [:set-name :string]
       [:token-name :string]
-      [:params [:maybe [:map [:token ::cto/token]]]]]]]])
+      [:token [:maybe ::cto/token]]]]]])
 
 (def schema:changes
   [:sequential {:gen/max 5 :gen/min 1} schema:change])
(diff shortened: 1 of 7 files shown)


common/src/app/common/files/changes.cljc
```diff
@@ -458,24 +458,12 @@
...
-      [:name :string]
-      [:token ::cto/token]]]
-
-    [:del-token
-     [:map {:title "DelTokenChange"}
-      [:type [:= :del-token]]
-      [:set-name :string]
-      [:name :string]]]]])
+      [:token-name :string]
+      [:params [:maybe [:map [:token ::cto/token]]]]]]]])
```
hirunatan: Why create a `params` map attribute and not directly having the token here?

floscr: We thought this was the common pattern, as other changes do it this way, but we can flatten it if you want
https://github.com/tokens-studio/penpot/blob/4913031a9a2c9dc7418b494d5802f583a1a5a833/common/src/app/common/files/changes.cljc#L637-L647

not-in-stock: It was done to maintain the similarly to previous API function signatures.
If having such structure isn’t necessary, we can change it to e.g. optional argument.

not-in-stock: @hirunatan 
Would it make sense to flatten the parameters structure for this method and any others with similar signature within the scope of this refactoring?

niwinz: I think we should flatten and pass `token` instead of `params {token}`
The pattern you mention is not directly applicable here because params on grid, means the grid-params (can be called `grid`  but is called `params` because we can assign partially object. But here, with tokens we should always assign a complete token instance

hirunatan: > @hirunatan Would it make sense to flatten the parameters structure for this method and any others with similar signature within the scope of this refactoring?

I don't think so. If we need to refactor other calls, better to do it in future changes (we need to review backward compatibility and perhaps other issues). But the tokens calls should not use the `params` name, in my opinion.

not-in-stock: @niwinz @hirunatan Fixed! ✅

PR #5766, opened 2025-02-04
:sparkles: Several changes to tokens 

niwinz:
Should be merged with "Merge Commit", please do NOT SQUASH


frontend/src/app/main/ui/workspace/tokens/token_pill.cljs
```diff
@@ -144,11 +144,13 @@
   {::mf/wrap-props false}
   [{:keys [on-click token theme-token full-applied on-context-menu half-applied selected-shapes token-type-props active-theme-tokens]}]
   (let [{:keys [name value errors]} token
-        is-reference (some #(= % "{") value)
+
+        is-reference? (str/includes? value "{")
```
hirunatan: Won't it be `reference?` without `is-`?

niwinz: It can be both, I mean, it is clear in both way

hirunatan: Yes, it's clear. But I think we tend to use only the ? as convention.

niwinz: I think this issue has near no relevance to the PR. It was `is-reference` and i changed it to `is-reference?` adding `?`
This code will be removed because i have plans to move this into the Token type because we should not calculate it every time.

And in any case, there are no exact convention that says that should be a **single-word and end with `?`**. So I think we are discussing something that is pretty irrelevant.

PR #5773, opened 2025-02-05
:bug: Fix several bugs for the combination of status and share on workspace

pabloalba:
Fix several bugs for the combination of status and share on workspace

### DIFF
--- a/frontend/src/app/main/ui/workspace/left_header.cljs
+++ b/frontend/src/app/main/ui/workspace/left_header.cljs
@@ -22,7 +22,11 @@
    [app.util.i18n :as i18n :refer [tr]]
    [app.util.keyboard :as kbd]
    [cuerdas.core :as str]
-   [rumext.v2 :as mf]))
+   [rumext.v2 :as mf]
+   [okulary.core :as l]))
+
+(def ref:persistence-status
+  (l/derived :status refs/persistence))
 
 ;; --- Header Component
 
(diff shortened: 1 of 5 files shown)


frontend/src/app/main/ui/workspace/left_header.cljs
```diff
@@ -108,6 +112,25 @@
          {:class (stl/css :file-name)
           :title file-name
           :on-double-click start-editing-name}
+          ;;-- Persistende state widget
+         [:div {:class (case (mf/deref ref:persistence-status)
```
niwinz: mf/deref is a react hook and should be statically defined in the let.
If this code on a future, for some unrelated reason is put under a conditional, it will break the app; hooks should be always defined statically.

frontend/src/app/main/ui/workspace/left_header.cljs
```diff
@@ -108,6 +112,25 @@
...
+                         :pending (stl/css :status-notification :pending-status)
+                         :saving (stl/css :status-notification :saving-status)
+                         :saved (stl/css :status-notification :saved-status)
+                         :error (stl/css :status-notification :error-status)
+                         (stl/css :status-notification))
+                :title (case (mf/deref ref:persistence-status)
```
niwinz: The same here

niwinz: Additionally you are derefing and watching the same value twice, which is also redundant operation

frontend/src/app/main/ui/workspace/left_header.cljs
```diff
@@ -108,6 +112,25 @@
...
+                         :pending (tr "workspace.header.saving")
+                         :saving (tr "workspace.header.saving")
+                         :saved (tr "workspace.header.saved")
+                         :error (tr "workspace.header.save-error")
+                         nil)}
+          (case (mf/deref ref:persistence-status)
```
niwinz: This is the third mf/deref of the same atom, and applies the same

PR #5784, opened 2025-02-06
:sparkles: Several performance oriented changes to tokens code (part 2)

niwinz:


frontend/src/app/main/ui/workspace/tokens/sets.cljs
```diff
@@ -445,23 +445,32 @@
          :on-edit-submit-set on-update-token-set
          :on-edit-submit-group on-update-token-set-group}])]))
 
+(mf/defc sets-list*
```
not-in-stock, line 512: Small question on the component naming. 
What does the asterisk at the end of the name mean?

floscr:
Here are some issues I found, I can test further when the first one is resolved.
I've tried to compare against early and develop so these seem to be introduced here / in the other PR


# New tokens dont get created in the selected set


## Steps to reproduce

1.  Create two sets
2.  Select the bottom set
3.  Create a token
4.  It gets stored in the first set although the last one is selected


# The token context menu doesnt close when editing a token


https://github.com/user-attachments/assets/3fe1ce14-c18a-4ace-9aca-17a872d86500


# Renaming doesnt clear renaming state

I think this only happens when renaming a set to be a nested group path


https://github.com/user-attachments/assets/f84d5894-f7e5-4aad-9a6a-49584f8750ac



## Steps to reproduce

1.  Create set `foo`
2.  Rename the set via context menu or double click to `foo/bar/baz`

frontend/src/app/main/ui/workspace/tokens/sidebar.cljs
```diff
@@ -307,6 +305,31 @@
...
+                                               (let [path (ctob/split-token-set-name name)]
+                                                 (= (peek path) selected-name))))
+                                     (first))]
+              (when match
+                (st/emit! (dt/set-selected-token-set-name match)))))
+
+          (let [match (->> (ctob/get-sets tokens-lib)
+                           (first)
+                           (:name))]
+            (st/emit! (dt/set-selected-token-set-name match))))))
```
floscr: This effect seems a bit much to resolve just for the undo case and on top of that being a guess (that could also be wrong if there's a set with matching name I think).
I would just fall back to the first set.

niwinz: Ok we can fallback to the first.

frontend/src/app/main/ui/workspace/tokens/sets.cljs
niwinz: It has two meanings:
1. Convention: a clear visual distinction between components and functions.
2. Call convention: makes it mandatory to use `:>` for instantiate that components to react elements and dictates how props are destructured. In this case it means `::mf/props :react`, so props are received as js object with react friendly casing (they can be destructured using clojure style, using kebab-kase, or as-is with camelCase, native to react style props). This allows destructure part of props, and pass the rest or transformed with (mf/spread-props...) to the underlying components. It also facilitates expose components to external systems like storybook and make it usable with JSX (which pases props using react camelCase style)

frontend/src/app/main/ui/workspace/tokens/sidebar.cljs
niwinz: Fixed

niwinz:
All reported issues fixed, thanks

floscr:
> All reported issues fixed, thanks 

Great! I'll test it again soon

frontend/src/app/main/ui/workspace/tokens/sidebar.cljs
```diff
@@ -307,6 +304,16 @@
...
+    (mf/with-effect [tokens-lib selected-token-set-name]
+      (when (and tokens-lib
+                 (or (nil? selected-token-set-name)
+                     (and selected-token-set-name
+                          (not (ctob/get-set tokens-lib selected-token-set-name)))))
+        (let [match (->> (ctob/get-sets tokens-lib)
+                         (first)
+                         (:name))]
+          (st/emit! (dt/set-selected-token-set-name match)))))
+
```
floscr, line 316: Does this have to be an effect? In previous versions we used a computed fallback where the it was something like `(or selected-token-set-name (:name (first sets))`.

An effect makes things more complicated to track down imo.

And it introduces a bug as well right now:
When you undo a set deletion the first set gets auto selected:


https://github.com/user-attachments/assets/1a6bdf7b-b204-4684-b333-0d0b46235441

Steps to reproduce:
1. Create two sets `a` `b`
2. Select `b`
3. Delete `a`
4. Undo

floscr:
Only found one more issue that is similar to the other one with the renaming:


https://github.com/user-attachments/assets/fbd07b84-9036-41f9-ab79-60a143d46ae6

Steps to reproduce:

1. Create  a nested set `a/b`
2. Rename set `a` as a nested set `a/c`

Otherwise couldn't find any issues :raised_hands:

frontend/src/app/main/ui/workspace/tokens/sidebar.cljs
niwinz: An effect is a good fallback when you start with no state, having the computed fallback makes a simple get operation have many dependencies, the current ephemeral state of selection, and the whole token lib.

And i'm not clearly understand the issue, right now on early:

- when you delete a selected set, no set becomes selected -> bug
- when you delete a not selected set, and then undo, the previously set comes back in a different order -> bug

On this PR:

- when you delete a selected set an other (the first one of the list) is automatically selected (I think this is a correct behavior)
- when you delete a non selected set, then it behaves in the same way as in early -> bug

niwinz:
> Only found one more issue that is similar to the other one with the renaming:
> 
>  2025-02-11-15-21.22.mp4 
> Steps to reproduce:
> 
> 1. Create  a nested set `a/b`
> 2. Rename set `a` as a nested set `a/c`
> 
> Otherwise couldn't find any issues 🙌

This issue is fixed

frontend/src/app/main/ui/workspace/tokens/sidebar.cljs
floscr: > An effect is a good fallback when you start with no state, having the computed fallback makes a simple get operation have many dependencies, the current ephemeral state of selection, and the whole token lib.

Yeah that true :+1: 

---

I just noticed that the bug from above is pre-existing, so I'll create a ticket for it.

floscr:
Thank you! I couldn't find any more issues and the changes look good to me as well :clap:

frontend/src/app/main/data/tokens.cljs
```diff
@@ -224,6 +231,7 @@
           (rx/of
            (drop-error (ex-data e))))))))
 
+;; FIXME: the the name is very confusing
```
hirunatan, line 236: I don't think the name is confusing... `update-create` is a common pattern for udating an object if it exists, or creating it if not...

niwinz: The name is clearly confusing, the common pattern is `create-or-update-<subject>`.
In any case this is already refactored and changed  on part3 PR

PR #5792, opened 2025-02-06
:sparkles: Enable Enter key for button activation in tokens modal

juanfran:
https://tree.taiga.io/project/penpot/issue/10062

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
@@ -28,6 +28,7 @@
    [app.main.ui.workspace.tokens.sets-context :as sets-context]
    [app.util.dom :as dom]
    [app.util.i18n :refer [tr]]
+   [app.util.keyboard :as k]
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
```diff
@@ -197,16 +198,26 @@
...
-                :type "button"
-                :on-click close-modal}
-    (tr "labels.cancel")]
-   [:> button* {:variant "primary"
-                :type "submit"
-                :on-click on-save-form
-                :disabled disabled?}
-    (tr "workspace.token.save-theme")]])
+  (let [handle-key-down
+        (mf/use-fn
```
niwinz, line 202: This use-fn (React.useCallback) has no real effect, because immediately afterwards you are calling it for generating new function instances.

You probably should define two separated handlers.

juanfran: done

PR #5803, opened 2025-02-07
:books: Add info for copy/paste options

myfunnyandy:
https://tree.taiga.io/project/penpot/task/9997


docs/user-guide/layer-basics/index.njk
```diff
@@ -204,6 +201,18 @@ title: 04· Layer basics
   </video>
 </figure>
 
+<h2 id="scale-elements">Copy CSS properties</h2>
+<p>To copy CSS properties from layers:</p>
+<ol>
+  <li>Select one or more layers.</li>
+  <li>Right click to show the layer menu.</li>
+  <li>Press <strong>Copy/Paste as... > Copy as CSS</strong> in case that you only want to get the CSS properties from the selected layer/s.</li>
+  <li>Press <strong>Copy/Paste as... > Copy as CSS (nested layers)</strong> in case that you only want to get the CSS properties from the selected layer/s and all the contained layers.</li>
```
madalenapmelo-kp: I would remove "that" from both sentences;
```suggestion
  <li>Press <strong>Copy/Paste as... > Copy as CSS (nested layers)</strong> in case you only want to get the CSS properties from the selected layer/s and all the contained layers.</li>
```

myfunnyandy: agree! this should be fixed already

madalenapmelo-kp (approved):

PR #5805, opened 2025-02-10
:sparkles: Mark all notifications as read from dashboard

mavalroot:
https://tree.taiga.io/project/penpot/us/9235

### DIFF
--- a/backend/src/app/rpc/commands/comments.clj
+++ b/backend/src/app/rpc/commands/comments.clj
@@ -797,3 +797,18 @@
                 {:id id}
                 {::db/return-keys false})
     nil))
+
+(def ^:private
+  schema:mark-all-threads-as-read
+  [:map {:title "mark-all-threads-as-read"}
+   [:threads [:vector ::sm/uuid]]])
+
+(sv/defmethod ::mark-all-threads-as-read
+  {::doc/added "1.15"
+   ::sm/params schema:mark-all-threads-as-read}
+  [cfg {:keys [::rpc/profile-id threads] :as params}]
+  (db/run!
+   cfg
+   (fn [{:keys [::db/conn]}]
+     (doseq [thread-id threads]
+       (upsert-comment-thread-status! conn profile-id thread-id)))))
--- a/backend/src/app/rpc/commands/comments.clj
+++ b/backend/src/app/rpc/commands/comments.clj
@@ -807,7 +807,7 @@
   {::doc/added "1.15"
    ::sm/params schema:mark-all-threads-as-read}
   [cfg {:keys [::rpc/profile-id threads] :as params}]
-  (db/run!
+  (db/tx-run!
    cfg
    (fn [{:keys [::db/conn]}]
      (doseq [thread-id threads]
--- a/frontend/src/app/main/data/comments.cljs
+++ b/frontend/src/app/main/data/comments.cljs
@@ -15,6 +15,7 @@
    [app.main.data.event :as ev]
    [app.main.data.helpers :as dsh]
    [app.main.data.team :as dtm]
+   [app.main.refs :as refs]
    [app.main.repo :as rp]
    [beicon.v2.core :as rx]
    [potok.v2.core :as ptk]))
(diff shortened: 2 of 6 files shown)


backend/src/app/rpc/commands/comments.clj
```diff
@@ -795,3 +795,18 @@
...
+(def ^:private
+  schema:mark-all-threads-as-read
+  [:map {:title "mark-all-threads-as-read"}
+   [:threads [:vector ::sm/uuid]]])
+
+(sv/defmethod ::mark-all-threads-as-read
+  {::doc/added "1.15"
+   ::sm/params schema:mark-all-threads-as-read}
+  [cfg {:keys [::rpc/profile-id threads] :as params}]
+  (db/run!
```
niwinz: I think this should be tx-run

frontend/src/app/main/data/comments.cljs
```diff
@@ -450,6 +451,22 @@
...
 
+(defn mark-all-threads-as-read
+  "Mark all threads as read"
+  [team-id on-success]
+  (ptk/reify ::mark-all-threads-as-read
+    ev/Event
+    (-data [_] {})
+    ptk/WatchEvent
+    (watch [_ _ _]
+      (let [threads (vals @refs/comment-threads)]
```
niwinz: Refs should not be used on events (at least as general advice). In this case, refs/comment-threads access the `:comment-threads` on the state. A watch event receives state as second argument so you can access the same data directly from state.

frontend/src/app/main/ui/dashboard/comments.cljs
```diff
@@ -63,7 +64,18 @@
...
-                         (with-meta {::ev/origin "dashboard"})))))]
+                         (with-meta {::ev/origin "dashboard"})))))
+
+        on-read-all
+        (mf/use-callback
+         (mf/deps team-id)
+         (fn []
+           (st/emit! (dcm/mark-all-threads-as-read
+                      team-id
+                      #(st/emit! (ntf/show {:level :info
```
niwinz: This can be directly handled on the `mark-all-threads-as-read` event and avoid this callback

niwinz (changes requested):

frontend/src/app/main/data/comments.cljs
```diff
@@ -454,6 +457,25 @@
...
+  [team-id]
+  (ptk/reify ::mark-all-threads-as-read
+    ev/Event
+    (-data [_] {})
+    ptk/WatchEvent
+    (watch [_ state _]
+      (let [threads (->> state :comment-threads vals)]
+        (->> (rp/cmd! :mark-all-threads-as-read {:threads (mapv :id threads)})
+             (rx/map #(retrieve-unread-comment-threads team-id))
+             (rx/tap #(st/emit! (ntf/show {:level :info
```
niwinz: `st/emit!` should not be used here. I think this can be a better approach for returning multiple events combined in one stream with the condition tha if the first fails, the second will be canceled:

```clojure
(let [threads (-> state :comment-threads vals)]
  (rx/concat
   (->> (rp/cmd! :mark-all-threads-as-read {:threads (mapv :id threads)})
        (rx/map #(retrieve-unread-comment-threads team-id))
        (rx/catch #(rx/throw {:type :comment-error})))
   (rx/of (ntf/show {:level :info
                     :type :toast
                     :content (tr "dashboard.mark-all-as-read.success")
                     :timeout 7000}))))
```

frontend/translations/en.po
```diff
@@ -6987,3 +6987,9 @@ msgstr "Autosaved versions will be kept for %s days."
 #, unused
 msgid "workspace.viewport.click-to-close-path"
 msgstr "Click to close the path"
+
+msgid "dashboard.mark-all-as-read.success"
+msgstr "Marked all notifications as read"
+
+msgid "labels.read"
```
niwinz: The translation strings under `label` namespace should be a bit more literal, because the represent the exact message. So I suggest using `label.mark-all-as-read` for this case

niwinz (changes requested):

frontend/src/app/main/data/comments.cljs
mavalroot: thanks for the code, I understand it a bit better now

PR #5806, opened 2025-02-10
:sparkles: Include active sets and themes in exported tokens json file

hirunatan:
Implements https://tree.taiga.io/project/penpot/task/9885

### DIFF
--- a/common/src/app/common/types/tokens_lib.cljc
+++ b/common/src/app/common/types/tokens_lib.cljc
@@ -1251,7 +1251,8 @@ Will return a value that matches this schema:
                                       (update "sets" keys)))))
           active-sets (get metadata "activeSets")
           active-themes (get metadata "activeThemes")
-          themes-data (if (empty? active-themes)
+          themes-data (if (and (seq active-sets)
+                               (empty? active-themes))
                         (conj themes-data {"name" hidden-token-theme-name
                                            "group" ""
                                            "description" nil
--- a/common/src/app/common/types/tokens_lib.cljc
+++ b/common/src/app/common/types/tokens_lib.cljc
@@ -26,15 +26,18 @@
 (def valid-groupable-item?
   (sm/validator schema:groupable-item))
 
-(def ^:private xf:clean-string
-  (comp (map str/trim)
-        (remove str/empty?)))
-
 (defn split-path
   "Decompose a string in the form 'one.two.three' into a vector of strings, removing spaces."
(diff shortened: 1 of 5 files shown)


common/src/app/common/types/tokens_lib.cljc
```diff
@@ -30,7 +30,7 @@
   "Decompose a string in the form 'one.two.three' into a vector of strings, removing spaces."
   [path separator]
   (let [xf (comp (map str/trim)
-                 (remove str/empty?))]
+                 #_(remove str/empty?))]
```
hirunatan: This causes that if a path has no groups, the name is put in the first element, and ends up being the "group" name. We need to think a way to avoid this.

common/src/app/common/types/tokens_lib.cljc
```diff
@@ -1222,16 +1225,21 @@ Will return a value that matches this schema:
                                                                           [s "enabled"])
                                                                         (into {})))))))))
                        (tree-seq d/ordered-map? vals themes))
+          ;; Active themes without exposing hidden penpot theme
+          active-themes-clear (disj active-themes hidden-token-theme-path)
```
hirunatan, line 1229: This is not necessary. Hidden theme is already filtered in line 1216

common/test/common_tests/types/tokens_lib_test.cljc
```diff
@@ -1241,7 +1241,8 @@
                                    "modified-at" now
                                    "name" "theme-1"
                                    "selectedTokenSets" {"core" "enabled"}}]
-                       "$metadata" {"tokenSetOrder" ["core"]}
+                       "$metadata" {"tokenSetOrder" ["core"]
+                                    "activeSets" #{},  "activeThemes" #{}}
                        "core"
                        {"colors" {"red" {"600" {"$value" "#e53e3e"
                                                 "$type" "color"
```
hirunatan, line 1269: Perhaps you could add some more tests to encode/decode files with active sets, both with or without non hidden themes.

common/src/app/common/types/tokens_lib.cljc
EvaMarco: But is being filtered from "themes", and that line removes the hidden theme from "active-themes"

hirunatan: Ok, my fault.

PR #5830, opened 2025-02-12
:bug: Fix Action Buttons size and add example to Storybook

elenatorro:
https://tree.taiga.io/project/penpot/issue/10118 

Fix:
- Add `--button-width`, `--button-height`, and `--button-padding` to the `%base-button` placeholder selector
- Apply 32px size to regular icon buttons and 24px size to the action variant, as reflected in the design
- Adjust the padding for the action variant
- Add example to Storybook

Action vs Ghost variants

image

image

Refactor:
- Remove `%base-button-action` placeholder selector since it's only being used for the icon button action variant for now

### DIFF
--- a/frontend/src/app/main/ui/ds/buttons/_buttons.scss
+++ b/frontend/src/app/main/ui/ds/buttons/_buttons.scss
@@ -20,14 +20,18 @@
   --button-focus-inner-ring-color: initial;
   --button-focus-outer-ring-color: initial;
 
+  --button-width: initial;
+  --button-height: #{$sz-32};
+
   appearance: none;
-  height: $sz-32;
-  border: none;
-  border-radius: $br-8;
+
+  width: var(--button-width);
+  height: var(--button-height);
 
   background: var(--button-bg-color);
   color: var(--button-fg-color);
   border: $b-1 solid var(--button-border-color);
+  border-radius: $br-8;
 
   &:hover {
     --button-bg-color: var(--button-hover-bg-color);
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/ui/ds/_sizes.scss
```diff
@@ -7,6 +7,8 @@
 @use "./utils.scss" as *;
 
 // TODO: create actual tokens once we have them from design
+$sz-2: px2rem(2);
+$sz-4: px2rem(4);
```
belen-albeza: These tokens are not needed when we remove the padding. Also, for reference:
- Size tokens are meant for things like `width`, `height`, etc. Paddings and margins are handled as spacing tokens (in `spacing.scss`)

frontend/src/app/main/ui/ds/buttons/_buttons.scss
```diff
@@ -20,14 +20,21 @@
   --button-focus-inner-ring-color: initial;
   --button-focus-outer-ring-color: initial;
 
+  --button-width: initial;
+  --button-height: #{$sz-32};
+  --button-padding: #{$sz-4};
```
belen-albeza: We don't need to set the padding if we set both `width` and `height` (the component has a `display:flex` that centers its contents), so we can remove this component token.

frontend/src/app/main/ui/ds/buttons/icon_button.scss
```diff
@@ -33,5 +36,24 @@
...
+
+  --button-focus-bg-color: transparent;
+  --button-focus-fg-color: var(--color-accent-primary);
+  --button-focus-inner-ring-color: transparent;
+  --button-focus-outer-ring-color: var(--color-accent-primary);
+
+  --button-width: #{$sz-24};
+  --button-height: #{$sz-24};
+
+  --button-padding: #{$sz-2};
```
belen-albeza: Let's remove this too

belen-albeza (changes requested):
Thanks @elenatorro ! Just a small fix: we don't actually need to set the padding for this component

frontend/src/app/main/ui/ds/_sizes.scss
elenatorro: Got it, thanks!!

belen-albeza (approved):

elenatorro:
@belen-albeza and I reviewed an issue together since it seems that when using `display: flex`, the svg icon was not correctly aligned to the center. It _sometimes_ applies flex-shrink to the svg, so following Belén's suggestion, we changed it to `display: grid` and now the icon is always properly centered.

PR #5835, opened 2025-02-12
:bug: Fix incorrect numbering files when new team

superalex:
https://tree.taiga.io/project/penpot/issue/10092

### DIFF
--- a/backend/src/app/rpc/commands/files.clj
+++ b/backend/src/app/rpc/commands/files.clj
@@ -384,8 +384,10 @@
           f.revn,
           f.vern,
           f.is_shared,
-          ft.media_id AS thumbnail_id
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/data/team.cljs
```diff
@@ -282,7 +282,11 @@
   [team]
   (ptk/reify ::team-created
     IDeref
-    (-deref [_] team)))
+    (-deref [_] team)
+
+    ptk/UpdateEvent
+    (update [_ state]
+      (dissoc state :files))))
```
niwinz: This is not a good fix, because it removes all the pre "cached" files

superalex: You are right <3

PR #5839, opened 2025-02-12
:bug: Fix problem with team permissions redirection

Alotor:
https://tree.taiga.io/project/penpot/issue/10119

### DIFF
--- a/frontend/src/app/main/data/team.cljs
+++ b/frontend/src/app/main/data/team.cljs
@@ -8,6 +8,7 @@
   (:require
    [app.common.data :as d]
    [app.common.data.macros :as dm]
+   [app.common.exceptions :as ex]
    [app.common.logging :as log]
    [app.common.schema :as sm]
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/data/team.cljs
```diff
@@ -118,8 +119,10 @@
       (let [team-id (:current-team-id state)
             teams   (get state :teams)
             team    (get teams team-id)]
-        (rx/of (set-current-team team)
-               (fetch-members))))))
+        (if (not team)
+          (rx/of (ex/raise :type :authentication))
```
niwinz: `rx/of` here is useless,. because ex/raise throws an exception.

I think this should be like `(when-not team (rx/raise...))`

Alotor: You're right. I'll chang it to `(rx/throw (ex/error...`

PR #5844, opened 2025-02-13
:recycle: Use constant for zero width space

luisddm:
Make the code more readable by using a constant for the [Zero-width space](https://en.wikipedia.org/wiki/Zero-width_space) character.

### DIFF
--- a/frontend/src/app/main/ui/comments.cljs
+++ b/frontend/src/app/main/ui/comments.cljs
@@ -43,6 +43,7 @@
 (def mentions-context (mf/create-context nil))
 (def r-mentions-split #"@\[[^\]]*\]\([^\)]*\)")
 (def r-mentions #"@\[([^\]]*)\]\(([^\)]*)\)")
+(def zero-width-space \u200B)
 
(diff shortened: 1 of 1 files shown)


belen-albeza:
How can we test this? (i.e. trigger the rendering of this). Is it by submitting an empty comment? Also, a question: why do we need this zero-width space? Is it a CSS thing? Thanks!

Alotor:
This is used in comments with mentions. 

The input for creating comments is a content editable that when a special character `@` is input it adds the mention as a `<span>` element.

The zero-width space is a "trick" so it's easier to handle the content editable. Mentions can be together with normal text and the handling them it's very tricky. Adding a "zero-width-space" surrounding the mentions eases it a lot because there is always some characters around it and you can check the limits. However, they need to be zero-space because otherwise the user will realize there is space.
 
As an example, imagine the following mention: `Hello @Belen, how are you?` there is a "zero-space" space between `n` and `,`.

Hope this clarifies.

belen-albeza (approved):
Thanks!

PR #5869, opened 2025-02-17
:tada: Add AB test for empty workspace set board tool by default

cocotime:
https://tree.taiga.io/project/penpot/us/9425?milestone=427670

### DIFF
--- a/frontend/src/app/main/data/workspace.cljs
+++ b/frontend/src/app/main/data/workspace.cljs
@@ -270,6 +270,16 @@
         (rx/of (dws/select-shapes frames-id)
                dwz/zoom-to-selected-shape)))))
 
+(defn- select-frame-tool
+  [page-id file-id]
+  (ptk/reify ::select-frame-tool
+    ptk/WatchEvent
+    (watch [_ state _]
+      (let [objects          (dsh/lookup-page-objects state file-id page-id)
+            is-page-empty?   (= (count objects) 1)]
+        (when is-page-empty?
+          (rx/of (dwd/select-for-drawing :frame)))))))
+
 (defn- fetch-bundle
   "Multi-stage file bundle fetch coordinator"
   [file-id]
@@ -463,7 +473,9 @@
       (if-let [page (dsh/lookup-page state file-id page-id)]
         (rx/of (initialize-page* file-id page-id page)
                (dwth/watch-state-changes file-id page-id)
-               (dwl/watch-component-changes))
+               (dwl/watch-component-changes)
+               (when (cf/external-feature-flag "boards-02" "test")
+                 (select-frame-tool page-id file-id)))
         (rx/of (dcm/go-to-workspace :file-id file-id ::rt/replace true))))))
 
 (defn finalize-page
--- a/frontend/src/app/main/data/workspace.cljs
+++ b/frontend/src/app/main/data/workspace.cljs
@@ -29,6 +29,7 @@
    [app.common.types.components-list :as ctkl]
    [app.common.types.container :as ctn]
    [app.common.types.file :as ctf]
+   [app.common.types.page :as ctp]
    [app.common.types.shape :as cts]
    [app.common.types.shape-tree :as ctst]
    [app.common.types.shape.layout :as ctl]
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/data/workspace.cljs
```diff
@@ -451,10 +452,13 @@
     ptk/WatchEvent
     (watch [_ state _]
       (if (dsh/lookup-page state page-id)
-        (let [file-id (:current-file-id state)]
+        (let [file-id (:current-file-id state)
+              objects        (deref refs/workspace-page-objects)
```
niwinz: Refs should almost never be used on events. Refs are derived atoms that points to state. But on events you already has the state.

In this case, you need `(dsh/lookup-page-objects state file-id page-id)`

frontend/src/app/main/data/workspace.cljs
```diff
@@ -451,10 +452,13 @@
...
+              is-page-empty? (= (count (keys objects)) 1)]
```
niwinz: Yo can execute `count` on objects directly, without the need to access the `(keys objects)`

frontend/src/app/main/data/workspace.cljs
```diff
@@ -451,10 +452,13 @@
...
           (rx/of (preload-data-uris page-id)
                  (dwth/watch-state-changes file-id page-id)
-                 (dwl/watch-component-changes)))
+                 (dwl/watch-component-changes)
+                 (when (and is-page-empty? (cf/external-feature-flag "boards-02" "test")) (dwd/select-for-drawing :frame))))
```
niwinz: I think that all this new code, can be encapsulated in a specific event and emit that event instead having all this logic executing all the time independently if the AB feature is activated or not.

niwinz (changes requested):

frontend/src/app/main/data/workspace.cljs
cocotime: fixed

frontend/src/app/main/data/workspace.cljs
cocotime: fixed

frontend/src/app/main/data/workspace.cljs
cocotime: I'm not sure if I did what you asked. I changed the code. Can you check if it's correct?

niwinz: Nice, yes, thanks

frontend/src/app/main/data/workspace.cljs
```diff
@@ -273,6 +273,17 @@
         (rx/of (dws/select-shapes frames-id)
                dwz/zoom-to-selected-shape)))))
 
+(defn select-frame-tool
+  [page-id]
```
niwinz: Pass the file-id that you already have on the other event

frontend/src/app/main/data/workspace.cljs
```diff
@@ -273,6 +273,17 @@
         (rx/of (dws/select-shapes frames-id)
                dwz/zoom-to-selected-shape)))))
 
+(defn select-frame-tool
```
niwinz: Use `defn-` for declare a private function

niwinz (changes requested):

frontend/src/app/main/data/workspace.cljs
cocotime: done

frontend/src/app/main/data/workspace.cljs
cocotime: done

frontend/src/app/main/data/workspace.cljs
```diff
@@ -273,6 +273,16 @@
         (rx/of (dws/select-shapes frames-id)
                dwz/zoom-to-selected-shape)))))
 
+(defn- select-frame-tool
+  [{:keys [page-id file-id]}]
```
niwinz: I think we don't need here a map for passing two static positional params. Avoid unnecesary allocation with a function with two arguments

```clojure
(defn- select-frame-tool
  [file-id page-id]
       ...
```

niwinz (changes requested):

PR #5888, opened 2025-02-18
:sparkles: Add more integration tests for tokens

juanfran:
https://tree.taiga.io/project/penpot/task/10096
https://tree.taiga.io/project/penpot/task/10097

### DIFF
--- a/frontend/playwright/ui/pages/WorkspacePage.js
+++ b/frontend/playwright/ui/pages/WorkspacePage.js
@@ -260,4 +260,9 @@ export class WorkspacePage extends BaseWebSocketPage {
   async clickTogglePalettesVisibility(clickOptions = {}) {
     await this.togglePalettesVisibility.click(clickOptions);
   }
+
+  async openTokenThemesModal(clickOptions = {}) {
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
```diff
@@ -130,6 +130,7 @@
              [:div {:class (stl/css :theme-row-right)}
               (if-let [sets-count (some-> theme :sets seq count)]
                 [:> button* {:class (stl/css :sets-count-button)
+                             :data-testid "edit-theme"
```
EvaMarco: This test id can be changed by "aria-labelledby" and an id on the text component down below. 
In the test we can use getbyrole, with the text content as a name.

frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
```diff
@@ -149,6 +150,7 @@
 
               [:> icon-button* {:on-click delete-theme
                                 :variant "ghost"
+                                :data-testid "delete-theme"
```
EvaMarco: we can use get by role and aria label as name, to avoid using get by test id

PR #5889, opened 2025-02-18
:bug: Fix library button condition and copy

cocotime:
https://tree.taiga.io/project/penpot/issue/10210

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/assets.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/assets.cljs
@@ -74,7 +74,7 @@
 (mf/defc assets-toolbox
   {::mf/wrap [mf/memo]
    ::mf/wrap-props false}
-  [{:keys [size]}]
+  [{:keys [size file-id]}]
   (let [components-v2  (mf/use-ctx ctx/components-v2)
         read-only?     (mf/use-ctx ctx/workspace-read-only?)
         filters*       (mf/use-state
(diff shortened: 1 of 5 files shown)


frontend/src/app/main/ui/workspace/sidebar/assets.cljs
```diff
@@ -90,6 +91,8 @@
         ordering       (:ordering filters)
         reverse-sort?  (= :desc ordering)
         num-libs       (count (mf/deref refs/libraries))
+        file           (mf/deref refs/file)
+        components     (ctkl/components (:data file))
```
niwinz: ctkl/components is an operation that not just returns components, it processes the map and returns the not-deleted components. This means that this function is candidate to use `mf/with-memo`.

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
```diff
@@ -90,6 +91,8 @@
         ordering       (:ordering filters)
         reverse-sort?  (= :desc ordering)
         num-libs       (count (mf/deref refs/libraries))
+        file           (mf/deref refs/file)
```
niwinz: The file is already available on refs/libraries, so I think we can reuse the libraries for access the file.
The current file id can be passed down from parent component that already has the file ( a file without data)

niwinz: I think we can implement further optimizations here i think but we can do it in other PR

niwinz (changes requested):

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
cocotime: done

frontend/src/app/main/ui/workspace/sidebar/assets.cljs
cocotime: done

PR #5890, opened 2025-02-18
:bug: Pass color format to css formatter

elenatorro:
Related tickets 

* https://tree.taiga.io/project/penpot/issue/9116

## Context

When going to the "INSPECT" bar and changing the fill color, it's expected to get the color in the selected format. Right now, the background CSS property is only being set as HEX, which is the default

## Changes

By passing the "format" value to the "copy-data" method, which holds the selected color format, we can parse it correctly in the color utils method, that returns the parsed color in the selected format.

Additionally, I've added a couple of roles and aria attributes to the elements to be able to test the flow properly using Playwright

## Demo

[screen-recorder-tue-feb-18-2025-15-41-52.webm](https://github.com/user-attachments/assets/1cd896e2-9cb6-4a8e-a49c-651819efa452)

## Next

While this works, I've seen a couple of things that can be fixed, but after talking with design, we're leaving them out of scope:

1. The `background:` CSS property is not updated in the code, but it's fine like this for now
2. The selected format is not saved. If the shape being inspected is not selected anymore, or if the tab changes, then the selected format goes back to HEX. We can work on preserve the state in future PRs

### DIFF
--- a/frontend/src/app/util/code_gen/style_css_formats.cljs
+++ b/frontend/src/app/util/code_gen/style_css_formats.cljs
@@ -53,28 +53,30 @@
     :else value))
 
 (defn format-color
-  [value _options]
-  (cond
-    (:image value)
-    (let [image-url (cfg/resolve-file-media (:image value))
-          opacity-color (when (not= (:opacity value) 1)
-                          (uc/gradient->css {:type :linear
-                                             :stops [{:color "#FFFFFF" :opacity (:opacity value)}
-                                                     {:color "#FFFFFF" :opacity (:opacity value)}]}))]
-      (if opacity-color
-        ;; CSS doesn't allow setting directly opacity to background image, we should add a dummy gradient to get it
-        (dm/fmt "%, url(%) no-repeat center center / cover" opacity-color image-url)
-        (dm/fmt "url(%) no-repeat center center / cover" image-url)))
+  [value options]
+  (let [format (or (:format options) :hex)]
+    (cond
+      (:image value)
+      (let [image-url (cfg/resolve-file-media (:image value))
+            opacity-color (when (not= (:opacity value) 1)
+                            (uc/gradient->css {:type :linear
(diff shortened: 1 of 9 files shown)


frontend/src/app/util/code_gen/style_css_formats.cljs
```diff
@@ -53,28 +53,30 @@
...
-        (dm/fmt "%, url(%) no-repeat center center / cover" opacity-color image-url)
-        (dm/fmt "url(%) no-repeat center center / cover" image-url)))
-
-    (not= (:opacity value) 1)
-    (uc/color->background value)
-
-    :else
-    (str/upper (:color value))))
+  [value options]
+  (let [format (or (:format options) :hex)]
```
niwinz: in this case i think it can be
```
(get options :format :hex)
```

frontend/src/app/util/code_gen/style_css_formats.cljs
```diff
@@ -53,28 +53,30 @@
...
+      (not= (:opacity value) 1)
+      (uc/color->format->background value format)
+
+      :else
+      (uc/color->format->background value format))))
 
 (defmethod format-value :color
   [_ value options]
-  (format-color value options))
+  (let [format (or (:format options) :hex)]
```
niwinz: The same here

niwinz (changes requested):
_

frontend/src/app/util/code_gen/style_css_formats.cljs
elenatorro: Awesome!! I'll update it, thanks for the tip

frontend/src/app/util/code_gen/style_css_formats.cljs
elenatorro: done!

EvaMarco (approved):

PR #5894, opened 2025-02-18
:sparkles: Add constraints calculation on WASM

Alotor:
- Calculates constraints inside the WASM code
- Added some geometry utilities and use the skia primitives for things such as Rects and Points
- Fix problem with clip_bounds when the parent is rotated/transformed.
- Send the type of the shape to the wasm

### DIFF
--- a/frontend/src/app/render_wasm/api.cljs
+++ b/frontend/src/app/render_wasm/api.cljs
@@ -106,8 +106,22 @@
   [clip-content]
   (h/call internal-module "_set_shape_clip_content" clip-content))
 
+(defn- translate-shape-type
+  [type]
+  (case type
+    :frame   0
+    :group   1
+    :bool    2
+    :rect    3
+    :path    4
+    :text    5
+    :circle  6
+    :svg-raw 7
+    :image   8))
+
 (defn set-shape-type
   [type {:keys [masked]}]
+  (h/call internal-module "_set_shape_type" (translate-shape-type type))
   (cond
     (= type :circle)
     (h/call internal-module "_set_shape_kind_circle")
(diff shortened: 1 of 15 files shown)


frontend/src/app/render_wasm/api.cljs
```diff
@@ -106,8 +106,22 @@
   [clip-content]
   (h/call internal-module "_set_shape_clip_content" clip-content))
 
+(defn- translate-shape-type
+  [type]
+  (case type
```
superalex, line 111: Probably we should include some doc about this in  render-wasm/docs/serialization.md

render-wasm/src/math.rs
```diff
@@ -1,4 +1,337 @@
...
+    origin: Point,
+    direction: Vector,
+}
+
+impl Ray {
+    pub fn new(origin: Point, direction: Vector) -> Self {
+        Self { origin, direction }
+    }
+
+    // pub fn new_points(from: Point, to: Point) -> Self {
```
superalex: Should we keep this code?

render-wasm/src/render.rs
```diff
@@ -273,27 +271,40 @@ impl RenderState {
...
-        }
+        if let Some((bounds, transform)) = clip_bounds {
+            self.drawing_surface.canvas().concat(&transform);
+            self.drawing_surface
+                .canvas()
+                .clip_rect(bounds, skia::ClipOp::Intersect, true);
 
-        let center = shape.bounds().center();
+            // Uncomment this to debug the bounds for the clip
+            /*
```
superalex: Could we enable this just in debug mode instead of leaving it commented?

render-wasm/src/shapes.rs
```diff
@@ -339,10 +388,36 @@ impl Shape {
         self.hidden
     }
 
-    pub fn bounds(&self) -> math::Rect {
+    // TODO: Maybe store this inside the shape
```
superalex, line 391: Are we leaving the TODO here?

Alotor: Yes, why not? I think it's a good idea to keep todos for possible future improvements even if we don't do them now.

render-wasm/src/render.rs
Alotor: ok

PR #5897, opened 2025-02-19
:wrench: Add filtering to logs in component sync algorithm

hirunatan:
For big files, allow the possibility of showing debugging traces of the components synchronization process only for a subtree of shapes (to avoid having thousands of traces).


niwinz (approved):
Ready to merge

PR #5908, opened 2025-02-19
:books: Document how to use a proxy

yamila-moreno:
This PR adds examples of reverse proxy with nginx and caddy server. They are not meant as comprehensive examples about nginx or caddy (or how a proxy works), just to illustrate how to proxy Penpot in a docker deployment.

To test it:
- run penpot with docker compose as explained in the documentation
- install nginx locally and use the given configuration - test it works as expected (including exporting and websockets)
- stop nginx, install caddy and use the configuration - test it works as expected (including exporting and websockets)

### DIFF
--- a/docs/technical-guide/getting-started.md
+++ b/docs/technical-guide/getting-started.md
@@ -280,6 +280,55 @@ Postgres database and another one for the assets uploaded by your users (images 
 clips). There may be more volumes if you enable other features, as explained in the file
 itself.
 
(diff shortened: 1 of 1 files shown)


niwinz:
I think the examples we provide handle correctly websocket connections: https://nginx.org/en/docs/http/websocket.html

bameda (approved):
Works perfectly!!

PR #5910, opened 2025-02-19
:bug: Don't send invitation email to profiles that explicitly disallow that

niwinz:
https://tree.taiga.io/project/penpot/issue/10252

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -70,7 +70,7 @@ is a number of cores)
 - Fix rename locked boards [Taiga #10174](https://tree.taiga.io/project/penpot/issue/10174)
 - Fix update-libraries dialog disappear when clicking outside [Taiga #10238](https://tree.taiga.io/project/penpot/issue/10238)
 - Fix incorrect handling of team access requests with deleted/recreated users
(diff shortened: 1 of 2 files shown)


superalex:
But with this solution if:
- User A has email notifications disabled
- User B is trying to invite User A

In the interface when User B creates the invitation it doesn't appear in the interface or no information is sent about what is happening (because the invitation is not created)

niwinz:
Now, invitation is created but email not sent

CHANGES.md
```diff
@@ -70,6 +70,7 @@ is a number of cores)
 - Fix rename locked boards [Taiga #10174](https://tree.taiga.io/project/penpot/issue/10174)
 - Fix update-libraries dialog disappear when clicking outside [Taiga #10238](https://tree.taiga.io/project/penpot/issue/10238)
 - Fix incorrect handling of team access requests with deleted/recreated users
+- Fix inccorect handling of profile settings related to invitation notifications [Taiga #10252](https://tree.taiga.io/project/penpot/issue/10252)
```
superalex: Small type inccorect -> incorrect

niwinz: Fixed

PR #5914, opened 2025-02-20
:tada: Tile rendering

AzazelN28:
This PR provides a rendering system that allows documents to be rendered in patches called tiles.

Those tiles allows us to cache parts of the document so we can rerender parts faster.

image


render-wasm/_build_env
```diff
@@ -28,8 +27,13 @@ _CARGO_PARAMS="--target=wasm32-unknown-emscripten";
 
 if [ "$_BUILD_MODE" = "release" ]; then
     _CARGO_PARAMS="--release $_CARGO_PARAMS"
+    EMCC_CFLAGS="-Os $EMCC_FLAGS"
 else
-    EMCC_CFLAGS="$EMCC_CFLAGS -sMALLOC=emmalloc-debug"
+    # TODO: Extra parameters that would be could to look into:
```
superalex: @AzazelN28 may be we should review this comment

render-wasm/src/shapes.rs
```diff
@@ -656,12 +685,17 @@ impl Shape {
         let width = bounds.width();
         let height = bounds.height();
 
-        self.selrect = math::Rect::from_xywh(
+        // NOTA: Viendo las trazas esto no hace nada, no está
```
superalex: We should clean these comments

render-wasm/src/shapes.rs
```diff
@@ -656,12 +685,17 @@ impl Shape {
...
+        // transformando el selrect.
+        let new_selrect = math::Rect::from_xywh(
             center.x - width / 2.0,
             center.y - height / 2.0,
             width,
             height,
         );
+        // let new_selrect = self.selrect.clone();
```
superalex: We should clean these comments

render-wasm/src/main.rs
```diff
@@ -169,11 +138,12 @@ pub extern "C" fn use_shape(a: u32, b: u32, c: u32, d: u32) {
 
 #[no_mangle]
 pub unsafe extern "C" fn set_parent(a: u32, b: u32, c: u32, d: u32) {
-    let state = unsafe { STATE.as_mut() }.expect("Got an invalid state pointer");
-    let id = uuid_from_u32_quartet(a, b, c, d);
-    if let Some(shape) = state.current_shape() {
-        shape.set_parent(id);
-    }
+    with_state!(state, {
```
elenatorro, line 141: Minor thing: we can use `with_current_shape` macro directly when manipulating the current shape

render-wasm/src/shapes.rs
```diff
@@ -574,7 +601,9 @@ impl Shape {
     }
 
     pub fn is_recursive(&self) -> bool {
-        !matches!(self.shape_type, Type::SVGRaw(_))
+        matches!(self.shape_type, Type::Frame(_))
+            || matches!(self.shape_type, Type::Group(_))
+            || matches!(self.shape_type, Type::Bool(_))
```
elenatorro:
```suggestion
        matches!(self.shape_type, Type::Frame(_) | Type::Group(_) | Type::Bool(_))
```

render-wasm/src/main.rs
superalex: Thanks!

render-wasm/src/shapes.rs
superalex: Thanks!

elenatorro (approved):

PR #5926, opened 2025-02-21
:tada: Add translations to token errors

EvaMarco:
This PR closes this task https://tree.taiga.io/project/penpot/task/9734

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/errors.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/errors.cljs
@@ -1,48 +1,48 @@
 (ns app.main.ui.workspace.tokens.errors
   (:require
+   [app.util.i18n :refer [tr]]
    [cuerdas.core :as str]))
 
 (def error-codes
   {:error.import/json-parse-error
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/ui/workspace/tokens/errors.cljs
```diff
@@ -1,48 +1,48 @@
 (ns app.main.ui.workspace.tokens.errors
   (:require
+   [app.util.i18n :refer [tr]]
    [cuerdas.core :as str]))
 
 (def error-codes
   {:error.import/json-parse-error
    {:error/code :error.import/json-parse-error
-    :error/message "Import Error: Could not parse json"}
+    :error/message (tr "workspace.token.error-parse")}
```
niwinz: This translations will not work correctly, because they are defined on top level. This means this tr is executed the application loading once, and will never be executed again, so if you change language, they will not reflect never the changed language.

EvaMarco: ohhh ok, i'll check it.

PR #5928, opened 2025-02-21
:sparkles: Remove menu options and shortcuts actions over variants

pabloalba:
Implements https://tree.taiga.io/project/penpot/task/10130

### DIFF
--- a/frontend/src/app/main/data/workspace/bool.cljs
+++ b/frontend/src/app/main/data/workspace/bool.cljs
@@ -11,6 +11,7 @@
    [app.common.files.helpers :as cph]
    [app.common.geom.shapes :as gsh]
    [app.common.svg.path.shapes-to-path :as stp]
+   [app.common.types.component :as ctc]
    [app.common.types.container :as ctn]
    [app.common.types.shape :as cts]
    [app.common.types.shape.layout :as ctl]
@@ -99,7 +100,10 @@
              shapes (->> ordered-indexes
                          (map (d/getf objects))
                          (remove cph/frame-shape?)
-                         (remove #(ctn/has-any-copy-parent? objects %)))]
+                         (remove #(ctn/has-any-copy-parent? objects %))
+                         (remove #(->> %
+                                       (get objects)
+                                       (ctc/is-variant?))))]
 
          (when-not (empty? shapes)
            (let [[boolean-data index] (create-bool-data bool-type name (reverse shapes) objects)
--- a/frontend/src/app/main/data/workspace/bool.cljs
+++ b/frontend/src/app/main/data/workspace/bool.cljs
@@ -100,10 +100,8 @@
              shapes (->> ordered-indexes
                          (map (d/getf objects))
                          (remove cph/frame-shape?)
-                         (remove #(ctn/has-any-copy-parent? objects %))
-                         (remove #(->> %
-                                       (get objects)
-                                       (ctc/is-variant?))))]
+                         (remove ctc/is-variant?)
+                         (remove #(ctn/has-any-copy-parent? objects %)))]
 
          (when-not (empty? shapes)
            (let [[boolean-data index] (create-bool-data bool-type name (reverse shapes) objects)
--- a/frontend/src/app/main/data/workspace/groups.cljs
+++ b/frontend/src/app/main/data/workspace/groups.cljs
@@ -188,7 +188,9 @@
             (->> ids
                  (cfh/clean-loops objects)
                  (remove #(ctn/has-any-copy-parent? objects (get objects %)))
-                 (shapes-for-grouping objects))
+                 (shapes-for-grouping objects)
+                 (remove #(ctk/is-variant? %)))
+
             parents  (into #{} (map :parent-id) shapes)]
         (when-not (empty? shapes)
           (let [[group changes]
--- a/frontend/src/app/main/data/workspace/groups.cljs
+++ b/frontend/src/app/main/data/workspace/groups.cljs
@@ -189,7 +189,7 @@
                  (cfh/clean-loops objects)
                  (remove #(ctn/has-any-copy-parent? objects (get objects %)))
                  (shapes-for-grouping objects)
-                 (remove #(ctk/is-variant? %)))
+                 (remove ctk/is-variant?))
 
             parents  (into #{} (map :parent-id) shapes)]
         (when-not (empty? shapes)
--- a/frontend/src/app/main/data/workspace/shape_layout.cljs
+++ b/frontend/src/app/main/data/workspace/shape_layout.cljs
@@ -172,8 +172,7 @@
             is-mask?        (and single? has-mask?)
             has-component?  (some true? (map ctc/instance-root? selected-shapes))
             is-component?   (and single? has-component?)
-
-            has-variant?    (some true? (map ctc/is-variant? selected-shapes))
+            has-variant?    (some ctc/is-variant? selected-shapes)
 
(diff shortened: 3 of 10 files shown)


frontend/src/app/main/data/workspace/bool.cljs
```diff
@@ -99,7 +100,10 @@
              shapes (->> ordered-indexes
                          (map (d/getf objects))
                          (remove cph/frame-shape?)
-                         (remove #(ctn/has-any-copy-parent? objects %)))]
+                         (remove #(ctn/has-any-copy-parent? objects %))
+                         (remove #(->> %
+                                       (get objects)
```
niwinz: Is this correct, i mean, on the line 101 you already resolve the id to shape, so this `get  objects` looks redundant and incorrect

frontend/src/app/main/data/workspace/groups.cljs
```diff
@@ -188,7 +188,9 @@
             (->> ids
                  (cfh/clean-loops objects)
                  (remove #(ctn/has-any-copy-parent? objects (get objects %)))
-                 (shapes-for-grouping objects))
+                 (shapes-for-grouping objects)
+                 (remove #(ctk/is-variant? %)))
```
niwinz: Don't need to create an anonymous function, just pass `ctk/is-variant?` to remove.

frontend/src/app/main/data/workspace/shape_layout.cljs
```diff
@@ -173,46 +173,55 @@
             has-component?  (some true? (map ctc/instance-root? selected-shapes))
             is-component?   (and single? has-component?)
 
+            has-variant?    (some true? (map ctc/is-variant? selected-shapes))
```
niwinz: Can this be expressed just like `(some ctc/is-variant? selected-shapes)` ?

frontend/src/app/main/data/workspace/transforms.cljs
```diff
@@ -968,7 +968,10 @@
      (watch [_ state _]
        (let [objects   (dsh/lookup-page-objects state)
              selected  (or ids (dsh/lookup-selected state {:omit-blocked? true}))
-             shapes    (map #(get objects %) selected)
+             shapes    (->> selected
+                            (map #(get objects %))
```
niwinz: use `(d/getf objects)`

frontend/src/app/main/ui/workspace/context_menu.cljs
```diff
@@ -317,11 +317,12 @@
...
   [{:keys [shapes]}]
-  (let [multiple?    (> (count shapes) 1)
-        single?      (= (count shapes) 1)
+  (let [multiple?       (> (count shapes) 1)
+        single?         (= (count shapes) 1)
 
-        objects      (deref refs/workspace-page-objects)
-        any-in-copy? (some true? (map #(ctn/has-any-copy-parent? objects %) shapes))
+        objects         (deref refs/workspace-page-objects)
+        any-in-copy?    (some true? (map #(ctn/has-any-copy-parent? objects %) shapes))
```
niwinz: This can be defined as `(some #(ctn/has-any-copy-parent? objects %) shapes)` ?

frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
```diff
@@ -839,13 +838,14 @@
...
-             [:div {:key (str (:id shape) (:name property)) :class (stl/css :variant-property-row)}
-              [:> input-with-values* {:name (:name property) :values val :on-blur (partial update-property-name pos)}]
-              [:> icon-button* {:variant "ghost"
-                                :aria-label (tr "workspace.shape.menu.remove-variant-property")
-                                :on-click (partial remove-property pos)
-                                :icon "remove"
-                                :disabled (<= (count properties) 1)}]]))]]])))
+        (when-not multi
+          [:*
+           (for [[pos property] (map vector (range) properties)]
```
niwinz: We already have the `(d/enumerate)` for this use case

niwinz: But instead of enumerating on each rerender, i think this should be handled when properties is defined.
Firstly, `properties` should be using `with-memo`
The index can be calcultated directly on that operation, so you don't need to iterate over it again for enumeration:

```clojure
(mf/with-memo [data objects variant-id]
  (->> (dwv/find-related-components data objects variant-id)
       (mapcat :variant-properties)
       (group-by :name)
       (map-indexed (fn [index [k v]]
                      {:name k
                       :pos index
                       :values (distinct
                                (map #(if (str/empty? (:value %)) "--" (:value %)) v))}))))
```
Son enumerating it here, just use `(:pos prop)` when you need position

frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
```diff
@@ -839,13 +838,14 @@
...
+             (let [val (str/join ", " (:values property))]
+               [:div {:key (str (:id shape) (:name property)) :class (stl/css :variant-property-row)}
+                [:> input-with-values* {:name (:name property) :values val :on-blur (partial update-property-name pos)}]
+                [:> icon-button* {:variant "ghost"
+                                  :aria-label (tr "workspace.shape.menu.remove-variant-property")
+                                  :on-click (partial remove-property pos)
```
niwinz: This partial creates a new callback on each render. Should be avoided. There are approaches that I have explained  in several sessions on how to handle this cases.

frontend/src/app/main/ui/workspace/sidebar/options/shapes/frame.cljs
```diff
@@ -71,58 +71,59 @@
...
+                         :shape shape}]
+
+     [:& component-menu {:shapes [shape]}]
+
+     (when is-variant?
+       [:> variant-menu* {:shapes [shape]}])
+
+     [:& layout-container-menu
+      {:type shape-type
+       :ids [(:id shape)]
```
niwinz: This should use the `ids` binding instead of creating a new vector on each render

frontend/src/app/main/ui/workspace/sidebar/options/shapes/frame.cljs
```diff
@@ -71,58 +71,59 @@
...
+                     :values layer-values}]
+     [:> measures-menu* {:ids ids
+                         :values measure-values
+                         :type shape-type
+                         :shape shape}]
+
+     [:& component-menu {:shapes [shape]}]
+
+     (when is-variant?
+       [:> variant-menu* {:shapes [shape]}])
```
niwinz: I think we should reuse the `[shape]`, avoiding creating a new vector with a single shape on each rerender, twice.

niwinz (changes requested):

frontend/src/app/main/ui/ds/controls/input_with_values.cljs
```diff
@@ -39,11 +39,10 @@
...
          (fn [event]
-           (let [new-name (dom/get-target-val event)]
-             (dom/stop-propagation event)
-             (reset! editing* false)
-             (when on-blur
-               (on-blur new-name)))))
+           (dom/stop-propagation event)
+           (reset! editing* false)
+           (when on-blur
+             (on-blur event))))
```
niwinz, line 45: This looks breaking change. I mean, all component usage is expecting name and not event. Are we sure that all usage is modified for handle this change?

PR #5931, opened 2025-02-21
:lipstick: Improve theme modal with UX enhancements

Xaviju:
This modal enhances tokens theme modals UX as requested in [tg-10279](https://tree.taiga.io/project/penpot/issue/10279)

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
@@ -37,10 +37,13 @@
   (let [create-theme
         (mf/use-fn
          (mf/deps set-state)
-         #(set-state (fn [_] {:type :create-theme})))]
+         #(set-state (fn [_] {:type :create-theme})))
+        close-modal
+        (mf/use-fn
+         #(st/emit! (modal/hide)))]
     [:div {:class (stl/css :themes-modal-wrapper)}
      [:> heading* {:level 2 :typography "headline-medium" :class (stl/css :themes-modal-title)}
-      (tr "workspace.token.themes")]
+      (tr "workspace.token.themes-list")]
      [:div {:class (stl/css :empty-themes-wrapper)}
       [:div {:class (stl/css :empty-themes-message)}
        [:> text* {:as "span" :typography "title-medium" :class (stl/css :empty-theme-title)}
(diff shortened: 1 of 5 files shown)


frontend/translations/en.po
```diff
@@ -6577,7 +6577,7 @@ msgstr "Add a theme (i.e. Light)"
 
 #: src/app/main/ui/workspace/tokens/modals/themes.cljs:56
 msgid "workspace.token.new-theme"
-msgstr "New theme"
+msgstr "Add new theme"
```
EvaMarco: There is a problem with editing an old translation. 
This string was created with a sentence “new-theme” and it was related to the id “workspace.token.new-theme”.
And users could start translating it, in fact this string has already been translated to 10 languages. 
Now you change the string, but you don't change the id, so those 10 lang are going to be translated with the old meaning. 
To avoid this you must create a new pair id - tr

frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
```diff
@@ -81,19 +88,25 @@
          (fn [e]
            (dom/prevent-default e)
            (dom/stop-propagation e)
-           (set-state (fn [_] {:type :create-theme}))))]
+           (set-state (fn [_] {:type :create-theme}))))
+
+        close-modal
+        (mf/use-fn (fn []
```
EvaMarco: Is not an error, and can be ignored, but this can be writen with an annonimous fn  #(st/emit...)

frontend/translations/en.po
Xaviju: Changed from `new-theme` to `add-theme`

frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
Xaviju: Fixed as suggested

frontend/translations/en.po
```diff
@@ -6652,7 +6652,15 @@ msgstr "Theme %s"
 
 #: src/app/main/ui/workspace/tokens/modals/themes.cljs:43, src/app/main/ui/workspace/tokens/modals/themes.cljs:88
 msgid "workspace.token.themes"
-msgstr "Themes"
+msgstr "Themes list"
```
EvaMarco, line 6651: The same happens with this translation

EvaMarco (approved):

PR #5936, opened 2025-02-24
:tada: Add AB test descriptive board tooltip

cocotime:
https://tree.taiga.io/project/penpot/us/9426?milestone=431051

### DIFF
--- a/frontend/src/app/main/data/workspace/thumbnails.cljs
+++ b/frontend/src/app/main/data/workspace/thumbnails.cljs
@@ -296,6 +296,6 @@
              ;; find a better way to handle this.
              (->> notifier-s
                   (rx/take 1)
-                  (rx/of (dwc/set-workspace-visited)))
+                  (rx/map dwc/set-workspace-visited))
 
              (rx/take-until stopper-s))))))
--- a/frontend/src/app/main/data/workspace/thumbnails.cljs
+++ b/frontend/src/app/main/data/workspace/thumbnails.cljs
@@ -14,6 +14,7 @@
    [app.main.data.changes :as dch]
    [app.main.data.helpers :as dsh]
    [app.main.data.persistence :as-alias dps]
+   [app.main.data.workspace.common :as dwc]
    [app.main.data.workspace.notifications :as-alias wnt]
    [app.main.rasterizer :as thr]
    [app.main.refs :as refs]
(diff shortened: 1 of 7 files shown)


frontend/src/app/main/data/workspace/thumbnails.cljs
```diff
@@ -291,4 +292,10 @@
                    (rx/mapcat #(into #{} %))
                    (rx/map #(update-thumbnail file-id page-id % "frame" "watch-state-changes"))))
 
+             ;; WARNING: This is a workaround for an AB test, in case we consolidate this change we should
+             ;; find a better way to handle this.
+             (->> notifier-s
+                  (rx/take 1)
+                  (rx/of (dwc/set-workspace-visited)))
```
niwinz: This code is not doing what you think is doing: it creates a stream taking 1 from notifier-s, and then creates a stream with rx/of where first argument is the event `set-workspace-visited` and second argument is the previous stream.

What you really want to do is emit an event once the first and only notification from X arrives, in this case the code would be:

```clojure
(->> notifier-s
     (rx/take 1)
     (rx/map dwc/set-workspace-visited))
```

frontend/src/app/main/ui/workspace/top_toolbar.cljs
```diff
@@ -120,7 +121,11 @@
         (mf/use-fn
          (fn [event]
            (dom/blur! (dom/get-target event))
-           (st/emit! (dwc/toggle-toolbar-visibility))))]
+           (st/emit! (dwc/toggle-toolbar-visibility))))
+
+        profile (mf/deref refs/profile)
+        props   (get profile :props)
+        test-tooltip-board-text (if (and (cf/external-feature-flag "boards-03" "test") (not (:workspace-visited props))) (tr "workspace.toolbar.frame-first-time" (sc/get-tooltip :draw-frame)) (tr "workspace.toolbar.frame" (sc/get-tooltip :draw-frame)))]
```
niwinz: Consider formatting the code a bit better for make it more clear to read.

Example, following the same style that is already used in this component:

```clojure
(let [...
      {:keys [props]}
      (mf/deref refs/profile)

      button-title
      (if (and (cf/external-feature-flag "boards-03" "test") 
               (not (:workspace-visited props)))
        (tr "workspace.toolbar.frame-first-time" (sc/get-tooltip :draw-frame)) 
        (tr "workspace.toolbar.frame" (sc/get-tooltip :draw-frame)))]
```

niwinz (changes requested):

frontend/src/app/main/data/workspace/thumbnails.cljs
cocotime: changed

frontend/src/app/main/ui/workspace/top_toolbar.cljs
cocotime: changed

PR #5940, opened 2025-02-24
:sparkles: Add dbg panel to enable a team feature flag

elenatorro:
https://tree.taiga.io/project/penpot/issue/10344

### DIFF
--- a/backend/src/app/http/debug.clj
+++ b/backend/src/app/http/debug.clj
@@ -430,6 +430,50 @@
      ::yres/body    "OK"}))
 
 
+(defn- add-team-feature
+  [{:keys [params] :as request}]
+  (let [team-id (some-> params :team-id d/parse-uuid)
+        feature (some-> params :feature d/read-string)
+        debug   (contains? params :debug)]
+
+    (when-not (contains? params :force)
+      (ex/raise :type :validation
+                :code :missing-force
+                :hint "missing force checkbox"))
+
+    (when (nil? team-id)
+      (ex/raise :type :validation
+                :code :invalid-team-id
+                :hint "provided invalid team id"))
+
+    (srepl/enable-team-feature! team-id feature :debug debug)
+
+    {::yres/status  200
+     ::yres/headers {"content-type" "text/plain"}
+     ::yres/body    "OK"}))
+
+(defn- remove-team-feature
+  [{:keys [params] :as request}]
+  (let [team-id (some-> params :team-id d/parse-uuid)
+        feature (some-> params :feature d/read-string)
+        debug   (contains? params :debug)]
+
+    (when-not (contains? params :force)
+      (ex/raise :type :validation
+                :code :missing-force
+                :hint "missing force checkbox"))
+
+    (when (nil? team-id)
+      (ex/raise :type :validation
+                :code :invalid-team-id
+                :hint "provided invalid team id"))
+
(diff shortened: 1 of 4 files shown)


backend/src/app/srepl/helpers.clj
```diff
@@ -85,6 +88,46 @@
       WHERE file_id = ANY(?)
         AND id IS NOT NULL")
 
+(defn enable-team-feature!
+  [team-id feature & {:keys [debug] :or {debug false}}]
+  (when (and (not debug) (not (contains? cfeat/supported-features feature)))
```
niwinz: I think we can rename the `debug` to something more clear: `skip-check`

backend/src/app/http/debug.clj
```diff
@@ -430,6 +430,50 @@
      ::yres/body    "OK"}))
 
 
+(defn- add-team-feature
+  [{:keys [params] :as request}]
+  (let [team-id (some-> params :team-id d/parse-uuid)
+        feature (some-> params :feature d/read-string)
```
niwinz: I think we should not use d/read-string, because this interprets it as clojure data structure:

```clojure
user=> (d/read-string "foo/bar")
foo/bar
user=> (type (d/read-string "foo/bar"))
clojure.lang.Symbol
```
So here we just need to take the value as is.

backend/src/app/srepl/helpers.clj
```diff
@@ -85,6 +88,46 @@
...
+    (ex/raise :type :assertion
+              :code :feature-not-supported
+              :hint (str "feature '" feature "' not supported")))
+
+  (let [team-id (h/parse-uuid team-id)
+        feature (str feature)]
```
niwinz: This step is not necessary, I think is better to have an assertion for string than coercing all to string

backend/src/app/srepl/helpers.clj
```diff
@@ -85,6 +88,46 @@
...
+
+(defn disable-team-feature!
+  [team-id feature & {:keys [debug] :or {debug false}}]
+  (when (and (not debug) (not (contains? cfeat/supported-features feature)))
+    (ex/raise :type :assertion
+              :code :feature-not-supported
+              :hint (str "feature '" feature "' not supported")))
+
+  (let [team-id (h/parse-uuid team-id)
+        feature (str feature)]
```
niwinz: The same here

niwinz (changes requested):
:heart: Very nice, thanks!

I have put some minor suggestions on code.

backend/src/app/http/debug.clj
elenatorro: Got it!! I think that's the reason why the disjoint didn't work on the `disable-team-feature` and I had to cast it there, let me fix that!

backend/src/app/http/debug.clj
```diff
@@ -433,8 +433,8 @@
 (defn- add-team-feature
   [{:keys [params] :as request}]
   (let [team-id (some-> params :team-id d/parse-uuid)
-        feature (some-> params :feature d/read-string)
-        debug   (contains? params :debug)]
+        feature (some-> params :feature str)
```
niwinz, line 436: Minor: the `:feature` is already string so `str` is not necessary,

`(some-> params :feature)` is ok

PR #5946, opened 2025-02-25
:sparkles: Downgrade s3 sdk for compatibility with minio

niwinz:


niwinz:
The check is failing because github is unable to take the new action code for check this PR, so I need to make two separate requests. Lets merge it with failing check just for skip the github limitation.

PR #5951, opened 2025-02-25
:bug: Expand padding/margin values on apply token and show 'Mixed' placeholder

juanfran:
https://tree.taiga.io/project/penpot/issue/10171


EvaMarco (changes requested):
Applying margin to "All" should not expand margin values.

EvaMarco (approved):

PR #5958, opened 2025-02-25
:bug: Fix comment update fails in viewer

luisddm:
Fixes Taiga issue [#8271](https://tree.taiga.io/project/penpot/issue/8271)

### DIFF
--- a/frontend/src/app/main/data/comments.cljs
+++ b/frontend/src/app/main/data/comments.cljs
@@ -14,7 +14,6 @@
    [app.common.uuid :as uuid]
    [app.main.data.event :as ev]
    [app.main.data.helpers :as dsh]
-   [app.main.data.team :as dtm]
    [app.main.repo :as rp]
    [beicon.v2.core :as rx]
    [potok.v2.core :as ptk]))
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/data/comments.cljs
```diff
@@ -411,12 +410,8 @@
     ptk/WatchEvent
     (watch [_ state _]
       (let [share-id (-> state :viewer-local :share-id)]
-        (rx/merge
-         (->> (rp/cmd! :get-comment-threads {:file-id file-id :share-id share-id})
-              (rx/map comment-threads-fetched))
-
-         ;; Refresh team members
-         (rx/of (dtm/fetch-members)))))))
```
pabloalba, line 419: I don't like it very much... That removes the fetch team members also when this is called from the workspace.

Maybe we can do it with a `(when (:current-team-id state)` ?

luisddm: Per #5957, we have now that parameter loaded in the state of the viewer. Also, I think this is a quite obscure way to distinguish between modes. Maybe we can use any other parameter in the state, a more explicit one, to condition this fetching?

luisddm: I suggest using the presence of `:workspace-local` in the state.

pabloalba: Works for me

pabloalba (approved):

PR #5963, opened 2025-02-26
:bug: Fix problem when creating layouts

Alotor:
https://tree.taiga.io/project/penpot/issue/10243

### DIFF
--- a/common/src/app/common/types/shape/layout.cljc
+++ b/common/src/app/common/types/shape/layout.cljc
@@ -1666,3 +1666,12 @@
                 (-> cells
                     (d/update-when id d/patch-object (get-data target-cells id))))
               source-cells))))))
+
+(defn toggle-fix-if-auto
+  "Changes the sizing to fix if it's fill"
+  [shape]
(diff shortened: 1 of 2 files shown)


common/src/app/common/types/shape/layout.cljc
```diff
@@ -1666,3 +1666,12 @@
                 (-> cells
                     (d/update-when id d/patch-object (get-data target-cells id))))
               source-cells))))))
+
+(defn toggle-fix-if-auto
+  "Changes the sizing to fix if it's fill"
+  [shape]
+  (-> shape
```
pabloalba: Why the double cond?

(cond-> shape
    (= (:layout-item-h-sizing shape) :fill)
        (assoc :layout-item-h-sizing :fix)
    (= (:layout-item-v-sizing shape) :fill)
        (assoc :layout-item-v-sizing :fix)))

Alotor: Fixed

PR #5970, opened 2025-02-26
:bug: Fix rendering order of inner shadows when shape has no fills (wasm)

belen-albeza:
Closes https://tree.taiga.io/project/penpot/task/10199

Screenshot 2025-02-26 at 4 40 09 PM

### DIFF
--- a/render-wasm/src/render.rs
+++ b/render-wasm/src/render.rs
@@ -187,7 +187,7 @@ impl RenderState {
         );
     }
 
-    pub fn apply_drawing_to_render_canvas(&mut self) {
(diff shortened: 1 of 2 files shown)


superalex:
Thank you very much @belen-albeza

I've tested it and I think it's not working properly (I did a test with the different stroke/fills combinations and different kinds of inner/center/outer stroke):

Production:



Hourly:



PR:



I've included a zipped penpot file example
[test.penpot.zip](https://github.com/user-attachments/files/19003114/test.penpot.zip)

PR #5981, opened 2025-02-28
:bug: Fix relative position application for flex children

### LINKED ISSUE #81: bug: Dimension tokens applied to X or Y are not appied relative to containers but to the position on the canvas
### Steps To Reproduce

- Create a flex layout with some elements inside
- Place an element inside with an absolute position
- In the design tab, change values of X or Y to see how the object is positioned relative to the container layout. X positions the element at a distance relative to the left side of the container, and Y positions it relative to the top of the container.
- Apply a dimension token to the X property of this element.
- The token is applied relative to the center of the whole canvas, instead of relative to the container box.

---
with regular boards:
- create a board 
- place an element inside it
- apply a dimension token to the X property of this element.

### Expected behavior

The token should be applied relative to the container box, as penpot does when the X and Y values are changed manually.

This also applies to regular boards with elements inside. The X and Y position should be calculated relative to the board container.

### Actual behavior

- The token is applied relative to the center of the canvas.

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

not-in-stock:
Fixes dimension token application to a shape that is a child of flex layout.

**Before**
Application of a token changes position of a shape relatively to canvas.
**After**
Application of a token changes position of shape relatively to flex container, if shape is a child of it.

### DIFF
--- a/frontend/src/app/main/data/workspace/transforms.cljs
+++ b/frontend/src/app/main/data/workspace/transforms.cljs
@@ -859,10 +859,11 @@
           (rx/of (reorder-selected-layout-child direction))
           (rx/of (nudge-selected-shapes direction shift?)))))))
 
-(defn- get-relative-delta [position bbox parent]
-  (let [parent-bbox (-> parent :points grc/points->rect)
+(defn- get-relative-delta [position bbox frame]
+  (let [frame-bbox (-> frame :points grc/points->rect)
         relative-cpos (gpt/subtract (gpt/point (:x bbox) (:y bbox))
-                                    (gpt/point (:x parent-bbox) (:y parent-bbox)))
+                                    (gpt/point (:x frame-bbox)
+                                               (:y frame-bbox)))
         cpos (gpt/point (:x relative-cpos) (:y relative-cpos))
         pos (gpt/point (or (:x position) (:x relative-cpos))
                        (or (:y position) (:y relative-cpos)))]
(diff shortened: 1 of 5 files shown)


hirunatan:
The change looks good to me. But this addresses only the case of a shape inside a layout with fixed position. The general case of shapes inside boards is still failing as in https://github.com/tokens-studio/penpot/issues/81. Are you planning to fix it in other PR?

not-in-stock:
> The change looks good to me. But this addresses only the case of a shape inside a layout with fixed position. The general case of shapes inside boards is still failing as in [tokens-studio#81](https://github.com/tokens-studio/penpot/issues/81). Are you planning to fix it in other PR?

@hirunatan Thanks for pointing out. I'll amend this PR with changes regarding other cases.

hirunatan:
I'm not sure, but perhaps will be enough to have the change always relative, whether is a fixed element or not, and do it against the `frame-id` instead of the parent.

not-in-stock:
@hirunatan 
Alright, I've changed it to work this way in https://github.com/penpot/penpot/pull/5981/commits/f1ec10955c7d0913e7fbb9c7ee0123bd5997a3a0

frontend/src/app/main/data/workspace/transforms.cljs
```diff
@@ -847,12 +847,20 @@
           (rx/of (reorder-selected-layout-child direction))
           (rx/of (nudge-selected-shapes direction shift?)))))))
 
+(defn- get-relative-delta [position bbox parent]
```
hirunatan: I'll call the third parameter `frame`, because this not the direct parent, but the nearest frame (the direct parent may be a group, for example).

hirunatan: Same for `parent-bbox` below

PR #5982, opened 2025-02-28
:wrench: Add render-wasm performance tests setup

elenatorro:
Related to https://tree.taiga.io/project/penpot/task/9719


.circleci/config.yml
```diff
@@ -161,6 +161,37 @@ jobs:
             yarn run playwright install --with-deps chromium
             yarn run test:e2e -x --workers=4
 
+  test-performance:
```
elenatorro, line 164: I'm not running them for now on CI since I think it's not needed until we have more complex test comparison (and not just a rect), but we can [see it run successfully](https://app.circleci.com/pipelines/github/penpot/penpot/26436/workflows/b86029b2-9374-4929-b490-e2de2ef48ad6/jobs/47102)

Issue #5984, opened 2025-03-01
feature: Automatic Incremental component naming convention


Kamal007OLica:
### **Concern: Automatic Incremental Component Naming Convention**

The issue arises from duplicate component names (e.g., **Component-01** appearing twice), causing confusion. To resolve this, an **automatic incremental naming convention** can ensure unique and meaningful names for each component.

![Image](https://github.com/user-attachments/assets/3856451c-8357-493f-9278-d5b3055a07cc)

### **Proposed Solution**

1. **Unique Naming**:
   - Assign unique identifiers like `Component-01`, `Component-02`, etc.

2. **Incremental Logic**:
   - Automatically increment the numeric suffix for new components (e.g., `Component-01`, `Component-02`).

3. **Meaningful Names**:
   - Add context (e.g., `Button-01`, `Header-02`) for clarity.

4. **Error Prevention**:
   - Block duplicate names and display warnings.

### **Benefits**
- **Reduced Confusion**: Unique names make components easier to identify.
- **Improved Workflow**: Faster component management and referencing.
- **Scalability**: Handles growing numbers of components without conflicts.

---

### **Implementation Example**
1. **Initial State**:
   - `Component-01`, `Component-02`

2. **New Component**:
   - System names it `Component-03`.

3. **Result**:
   - No duplicates; each component is uniquely identifiable.

PR #5985, opened 2025-03-01
Incremental-Component naming-feature Addressing Issue #5984

Kamal007OLica:
### What’s New?
- Added automatic incremental naming for components (e.g., `Component-01`, `Component-02`).
- Prevented duplicate component names.
- Added contextual naming (e.g., `Button-01`, `Header-02`).

### Why This Matters?
- Reduces confusion caused by duplicate component names.
- Improves workflow efficiency and scalability.

### How to Test?
1. Create a new component and check if the name is auto-generated.
2. Try renaming a component to an existing name and verify that it’s blocked.

### Tests Added
- Backend unit tests for `generate-component-name` and `is-name-unique?`.
- Frontend tests for component name input.

### DIFF
--- a/backend/src/app/main.clj
+++ b/backend/src/app/main.clj
@@ -613,3 +613,9 @@
       (ex/print-throwable cause)
       (px/sleep 500)
       (System/exit -1))))
+
+
+(defn generate-component-name [component-type]
+  (let [existing-names (get-existing-component-names)
+        next-id (inc (count existing-names))]
+    (str component-type "-" (format "%02d" next-id))))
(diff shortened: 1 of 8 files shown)


miryamgduque:
Hi, @Kamal007OLica!
We really appreciate your collaboration, but it turns out this issue conflicts with something we are working on: variants. We still don't know what the final solution will be; we may even end up doing it this way, but until we are sure, we will not make any changes in this regard. This process may take a few months, so I'm going to reject the MR.

Thanks again 😊

Cheers,

PR #5992, opened 2025-03-03
:wrench: Do not use global static mut variables when possible

elenatorro:
https://tree.taiga.io/project/penpot/task/10314

## Context

We’re planning to upgrade to newer versions of rust and keep them as much up to date as possible. In order to accomplish this, one of the changes we need to make is not use anymore global static mutable variables. As explained in the following post, this warning will be moved to "deny" level:

https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html

```sh
error: creating a mutable reference to mutable static is discouraged
let buffer = unsafe { BUFFERU8.take() }.expect("uninitialized buffer");
                      ^^^^^^^^^^^^^^^ mutable reference to mutable static
```

#### Optional `lint` script

I've configured a lint script that includes the warnings we're currently allowing. It isn't running on CI yet, but the goal is to use it in the future to improve our code and facilitate migration to newer versions.

#### `BUFFERU8`

`BUFFERU8 → Mutex` (mem.rs)

In this case we can simply make the variable immutable and use a Mutex.

#### `State`

In this case, we can’t simply use a Mutex due to this limitation from the rust-skia bindings library:

```
25  | static STATE: Mutex<Option<Box<State>>> = Mutex::new(None);
	|           	^^^^^^^^^^^^^^^^^^^^^^^^^ `RCHandle<skia_bindings::SkSVGDOM>` cannot be sent between threads safely
```

Therefore, after some research, I thought using thread_local along with RefCell could help us in this scenario. It’s a way of manipulating a global static value ensuring it’s thread-safe, although for our case this is not a current issue since our approach is single-threaded.

However, I had some issues related to the lifetime access to `current_shape`  that I couldn't manage to solve, so in order to unblock this PR, we've decided to move forward with this warning and investigate it in the future.

What we decided to keep are the `with_sate` and `with_current_shape` macros to manipulate the STATE, which improves readability and extensibility, and we're enforcing in `main` to always access State properties through getters and setters.

### DIFF
--- a/render-wasm/lint
+++ b/render-wasm/lint
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+# Enable debugging if the script is run with --debug
+if [[ "$1" == "--debug" ]]; then
+    set -x
+fi
+
+. ./_build_env
+
+ALLOWED_RULES="
+        -A clippy::box_collection \
+        -A clippy::clone_on_copy \
+        -A clippy::derivable_impls \
+        -A clippy::enum_variant_names \
+        -A clippy::field_reassign_with_default \
+        -A clippy::from_over_into \
+        -A clippy::len_zero \
+        -A clippy::manual_map \
+        -A clippy::map_entry \
+        -A clippy::missing_safety_doc \
+        -A clippy::missing_transmute_annotations \
+        -A clippy::needless_borrow \
+        -A clippy::needless_borrows_for_generic_args \
+        -A clippy::needless_range_loop \
+        -A clippy::needless_return \
+        -A clippy::redundant_closure \
+        -A clippy::redundant_field_names \
+        -A clippy::single_match \
+        -A clippy::slow_vector_initialization \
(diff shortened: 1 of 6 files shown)


render-wasm/lint
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+
+# Enable debugging if the script is run with --debug
+if [[ "$1" == "--debug" ]]; then
+    set -x
+fi
+
+export SKIA_BINARIES_URL="https://github.com/penpot/skia-binaries/releases/download/0.80.1-1/skia-binaries-9e7d2684a17084095aef-wasm32-unknown-emscripten-gl-svg-textlayout.tar.gz"
```
superalex: this url is also defined in the _build_env it would be great if we could define it in just one place

elenatorro: sure! let me refactor this 👍🏽

belen-albeza (approved):

PR #6032, opened 2025-03-07
:sparkles: Create variant from component with path

pabloalba:
Implements https://tree.taiga.io/project/penpot/task/10444

### DIFF
--- a/frontend/src/app/main/data/workspace/variants.cljs
+++ b/frontend/src/app/main/data/workspace/variants.cljs
@@ -10,6 +10,7 @@
    [app.common.data :as d]
    [app.common.data.macros :as dm]
    [app.common.files.changes-builder :as pcb]
+   [app.common.files.helpers :as cfh]
    [app.common.logic.libraries :as cll]
    [app.common.logic.variants :as clv]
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -211,40 +212,68 @@
...
+
+         ;; When the component has path, set the path items as properties values
+         (when (> (count cpath) 1)
+           (rx/from
+            (map
+             #(update-property-value component-id % (nth cpath (inc %)))
+             (range num-props))))
+
+         (rx/of
+          (add-new-variant main-instance-id)
```
hirunatan, line 275: I think an event should not call another event this way.

It's best to have all logic in `generate-xxx` functions and that the events limit to gather things from the global state, call the function to create the changes, and commit them in a transaction.

This way the `generate-transform-in-variant` for example could call `generate-add-new-variant` without having nested undo transactions or having to read again all info from global state.

hirunatan: Ok, I see this needs a big refactor in other places. Better to do it in a different PR.

PR #6033, opened 2025-03-07
:sparkles: Add a default appearance to context notifications

Xaviju:
This PR adds a new `default` status to context notification component as requested by 
[tg-10392](https://tree.taiga.io/project/penpot/issue/10392)

### DIFF
--- a/frontend/src/app/main/ui/ds/notifications/context-notification.stories.jsx
+++ b/frontend/src/app/main/ui/ds/notifications/context-notification.stories.jsx
@@ -40,7 +40,7 @@ export default {
   render: ({ ...args }) => <ContextNotification {...args} />,
 };
 
-export const Default = {};
+export const Base = {};
 
 export const WithLongerText = {
   args: {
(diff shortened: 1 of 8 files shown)


frontend/src/app/main/ui/ds/notifications/context-notification.stories.jsx
```diff
@@ -30,7 +30,7 @@ export default {
     isHtml: false,
     type: "context",
     appearance: "neutral",
-    level: "info",
```
hirunatan, line 33: Shouldn't be interesting to also add a new Story for the default level, next to the Info, Error, etc. stories below?

hirunatan:
Looking at the design file it seems that the same thing should be done with the Toast notification component, could it be? It also has a Default level.

frontend/src/app/main/ui/ds/notifications/context-notification.stories.jsx
Xaviju: You are right, otherwise might be confusing

Xaviju:
Added a new commit with the suggested changes, please review it @hirunatan

hirunatan (approved):

PR #6045, opened 2025-03-10
:recycle: Simplifies RPC pattern for token sets

### LINKED ISSUE #73: Set

not-in-stock:
Closes https://github.com/tokens-studio/penpot/issues/73 of parent issue: https://github.com/tokens-studio/penpot/issues/66

### DIFF
--- a/common/src/app/common/files/changes_builder.cljc
+++ b/common/src/app/common/files/changes_builder.cljc
@@ -908,6 +908,32 @@
                                       :token nil}))
         (apply-changes-local))))
 
+(defn set-token-set [changes set-name group? token-set]
+  (assert-library! changes)
+  (let [library-data (::library-data (meta changes))
+        prev-token-set (some-> (get library-data :tokens-lib)
+                               (ctob/get-set set-name))]
+    (-> changes
+        (update :redo-changes conj {:type :set-token-set
+                                    :set-name set-name
+                                    :token-set token-set
+                                    :group? group?})
+        (update :undo-changes conj (if prev-token-set
+                                     {:type :set-token-set
+                                      :set-name (or
+                                                 ;; Undo of edit
+                                                 (:name token-set)
+                                                 ;; Undo of delete
+                                                 set-name)
+                                      :token-set prev-token-set
+                                      :group? group?}
+                                     ;; Undo of create
+                                     {:type :set-token-set
+                                      :set-name set-name
+                                      :token-set nil
+                                      :group? group?}))
+        (apply-changes-local))))
+
 (defn add-component
   ([changes id path name new-shapes updated-shapes main-instance-id main-instance-page]
    (add-component changes id path name new-shapes updated-shapes main-instance-id main-instance-page nil nil nil))
--- a/common/src/app/common/files/changes_builder.cljc
+++ b/common/src/app/common/files/changes_builder.cljc
@@ -803,13 +803,6 @@
         (update :undo-changes conj {:type :add-token-theme :token-theme prev-token-theme})
         (apply-changes-local))))
 
-(defn add-token-set
-  [changes token-set]
-  (-> changes
-      (update :redo-changes conj {:type :add-token-set :token-set token-set})
-      (update :undo-changes conj {:type :del-token-set :name (:name token-set)})
-      (apply-changes-local)))
-
 (defn rename-token-set-group
   [changes set-group-path set-group-fname]
   (let [undo-path (ctob/replace-last-path-name set-group-path set-group-fname)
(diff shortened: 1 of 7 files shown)


common/src/app/common/files/changes_builder.cljc
```diff
@@ -908,6 +908,32 @@
...
+    (-> changes
+        (update :redo-changes conj {:type :set-token-set
+                                    :set-name set-name
+                                    :token-set token-set
+                                    :group? group?})
+        (update :undo-changes conj (if prev-token-set
+                                     {:type :set-token-set
+                                      :set-name (or
+                                                 ;; Undo of edit
+                                                 (:name token-set)
```
hirunatan, line 895: Shouldn't this be `(:name prev-token-set)`?

The provided token set may already have the new name.

not-in-stock: Good catch, thanks!

not-in-stock: Well, if case of using the `:name` of `prev-token-set` application of changes would create a duplicate set in the undo-library

hirunatan:
I get an error when renaming a set from the left sidebar:
```
Internal Assertion Error
errors.cljs:48 Error: Call (with-library-data) before using this function
  at new cljs$core$ExceptionInfo (core.cljs:11778:11)
  at Function.eval [as cljs$core$IFn$_invoke$arity$3] (core.cljs:11810:5)
  at Object.app$common$data$macros$runtime_assert [as runtime_assert] (macros.cljc:130:17)
  at Object.app$common$files$changes_builder$assert_library_BANG_ [as assert_library_BANG_] (changes_builder.cljc:148:3)
  at Object.app$common$files$changes_builder$set_token_set [as set_token_set] (changes_builder.cljc:882:4)
  at Object.eval [as potok$v2$core$WatchEvent$watch$arity$3] (tokens.cljs:157:26)
```

not-in-stock:
> I get an error when renaming a set from the left sidebar:
> 
> ```
> Internal Assertion Error
> errors.cljs:48 Error: Call (with-library-data) before using this function
>   at new cljs$core$ExceptionInfo (core.cljs:11778:11)
>   at Function.eval [as cljs$core$IFn$_invoke$arity$3] (core.cljs:11810:5)
>   at Object.app$common$data$macros$runtime_assert [as runtime_assert] (macros.cljc:130:17)
>   at Object.app$common$files$changes_builder$assert_library_BANG_ [as assert_library_BANG_] (changes_builder.cljc:148:3)
>   at Object.app$common$files$changes_builder$set_token_set [as set_token_set] (changes_builder.cljc:882:4)
>   at Object.eval [as potok$v2$core$WatchEvent$watch$arity$3] (tokens.cljs:157:26)
> ```

Alright I'll take a look

common/src/app/common/files/changes_builder.cljc
hirunatan: I don't understand. Why would it create a duplicate set?

hirunatan:
I've found that in the e2e tests the rename does not fail because it's run in release mode and the assertions are removed. I think that the issue here is the assert checks a condition that if not true will break the undo, but the undo is not being checked in the test.

common/src/app/common/files/changes_builder.cljc
not-in-stock: This is what those test assertions catch:
https://github.com/penpot/penpot/pull/6045/files#diff-103c6efec858c177e024e9d7c9234cda76606c130eadfa61e6201516096c93b6R187-R188
For the library with applied undo changes (`undo-lib` in the test code):

- The first assertion checks that after changing the set name, we can still find a token in the set with its old name.
- The second assertion checks the opposite - that we cannot find that token in the set with the new name.

When the suggested change to `(:name prev-token-set)` is made, the second assertion fails, showing that the token can still be found in the set with the new name in the `undo-lib`, which is incorrect. This is what I referred to as a duplicate - the token can be found in both sets (by both the new name and old name).

not-in-stock:
> I've found that in the e2e tests the rename does not fail because it's run in release mode and the assertions are removed. I think that the issue here is the assert checks a condition that if not true will break the undo, but the undo is not being checked in the test.

Yeah, it looks like my recent commits broke it, so I'll check when exactly it happened

common/src/app/common/files/changes_builder.cljc
not-in-stock: And the way the name is handled is almost the same as in `set-token`:
https://github.com/penpot/penpot/blob/develop/common/src/app/common/files/changes_builder.cljc#L884-L909

hirunatan: Ah, I understand. In the undo call, you tell that the modified set (with the new name) is changed back to be equal to the previous set (with the old name). It's correct, then.

PR #6051, opened 2025-03-10
:tada: Add variants properties to inspect panel

pabloalba:
Implements https://tree.taiga.io/project/penpot/task/10129

### DIFF
--- a/common/src/app/common/logic/variants.cljc
+++ b/common/src/app/common/logic/variants.cljc
@@ -60,6 +60,11 @@
                                                :value v}) remaining)]
     (into assigned new-properties)))
 
+(defn- dashes-to-end
+  [property-values]
+  (let [dashes (if (some #(= % "--") property-values) ["--"] [])]
+    (concat (remove #(= % "--") property-values) dashes)))
+
 
 (defn extract-properties-values
   [data objects variant-id]
@@ -68,8 +73,10 @@
        (group-by :name)
        (map (fn [[k v]]
               {:name k
-               :values (distinct
-                        (map #(if (str/empty? (:value %)) "--" (:value %)) v))}))))
+               :value (->> v
+                           (map #(if (str/empty? (:value %)) "--" (:value %)))
+                           distinct
+                           dashes-to-end)}))))
 
 
 (defn generate-update-property-name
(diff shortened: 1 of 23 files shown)


common/src/app/common/logic/variants.cljc
```diff
@@ -60,6 +60,11 @@
                                                :value v}) remaining)]
     (into assigned new-properties)))
 
+(defn- dashes-to-end
+  [data]
```
hirunatan: This parameter should have a meaningful name. In general, "data" is a too abstract name. In Penpot we have adopted the convention of using this name for a file-data object (what it's not the case here).

Perhaps it could be "values" or "property-values".

pabloalba: Fixed

frontend/src/app/main/ui/inspect/attributes/variant.cljs
```diff
@@ -0,0 +1,41 @@
...
+  [:div {:title value
+         :class (stl/css :variant-row)}
+   [:div {:class (stl/css :global/attr-label)} name]
+   [:div {:class (stl/css :global/attr-value)}
+    [:& copy-button {:data value}
+     [:div {:class (stl/css :button-children)} value]]]])
+
+
+(mf/defc variant-panel
+  [{:keys [objects shapes libraries fid]}]
```
hirunatan: Plz don't use random abbreviations... :stuck_out_tongue: 

Name this `file-id`. Memory and disk are cheap, and you'l spare the future reader some time trying to figure out what a `fid` is...

frontend/src/app/main/ui/inspect/attributes/variant.cljs
```diff
@@ -0,0 +1,41 @@
...
+  (let [data            (dm/get-in libraries [fid :data])
+        variant-id      (->> shapes first :id)
```
hirunatan: Those two may also be wrapped in a `with-memo` (I'm not sure of it)

pabloalba: Technically, all can be surrounded by with-memo :P

But those two aren't calculation-intensive, only access to properties inside maps. I think with-memo will produce overhead instead of helping here

pabloalba: done

hirunatan:
I'm missing this part of the tab heading:
image

hirunatan (approved):

PR #6052, opened 2025-03-10
:recycle: Simplifies RPC pattern for token themes

### LINKED ISSUE #72: Theme

not-in-stock:
Closes https://github.com/tokens-studio/penpot/issues/72 of parent issue: https://github.com/tokens-studio/penpot/issues/66

### DIFF
--- a/common/src/app/common/files/changes.cljc
+++ b/common/src/app/common/files/changes.cljc
@@ -1045,6 +1045,24 @@
                 (ctob/update-set lib' set-name (fn [prev-token-set]
                                                  (ctob/make-token-set (merge prev-token-set token-set)))))))))
 
+(defmethod process-change :set-token-theme
+  [data {:keys [group theme-name theme]}]
+  (update data :tokens-lib
+          (fn [lib]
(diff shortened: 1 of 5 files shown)


common/src/app/common/types/tokens_lib.cljc
```diff
@@ -590,6 +589,14 @@
...
       (map->TokenTheme)))
 
+(defn make-token-theme
+  [& {:as attrs}]
+  (make-token-theme-impl attrs))
+
+(defn make-theme
+  [attrs]
+  (make-token-theme-impl attrs))
+
```
hirunatan: This is not necessary. In clojure you can define a function with keyword arguments and call it with a single map:

https://clojure.org/news/2021/03/18/apis-serving-people-and-programs

In fact, other `make-xxx` functions in this module are being used this way.

not-in-stock: Wow, I didn't know that actually, thanks!
I'll revert it back to the original signature. A very neat language update BTW.

PR #6053, opened 2025-03-11
:sparkles: Support exporting WebP images

dfelinto:
It is very convenient to be able to export WEBP right from penpot. Otherwise users have to first download to PNG then convert it locally.

---

Playwright only supports JPEG and PNG. So in order to support WEBP I had to first generate a PNG and then convert it afterwards. The webp is converted from the PNG with quality 100.

### Patch review notes:

1. I don't delete the generated PNG afterwards, not sure if it is needed or if I can trust it to be auto-cleaned since it is in the temp folder.

2. Also I'm not sure if the sh/cmd! is blocking. On my tests it worked well, but if it is not blocking, there is a chance it won't be finished by the time penpot tries to move the WEBP file to its final location.

### Summary

Support to export WEBP images.

### Steps to reproduce 

1. Select any element (e.g., a frame).
2. Click on Export.
3. Pick WEBP from the format list.

https://github.com/user-attachments/assets/fa79bc81-20f9-4e2b-a785-252594463122

### Checklist

- [ ] ~Add or modify existing integration tests in case of bugs or new features, if applicable.~
- [ ] ~Update the `CHANGES.md` file, referencing the related GitHub issue, if applicable.~

### DIFF
--- a/common/src/app/common/types/shape/export.cljc
+++ b/common/src/app/common/types/shape/export.cljc
@@ -8,7 +8,7 @@
   (:require
    [app.common.schema :as sm]))
 
-(def types #{:png :jpeg :svg :pdf})
+(def types #{:png :jpeg :webp :svg :pdf})
 
 (def schema:export
   [:map {:title "ShapeExport"}
--- a/docs/user-guide/exporting/index.njk
+++ b/docs/user-guide/exporting/index.njk
@@ -27,7 +27,7 @@ title: 07· Exporting objects
 <ul>
   <li><strong>Size</strong> - Options for the most common sizing scales.</li>
   <li><strong>Suffix</strong> - Especially useful if you are exporting at different scales.</li>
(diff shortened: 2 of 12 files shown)


dfelinto:
For the records, my original intend was for the WEBP to show up after JPG (so all the bitmap ones are grouped together). But for whatever reason it is showing at the end. I still need to investigate why.

image

dfelinto:
Did a fresh build here in a different computer, and the order seems correct actually :) so maybe it was just cached wrongly in my PC:

image

niwinz:
Awesome.
The PR looks good. After internal conversation, we have decided to port it to the current version (in order to be able to release it quickly to all users), so I created a PR with the same commit (with changelog updated) targeting the `staging` branch.

dfelinto:
I'm glad to hear it. I use the officially deployed penpot as well, so that also means I also get to use this feature sooner. Thanks for reviewing it :)

PR #6064, opened 2025-03-12
:bug: Fix clicking on a comment at the viewer's sidebar is not opening threads

luisddm:
### Related Ticket

Fixes Taiga issue [#10465](https://tree.taiga.io/project/penpot/issue/10465)

### Summary

Clicking on a comment in the viewer's comment sidebar causes an error in the application (it shows the screen to request access). This is because it tries to redirect the user to the workspace, and it does not have any info about the current user and their privileges.

With this fix, the redirection is avoided and  the thread is opened in the viewer, exactly the same way as it is done in the workspace.

### DIFF
--- a/frontend/src/app/main/ui/workspace/comments.cljs
+++ b/frontend/src/app/main/ui/workspace/comments.cljs
@@ -8,6 +8,7 @@
   (:require-macros [app.main.style :as stl])
   (:require
    [app.main.data.comments :as dcmt]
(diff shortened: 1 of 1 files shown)


niwinz:
Is this should be merged on staging? I mean, this bug happens on current release?

luisddm:
Yes, it should be merged on staging. Done in https://github.com/penpot/penpot/pull/6083

PR #6066, opened 2025-03-12
:sparkles: Support emoji default font in text rendering

elenatorro:
### Related Ticket

https://tree.taiga.io/project/penpot/task/10387

### Summary

To support emoji in the new render we need to make use of an emoji font. This PR adds a way of using the `Noto Emoji Color` font, which also support emoji sequences.

At the beginning I thought I needed to spilt by char to select the correct font, but by sending multiple fonts, it seems to use the fallback correctly, so there's no need to use external libraries.

### Demo

* Emoji fallback

Screenshot 2025-03-12 at 12-25-42 emoji - Penpot

* Emoji fallback with custom fonts

Screenshot from 2025-03-17 15-08-45

### Checklist

### DIFF
--- a/render-wasm/src/render.rs
+++ b/render-wasm/src/render.rs
@@ -86,10 +86,6 @@ pub(crate) struct RenderState {
     pub options: RenderOptions,
     pub surfaces: Surfaces,
     fonts: FontStore,
-    // TODO: we should probably have only one of these
-    // pub font_provider: skia::textlayout::TypefaceFontProvider,
-    // pub font_collection: skia::textlayout::FontCollection,
-    // ----
     pub cached_surface_image: Option<CachedSurfaceImage>,
     pub viewbox: Viewbox,
     pub images: ImageStore,
@@ -117,7 +113,6 @@ impl RenderState {
 
         // This is used multiple times everywhere so instead of creating new instances every
         // time we reuse this one.
-
         RenderState {
             gpu_state,
             surfaces,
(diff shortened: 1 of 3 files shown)


render-wasm/src/render.rs
```diff
@@ -113,11 +109,10 @@ impl RenderState {
             skia::SamplingOptions::new(skia::FilterMode::Linear, skia::MipmapMode::Nearest);
         let surfaces = Surfaces::new(&mut gpu_state, (width, height), sampling_options);
 
-        let fonts = FontStore::new();
+        let fonts: FontStore = FontStore::new();
```
Alotor: do you need to have the type explicitely setted?

render-wasm/src/shapes/text.rs
```diff
@@ -57,22 +60,25 @@ impl TextContent {
...
+                    let text_style = leaf.to_style();
+                    builder.push_style(&text_style);
                     builder.add_text(&leaf.text);
                     builder.pop();
                 }
+
                 builder.build()
             })
-            .collect()
+            .collect::<Vec<_>>() // Collect the built paragraphs into a Vec
```
Alotor: Do you need to have the type here? I think i can be inferred by the return type

render-wasm/src/render.rs
elenatorro: not needed, let me remove it! 👍🏽

Alotor (approved):

PR #6112, opened 2025-03-19
:books: Update changelog

myfunnyandy:
https://tree.taiga.io/project/penpot/task/10523

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 ### :rocket: Epics and highlights
 
+- Design Tokens
+
 ### :boom: Breaking changes & Deprecations
 
 ### :heart: Community contributions (Thank you!)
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -13,6 +15,15 @@
...
 - Change templates carrousel [Taiga #9803](https://tree.taiga.io/project/penpot/us/9803)
+- [DESIGN TOKENS] Tokens CRUD. Types added: Color, Opacity, Border radius, Dimension, Sizing, Spacing, Rotation and Stroke.
+- [DESIGN TOKENS] Create references (alias) that point to other tokens.
+- [DESIGN TOKENS] Math operations in token values.
+- [DESIGN TOKENS] Sets CRUD, grouping and reordering.
+- [DESIGN TOKENS] Multidimensional Themes and Sets management.
+- [DESIGN TOKENS] Apply/Remove tokens to/from elements from the Tokens tab.
+- [DESIGN TOKENS] Integration with components.
+- [DESIGN TOKENS] Import and export tokens from a JSON file.
+- [DESIGN TOKENS] Apply Themes and Sets at document level.
```
niwinz, line 26: I think this entries are not very relevant, because from the user perspective, they receive the tokens feature and not all the underlying tasks.

hirunatan: On the contrary, I find this interesting. Design Tokens is a quite large matter and we are only deploying an MVP. There are many functions that perhaps people may expect but are not done, like syncing tokens with external sources, complex tokens or applying themes to page and board level. Those things may be added to the changes log in the future, when we add them.

niwinz: They may be interesting, but changelog is for relevant changes, or we have a single entry for design tokens or we put all the underlying "task/us", but both are redundant.
The purpose of changelog is tell the user of on-premise or any external user of what is coming or what changes are made. The design tokens didnt existed previously, so the change is clear: Add design tokens.

Explain what exactly means design tokens on this version should go to the documentation, not changelog.

niwinz: After thinking a bit more on it. I think you are right, and that not make sense is put the EPIC on the changelog. Changelog should be traceable along the time, and the EPIC is a changing (the implemented US's on EPIC can change over time). So on the end, I just suggest don't link the EPIC. And leave the rest as is.

myfunnyandy: I agree with this last thought, I haven't thought about the (un)convenience of adding the EPIC to the changelog. I'll remove the link.

PR #6115, opened 2025-03-19
:bug: Fix incorrect absolute frame positioning with measures sidebar

niwinz:
### Related Ticket

- https://tree.taiga.io/project/penpot/issue/10538
- https://tree.taiga.io/project/penpot/issue/10569

No changelog required because it fixes a regression that happened in the same release.

**NO SQUASH**

### DIFF
--- a/frontend/src/app/main/data/workspace/transforms.cljs
+++ b/frontend/src/app/main/data/workspace/transforms.cljs
@@ -859,42 +859,44 @@
           (rx/of (reorder-selected-layout-child direction))
           (rx/of (nudge-selected-shapes direction shift?)))))))
 
-(defn- get-delta [position bbox]
-  (let [cpos (gpt/point (:x bbox) (:y bbox))
-        pos (gpt/point (or (:x position) (:x bbox))
-                       (or (:y position) (:y bbox)))]
-    (gpt/subtract pos cpos)))
-
-(defn- get-relative-delta [position bbox frame]
-  (let [frame-bbox (-> frame :points grc/points->rect)
-        relative-cpos (gpt/subtract (gpt/point (:x bbox) (:y bbox))
-                                    (gpt/point (:x frame-bbox)
-                                               (:y frame-bbox)))
-        cpos (gpt/point (:x relative-cpos) (:y relative-cpos))
-        pos (gpt/point (or (:x position) (:x relative-cpos))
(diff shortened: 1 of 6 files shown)


frontend/src/app/main/data/workspace/transforms.cljs
```diff
@@ -859,42 +859,41 @@
...
-                          (get-relative-delta position bbox frame))
-             modif-tree (dwm/create-modif-tree [id] (ctm/move-modifiers delta))]
-         (rx/of (dwm/apply-modifiers {:modifiers modif-tree
+       (let [page-id   (or (get options :page-id)
+                           (get state :current-page-id))
+             objects   (dsh/lookup-page-objects state page-id)
+             shape     (get objects id)
+
+             bbox      (-> shape :points grc/points->rect)
+             frame     (if (:absolute? options)
```
hirunatan, line 892: It's confusing to have an `absolute?` option if now the update is always absolute.

Perhaps it may be complex to correctly fix this now. If it's the case, we can at least add a `FIXME` comment here, what do you think?

niwinz: It always relative to frame, not absolute. It is handled in absolute way on plugins API, and this is the reason to exist that option.

niwinz: We can think on have two different events instead of an option with better name. In any case this should not block the merge. We can approach this different PR.

hirunatan: Ok, so there is one case (API calls) where the position needs to be absolute respect to the workspace canvas, and not relative to the containing board. So we can just leave this explained in the docstring of the function. E.g. 

 "Move shapes to a new position. It will resolve to the current frame of the shape,
  unless given the `absolute` option. In this case it will resolve to the root frame of the page.
 
  The `position` is a map that can have a partial position (it means it can receive {:x 10}."

hirunatan: With this, It's ok for me to proceed with the merge.

hirunatan (approved):

frontend/src/app/main/data/workspace/transforms.cljs
niwinz: Fixed

PR #6126, opened 2025-03-20
:bug: Fix stroke shadows

elenatorro:
### Related Ticket

https://tree.taiga.io/project/penpot/task/10516

### Summary

This is the third step for fixing shadows rendering. We fixed [Fill drop shadows](https://github.com/penpot/penpot/pull/6077) and [Fill inner shadows](https://github.com/penpot/penpot/pull/6122) previously.

### Changes

- This PR deprecates the old `Shadow` and `Overlay` surfaces, that we kept to be able to fix this issue in different PRs.
- We now use `DropShadows` and `InnerShadows` surfaces for "shape" shadows (fills, not strokes), while the `Strokes` surface is being used for stroke shadows, which makes it easier to make sure the different stroke "layers" behave correctly when having shadows.

### What does PR this fix?

- [x] Drop and Inner stroke shadows in borders are not clipped anymore when the shape intersects the viewport or the tiles

current_inner_stroke_shadows
new_inner_stroke_shadows

- [x] Inner stroke shadows work correctly when a shape has multiple strokes.

This final point requires a more detailed explanation. To test strokes and inner shadows, we must consider that there are three types of stroke positioning: inside, outside, and center. In the following screenshots, I applied these strokes in different orders and compared the individual strokes to the final solution to clarify how it should appear in certain cases:

#### Setup

- Strokes: 
Screenshot 2025-03-25 at 14-55-02 shadows_default - Penpot

- Inner shadow:
Screenshot 2025-03-25 at 14-48-39 shadows_default - Penpot

##### Case 1: outside + inside 

Screenshot 2025-03-25 at 14-49-59 shadows_default - Penpot

##### Case 2: outside + inside + center

Screenshot 2025-03-25 at 14-50-04 shadows_default - Penpot

#### Case 3: outside + center + inside

Screenshot 2025-03-25 at 14-50-08 shadows_default - Penpot

#### Case 4: center + outside + inside

Screenshot 2025-03-25 at 14-50-14 shadows_default - Penpot

This approach differs from the current production render, so we should discuss if this is the correct approach we'd like to follow.

### DIFF
--- a/render-wasm/src/render.rs
+++ b/render-wasm/src/render.rs
@@ -211,13 +211,7 @@ impl RenderState {
 
     pub fn apply_drawing_to_render_canvas(&mut self, shape: Option<&Shape>) {
         self.surfaces
-            .flush_and_submit(&mut self.gpu_state, SurfaceId::Fills);
-
-        self.surfaces
             .flush_and_submit(&mut self.gpu_state, SurfaceId::DropShadows);
-
-        self.surfaces
-            .flush_and_submit(&mut self.gpu_state, SurfaceId::InnerShadows);
 
         self.surfaces.draw_into(
             SurfaceId::DropShadows,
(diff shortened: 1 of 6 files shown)


superalex:
I think this MR breaks some strokes with images as fills. For example in this file: https://design.penpot.dev/#/workspace?team-id=b62b19c3-5105-81f2-8005-448a48df550b&project-id=7f0e152c-fb26-81a6-8005-8de0e641f1af&file-id=486fe96e-4e62-810b-8005-8de1964a4af6&page-id=486fe96e-4e62-810b-8005-8de1964a4af7

Current develop branch:



Pull request:

elenatorro:
Thanks @superalex!! Good catch!! I've fixed some issues:
- As you mentioned, I was not drawing images on strokes cause I forgot to remove a check on the shadow parameter, this is fixed
- I fixed the shadows on stroke images, which were being rendered in the wrong order because the stroke images were being painted in the Fill surface


Screenshot 2025-03-28 at 08-43-51 Render Test (v3) - Penpot

While testing this I noticed that images on strokes seem to be cropped, I'll review that separately since this is happening in develop as well 👍🏽

superalex:
Thank you very much @elenatorro, it works great!

PR #6136, opened 2025-03-24
:bug: Disable svg viewport texts for wasm render

superalex:
Related to: https://tree.taiga.io/project/penpot/task/10596


frontend/src/app/main/ui/workspace/viewport_wasm.cljs
```diff
@@ -326,21 +325,6 @@
      (when (:can-edit permissions)
        [:& top-bar/top-bar {:layout layout}])
      [:div {:class (stl/css :viewport-overlays)}
-      ;; The behaviour inside a foreign object is a bit different that in plain HTML so we wrap
```
AzazelN28, line 329: Instead of removing this, maybe we should add a condition that tests if render-v2 is enabled, shouldn't we?

PR #6138, opened 2025-03-24
:bug: Fix calculate zoom to avoid bubbles to get outside viewbox

luisddm:
### Related Ticket

Fixes Taiga issue [#10563](https://tree.taiga.io/project/penpot/issue/10563)

### Summary

Sometimes, when clicking on a bubble group,  the zoom scale we need to apply in order to separate it into bubbles is so big that some or all the bubbles go outside the viewbox, and we need to zoom out again to see any of them, making this confusing for the user. This typically happens when some of the bubbles of the group are very close and some others are far away.

We have adapted the algorithm that calculates the zoom level. In each iteration, we also check that all the bubbles remain inside the viewbox. If any of them gets outside, then we stop the iteration and take the zoom level, even if there are still grouped bubbles. The user can still click in the remaining groups to get them separated recursively.

This may sometimes make the process of separating bubbles a multi-step process rather of doing it at once, but the result is more intuitive for the user.

### DIFF
--- a/frontend/src/app/main/data/workspace/viewport.cljs
+++ b/frontend/src/app/main/data/workspace/viewport.cljs
@@ -96,17 +96,12 @@
 
 (defn update-viewport-position-center
   [position]
-
-  (dm/assert!
-   "expected a point instance for `position` param"
-   (gpt/point? position))
+  (assert (gpt/point? position) "expected a point instance for `position` param")
 
   (ptk/reify ::update-viewport-position-center
     ptk/UpdateEvent
     (update [_ state]
-      (update state :workspace-local
-              (fn [local]
-                (calculate-centered-viewbox local position))))))
+      (update state :workspace-local calculate-centered-viewbox position))))
 
 (defn update-viewport-position
   [{:keys [x y] :or {x identity y identity}}]
--- a/frontend/src/app/main/data/workspace/viewport.cljs
+++ b/frontend/src/app/main/data/workspace/viewport.cljs
@@ -10,6 +10,7 @@
    [app.common.data.macros :as dm]
    [app.common.files.helpers :as cfh]
    [app.common.geom.align :as gal]
+   [app.common.geom.point :as gpt]
    [app.common.geom.rect :as gpr]
    [app.common.geom.shapes :as gsh]
    [app.common.math :as mth]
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/data/workspace/viewport.cljs
```diff
@@ -83,6 +84,30 @@
...
+  (dm/assert!
+   "expected a point instance for `position` param"
+   (gpt/point? position))
+
+  (ptk/reify ::update-viewport-position-center
+    ptk/UpdateEvent
+    (update [_ state]
+      (update state :workspace-local
+              (fn [local]
+                (calculate-centered-viewbox local position))))))
```
niwinz: This can be expressed this way, without crating an additional anonymous function:

```clojure
(update state :workspace-local calculate-centered-viewbox position)
```

frontend/src/app/main/data/workspace/viewport.cljs
```diff
@@ -83,6 +84,30 @@
...
+        nx   (- (:x position) nw)
+        ny   (- (:y position) nh)]
+    (update local :vbox assoc :x nx :y ny)))
+
+(defn update-viewport-position-center
+  [position]
+
+  (dm/assert!
+   "expected a point instance for `position` param"
+   (gpt/point? position))
```
niwinz: We are deprecating `dm/assert!`, so consider using that instead:

```clojure
(assert (gpt/point? position) "expected a point instance for `position` param")
```

frontend/src/app/main/ui/comments.cljs
```diff
@@ -1097,15 +1098,34 @@
...
-        num-grouped-threads (count (group-bubbles zoom threads))
-        zoom-scale-step     1.75]
-    (if (= num-threads num-grouped-threads)
+        grouped-threads     (group-bubbles zoom threads)
+        num-grouped-threads (count grouped-threads)
+        zoom-scale-step     1.75
+        scaled-zoom         (* zoom zoom-scale-step)
+        zoomed-wl           (dwz/impl-update-zoom wl position scaled-zoom)]
+    (if (or (= num-threads num-grouped-threads)
+            (some #((comp not inside-vbox?) % zoomed-wl) grouped-threads))
```
niwinz: Consider define `(comp not inside-vbox?)` in a let binding and reuse, because with this code you are creating a new function on each iteration.

Also, this also can be simplified to `(complement inside-vbox?)`

frontend/src/app/main/ui/comments.cljs
```diff
@@ -1124,13 +1144,17 @@
 
         test-id     (str/join "-" (map :seqn (sort-by :seqn thread-group)))
 
+        wl          (mf/deref refs/workspace-local)
```
niwinz: if the `wl` value is only needed on time of executing the click event, the value can be derefed with plain deref on the click handler instead of creating this reactive deref that adds a watcher on workspace-local ref.

niwinz (changes requested):

PR #6142, opened 2025-03-24
:bug: Fix asynchronous content dependant rendering

superalex:
We broke that when merging tiles feature.

Also related to https://tree.taiga.io/project/penpot/task/10579


belen-albeza:
Still does not work for me :( I'll keep looking if there are more changes that got lost.

PR #6155, opened 2025-03-25
:tada: Cache extra tiles

AzazelN28:
This PR simply adds more tiles around the viewport tiles to have those prerendered when you pan.

### DIFF
--- a/render-wasm/src/render.rs
+++ b/render-wasm/src/render.rs
@@ -28,6 +28,8 @@ pub use blend::BlendMode;
 pub use fonts::*;
 pub use images::*;
 
+// This is the extra are used for tile rendering.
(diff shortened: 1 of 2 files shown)


render-wasm/src/render.rs
```diff
@@ -454,17 +454,9 @@ impl RenderState {
             },
         );
 
-        let (sx, sy, ex, ey) = tiles::get_tiles_for_viewbox(self.viewbox);
+        // TODO: Maybe we should calculate the interest area based on the actual viewport. See how.
+        let (sx, sy, ex, ey) = tiles::get_tiles_for_viewbox_with_interest(self.viewbox, 1);
```
superalex: May be we should move this "1" to a constant?

PR #6160, opened 2025-03-26
:tada: Add PathData data type

niwinz:
### Summary

It encapsulates all the interaction with binary encoded path data. It implements all the necessary clojure protocols for preserve compatibility with the previous format that will be called `plain` from now.

Is a first step before approach a migration and make all path content to be stored in binary format on database and transferred in binary format from frontend to backend.

### How to test

This right now only affects partially to wasm render, because it replaces the old ad-hoc functions with the new type with its own API.

- Draw a path
- Activate render wasm
- Check that path is still rendered in the same way as before applying the commits of this PR.


elenatorro (approved):
I tested it in the new render and works fine 👍🏽

PR #6169, opened 2025-03-26
:bug: Fixes bugs from design review

pabloalba:
Implements https://tree.taiga.io/project/penpot/task/10435

* :bug: Fix create variant on container without flex doesn't resize container
*  :bug: Fix when transform component in variant it moves
*  :bug: Fix validator for deleted components
*  :bug: Fix system error when moving a variant container ourside a frame
*  :bug: Fix can't add a flex layout to a variant container
*  :bug: Fix bad color on variant container name and outlines
*  :bug: Fix variants don't rename when the variant container is renamed
*  :bug: Fix default variant value should be "Value 1"
*  :bug: Fix component is not shown when selected "show in asset panel" over a variant
*  :bug: In assets tab variants should have a mark
*  :bug: Fix can't rename nor group variants from the assets panel
*  :bug: Fix different title for container/variant on inspect tab
*  :bug: Fix scroll overlaps variant properties on design tab

### DIFF
--- a/common/src/app/common/files/variant.cljc
+++ b/common/src/app/common/files/variant.cljc
@@ -6,6 +6,7 @@
 (ns app.common.files.variant
   (:require
    [app.common.data.macros :as dm]
+   [app.common.types.component :as ctc]
    [app.common.types.components-list :as ctcl]
    [app.common.types.variant :as ctv]
    [cuerdas.core :as str]))
@@ -61,3 +62,22 @@
     (and (seq shapes)
          (not= (:main-instance-id component) (last shapes)))))
 
+(defn get-primary-variant
+  [data component]
+  (let [page-id    (:main-instance-page component)
+        objects    (-> (dm/get-in data [:pages-index page-id])
+                       (get :objects))
+        variant-id (:variant-id component)]
+    (->> (dm/get-in objects [variant-id :shapes])
+         peek
+         (get objects))))
+
+(defn get-primary-component
+  [data component-id]
+  (when-let [component (ctcl/get-component data component-id)]
+    (if (ctc/is-variant? component)
+      (->> component
+           (get-primary-variant data)
+           :component-id
+           (ctcl/get-component data))
+      component)))
--- a/frontend/src/app/main/data/workspace.cljs
+++ b/frontend/src/app/main/data/workspace.cljs
@@ -13,6 +13,7 @@
    [app.common.features :as cfeat]
    [app.common.files.changes-builder :as pcb]
    [app.common.files.helpers :as cfh]
+   [app.common.files.variant :as cfv]
    [app.common.geom.align :as gal]
    [app.common.geom.point :as gpt]
    [app.common.geom.proportions :as gpp]
@@ -25,7 +26,7 @@
    [app.common.schema :as sm]
    [app.common.text :as txt]
    [app.common.transit :as t]
-   [app.common.types.component :as ctk]
+   [app.common.types.component :as ctc]
    [app.common.types.components-list :as ctkl]
    [app.common.types.container :as ctn]
    [app.common.types.file :as ctf]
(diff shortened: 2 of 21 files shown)


niwinz:
I will look only the code part, FYI

common/src/app/common/files/variant.cljc
```diff
@@ -61,3 +62,22 @@
...
          (not= (:main-instance-id component) (last shapes)))))
 
+(defn get-primary-variant
+  [data component]
+  (let [page-id    (:main-instance-page component)
+        objects    (-> (dm/get-in data [:pages-index page-id])
+                       (get :objects))
+        variant-id (:variant-id component)]
+    (->> (dm/get-in objects [variant-id :shapes])
+         last
```
niwinz: Shapes is known to be a vector, in this case the most efficient way to get the last element is using `peek`

frontend/src/app/main/data/workspace.cljs
```diff
@@ -547,7 +549,7 @@
             name               (cfh/generate-unique-name base-name unames :suffix-fn suffix-fn)
             objects            (update-vals (:objects page) #(dissoc % :use-for-thumbnail))
 
-            main-instances-ids (set (keep #(when (ctk/main-instance? (val %)) (key %)) objects))
+            main-instances-ids (set (keep #(when (ctc/main-instance? (val %)) (key %)) objects))
```
niwinz, line 582: This is a minor one, and consider as you want. The most efficient way to write this code is:

```clojure
(reduce-kv (fn [result id shape]
             (if (ctc/main-instance? shape)
               (conj result id)
               result))
           #{}
           objects)        
```

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -144,6 +147,27 @@
...
+(defn- resposition-and-resize-variant
+  "Resize the variant container, and move the shape (that is a variant) to the right"
+  [shape-id]
+  (ptk/reify ::resposition-and-resize-variant
+    ptk/WatchEvent
+    (watch [_ state _]
+      (let [page-id   (:current-page-id state)
+            data      (dsh/lookup-file-data state)
+            objects   (-> (dsh/get-page data page-id)
+                          (get :objects))
```
niwinz: this can be reduced to `(dsh/lookup-page-objects state page-id)`

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -310,3 +347,38 @@
...
+
+(defn rename-variant
+  [variant-id name]
+  (ptk/reify ::rename-variant
+
+    ptk/WatchEvent
+    (watch [_ state _]
+      (let [data               (dsh/lookup-file-data state)
+            objects            (dsh/lookup-page-objects state)
+            variant-components (cfv/find-variant-components data objects variant-id)
```
niwinz, line 360: And here you should avoid duplicate lookups using `dsh/get-page` + `(get page :objects)` istead of `(dsh/lookup-page-objects state)`

frontend/src/app/main/ui/workspace/sidebar/assets/file_library.cljs
```diff
@@ -384,7 +385,15 @@
...
-           (st/emit! (dw/unselect-all-assets file-id))))]
+           (st/emit! (dw/unselect-all-assets file-id))))
+
+        count-variants
+        (mf/use-fn
+         (mf/deps library)
+         (fn [variant-id]
+           (->> (ctkl/components-seq library)
+                (filter #(= variant-id (:variant-id %)))
+                count)))]
```
niwinz, line 397: use filterv here, because you intention is counting it, and count will iterate again the seq for count, so if you generate a vector, then the count will be for free.

niwinz (changes requested):

PR #6174, opened 2025-03-27
:sparkles: Allow modifying property name when a variation is selected

luisddm:
### Related Ticket

Implements Taiga [#10649](https://tree.taiga.io/project/penpot/task/10649)

### Summary

When a variant is selected, the user should be able to edit the name of any of its properties by clicking on them on the right sidebar.

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/options/menus/component.scss
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/menus/component.scss
@@ -719,6 +719,7 @@
 .variant-property-name {
   color: var(--color-foreground-primary);
   height: var(--sp-xxxl);
+  width: $s-104;
 
   display: flex;
   align-items: center;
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/ui/workspace/sidebar/options/menus/component.scss
```diff
@@ -716,22 +716,19 @@
...
   align-items: center;
   justify-content: center;
-  height: var(--sp-xxxl);
 
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
 }
+
+.variant-property-name-width {
```
pabloalba: I don't think that we should use "width" as part of the name of a class...

Maybe call it `:variant-property-name-wrapper` or something similar...

Also, I will keep the width also on `variant-property-name`, and use one or the other

pabloalba (approved):

PR #6177, opened 2025-03-27
:tada: Sort viewport tiles by distance to center

AzazelN28:
Sort tiles by distance to the center of the viewport.

### DIFF
--- a/render-wasm/src/render.rs
+++ b/render-wasm/src/render.rs
@@ -94,7 +94,7 @@ pub(crate) struct RenderState {
     pub sampling_options: skia::SamplingOptions,
     pub render_area: Rect,
     pub tiles: tiles::TileHashMap,
-    pub pending_tiles: Vec<tiles::Tile>,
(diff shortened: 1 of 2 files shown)


render-wasm/src/render.rs
```diff
@@ -468,15 +468,19 @@ impl RenderState {
...
             for x in sx..=ex {
                 let tile = (x, y);
-                self.pending_tiles.push(tile);
+                let distance = tiles::manhattan_distance(tile, tile_center);
+                self.pending_tiles.push((x, y, distance));
                 self.surfaces.cache_visit(tile);
             }
         }
+        // reorder by distance to the center.
+        self.pending_tiles.sort_by(|a, b| b.2.cmp(&a.2));
```
elenatorro, line 476: I guess we want to sort in ascending order, to process closer tiles first, right?

```suggestion
        self.pending_tiles.sort_by(|a, b| a.2.cmp(&b.2));
```

elenatorro (approved):
Apart from the conflict file, LGTM 🚀

PR #6184, opened 2025-03-28
:bug: Fix single set import

### LINKED ISSUE #93: bug:  While importing a json file without $metadata, the file should not be broken down into sets
### Steps To Reproduce

Steps to reproduce:
1. Import a single json file, which is a single set without $metadata or $themes. This could be a json file which is not from the plugin, but could be user generated.
2. The file is broken down into sets taking the first key of the json as the set name.
3. This is not desired if the uploaded file is just a single set.

### Expected behavior

If a single set is uploaded without any $metadata or $themes, we should not break up the file into sets. But import it as a single set.

If a json file does not contain $metadata or $themes we should not break it up into sets. If the $metadata and $themes is present, we should continue with the existing behaviour.

### Actual behavior

Currently the single set is broken up into multiple sets using the first key.

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

not-in-stock:
Fixes https://github.com/tokens-studio/penpot/issues/93

Files to test the changes:
- Legacy single set - [legacy_single_set.json](https://github.com/user-attachments/files/19535508/legacy_single_set.json)
- DTCG single set - [single_set.json](https://github.com/user-attachments/files/19535528/single_set.json)
- Invalid JSON file - [invalid_json.json](https://github.com/user-attachments/files/19535534/invalid_json.json)
- Legacy tokens file - [legacy_tokens.json](https://github.com/user-attachments/files/19535539/legacy_tokens.json)
- DTCG tokens file - [tokens.json](https://github.com/user-attachments/files/19535542/tokens.json)


frontend/src/app/main/ui/workspace/tokens/style_dictionary.cljs
```diff
@@ -249,33 +250,44 @@
...
-                                       (throw err)))))
+                      (t/decode-str data)
                       (catch js/Error e
-                        (p/rejected (wte/error-ex-info :error.import/style-dictionary-unknown-error "" e))))))))
+                        (throw (wte/error-ex-info :error.import/json-parse-error data e))))))
+          (rx/map (fn [json-data]
+                    (try
+                      (cond
+                        (ctob/single-set? json-data)
+                        (ctob/decode-single-set-dtcg-json (deref refs/tokens-lib) file-name json-data)
```
hirunatan: This breaks when you import a single set file in a file without tokens (because `refs/tokens-lib` is nil). Why is this different to the other cases, where you use `ensure-tokens-lib`?

not-in-stock: I suppose `ensure-tokens-lib` should fix it.
I'll add call around `(deref refs/tokens-lib)`

not-in-stock: Done ✅ (https://github.com/penpot/penpot/pull/6184/commits/b6ca225d0c848d5fb44bb983ba16d374845bd881)

hirunatan: I understand that the difference is to allow incremental load (e.g. import several json files and get them added). Ok for me.

not-in-stock: Yes, you can try importing example sets sequentially and there won't be an override of all of the sets in the file
[single_set_2.json](https://github.com/user-attachments/files/19510238/single_set_2.json)
[single_set.json](https://github.com/user-attachments/files/19510239/single_set.json)

common/src/app/common/types/tokens_lib.cljc
```diff
@@ -665,20 +665,20 @@
      ["$value" :map]
      ["$type" :string]]]))
 
-(defn has-legacy-format?
+(defn get-json-format
```
not-in-stock: I've changed the predicate to a function that returns a file format keyword, since relying on the distinction between `true`, `false`, and `nil` in a language with nil punning seems error-prone.

hirunatan: ok

PR #6197, opened 2025-03-31
:sparkles: Fix single set import

not-in-stock:
Fixes https://github.com/tokens-studio/penpot/issues/93

Files to test the changes:
- Legacy single set - [legacy_single_set.json](https://github.com/user-attachments/files/19535508/legacy_single_set.json)
- DTCG single set - [single_set.json](https://github.com/user-attachments/files/19535528/single_set.json)
- Invalid JSON file - [invalid_json.json](https://github.com/user-attachments/files/19535534/invalid_json.json)
- Legacy tokens file - [legacy_tokens.json](https://github.com/user-attachments/files/19535539/legacy_tokens.json)
- DTCG tokens file - [tokens.json](https://github.com/user-attachments/files/19535542/tokens.json)

### DIFF
--- a/common/src/app/common/types/tokens_lib.cljc
+++ b/common/src/app/common/types/tokens_lib.cljc
@@ -665,20 +665,20 @@
      ["$value" :map]
      ["$type" :string]]]))
 
-(defn has-legacy-format?
+(defn get-json-format
   "Searches through parsed token file and returns:
-   - true when first node satisfies `legacy-node?` predicate
-   - false when first node satisfies `dtcg-node?` predicate
-   - nil if neither combination is found"
+   - `:json-format/legacy` when first node satisfies `legacy-node?` predicate
+   - `:json-format/dtcg` when first node satisfies `dtcg-node?` predicate
+   - `nil` if neither combination is found"
   ([data]
-   (has-legacy-format? data legacy-node? dtcg-node?))
+   (get-json-format data legacy-node? dtcg-node?))
   ([data legacy-node? dtcg-node?]
    (let [branch? map?
          children (fn [node] (vals node))
          check-node (fn [node]
                       (cond
-                        (legacy-node? node) true
-                        (dtcg-node? node) false
+                        (legacy-node? node) :json-format/legacy
+                        (dtcg-node? node) :json-format/dtcg
                         :else nil))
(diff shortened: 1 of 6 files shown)


common/src/app/common/types/tokens_lib.cljc
```diff
@@ -1287,6 +1311,17 @@ Will return a value that matches this schema:
           (assoc-in ["$metadata" "activeThemes"] active-themes-clear)
           (assoc-in ["$metadata" "activeSets"] active-sets))))
 
+  (decode-single-set-json [this set-name tokens]
```
niwinz, line 1314: This function does not requires access to anything internal of the data structure, so it should be implemented using a simple function and not a protocol method.

common/src/app/common/types/tokens_lib.cljc
```diff
@@ -1287,6 +1311,17 @@ Will return a value that matches this schema:
...
+    (assert (map? tokens) "expected a map data structure for `data`")
+
+    (add-set this (make-token-set :name (normalize-set-name set-name)
+                                  :tokens (flatten-nested-tokens-json tokens ""))))
+
+
+  (decode-single-set-legacy-json [this set-name tokens]
```
niwinz, line 1321: The same applies here

common/src/app/common/types/tokens_lib.cljc
niwinz: Also, the name indicates decode, but it also transforms the current tokensLib, adding the decoded data.
I think this should be defined on first instance as decode function  and then use already available methods on token lib for adding the decoded values.
This function is only used on tokens-dictionary ns, so inclusive, the helper that does both decode+add can be implemented there and not in the data structure.

niwinz (changes requested):

common/src/app/common/types/tokens_lib.cljc
hirunatan: @not-in-stock those changes are important, but we have agreed on doing them in a separate branch, for the next release.

hirunatan: done in https://github.com/penpot/penpot/pull/6221

PR #6200, opened 2025-04-01
:sparkles: Allow editing of sets in create theme dialog

### LINKED ISSUE #70: feature:  enhancement: set selection to be allowed during theme creation

When creating a new theme allow the set selection to be done immediately.

Currently we first create the theme and then have to go to the theme to select the sets. 

https://github.com/user-attachments/assets/4cd52837-962c-46fa-a0c7-6d23eecd0c2d

We should allow the selection of sets for the theme to be done in the theme creation process. 

_No response_

### Additional context

_No response_

floscr:
Closes https://github.com/tokens-studio/penpot/issues/70

https://github.com/user-attachments/assets/accbfdc9-d1fb-4470-9eef-e810a41e7c80

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
@@ -421,12 +421,12 @@
         state       (deref state*)
 
         change-view (mf/use-fn
-                     (fn [type & [state']]
+                     (fn [type & {:keys [theme-path]}]
                        (swap! state* (fn [current-state]
-                                       (merge
-                                        {:type type
-                                         :prev-type (:type current-state)}
-                                        state')))))
+                                       (cond-> current-state
+                                         :always (assoc :type type
+                                                        :prev-type (:type current-state))
+                                         :theme-path (assoc :theme-path theme-path))))))
 
         component (case (:type state)
                     :empty-themes empty-themes
--- a/frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
@@ -248,15 +248,13 @@
     (ctob/activate-theme tlib (:group theme) (:name theme))))
 
 (mf/defc edit-create-theme*
-  [{:keys [state change-view theme on-save is-editing]}]
+  [{:keys [change-view theme on-save is-editing has-prev-view]}]
   (let [ordered-token-sets (mf/deref refs/workspace-ordered-token-sets)
         token-sets (mf/deref refs/workspace-token-sets-tree)
 
         current-theme* (mf/use-state theme)
         current-theme (deref current-theme*)
         lib (make-lib-with-theme current-theme ordered-token-sets)
-
-        has-back-button? (contains? #{:empty-themes :themes-overview} (:prev-type state))
 
         ;; Form / Modal handlers
         on-back (mf/use-fn
@@ -339,7 +337,7 @@
 
      [:form {:on-submit on-save-form :class (stl/css :edit-theme-form)}
       [:div {:class (stl/css :edit-theme-wrapper)}
-       (when has-back-button?
+       (when has-prev-view
          [:button {:on-click on-back
                    :class (stl/css :back-btn)
                    :type "button"}
@@ -374,11 +372,15 @@
                                    :on-save-form on-save-form
                                    :disabled? disabled?}]]]]]]))
 
+(defn has-prev-view [prev-view-type]
+  (contains? #{:empty-themes :themes-overview} prev-view-type))
+
 (mf/defc edit-theme
   [{:keys [state change-view]}]
   (let [{:keys [theme-path]} state
         [_ theme-group theme-name] theme-path
         theme (mf/deref (refs/workspace-token-theme theme-group theme-name))
+        has-prev-view (has-prev-view (:prev-type state))
 
         on-save
         (mf/use-fn
@@ -387,11 +389,11 @@
            (st/emit! (wdt/update-token-theme [(:group theme) (:name theme)] theme'))))]
 
     [:> edit-create-theme*
-     {:state state
-      :change-view change-view
+     {:change-view change-view
       :theme theme
       :on-save on-save
-      :is-editing true}]))
+      :is-editing true
+      :has-prev-view has-prev-view}]))
 
 (mf/defc create-theme
   [{:keys [state change-view]}]
@@ -400,13 +402,14 @@
         (mf/use-fn
          (fn [theme]
            (st/emit! (ptk/event ::ev/event {::ev/name "create-tokens-theme"})
-                     (wdt/create-token-theme theme))))]
+                     (wdt/create-token-theme theme))))
+        has-prev-view (has-prev-view (:prev-type state))]
 
     [:> edit-create-theme*
-     {:state state
-      :change-view change-view
+     {:change-view change-view
       :theme theme
-      :on-save on-save}]))
+      :on-save on-save
+      :has-prev-view has-prev-view}]))
 
 (mf/defc themes-modal-body*
   {::mf/private true}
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
```diff
@@ -239,68 +239,17 @@
...
-        [_ theme-group theme-name] theme-path
-        ordered-token-sets (mf/deref refs/workspace-ordered-token-sets)
+(mf/defc edit-create-theme
+  [{:keys [set-state theme on-save edit-theme?]}]
+  (let [ordered-token-sets (mf/deref refs/workspace-ordered-token-sets)
         token-sets (mf/deref refs/workspace-token-sets-tree)
-        theme (mf/deref (refs/workspace-token-theme theme-group theme-name))
 
         theme-state* (mf/use-state theme)
         theme-state (deref theme-state*)
```
hirunatan: I suggest the name `current-theme`.

frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
```diff
@@ -239,68 +239,17 @@
...
         tlib (reduce ctob/add-set tlib sets)]
     (ctob/activate-theme tlib (:group theme) (:name theme))))
 
-(mf/defc controlled-edit-theme
-  [{:keys [state set-state]}]
-  (let [{:keys [theme-path]} @state
-        [_ theme-group theme-name] theme-path
-        ordered-token-sets (mf/deref refs/workspace-ordered-token-sets)
+(mf/defc edit-create-theme
+  [{:keys [set-state theme on-save edit-theme?]}]
```
hirunatan: I suggest the name `is-editing`. We should not use `?` in component property names, since they are converted to Javascript properties by react, and it doesn't handle special characters efficiently.

frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
```diff
@@ -379,17 +328,20 @@
...
      [:form {:on-submit on-save-form :class (stl/css :edit-theme-form)}
       [:div {:class (stl/css :edit-theme-wrapper)}
-       [:button {:on-click on-back
-                 :class (stl/css :back-btn)
-                 :type "button"}
-        [:> icon* {:icon-id ic/arrow-left :aria-hidden true}]
-        (tr "workspace.token.back-to-themes")]
-
-       [:> theme-inputs* {:theme theme
+       (when edit-theme?
```
hirunatan: Since you're here, I'll propose something to you: add some more checks so that this back link is only hidden when the form has been called from the left sidebar. If this comes from the main themes modal, do enable the link even for create theme.

frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
```diff
@@ -239,95 +239,44 @@
...
                  (ctob/add-theme theme))
         tlib (reduce ctob/add-set tlib sets)]
     (ctob/activate-theme tlib (:group theme) (:name theme))))
 
-(mf/defc controlled-edit-theme
-  [{:keys [state set-state]}]
-  (let [{:keys [theme-path]} @state
-        [_ theme-group theme-name] theme-path
-        ordered-token-sets (mf/deref refs/workspace-ordered-token-sets)
+(mf/defc edit-create-theme
```
hirunatan: You should use the new `rumext` component calling convention in all new ones (and if possible change old ones in the file, too):

- Add a `*` to the end of the component name.
- Use `:>` instead of `:&` when including a component inside other one.

frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
```diff
@@ -239,95 +239,44 @@
...
 
-        theme-state* (mf/use-state theme)
-        theme-state (deref theme-state*)
-        lib (make-lib-with-theme theme-state ordered-token-sets)
+        current-theme* (mf/use-state theme)
+        current-theme (deref current-theme*)
+        lib (make-lib-with-theme current-theme ordered-token-sets)
 
         ;; Form / Modal handlers
         on-back #(set-state (constantly {:type :themes-overview}))
```
hirunatan: Add a `use-fn` here.

frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
hirunatan: ok

frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
hirunatan: ok

frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
floscr: The logic is more about if there are any existing themes or not, as the initial view is defined by https://github.com/tokens-studio/penpot/blob/55a0067d01e1b5763f81bf77d9da2b2d64ea9aff/frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs#L409-L411

So you could have an back button when coming from the sidebar.
I would propose a mini router like state, where you have something like
```
set-type (mf/use-fn (fn [type] (set-state (assoc state :type type :prev-type  (:type state))))
````

Then you do checks like

```
has-back-button (some? (:prev-type state))

;; Or more explicit
has-back-button? (contains? #{:empty-themes :themes-overview} (:prev-type state))
```

This would be a bit more future-proof if you were to create a theme from anywhere else.

Comment (hirunatan) [reply frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs]: Interesting. Go ahead.

This way you don't specify where do you want to go back. It automatically goes to where it was before (if it's not the first screen).

Comment (floscr) [reply frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs]: Implemented in https://github.com/penpot/penpot/pull/6200/commits/b69df80d850e8c81bee27256b1ffb4af45d7023f

File: frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs [hunk 6]
```diff
@@ -249,16 +248,20 @@
...
-  [{:keys [set-state theme on-save is-editing]}]
+  [{:keys [state change-view theme on-save is-editing]}]
   (let [ordered-token-sets (mf/deref refs/workspace-ordered-token-sets)
         token-sets (mf/deref refs/workspace-token-sets-tree)
 
         current-theme* (mf/use-state theme)
         current-theme (deref current-theme*)
         lib (make-lib-with-theme current-theme ordered-token-sets)
 
+        has-back-button? (contains? #{:empty-themes :themes-overview} (:prev-type state))
```
hirunatan: I'd prefer not leaking state here. My proposal is to add a component parameter like `is-first-view` that is calculated by the caller, looking if `:prev-type` is empty. And here, show the back button only if it's not the first view.

frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
```diff
@@ -384,41 +387,51 @@
...
+                                     {:type :themes-overview}))
+        state       (deref state*)
+
+        change-view (mf/use-fn
+                     (fn [type & [state']]
+                       (swap! state* (fn [current-state]
+                                       (merge
+                                        {:type type
+                                         :prev-type (:type current-state)}
+                                        state')))))
```
hirunatan: This is a bit subtle... I don't like the idea that sub components need to know the structure of the state. Instead I'd prefer something like this:

```clojure
        change-view (mf/use-fn
                     (fn [type & {:keys [:theme-path]}]
                       (swap! state* (fn [current-state]
                                       (cond-> current-state
                                         :always
                                         (assoc :type type)

                                         (= type :edit-theme)
                                         (assoc :theme-path theme-path))))))
```

What do you think? I consider this to be more explicit, and to introduce less coupling between this component and the inner ones.

frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
floscr: Done, I've called it `has-prev-view` as it gets weird with negation with the predicate and consumer props

frontend/src/app/main/ui/workspace/tokens/modals/themes.cljs
floscr: Fine with me :+1:

PR #6203, opened 2025-04-01
:bug: Fix position problems cutting-pasting a component

pabloalba:
Fixes https://tree.taiga.io/project/penpot/issue/10677

### DIFF
--- a/common/src/app/common/files/changes.cljc
+++ b/common/src/app/common/files/changes.cljc
@@ -345,7 +345,9 @@
      [:map {:title "DelComponentChange"}
       [:type [:= :del-component]]
       [:id ::sm/uuid]
-      [:main-instance {:optional true} :any]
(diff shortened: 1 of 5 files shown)


common/src/app/common/files/changes.cljc
```diff
@@ -345,7 +345,7 @@
      [:map {:title "DelComponentChange"}
       [:type [:= :del-component]]
       [:id ::sm/uuid]
-      [:main-instance {:optional true} :any]
+      [:shapes {:optional true} [:vector {:gen/max 3} :any]]
```
niwinz: We should not introduce many more any on specs, this is a big source of problems.
And this looks it expects a vector of the whole shapes, that is bit strange.

niwinz (changes requested):

PR #6204, opened 2025-04-01
:recycle: Refactor heap usage

AzazelN28:
This PR introduces multiple changes in how data is rw from the heap.

### DIFF
--- a/frontend/src/app/render_wasm/api.cljs
+++ b/frontend/src/app/render_wasm/api.cljs
@@ -10,7 +10,6 @@
    ["react-dom/server" :as rds]
    [app.common.data :as d]
    [app.common.data.macros :as dm]
-   [app.common.geom.matrix :as gmt]
    [app.common.math :as mth]
    [app.common.types.shape.layout :as ctl]
    [app.common.types.shape.path :as path]
@@ -20,7 +19,9 @@
    [app.main.refs :as refs]
    [app.main.render :as render]
    [app.render-wasm.api.fonts :as f]
+   [app.render-wasm.deserializers :as dr]
    [app.render-wasm.helpers :as h]
+   [app.render-wasm.mem :as mem]
    [app.render-wasm.serializers :as sr]
    [app.render-wasm.wasm :as wasm]
    [app.util.debug :as dbg]
@@ -28,17 +29,50 @@
    [app.util.webapi :as wapi]
    [beicon.v2.core :as rx]
    [cuerdas.core :as str]
-   [goog.object :as gobj]
    [promesa.core :as p]
    [rumext.v2 :as mf]))
 
 ;; (defonce internal-frame-id nil)
-;; (defonce internal-module #js {})
+;; (defonce wasm/internal-module #js {})
 (defonce use-dpr? (contains? cf/flags :render-wasm-dpr))
+
+;;
+;; List of common entry sizes.
+;;
+;; All of these entries are in bytes so we need to adjust
+;; these values to work with TypedArrays of 32 bits.
+;;
+(def CHILD-ENTRY-SIZE 16)
+(def MODIFIER-ENTRY-SIZE 40)
+(def MODIFIER-ENTRY-TRANSFORM-OFFSET 16)
+(def GRID-LAYOUT-ROW-ENTRY-SIZE 5)
+(def GRID-LAYOUT-COLUMN-ENTRY-SIZE 5)
+(def GRID-LAYOUT-CELL-ENTRY-SIZE 37)
+(def GRADIENT-STOP-SIZE 5)
+
+(defn gradient-stop-get-entries-size
+  [stops]
+  (mem/get-list-size stops GRADIENT-STOP-SIZE))
+
+(defn modifier-get-entries-size
+  "Returns the list of a modifier list in bytes"
+  [modifiers]
+  (mem/get-list-size modifiers MODIFIER-ENTRY-SIZE))
+
+(defn grid-layout-get-row-entries-size
+  [rows]
+  (mem/get-list-size rows GRID-LAYOUT-ROW-ENTRY-SIZE))
+
+(defn grid-layout-get-column-entries-size
+  [columns]
+  (mem/get-list-size columns GRID-LAYOUT-COLUMN-ENTRY-SIZE))
+
+(defn grid-layout-get-cell-entries-size
+  [cells]
+  (mem/get-list-size cells GRID-LAYOUT-CELL-ENTRY-SIZE))
 
 (def dpr
   (if use-dpr? js/window.devicePixelRatio 1.0))
-
 
 ;; Based on app.main.render/object-svg
 (mf/defc object-svg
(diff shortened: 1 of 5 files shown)


frontend/src/app/render_wasm/api.cljs
```diff
@@ -154,28 +188,22 @@
...
-           offset  0]
-      (when-not (empty? entries)
-        (let [id (first entries)]
-          (.set heap (sr/uuid->u8 id) offset)
-          (recur (rest entries) (+ offset ENTRY-SIZE)))))
+        (loop [entries (seq shape-ids)
+               offset  ptr]
+          (when-not (empty? entries)
+            (let [id (first entries)]
+              (sr/uuid->heapu32 id heap (mem/ptr8->ptr32 offset))
```
niwinz: Function names like `uuid->heapu32` induces to think they return heapu32.
I think this kind of functions should be called something like `write-uuid`.

niwinz: I'm also not very convinced about the ptr8->ptr32 approach here. Because it works OK for calculate the correct offset on the u32 heap, but writting can write on non allocated memory, because the original pointer is not correctly aliged.

Imagine you have `((mem/alloc-bytes 4)` and this gives you ptr 5. Once you convert this ptr8 to ptr32, you have `1`. This means you will write the 4 bytes in this way (x represents heap unknown values, a represents the written value):

x x x x a a a a x x x x

But the allocator indicates you need to write this way

x x x x x a a a a x x x

Because the original ptr returned by the alloc operation is 5 and not 4. So, with the proposed approach you have overwritten one byte. Unless the mem/alloc-bytes always allocates with 4 bytes alignment, that is not expected but possible. Can you confirm that?

frontend/src/app/render_wasm/api.cljs
```diff
@@ -777,51 +801,27 @@
...
   [entries]
-  (let [entry-size 40
-        ptr (h/call wasm/internal-module "_alloc_bytes" (* entry-size (count entries)))
-
-        heap
-        (js/Uint8Array.
-         (.-buffer (gobj/get ^js wasm/internal-module "HEAPU8"))
-         ptr
-         (* entry-size (count entries)))]
+  (let [ptr (mem/alloc-bytes-32 (modifier-get-entries-size entries))
```
niwinz: I think we should be consistent, calling it `ptr` or `offset` I think.

frontend/src/app/render_wasm/mem.cljs
```diff
@@ -0,0 +1,53 @@
...
+(defn alloc-bytes-32
+  [size]
+  (when (= size 0)
+    (js/console.trace "tried to allocate 0 bytes"))
+  (let [ptr (h/call wasm/internal-module "_alloc_bytes" size)]
+    (ptr8->ptr32 ptr)))
+
+(defn get-heap-u8
+  []
+  (gobj/get ^js wasm/internal-module "HEAPU8"))
```
niwinz: on this case, lets use `unchecked-get`

frontend/src/app/render_wasm/mem.cljs
```diff
@@ -0,0 +1,53 @@
...
+
+(defn get-heap-u32
+  []
+  (let [heap (gobj/get ^js wasm/internal-module "HEAPU32")]
```
niwinz: The same here, and the `let` is not necessary

frontend/src/app/render_wasm/mem.cljs
```diff
@@ -0,0 +1,53 @@
...
+    heap))
+
+(defn get-heap-f32
+  []
+  (gobj/get ^js wasm/internal-module "HEAPF32"))
```
niwinz: Same here

frontend/src/app/render_wasm/serializers.cljs
```diff
@@ -43,16 +43,24 @@
...
+  [id heap offset]
+  (let [buffer (uuid/get-u32 id)]
+    (aset heap (+ offset 0) (aget buffer 0))
+    (aset heap (+ offset 1) (aget buffer 1))
+    (aset heap (+ offset 2) (aget buffer 2))
+    (aset heap (+ offset 3) (aget buffer 3))
+    buffer))
+
+(defn matrix->heapf32
+  [{:keys [a b c d e f]} heap offset]
```
niwinz: Use dm/get-prop on matrix for faster (0 overhead) access to its props

frontend/src/app/render_wasm/serializers.cljs
```diff
@@ -43,16 +43,24 @@
...
-    (aset f32-arr 0 a)
-    (aset f32-arr 1 b)
-    (aset f32-arr 2 c)
-    (aset f32-arr 3 d)
-    (aset f32-arr 4 e)
-    (aset f32-arr 5 f)
-    (js/Uint8Array. (.-buffer f32-arr))))
+(defn uuid->heapu32
+  [id heap offset]
+  (let [buffer (uuid/get-u32 id)]
```
niwinz, line 49: why not just `(.set mem buffer)` ?

render-wasm/src/mem.rs
```diff
@@ -10,11 +14,17 @@ pub extern "C" fn alloc_bytes(len: usize) -> *mut u8 {
...
+        let layout = Layout::from_size_align_unchecked(len, LAYOUT_ALIGN);
+        let ptr = alloc(layout) as *mut u8;
+        if ptr.is_null() {
+            panic!("Allocation failed");
+        }
+        // TODO: Esto quizá se podría eliminar.
+        ptr::write_bytes(ptr, 0, len);
+        *guard = Some(Box::new(Vec::from_raw_parts(ptr, len, len)));
+        ptr
+    }
```
niwinz, line 27: If I understand it correctly, this makes the allocation always 4 bytes aligned, yes?

niwinz (changes requested):

frontend/src/app/render_wasm/api.cljs
AzazelN28: The `mem/alloc-bytes` always returns a 4 bytes aligned pointer, so it is impossible that the returned `ptr` is 5.

frontend/src/app/render_wasm/api.cljs
AzazelN28: Ok, this is something that was already inconsistent but I think it would be better to stick to one. Do you have some preference? I think I prefer `ptr` or even `address`, what do you think?.

render-wasm/src/mem.rs
AzazelN28: Yes, that's it. This ensures that the allocated memory is always aligned so, when something is allocated, it address starts at a 4 byte aligned address and it has a 4 byte aligned size (allocating 1 byte always returns 4 bytes).

frontend/src/app/render_wasm/api.cljs
niwinz: For me, offset is more clear on meaning, because we are always treats with "array" sometimes is 8bits width sometimes is 32 bits width, and you want to access by offset on that arrays.

render-wasm/src/mem.rs
niwinz: nice, thanks

PR #6209, opened 2025-04-01
:zap: Make feature resolved on team load

niwinz:
### Summary

Mainly simplifies features retrieval to simple get, so we don't need to calculate the team features each time we need that.
There are no user perceived changes, everything should working in the samer way as before.

### How to test

There are no specific thing to test but the list of thinks that affects feature loading:
- Create a file from dashboard
- Open a workspace file
- Open a workspace file on other tab
- Add library to a file
- Open a workspace file with a library

### DIFF
--- a/frontend/src/app/main/data/team.cljs
+++ b/frontend/src/app/main/data/team.cljs
@@ -101,7 +101,7 @@
           (let [permissions (get team :permissions)
                 features    (get team :features)]
             (rx/of #(assoc % :permissions permissions)
-                   (features/initialize (or features #{}))
+                   (features/initialize features)
                    (fetch-members team-id))))))
 
     ptk/EffectEvent
@@ -255,12 +255,12 @@
   (dm/assert! (string? name))
   (ptk/reify ::create-team
     ptk/WatchEvent
-    (watch [it state _]
+    (watch [it _ _]
       (let [{:keys [on-success on-error]
              :or {on-success identity
                   on-error rx/throw}} (meta params)
-            features (features/get-enabled-features state)
-            params {:name name :features features}]
+            features features/global-enabled-features
+            params   {:name name :features features}]
         (->> (rp/cmd! :create-team (with-meta params (meta it)))
              (rx/tap on-success)
              (rx/map team-created)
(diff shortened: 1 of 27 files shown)


niwinz:
The tests are failing randomly for unknown reason of unable to fetch some dependencies. Looks unrelated to PR

frontend/src/app/main/data/team.cljs
```diff
@@ -255,12 +255,12 @@
   (dm/assert! (string? name))
   (ptk/reify ::create-team
     ptk/WatchEvent
-    (watch [it state _]
+    (watch [it _ _]
       (let [{:keys [on-success on-error]
              :or {on-success identity
                   on-error rx/throw}} (meta params)
-            features (features/get-enabled-features state)
```
hirunatan, line 262: I can't see how `features/global-enabled-features` is returning the same values as `features/get-enabled-features`

frontend/src/app/main/features.cljs
```diff
@@ -90,46 +76,28 @@
...
-       (when *assert*
-         (->> (rx/from cfeat/no-migration-features)
-              ;; text editor v2 isn't enabled by default even in devenv
-              ;; wasm render v1 isn't enabled by default even in devenv
-              (rx/filter #(not (or (contains? cfeat/backend-only-features %)
-                                   (= "text-editor/v2" %)
-                                   (= "render-wasm/v1" %)
-                                   (= "design-tokens/v1" %))))
-              (rx/observe-on :async)
-              (rx/map enable-feature))))
```
hirunatan, line 123: With this removed, we will no longer have `no-migration-features` automatically enabled in dev environment?

niwinz: Yes, thats it. It is continue source of confusion that why some feature is autoenabled. It is better just to remove this and if you work with a feature, just enable it. Just look, all ongoing features were excluded from activation

frontend/src/app/main/data/team.cljs
niwinz: They don't return the exact the same, but for team creation that difference has no real difference.

hirunatan: Ok, I'll take it with faith :D

frontend/src/app/main/features.cljs
hirunatan: ok. Only tell people about the change.

hirunatan (approved):

PR #6223, opened 2025-04-03
:sparkles: Optimize profile setup flow for better user experience

mavalroot:
### Related Ticket

Optimize profiling for better user experience [Taiga #10028](https://tree.taiga.io/project/penpot/us/10028)

### Summary

We've updated and simplified the onboarding process. The signup form and the team creation form within onboarding, which previously required two steps, are now combined into a single step

### How to test 

1. Create a new user
2. Fill the onboarding process
 2.1 If you select "work" as your expected use for penpot, there's an extra step (4 in total)
3. You shouldn't be asked to check your newsletter's preferences because you already did that in the creation form
4. Create a team

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,6 @@
 
 ### :sparkles: New features
 
-
 - Update board presets with a newer devices [Taiga #10610](https://tree.taiga.io/project/penpot/us/10610)
 - Propagate "sharing a prototype" to editors and viewers [Taiga #8853](https://tree.taiga.io/project/penpot/us/8853)
 - Design improvements to the Invitations page with an empty state [Taiga #4554](https://tree.taiga.io/project/penpot/us/4554)
@@ -18,7 +17,7 @@
 - Add set selection in create Token themes flow [Taiga #10746](https://tree.taiga.io/project/penpot/issue/10746)
 - Display indicator on not active sets [Taiga #10668](https://tree.taiga.io/project/penpot/issue/10668)
 - Create `input*` wrapper component, and `label*`, `input-field*` and `hint-message*` components [Taiga #10713](https://tree.taiga.io/project/penpot/us/10713)
-- Optimize profiling for better user experience [Taiga #10028](https://tree.taiga.io/project/penpot/us/10028)
+- Optimize profile setup flow for better user experience [Taiga #10028](https://tree.taiga.io/project/penpot/us/10028)
 
 ### :bug: Bugs fixed
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,20 @@
 # CHANGELOG
 
+## 2.8.0 (Next / Unreleased)
+
+### :rocket: Epics and highlights
+
+### :boom: Breaking changes & Deprecations
+
+### :heart: Community contributions (Thank you!)
+
+### :sparkles: New features
+
+- Optimize profile setup flow for better user experience [Taiga #10028](https://tree.taiga.io/project/penpot/us/10028)
+
+### :bug: Bugs fixed
+
+
 ## 2.7.0 (Unreleased)
 
 ### :rocket: Epics and highlights
@@ -17,7 +32,6 @@
 - Add set selection in create Token themes flow [Taiga #10746](https://tree.taiga.io/project/penpot/issue/10746)
 - Display indicator on not active sets [Taiga #10668](https://tree.taiga.io/project/penpot/issue/10668)
 - Create `input*` wrapper component, and `label*`, `input-field*` and `hint-message*` components [Taiga #10713](https://tree.taiga.io/project/penpot/us/10713)
-- Optimize profile setup flow for better user experience [Taiga #10028](https://tree.taiga.io/project/penpot/us/10028)
 
 ### :bug: Bugs fixed
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@
 - Add set selection in create Token themes flow [Taiga #10746](https://tree.taiga.io/project/penpot/issue/10746)
 - Display indicator on not active sets [Taiga #10668](https://tree.taiga.io/project/penpot/issue/10668)
 - Create `input*` wrapper component, and `label*`, `input-field*` and `hint-message*` components [Taiga #10713](https://tree.taiga.io/project/penpot/us/10713)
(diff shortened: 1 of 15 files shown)


CHANGES.md
```diff
@@ -10,6 +10,8 @@
 
 ### :sparkles: New features
 
+- Optimize profiling for better user experience [Taiga #10028](https://tree.taiga.io/project/penpot/us/10028)
```
niwinz: I think profiling has general meaning on performance, so I suggest to change how this named here. 
I suggest something like:  Optimize profile setup flow

niwinz:
The checkboxes looks unaligned.
Also they should be always optional, so when you select it and unselect it, the invalid data error message should not appear i guess.

image

backend/src/app/rpc/commands/auth.clj
```diff
@@ -275,7 +275,9 @@
                               :viewed-walkthrough? false
                               :nudge {:big 10 :small 1}
                               :v2-info-shown true
-                              :release-notes-viewed (:main cf/version)})
+                              :release-notes-viewed (:main cf/version)
+                              :newsletter-news? (or newsletter-news false)
+                              :newsletter-updates? (or newsletter-updates false)})
```
niwinz: This props already exists without `?` on the name, so I think we should respect the previous naming.

frontend/src/app/main/ui/auth/register.cljs
```diff
@@ -18,19 +18,68 @@
...
+
 (def ^:private schema:register-form
   [:map {:title "RegisterForm"}
    [:password ::sm/password]
+   [:fullname [::sm/text {:max 250}]]
    [:email ::sm/email]
-   [:invitation-token {:optional true} ::sm/text]])
+   [:accept-terms-and-privacy {:optional (not (contains? cf/flags :terms-and-privacy-checkbox))}
+    [:and :boolean [:= true]]]
+   [:accept-newsletter-updates {:optional true :default false} [:and :boolean [:= true]]]
```
niwinz: The `:default` has no meaning here on specs. 

Also, I think `[:= true]` is also not necessary, the user should be able to check it and uncheck it (setting false value) with no issue.

frontend/src/app/main/ui/auth/register.cljs
```diff
@@ -65,23 +114,60 @@
...
+           (reset! submitted? true)
+           (let [create-welcome-file?
+                 (cf/external-feature-flag "onboarding-03" "test")
+
+                 params
+                 (cond-> form
+                   create-welcome-file? (assoc :create-welcome-file true))]
+             (->> (rp/cmd! :register-profile params)
+                  (rx/finalize #(reset! submitted? false))
+                  (rx/subs! on-success on-error)))))
```
niwinz: I think you can pass the finalize callback as third argument to rx/subs! instead of rx/finalize here

frontend/src/app/main/ui/auth/register.cljs
```diff
@@ -18,19 +18,68 @@
...
    [app.main.ui.components.link :as lk]
-   [app.main.ui.icons :as i]
    [app.util.i18n :as i18n :refer [tr]]
    [app.util.storage :as storage]
    [beicon.v2.core :as rx]
    [rumext.v2 :as mf]))
 
 ;; --- PAGE: Register
 
+(mf/defc newsletter-options
```
niwinz: New components should follow the new call convention for react components (name ends with `*`, no need for explicit `{::mf/props :obj}`, is now a defaut with components with `*` in the name, and use `[:>` instead of `[:&` for instantiate it)

frontend/src/app/main/ui/onboarding/team_choice.cljs
```diff
@@ -54,27 +51,30 @@
...
+(def ^:private schema:team-form
+  [:map {:title "TeamForm"}
+   [:name [::sm/text {:max 250}]]
+   [:role :keyword]
+   [:emails {:optional true} [::sm/set ::sm/email]]])
+
+(mf/defc team-form
+  {::mf/props :obj
+   ::mf/private true}
+  [{:keys [go-to-team?]}]
```
niwinz: The same here for call conventions (also, no `?` is allowed in component props as convention)
I know that this is not new component, but this PR makes substantial changes to the component, so I think we also can improve it and migrate it to new style.

backend/src/app/rpc/commands/auth.clj
```diff
@@ -375,6 +377,8 @@
                                              (not (contains? cf/flags :email-verification)))
                                params    (-> params
                                              (assoc :is-active is-active)
+                                             (assoc :newsletter-updates accept-newsletter-updates)
+                                             (assoc :newsletter-news accept-newsletter-news)
```
niwinz: With the workflow change on the UI side, I think we need respect the backend API flow, so all this additional parameters can be and I think should be handled on the prepare-register.

niwinz (changes requested):

niwinz:
The registration process was divided into two steps because there are several flows, and not all of them follow the same steps. A clear example is the registration of users with social auth that don't have email verification report (GitHub, for example). With the registration flow via GitHub the profile first is redirected to github, authenticated and then the user is now back redirected to penpot on the second registration step screen, which with this PR no longer exists, and right now, with the changes in this PR, that flow is broken.

In my opinion, this wasn't consistent before either. I think it should have been the same for all social auths, but I don't remember why; it was only done that way for those who didn't verify their email.

niwinz:
I put this PR on draft until is ready again.

mavalroot:
All done!

I restored the second registration step for the social auth that need it, and adjusted it a bit to match the latest style changes. I agree it’d be great to make the whole social auth flow more consistent, but for now I just made sure the GitHub registration is working properly again.

Also cleaned up an extra step that didn’t belong there, and merged the two newsletter checkboxes into just one.

niwinz:
In my personal opinion, all this changes, I think they should be discussed previously to be made, because from start of first version fo this PR all changes indicate that not all login paths are contemplated and taken in account.

I'm still not sure of all this changes. I think further changes should be done in order to make the user experience uniform across all login paths.

Per example: when user goes through not OIDC provider that does not properly report user email verify status, now all that auth process path is very different from the "canonical" and it asks the user for a full name two times (the first register page now has the fullname field, so for github users we ask the fullname on first page and on return from github)

I think for not make this PR last longer unmerged, I think we should meet and properly decide how to proceed.

**IMPORTANT**:  I have rebased this on top of develop and force-pushed the changes

PR #6231, opened 2025-04-04
:sparkles: Don't override user provided color format

### LINKED ISSUE #69: Colors picked from color picker should preserve the users color space

When the user sets a non hex / rgba color space like `hsv`, changing it via the color picker will convert it to rgb and insert an rgb value as the token value instead of the previous color space.

## Steps to reproduce

1. Set a color via the value field like `hsv(1,1,1,0.3)
2. Change the color via the color picker
3. The color value will be updated in `rgba`

_No response_

_No response_

### Additional context

_No response_

### LINKED ISSUE #102: Bug: Out of Sync Token Value & Color Picker
Updating the token value while the color picker is open won't update the value inside the color picker:

https://github.com/user-attachments/assets/40ed373b-f797-4c8b-aad4-99dab9729622

This happens because we keep an internal `ref` in the [ramp](https://github.com/tokens-studio/penpot/blob/develop/frontend/src/app/main/ui/workspace/tokens/form.cljs#L176), as we need more information about hue of the color when picking from a color, as this doesn't get stored in hex format.

Another issue is that the resolved token value has to be computed async, so it's difficult to tell if the resolved async value or the range picker value should take precedence.

We have to resolve the value, as it might contain a reference to another token.

Here's the other way around, with changing the color from the picker, which works of course because the two states line up.

https://github.com/user-attachments/assets/40a00dca-ba55-4b52-8073-d600731bfd26

This gets even more complex when you introduce multiplayer with references:

https://github.com/user-attachments/assets/092e3ee4-32ef-45f1-902b-66a47aabaa36

floscr:
Respects the users color format when changing a color via the color picker for token colors.

https://github.com/user-attachments/assets/90b2d31c-3fbf-48df-9be0-8c96e8c4d876

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/components/controls/input_token_color_bullet.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/components/controls/input_token_color_bullet.cljs
@@ -20,7 +20,8 @@
   {::mf/props :obj
    ::mf/schema schema::input-token-color-bullet}
   [{:keys [color on-click]}]
-  [:div {:class (stl/css :input-token-color-bullet)
+  [:div {:data-testid "token-form-color-bullet"
+         :class (stl/css :input-token-color-bullet)
          :on-click on-click}
    (if-let [color' (wtt/color-bullet-color color)]
      [:> color-bullet {:color color' :mini true}]
--- a/frontend/src/app/main/ui/workspace/tokens/tinycolor.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/tinycolor.cljs
@@ -32,16 +32,23 @@
     (let [tc (tinycolor color-str)]
       (str/starts-with? (.getFormat tc) "hex"))))
 
-(defn ->string [^js tc]
-  (.toString tc))
+(defn ->string
+  "Stringify `tc` to `format`, uses `hex` as per default."
+  [^js tc format]
+  (let [format' (case format
+                  ;; Tinycolor `.toString` doesnt support the `a` suffix it gives you via `.getFormat`
+                  "rgba" "rgb"
+                  "hsva" "hsv"
+                  ;; Keep these formats
+                  "rgb"  "rgb"
+                  "hsv"  "hsv"
+                  ;; Fall back to hex as the default
+                  "hex")]
+    (.toString tc format')))
 
 (defn ->hex-string [^js tc]
   (assert (tinycolor? tc))
   (.toHexString tc))
-
-(defn ->rgba-string [^js tc]
-  (assert (tinycolor? tc))
-  (.toRgbString tc))
 
 (defn color-format [^js tc]
   (assert (tinycolor? tc))
--- a/frontend/playwright/ui/specs/tokens.spec.js
+++ b/frontend/playwright/ui/specs/tokens.spec.js
@@ -371,6 +371,59 @@ test.describe("Tokens: Tokens Tab", () => {
     await expect(tokensTabPanel.getByLabel("color.dark.primary")).toBeEnabled();
   });
 
+  test("User changes color token color while keeping custom color space", async ({
+    page,
+  }) => {
+    const { workspacePage, tokensUpdateCreateModal, tokenThemesSetsSidebar } =
+      await setupEmptyTokensFile(page);
+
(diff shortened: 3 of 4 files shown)


frontend/src/app/main/ui/workspace/tokens/components/controls/input_token_color_bullet.cljs
```diff
@@ -14,14 +14,16 @@
...
 
 (mf/defc input-token-color-bullet*
   {::mf/props :obj
    ::mf/schema schema::input-token-color-bullet}
-  [{:keys [color on-click]}]
+  [{:keys [color on-click data-testid]}]
   [:div {:class (stl/css :input-token-color-bullet)
-         :on-click on-click}
+         :on-click on-click
+         :data-testid data-testid}
```
floscr: This is inlined on components like: https://github.com/tokens-studio/penpot/blob/70f75259649ef2211795fb34bd80f3e7ab25e9d2/frontend/src/app/main/ui/workspace/colorpicker/ramp.cljs#L53
Should we inline it too instead of accepting a prop?

frontend/src/app/main/ui/workspace/tokens/tinycolor.cljs
```diff
@@ -32,17 +32,24 @@
...
+  `.toString` doesn't accept the alpha `a` suffix it gives you via `.getFormat`, so we remove it."
+  {"rgba" "rgb"
+   "rgb"  "rgb"
+   "hsva" "hsv"
+   "hsv"  "hsv"})
+
+(defn ->string
+  "Stringify `tc` to `format` with regards to `stringify-accepted-formats-table`, uses `hex` as per default."
+  [^js tc format]
+  (let [format' (get stringify-accepted-formats-table format "hex")]
```
niwinz: use `case` instead of `hash-map` + `get`

frontend/src/app/main/ui/workspace/tokens/components/controls/input_token_color_bullet.cljs
niwinz: It depends, it is inlined on a components where we expect that component appear only once and on specific condition (in that case, color picker open).
If this input is expected to appear multiple times and on test you want to select one of them, so inlining is not going to be a good solution.

niwinz:
I have rebased it on current develop and looks the issue is still not fixed
[Screencast from 2025-04-11 11-22-42.webm](https://github.com/user-attachments/assets/52d15afa-3a31-4483-a54a-5130c1fbad0f)

niwinz (commented):
-

niwinz (changes requested):
-

niwinz:
I put this PR on draft until is ready again.

floscr:
> I have rebased it on current develop and looks the issue is still not fixed [Screencast from 2025-04-11 11-22-42.webm](https://github.com/user-attachments/assets/52d15afa-3a31-4483-a54a-5130c1fbad0f)

Yeah this issue is because you opened the color picker and then typed your value.

This is a pre-existing & unrelated issue which I didn't want to touch & handle seperately, as this component has been updated & broken so many times in the last couple of weeks.

You can already see the color picker wont pick up new colors when you're updating the value when the picker is open, because the value doesn't get forwarded correctly to the color picker component (which uses a ref inside, so it's a bit complicated to keep them in sync + async resolving of tokens).

With the integration test now, we have some more confidence with fixing this and it wont get broken again, so I could tackle this issue next, but IMO would still keep it separate from this enhancement.

floscr:
I've outlined the issue from above here:
https://github.com/tokens-studio/penpot/issues/102

niwinz:
Still not clear to me if this PR is ready or not.
I have tried that says on the video and i'm unable to reproduce the implemented behavior with the current instructions.

niwinz:
And if you consider this PR is ready again, please, remove the draft. I have missed your responses for 5 days, just because this pr is still on draft and is out of my radar.

floscr:
Sorry for the noise, I'm going to incorporate the changes from above here as well, then everything should work as expected.

floscr:
I've fixed the issue now from above as well here!

PR #6233, opened 2025-04-06
WIP :sparkles: Tokens: Support for shadow opacity

dfelinto:
This patch splits the opacity token in two:

1. Layer (the existing opacity token);
2. Shadows (a new one);

I wasn't sure whether to call it Layer or Shape, but the user guide seemed to be calling it Layer, so I went with it,

Screenshot 2025-04-07 at 00 12 22

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 
 ### :heart: Community contributions (Thank you!)
 
+- [DESIGN TOKENS] Types added: Shadows opacity.
+
 ### :sparkles: New features
 
 ### :bug: Bugs fixed
--- a/common/src/app/common/types/shape/shadow.cljc
+++ b/common/src/app/common/types/shape/shadow.cljc
@@ -33,10 +33,10 @@
 
 (defn update-shadow-opacity [shape value]
   (let [updated-shadows (mapv
-                          (fn [shadow]
-                            (let [shadow-color (:color shadow)
-                                  updated-color (assoc shadow-color :opacity value)]
-                              (assoc shadow :color updated-color)))
-                          (:shadow shape))]
+                         (fn [shadow]
+                           (let [shadow-color (:color shadow)
+                                 updated-color (assoc shadow-color :opacity value)]
+                             (assoc shadow :color updated-color)))
+                         (:shadow shape))]
     ;; Update all shadows in the shape
     (assoc shape :shadow updated-shadows)))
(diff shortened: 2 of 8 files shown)


dfelinto:
This patch is close to complete, but I still need to:
* Test if previously created opacity token are still working.
* Update the tests (I imagine they to be failing at the moment since I renamed opacity to layer-opacity).

I could also leave the old token with its name untouched, but I personally find it a bit confusing.

mikekamminga:
Hi @dfelinto, first of all congrats on being the first one to raise a PR on the just launched design tokens features in Penpot. We didn't expect it so soon, but as a team we're stoked!

However, before you put in a lot of additional work, we'll add some additional context here later today. We are about to publish the roadmap of upcoming feature work which will provide more transparency on how we have currently planned the design tokens features, which includes for this case the shadow token. We have others coming up, typography being a big one.

We are also working towards interoperability and open standards. We're currently closely aligned to the W3C Design Tokens Community Group (DTCG) [design tokens format](https://tr.designtokens.org/format/). We might extend this format where needed, but with the goal for that to be adopted in the spec.

I'll come back later with some more details, specifically around what's planned for shadows. Hopefully that also solves your use-case.

Cheers,
Mike

dfelinto:
Hi @mikekamminga, thanks for your reply.

> We didn't expect it so soon, but as a team we're stoked!

I was actually planning to send as a PR while tokens were still on the `develop` branch, but couldn't set enough time aside. So I was the one surprised to see this landing into a release that soon :) Great job with the initial work.

> I'll come back later with some more details, specifically around what's planned for shadows. Hopefully that also solves your use-case.

When I first looked at the design tokens format I didn't dig enough to see if the types definition were part of the specs. It seems to be and indeed shadow itself is a type there: https://tr.designtokens.org/format/#composite-types

I guess shadow is particularly tricky because is one of the "composite" types that may apply to an array even.
Anyways, looking forward to your plans in that regard.

> Hopefully that also solves your use-case.

I'm sure it will. My [use-case](https://community.penpot.app/t/tokens-initial-impressions/8290) is rather simple compared to web-dev.

> However, before you put in a lot of additional work, we'll add some additional context here later today.  (...)

Don't worry much about the time I spent making this. It is fun to try to get something working, regardless.

I will just close it for now, so it doesn't pollute the list of existing patches. When you publish the design/roadmap I can always revisit this (and re-open) if relevant (and not already tackled by your team).

PR #6241, opened 2025-04-08
:recycle: Refactor how rAF/cAF are handled

AzazelN28:
### Summary

Instead of calling `run_script` or `run_script_int`, we should define our own JS API that's called from Rust code.

### Checklist

### DIFF
--- a/render-wasm/src/js/wapi.js
+++ b/render-wasm/src/js/wapi.js
@@ -1,0 +1,8 @@
+addToLibrary({
+  wapi_requestAnimationFrame: function wapi_requestAnimationFrame() {
+    return window.requestAnimationFrame(Module._process_animation_frame);
+  },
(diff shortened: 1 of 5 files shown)


render-wasm/src/js/wapi.js
```diff
@@ -0,0 +1,10 @@
+addToLibrary({
+  wapi_requestAnimationFrame: function wapi_requestAnimationFrame() {
+    console.log('wapi_requestAnimationFrame', Module._process_animation_frame)
```
belen-albeza: I think this console.log and the other one below are leftover, right ?

belen-albeza (approved):
Approving, pls note the CI needs fixing (it seems the format is off) and there's leftover console.log. Thanks!

PR #6252, opened 2025-04-09
:bug: Fix error when an external user tries to export something from a prototype using a shared link

superalex:
Related to: https://tree.taiga.io/project/penpot/issue/10251

### DIFF
--- a/backend/src/app/rpc/commands/files.clj
+++ b/backend/src/app/rpc/commands/files.clj
@@ -474,7 +474,7 @@
   (update page :objects update-vals #(dissoc % :thumbnail)))
 
 (defn get-page
-  [{:keys [::db/conn] :as cfg} {:keys [profile-id file-id page-id object-id] :as params}]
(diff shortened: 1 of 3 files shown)


backend/src/app/rpc/commands/files.clj
```diff
@@ -474,30 +474,40 @@
...
               :hint "page-id is required when object-id is provided"))
 
-  (let [team (teams/get-team conn
-                             :profile-id profile-id
-                             :file-id file-id)
+  (let [perms (get-permissions conn profile-id file-id share-id)
+
+        file  (get-file cfg file-id)
+
+        proj  (db/get* conn :project {:id (:project-id file)}
```
niwinz: I think this should be normal get instead of get* and without the `{::db/remove-deleted ...}`
The same for the team

superalex: Thank you very much!

PR #6257, opened 2025-04-09
:sparkles: Add, edit and delete variant properties from layer panel

luisddm:
### Related Ticket

Implements Taiga task [#10665](https://tree.taiga.io/project/penpot/task/10665).

### Summary

The user is now able to update the variant properties from the layers panel, using the formula `[property_name]=[value]`. This implies removing a value, editing a value and add a property with a value. 

If the formula is malformed, all the values of the properties for this component will be set to a blank value. The error management will be addressed in a different task.

### DIFF
--- a/common/src/app/common/logic/variant_properties.cljc
+++ b/common/src/app/common/logic/variant_properties.cljc
@@ -59,14 +59,14 @@
 
 
 (defn generate-add-new-property
-  [changes variant-id & {:keys [fill-values?]}]
+  [changes variant-id & {:keys [fill-values? property-name]}]
   (let [data               (pcb/get-library-data changes)
         objects            (pcb/get-objects changes)
         related-components (cfv/find-variant-components data objects variant-id)
 
         props              (-> related-components first :variant-properties)
         next-prop-num      (ctv/next-property-number props)
-        property-name      (str ctv/property-prefix next-prop-num)
+        property-name      (d/nilv property-name (str ctv/property-prefix next-prop-num))
 
         [_ changes]
         (reduce (fn [[num changes] component]
--- a/common/src/app/common/logic/variant_properties.cljc
+++ b/common/src/app/common/logic/variant_properties.cljc
@@ -66,7 +66,7 @@
 
         props              (-> related-components first :variant-properties)
         next-prop-num      (ctv/next-property-number props)
-        property-name      (d/nilv property-name (str ctv/property-prefix next-prop-num))
+        property-name      (or property-name (str ctv/property-prefix next-prop-num))
 
         [_ changes]
         (reduce (fn [[num changes] component]
--- a/frontend/src/app/main/data/workspace/variants.cljs
+++ b/frontend/src/app/main/data/workspace/variants.cljs
@@ -32,6 +32,78 @@
    [beicon.v2.core :as rx]
    [potok.v2.core :as ptk]))
 
+(defn- find-properties-to-remove
+  [prev-props upd-props]
+  (filterv #(not (some (fn [prop] (= (:name %) (:name prop))) upd-props)) prev-props))
+
+(defn- find-properties-to-update
+  [prev-props upd-props]
+  (filterv #(some (fn [prop] (and (= (:name %) (:name prop))
+                                  (not= (:value %) (:value prop)))) prev-props) upd-props))
+
+(defn- find-properties-to-add
+  [prev-props upd-props]
+  (filterv #(not (some (fn [prop] (= (:name %) (:name prop))) prev-props)) upd-props))
+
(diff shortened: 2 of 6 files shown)


common/src/app/common/logic/variant_properties.cljc
```diff
@@ -59,14 +59,14 @@
...
-  [changes variant-id & {:keys [fill-values?]}]
+  [changes variant-id & {:keys [fill-values? property-name]}]
   (let [data               (pcb/get-library-data changes)
         objects            (pcb/get-objects changes)
         related-components (cfv/find-variant-components data objects variant-id)
 
         props              (-> related-components first :variant-properties)
         next-prop-num      (ctv/next-property-number props)
-        property-name      (str ctv/property-prefix next-prop-num)
+        property-name      (d/nilv property-name (str ctv/property-prefix next-prop-num))
```
pabloalba: For performance, d/nilv should be used only when the second value is a constant
Here is better an or:
```
(or property-name (str ctv/property-prefix next-prop-num))
```

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -32,6 +32,78 @@
    [beicon.v2.core :as rx]
    [potok.v2.core :as ptk]))
 
+(defn- find-properties-to-remove
+  [prev-props upd-props]
+  (filterv #(not (some (fn [prop] (= (:name %) (:name prop))) upd-props)) prev-props))
```
pabloalba: This isn't very efficient, because it needs to go trough all the upd-props for each prev-prop.
I would do it with a set, something like:

```
(let [upd-names (set (map :name upd-props))]
    (filterv #(not (contains? upd-names (:name %))) prev-props)))
```

pabloalba: Same for the other two helpers

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -32,6 +32,78 @@
...
+
+(defn- find-properties-to-add
+  [prev-props upd-props]
+  (filterv #(not (some (fn [prop] (= (:name %) (:name prop))) prev-props)) upd-props))
+
+(defn- find-property-index [props name]
+  (some (fn [[idx prop]]
+          (when (= (:name prop) name)
+            idx))
+        (keep-indexed vector props)))
```
pabloalba: keep-indexed is meant for filtering and transforming at the same time
this doesn't transform, so it is better to use map-indexed

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -32,6 +32,78 @@
...
+            changes (-> (pcb/empty-changes it page-id)
+                        (pcb/with-objects objects)
+                        (pcb/with-library-data data))
+
+            changes (reduce
+                     (fn [changes [_ {:keys [name]}]]
+                       (-> changes
+                           (clvp/generate-update-property-value component-id (find-property-index previous-properties name) "")))
+                     changes
+                     (map-indexed vector properties-to-remove))
```
pabloalba: Why add an index, just to ignore it?

frontend/src/app/main/ui/workspace/sidebar/layer_item.cljs
```diff
@@ -54,10 +55,16 @@
...
         main-instance?        (:main-instance item)
+
         variants?             (features/use-feature "variants/v1")
         is-variant?           (when variants? (ctk/is-variant? item))
+        is-variant-container? (when variants? (ctk/is-variant-container? item))
+        variant-id            (when is-variant? (:variant-id item))
         variant-name          (when is-variant? (:variant-name item))
-        is-variant-container? (when variants? (ctk/is-variant-container? item))]
+
+        fdata                 (:data (deref refs/file))
```
pabloalba: Is better: `(deref refs/workspace-data)`

frontend/src/app/main/ui/workspace/sidebar/layer_name.cljs
```diff
@@ -54,13 +60,17 @@
 
         accept-edit
         (mf/use-fn
-         (mf/deps shape-id on-stop-edit)
+         (mf/deps shape-id on-stop-edit variant-properties)
```
pabloalba: Also `variant-name`, `component-id`, `variant-id`

frontend/src/app/main/data/workspace/variants.cljs
luisddm: I applied that optimization for both remove and add helpers, but didn't know how to improve it for the update helper.

PR #6263, opened 2025-04-09
:zap: Performance optimizations to path related functions

niwinz:
### Summary

This PR includes a several optimizations and code cleaning related to path data type functions.
It is a continuation of: https://github.com/penpot/penpot/pull/6170

**Relevant changes introduced on PathData type:**

- Add the ability to walk, reduce and lookup segmens with internal reduction (avoid the creation of garbage, and  allow implement algorithms that traverse or access segment data directly without converting it to a plain format)
- Add PathData sepecific cache mechanism (with GC driven eviction mechanism using custom Map type based internaly on WeakRef and FinalizationRegistry)
- Coerce all PathData returned values on both, plain format and internal iteration that removes strange corner cases of hash and equality missmatch on JVM)

**Relevant changes introduced to PathData type surrounding functions and UI:**

- Efficiency refactor of path editor components
- Performance improvement on `get-points` (replacement for content->points, heavy used on path edition, near 80% speed improvement without taking in account the cache)
- Performance improvement on `calculate-extremities` (heavy use on path editor, nearly 50% speed improvement)
- Performance improvement on `handler->point` (heavy use on path editor, nearly 50% speed improvement and reduction of gc generation)
- Performance improvement on `get-handlers` (heavy use on path editor, nearly 65% speed improvement and reduction of gc generation)
- Remove several functions and namespaces with legacy code
- Apply PathData internal formatting mechanism on outline and path editor taking advantage of performance improvements of the path data string formatting mechanism of the new type instead of recurring to plain format (reduces the generation of gc) 

### How to test it

It involves the path creation and edition, so any operation that does that is susceptible to be broken. Lets enumerate a several actions:

- Create a path drawing a curve
- Create a path by path tool, settings points and curves
- Creating modify a bool shape
- Update path by editing points or adding new points
- Exportation of path shapes should work as expected
- Exporting and importing files with path and bool shapes


backend/src/app/features/fdata.clj
```diff
@@ -30,7 +32,7 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (defn enable-objects-map
-  [file]
+  [file & _opts]
```
Alotor, line 36: Why add a new parameter if you're not using it?

backend/src/app/features/fdata.clj
```diff
@@ -136,10 +138,39 @@
 
 (defn enable-pointer-map
   "Enable the fdata/pointer-map feature on the file."
-  [file]
+  [file & _opts]
```
Alotor, line 142: Idem

common/src/app/common/geom/point.cljc
```diff
@@ -470,6 +470,11 @@
   (and ^boolean (mth/almost-zero? (dm/get-prop p :x))
        ^boolean (mth/almost-zero? (dm/get-prop p :y))))
 
+(defn zero?
+  [p]
+  (and ^boolean (= 0 (dm/get-prop p :x))
```
Alotor: Don't use equality it's going to be a problem with floats. There is a method in `math` called `almost-zero?` that checks for values very close to zero

frontend/src/app/main/data/workspace/path/edition.cljs
```diff
@@ -256,22 +258,22 @@
             start-delta-y (dm/get-in modifiers [index cy] 0)
 
             content (st/get-path state :content)
-            points (upg/content->points content)
+            points  (path.segment/get-points content)
 
-            point (-> content (get (if (= prefix :c1) (dec index) index)) (upc/command->point))
-            handler (-> content (get index) (upc/get-handler prefix))
+            point (-> content (nth (if (= prefix :c1) (dec index) index)) (path.helpers/segment->point))
```
Alotor, line 262: You've changed `get` for `nth`. It's fine but `nth` is more prone to throw an exception. `get` if the index is "out of bounds" will return null, but nth will crash.

frontend/src/app/main/ui/shapes/path.cljs
```diff
@@ -7,28 +7,35 @@
 (ns app.main.ui.shapes.path
   (:require
    [app.common.logging :as log]
+   [app.common.types.path :as path]
    [app.main.ui.shapes.custom-stroke :refer [shape-custom-strokes]]
-   [app.util.object :as obj]
-   [app.util.path.format :as upf]
    [rumext.v2 :as mf]))
 
+(defn- content->string
```
Alotor, line 14: Is this the better place for this function? don't we need it elsewhere?

frontend/src/app/plugins/shape.cljs
```diff
@@ -1018,7 +1018,7 @@
                  (u/display-not-valid :makeMask (:type shape))
 
                  :else
-                 (upf/format-path (:content shape)))))
+                 (str (:content shape)))))
```
Alotor: better to use `dm/str`?? in other places you've used `.toString` instead of plain `str`

frontend/src/app/plugins/shape.cljs
```diff
@@ -1309,21 +1309,22 @@
          (cond-> (or (cfh/path-shape? data) (cfh/bool-shape? data))
            (crc/add-properties!
             {:name "content"
-             :get #(-> % u/proxy->shape :content upf/format-path)
+             :get #(-> % u/proxy->shape :content str)
```
Alotor: same here. should we call to `.toString` instead.

backend/src/app/features/fdata.clj
niwinz: Because this functions are designed to be used in srepl, with process-file or process-files and this srepl helpers pass an extra parameter: `opts` that in this case is unused.

backend/src/app/features/fdata.clj
niwinz: The same

common/src/app/common/geom/point.cljc
niwinz: I will check it. Thanks

frontend/src/app/main/data/workspace/path/edition.cljs
niwinz: The PathData structure does not implement IPersistentCollection which allows the usage of `get`. This is because some file operations uses walk to iterate over all objects in file data. And we don't want PathData treated like a persistent collection so walking traversal is not going deeper on PathData segments on walk operation.
For compensate this, the `nth` on PathData does not trigger index error.

In any case, i will recheck it again for ensure this works as I'm saying

frontend/src/app/main/ui/shapes/path.cljs
niwinz: This function is specific on this namespace and only used here, because the normal process is just doing the `.toString`. This function add some safety checks for the render of the component purpose.

frontend/src/app/plugins/shape.cljs
niwinz: Just used str for cosmetic look and feel. I will revisit the usage of str and .toString for make it consistent on the code, maybe making a specific function for it just for syntactic convenience...

common/src/app/common/geom/point.cljc
niwinz: In this concrete case, i need an exact 0 checking. I fixed the jvm float vs long compare using type independent equality function: https://clojure.github.io/clojure/clojure.core-api.html#clojure.core/==

frontend/src/app/main/data/workspace/path/edition.cljs
niwinz: I confirm, nth does not throw on Path Data

frontend/src/app/plugins/shape.cljs
niwinz: Fixed

frontend/src/app/plugins/shape.cljs
niwinz: Fixed

PR #6266, opened 2025-04-10
:bug: Fix unions for new render

superalex:
Related to: 
- https://tree.taiga.io/project/penpot/task/10736
- https://tree.taiga.io/project/penpot/task/10819

### DIFF
--- a/render-wasm/src/shapes/paths.rs
+++ b/render-wasm/src/shapes/paths.rs
@@ -76,18 +76,6 @@ pub struct Path {
     open: bool,
 }
 
-fn starts_and_ends_at_same_point(path: &skia::Path) -> bool {
-    if path.count_points() < 2 {
-        return false; // A path with fewer than 2 points cannot be closed
-    }
-
-    let start_point = path.get_point(0); // First point of the path
-    let end_point = path.get_point(path.count_points() - 1); // Last point of the path
-
-    // Compare the start and end points
-    start_point == end_point
-}
-
 impl Default for Path {
     fn default() -> Self {
         Path::try_from(Vec::new()).unwrap()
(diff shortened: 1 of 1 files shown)


render-wasm/src/shapes/paths.rs
```diff
@@ -76,24 +76,21 @@ pub struct Path {
...
-    start_point == end_point
-}
-
 impl Default for Path {
     fn default() -> Self {
         Path::try_from(Vec::new()).unwrap()
     }
 }
 
+fn maybe_close(skia_path: &mut skia::Path, start: Option<Point>, destiny: Point, open: &mut bool) {
```
belen-albeza: typo: `destiny` -> `destination` (or `end`)

superalex: Thank you very much! <3

render-wasm/src/shapes/paths.rs
```diff
@@ -76,24 +76,26 @@ pub struct Path {
...
     fn default() -> Self {
         Path::try_from(Vec::new()).unwrap()
     }
 }
 
+fn maybe_close(
+    skia_path: &mut skia::Path,
+    start: Option<Point>,
+    destination: Point,
+    open: &mut bool,
```
belen-albeza: This is a bit of a nitpick but mutating two different things feels odd. Could we we mutate just the `skia_path` and return the new `open` value?

belen-albeza (approved):

render-wasm/src/shapes/paths.rs
superalex: We could but I was trying to avoid something like this (same if condition, open must be true until one of the calls to maybe_close returns false):
```
        Segment::LineTo(xy) => {
            skia_path.line_to(xy);
            if !maybe_close(&mut skia_path, start, xy) {
                open = false;
            }
        }
        Segment::CurveTo((c1, c2, xy)) => {
            skia_path.cubic_to(c1, c2, xy);
            if !maybe_close(&mut skia_path, start, xy) {
                open = false;
            }
        }
```

Another option to avoid the duplicated if could be having something like this (but it has an anidated match):
```
        Segment::LineTo(xy) | Segment::CurveTo((_, _, xy)) => {
            match *segment {
                Segment::LineTo(xy) => skia_path.line_to(xy),
                Segment::CurveTo((c1, c2, xy)) => skia_path.cubic_to(c1, c2, xy),
                _ => unreachable!(),
            }
            if !maybe_close(&mut skia_path, start, xy) {
                open = false;
            }
        }
```

Another less idiomatic one could be having an extra function
```
fn handle_draw(
    skia_path: &mut skia::Path,
    start: Option<Point>,
    destination: Point,
    draw: impl FnOnce(&mut skia::Path),
) -> bool {
    draw(skia_path);
    maybe_close(skia_path, start, destination)
}
```
and then something like:
```
        Segment::LineTo(xy) => {
            if !handle_draw(&mut skia_path, start, xy, |p| p.line_to(xy)) {
                open = false;
            }
        }
        Segment::CurveTo((c1, c2, xy)) => {
            if !handle_draw(&mut skia_path, start, xy, |p| p.cubic_to(c1, c2, xy)) {
                open = false;
            }
        }
```

But all these options require more code or adding duplicated one :(

superalex: I think I found a solution that I like!, let me know what you think, thank you very much! :)

PR #6277, opened 2025-04-10
:bug: Fix token color reference issue

floscr:
### Related Ticket

https://github.com/tokens-studio/penpot/issues/58

### Summary

The [sd-transforms](https://github.com/tokens-studio/sd-transforms) package comes with some default transforms that are buggy for our use-case of a runtime resolver.
We'll fix it in the package as well, but we don't need the CSS transforms for our needs, so I've removed it from the transforms.

Here's the Issue in sd-transforms: https://github.com/tokens-studio/sd-transforms/issues/353

I've Updated sd-transforms package as well while visiting it.

### Steps to reproduce 

1. Create two sets `Foo` & `Bar`
2. Create a color token in `Foo` called `color.foo` with the value of `black`
3. Create a color token in `Bar` called `background` with the value of `{color.foo}`
4. Try to rename `color.foo` to `color.foo.100` or something else after the `foo` group and it wouldn't work

### Checklist

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -105,7 +105,7 @@
     "@penpot/mousetrap": "portal:./vendor/mousetrap",
     "@penpot/svgo": "penpot/svgo#v3.1",
     "@penpot/text-editor": "portal:./text-editor",
-    "@tokens-studio/sd-transforms": "1.2.11",
+    "@tokens-studio/sd-transforms": "1.2.12",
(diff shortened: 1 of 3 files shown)


floscr:
I'm actually closing this in favor of upgrading StyleDictionary to v5.0.0, as it resolves this issue and adds better error handling for such cases.

PR #6278, opened 2025-04-10
:wrench: Parse text leaves all at once

elenatorro:
### Related Ticket

https://tree.taiga.io/project/penpot/task/10382

### Summary

This PR parses all text leaves at once. It adds the basic functionality to parse current leaf attributes, except fills and not taking into account some paragraph attributes such strokes. This will be iterated in upcoming steps.

[screen-recorder-wed-apr-16-2025-11-46-27.webm](https://github.com/user-attachments/assets/8fce638f-f2d9-4dbb-8812-d987bc34cb46)

### Checklist

### DIFF
--- a/frontend/src/app/render_wasm/api/texts.cljs
+++ b/frontend/src/app/render_wasm/api/texts.cljs
@@ -1,0 +1,105 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC
+
+(ns app.render-wasm.api.texts
+  (:require
+   [app.render-wasm.api.fonts :as f]
+   [app.render-wasm.helpers :as h]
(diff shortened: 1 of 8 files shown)


frontend/src/app/render_wasm/api/texts.cljs
```diff
@@ -0,0 +1,105 @@
...
+        paragraph-attr-size 48
+        leaf-attr-size 52
+        metadata-size (+ 1 paragraph-attr-size (* num-leaves leaf-attr-size))
+        text (apply str (map :text leaves))
+        text-buffer (utf8->buffer text)
+        text-size (.-byteLength text-buffer)
+        buffer (js/ArrayBuffer. (+ metadata-size text-size))
+        dview (js/DataView. buffer)]
+
+    (.setUint8 dview 0 num-leaves)
```
AzazelN28: I think 255 aren't enough leaves for complex texts, I think it would be better to make this a Uint32 value.

This will also make the following memory 4-byte aligned, making it easier (in the future) to use the heap32 direct access instead of using a DataView.

elenatorro: Yes! You're right, done ✅ much better now

PR #6288, opened 2025-04-11
:sparkles: Copy values of same named properties moving a variant into…

pabloalba:
Implements https://tree.taiga.io/project/penpot/task/10785

### Checklist

### DIFF
--- a/common/src/app/common/logic/variant_properties.cljc
+++ b/common/src/app/common/logic/variant_properties.cljc
@@ -167,8 +167,10 @@
             (generate-new-properties-from-variant shape min-props data container-name base-props)
             (generate-new-properties-from-non-variant shape min-props container-name base-props)))
 
-        total-props    (apply max (map #(-> (generate-new-properties % num-base-props) count)
-                                       shapes))
+        total-props    (reduce (fn [m shape]
+                                 (max m (count (generate-new-properties shape num-base-props))))
+                               0
+                               shapes)
 
         num-new-props  (if (or (zero? num-base-props)
                                (< total-props num-base-props))
@@ -200,4 +202,3 @@
                                   #(assoc % :variant-name variant-name))))))
      changes
      shapes)))
-
--- a/common/src/app/common/logic/variant_properties.cljc
+++ b/common/src/app/common/logic/variant_properties.cljc
@@ -117,86 +117,81 @@
   [changes shapes]
   (reduce generate-make-shape-no-variant changes shapes))
 
+
+(defn- generate-new-properties-from-variant
+  [shape min-props data container-name base-properties]
+  (let [component (ctcl/get-component data (:component-id shape) true)
+        add-name? (not= (:name component) container-name)
+        props     (ctv/merge-properties base-properties
+                                        (:variant-properties component))
+        new-props (- min-props
+                     (+ (count props)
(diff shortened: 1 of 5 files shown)


common/src/app/common/types/variant.cljc
```diff
@@ -74,22 +74,33 @@
...
+  (conj props {:name (str property-prefix (next-property-number props))
+               :value value}))
+
+(defn add-new-props
+  "Adds new properties with generated names and provided values to the existing props list."
+  [props values]
+  (let [next-prop-num (next-property-number props)
+        new-props (map-indexed (fn [i v] {:name (str property-prefix (+ next-prop-num i))
+                                          :value v}) values)]
+    (into props new-props)))
```
niwinz: Can be easily implemented as transducer.

common/src/app/common/types/variant.cljc
```diff
@@ -147,3 +158,73 @@
...
+
+(defn- matching-indices
+  "Returns a set of indices in props1 where the property name exists in props2."
+  [props1 props2]
+  (let [names-in-p2 (set (map :name props2))]
+    (->> props1
+         (map-indexed vector)
+         (filter #(contains? names-in-p2 (:name (second %))))
+         (map first)
+         set)))
```
niwinz: This is a bit more verbose, but more efficient impl for this:

```clojure
(def ^:private xf:map-name)

(defn- matching-indices
  "Returns a set of indices in props1 where the property name exists in props2."
  [props1 props2]
  (let [names-in-p2
        (into #{} xf:map-name props2)

        xform
        (comp
         xf:map-name
         (map-indexed (fn [[index name]]
                        (when (contains? names-in-p2 name)
                          index)))
         (filter some?))]

    (into #{} xform props1)))
```

common/src/app/common/logic/variant_properties.cljc
```diff
@@ -117,86 +117,81 @@
...
+        [cpath cname]  (cfh/parse-path-name (:name variant-container))
+        container-name (:name variant-container)
+
+        generate-new-properties
+        (fn [shape min-props]
+          (if (ctk/is-variant? shape)
+            (generate-new-properties-from-variant shape min-props data container-name base-props)
+            (generate-new-properties-from-non-variant shape min-props container-name base-props)))
+
+        total-props    (apply max (map #(-> (generate-new-properties % num-base-props) count)
```
niwinz: is more efficient to implement it using reduce.

common/src/app/common/logic/variant_properties.cljc
```diff
@@ -117,86 +117,81 @@
...
+                                       shapes))
+
+        num-new-props  (if (or (zero? num-base-props)
+                               (< total-props num-base-props))
+                         0
+                         (- total-props num-base-props))
+
+        changes        (nth
+                        (iterate #(generate-add-new-property % variant-id) changes)
+                        num-new-props)
```
niwinz, line 182: `nth` on sequence will iterate over all sequence until the index. This is just information, I don't know the algorithm behind this code so just comment it to you to be aware of it.

niwinz (commented):
I added several performance oriented comments, but as I don't have a big context of this feature so I just suggest them. They can be done on this PR, an other PR later or revisited later in a unknown future.

PR #6298, opened 2025-04-14
:sparkles: Send fill + Stops in a single wasm call

belen-albeza:
Closes https://tree.taiga.io/project/penpot/task/10752

Screenshot 2025-04-15 at 12 43 52 PM

### DIFF
--- a/frontend/src/app/render_wasm/serializers/color.cljs
+++ b/frontend/src/app/render_wasm/serializers/color.cljs
@@ -1,0 +1,22 @@
+(ns app.render-wasm.serializers.color
+  (:require
+   [app.common.math :as mth]))
+
+(defn hex->u32argb
+  "Takes a hex color in #rrggbb format, and an opacity value from 0 to 1 and returns its 32-bit argb representation"
+  [hex opacity]
+  (let [rgb (js/parseInt (subs hex 1) 16)
+        a (mth/floor (* (or opacity 1) 0xff))]
+        ;; rgba >>> 0 so we have an unsigned representation
+    (unsigned-bit-shift-right (bit-or (bit-shift-left a 24) rgb) 0)))
+
+(defn rgba-bytes-from-hex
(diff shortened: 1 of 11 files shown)


frontend/src/app/render_wasm/serializers/fills.cljs
```diff
@@ -0,0 +1,33 @@
...
+   [app.common.data.macros :as dm]
+   [app.render-wasm.serializers.color :as clr]))
+
+(def GRADIENT-STOP-SIZE 8)
+(def GRADIENT-BASE-SIZE 24)
+
+(defn serialize-gradient-fill
+  [gradient opacity heap offset]
+  (let [dview (js/DataView. (.-buffer heap))
+        start-x (dm/get-prop gradient :start-x)
```
niwinz: Don't use dm/get-prop here, is known that gradient is plain map so `get` is better here

frontend/src/app/render_wasm/serializers/color.cljs
```diff
@@ -0,0 +1,22 @@
...
+
+(defn hex->u32argb
+  "Takes a hex color in #rrggbb format, and an opacity value from 0 to 1 and returns its 32-bit argb representation"
+  [hex opacity]
+  (let [rgb (js/parseInt (subs hex 1) 16)
+        a (mth/floor (* (or opacity 1) 0xff))]
+        ;; rgba >>> 0 so we have an unsigned representation
+    (unsigned-bit-shift-right (bit-or (bit-shift-left a 24) rgb) 0)))
+
+(defn rgba-bytes-from-hex
```
niwinz: This function was private and now looks unused. I'm wrong?

belen-albeza: Yep, just removed it in a follow up commit 😅

belen-albeza:
Updated the PR with the latest changes discussed in this morning's call re: naming and args convention

PR #6336, opened 2025-04-21
:sparkles: Selection of several components of a variant ar the same time

luisddm:
### Related Ticket

Implements Taiga [#10664](https://tree.taiga.io/project/penpot/task/10664)

### Summary

When several variant components are selected, the view in the design tab shows all of the properties and their values. If for one property there have mixed values, the combobox showcases the “Mixed” label, and the user can override both by typing a new value or by selecting one from the selector.

This only works when the user is selecting multiple main components belonging to the same variant.

### DIFF
--- a/common/src/app/common/types/variant.cljc
+++ b/common/src/app/common/types/variant.cljc
@@ -231,3 +231,19 @@
          {:props (vec props1) :used-pos (matching-indices props1 props2)}
          props2)
         :props)))
+
+(defn compare-properties
+  "Compares vectors of properties keeping the value if it is the same for all
+   or setting a custom value where their values do not coincide"
+  ([props-list]
+   (compare-properties props-list nil))
+
+  ([props-list distinct-mark]
+   (let [grouped (group-by :name (apply concat props-list))
+         check-values (fn [values]
+                        (if (apply = (map :value values))
+                          (first (map :value values))
+                          distinct-mark))]
+     (mapv (fn [[name values]]
+             {:name name :value (check-values values)})
+           grouped))))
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
```diff
@@ -641,6 +665,12 @@
         is-variant?     (when variants? (ctk/is-variant? component))
         main-instance?  (ctk/main-instance? shape)
 
+        components      (mapv #(ctf/resolve-component %
+                                                      current-file
+                                                      libraries
+                                                      {:include-deleted? true}) shapes)
+        same-variant?   (= 1 (count (distinct (map :variant-id components))))
```
pabloalba: The same-variant check is used twice, and may be useful in other places. I would move it to types/variant

common/src/app/common/types/variant.cljc
```diff
@@ -230,4 +230,20 @@
...
+  "Compares vectors of properties keeping the value if it is the same for all
+   or setting a custom value where their values do not coincide"
+  ([props-list]
+   (compare-properties props-list nil))
+
+  ([props-list distinct-mark]
+   (let [grouped (group-by :name (apply concat props-list))
+         check-values (fn [values]
+                        (if (apply = (map :value values))
+                          (first (map :value values))
```
pabloalba: You're calling (map :value values) twice. 

(let [vals (map :value values)]
  (if (apply = vals)
    (first vals)
    distinct-mark))

frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
luisddm: Extracted to types/variant with an additional check and added unit test.

pabloalba (approved):

PR #6368, opened 2025-04-24
🐛 Handle swapped nested instances when detaching

hirunatan:
**DO NOT SQUASH COMMITS WHEN MERGING**

### Related Ticket

[https://tree.taiga.io/project/penpot/issue/10837](https://tree.taiga.io/project/penpot/issue/10837)

### Summary

A bug occurs when user tries to detach a copy and all those things occur:
- The copy has the main instance in a library that is not attached to the file (so the main instance cannot be located).
- There are two levels of nested copies.
- The copies of the second level have been swapped to a component in a library that is attached to the file (so the new main instance can be located).

In this case the algorithm will try to keep the second level instances as copies, but it fails because they are not root copies.

### Steps to reproduce 

See Taiga issue.

### Checklist

- [*] Choose the correct target branch; use `develop` by default.
- [*] Provide a brief summary of the changes introduced.
- [*] Add a detailed explanation of how to reproduce the issue and/or verify the fix, if applicable.
- [*] Include screenshots or videos, if applicable.
- [*] Check CI passes successfully.
- [*] Update the `CHANGES.md` file, referencing the related GitHub issue, if applicable.

### DIFF
--- a/common/src/app/common/logic/libraries.cljc
+++ b/common/src/app/common/logic/libraries.cljc
@@ -36,6 +36,8 @@
 ;; Change this to :info :debug or :trace to debug this module, or :warn to reset to default
 (log/set-level! :warn)
 
+;; Add uuids here to filter logs to only show specific shapes or containers (and all shapes
+;; contained in them).
 (def log-shape-ids #{})
 (def log-container-ids #{})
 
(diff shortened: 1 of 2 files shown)


common/src/app/common/logic/libraries.cljc
```diff
@@ -338,42 +339,54 @@
 (defn- generate-advance-nesting-level
   [changes file container libraries shape-id]
   (log/trace :msg "  -> advance-nesting-level")
-  (let [children  (cfh/get-children-with-self (:objects container) shape-id)
+  (let [detached-ids (transient #{})
```
Alotor: I think for what you're using the transient is better to have an `atom`

common/src/app/common/logic/libraries.cljc
```diff
@@ -338,42 +339,54 @@
...
+                          (nil? ref-shape)
+                          (generate-detach-immediate container (:id shape) detached-ids)))))]
 
     (reduce skip-near changes children)))
 
+(defn- generate-detach-immediate
+  [changes container shape-id detached-ids]
+  (let [shape-and-children (cfh/get-children-ids-with-self (:objects container) shape-id)]
+    (log/trace :msg "       (cannot advance; detach shape and children)")
+    (reduce conj! detached-ids shape-and-children)
```
Alotor: Does this work? reduce is lazy... and you're not even doing anything with the result so shouldn't be ever realized

pabloalba: In my tests, it works. Maybe it would be realized later, when in `generate-advance-nesting-level` it does `(contains? detached-ids (:id shape))`. I agree that this is a bit weird...

common/src/app/common/logic/libraries.cljc
hirunatan: ok

common/src/app/common/logic/libraries.cljc
hirunatan: changed

PR #6379, opened 2025-04-29
:tada: Serialize as bytes all fill kinds

belen-albeza:
### Related Ticket

https://tree.taiga.io/project/penpot/task/10753

### How to test

Open or create a file with different fill types (also fills for strokes).

### Summary

Screenshot 2025-05-06 at 12 54 41 PM

This creates a unified DTO to deserialize in Wasm fill data. Now all fills have the same size count (`160` bytes) and gradients have a cap of `16` stops.

We switched to use `mem::transmute` to do deserialization (it is a raw byte copy). To be able to do that, we need to use `repr(C)`, `repr(4)` and `repr(u8)` to control both the alignment and layout, as well as the actual values to serialize the discriminant byte for the enum variant.

### Checklist

### DIFF
--- a/frontend/src/app/render_wasm/api.cljs
+++ b/frontend/src/app/render_wasm/api.cljs
@@ -297,8 +297,12 @@
                   (store-image id)))
 
               (some? color)
-              (let [rgba (sr-clr/hex->u32argb color opacity)]
-                (h/call wasm/internal-module "_add_shape_stroke_solid_fill" rgba)))))
+              (let [size   sr-fills/SOLID-BYTE-SIZE
+                    offset (mem/alloc-bytes size)
+                    heap   (mem/get-heap-u32)
+                    argb (sr-clr/hex->u32argb color opacity)]
+                (sr-fills/write-solid-fill! offset heap argb)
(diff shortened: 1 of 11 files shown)


frontend/src/app/render_wasm/api.cljs
```diff
@@ -214,35 +214,30 @@
...
-                (h/call wasm/internal-module "_add_shape_image_fill"
-                        (aget buffer 0)
-                        (aget buffer 1)
-                        (aget buffer 2)
-                        (aget buffer 3)
-                        opacity
-                        (dm/get-prop image :width)
-                        (dm/get-prop image :height))
+                (sr-fills/write-image-fill! offset heap id opacity (dm/get-prop image :width) (dm/get-prop image :height))
+                (h/call wasm/internal-module "_add_shape_fill")
```
elenatorro, line 236: Now that we're calling "_add_shape_fill" for all fill types, maybe we can just call it once at the end?

belen-albeza: I just tried and it turns out we need to call `_add_shape_fill` before `(store-image)` :( I did remove the redundant calls in gradients

PR #6386, opened 2025-04-30
:recycle: Refactor update attrs

pabloalba:


common/src/app/common/logic/libraries.cljc
```diff
@@ -1624,6 +1624,56 @@
       (seq applied-tokens)
       (update-tokens container dest-shape origin-shape applied-tokens))))
 
+(defn- generate-update-attrs
```
hirunatan: I propose the name `add-update-attr-changes`. Although it ends generating some changes objects, I think this is not a typical `generate-xx` function, because this one is lower level (it receives directly the operations it needs to add).

common/src/app/common/logic/libraries.cljc
```diff
@@ -1624,6 +1624,56 @@
...
+                                    container
+                                    {:type :mod-obj
+                                     :id (:id dest-shape)
+                                     :operations (vec uoperations)}))
+        (update :undo-changes concat [(make-change
+                                       container
+                                       {:type :reg-objects
+                                        :shapes all-parents})]))))
+
+(defn- update-attr-operations
```
hirunatan: Here I suggest `add-update-attr-operations`, to start with a verb and be consistent with previous one.

hirunatan (approved):

PR #6387, opened 2025-04-30
:recycle: Rename, move and refactor the input-with-values component

luisddm:
### Related Ticket

Taiga [#10935](https://tree.taiga.io/project/penpot/task/10935)

### Summary

- Change pointer and background on hover so it mimics the behavior of the normal input.
- Rename the component and their properties to make it more agnostic and easy to understand.
- Move the component in the Storybook to the Product section since it is very specific for variants.


frontend/src/app/main/ui/ds/product/input_with_meta.scss
```diff
@@ -9,10 +9,10 @@
 @use "../_sizes.scss" as *;
 @use "../typography.scss" as t;
 
-.input-with-values-container {
+.input-with-meta-container {
```
Xaviju, line 12: This is just a enhancement proposal:

In DS components we are not using semantic tokens directly (as `--color-background-tertiary`). We are trying to create a new layer of component tokens on top, referencing our semantic tokens. 
Example:

```css
.input-with-meta-container {
--input-meta-background: var(--color-background-tertiary);

background-color: var(--input-meta-background);

&:hover {
--input-meta-background: var(--color-background-quaternary);
}
```

Xaviju (approved):

PR #6406, opened 2025-05-05
:sparkles: Manage overrides on variants switch

pabloalba:
Implements https://tree.taiga.io/project/penpot/task/10838

### DIFF
--- a/common/src/app/common/logic/libraries.cljc
+++ b/common/src/app/common/logic/libraries.cljc
@@ -1745,36 +1745,34 @@
     (loop [attrs       updatable-attrs
            roperations [{:type :set-touched :touched (:touched origin-shape)}]
            uoperations (list {:type :set-touched :touched (:touched dest-shape)})]
-      (let [attr (first attrs)]
-        (if (nil? attr)
-          (cond-> changes
-            (> (count roperations) 1)
-            (add-update-attr-changes dest-shape container roperations uoperations)
+      (if-let [attr (first attrs)]
+        (let [attr-group (get ctk/sync-attrs attr)
+              [roperations' uoperations']
+              (if (or
+                             ;; If the attribute is not valid for the destiny, don't copy it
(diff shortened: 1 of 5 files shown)


common/src/app/common/logic/libraries.cljc
```diff
@@ -1730,6 +1733,49 @@
...
+        ;; (see update-attrs for a full explanation)
+        origin-shape   (reposition-shape origin-shape origin-root dest-root)
+        touched        (get dest-shape :touched #{})
+        touched-origin (get origin-shape :touched #{})]
+
+    (loop [attrs       updatable-attrs
+           roperations [{:type :set-touched :touched (:touched origin-shape)}]
+           uoperations (list {:type :set-touched :touched (:touched dest-shape)})]
+      (let [attr (first attrs)]
+        (if (nil? attr)
```
niwinz: You can use `if-let` here

common/src/app/common/types/shape.cljc
```diff
@@ -396,6 +396,50 @@
...
+(def ^:private valid-bool-attrs #{:shapes :bool-type :content})
+(def ^:private valid-group-attrs #{:shapes})
+(def ^:private valid-frame-attrs #{:shapes :hide-fill-on-export :show-content :hide-in-viewer})
+(def ^:private valid-image-attrs #{:metadata})
+(def ^:private valid-svg-attrs #{:content})
+(def ^:private valid-path-attrs #{:content})
+(def ^:private valid-text-attrs #{:content})
+(def ^:private valid-generic-attrs (set/union valid-shape-attrs valid-shape-geom-attrs valid-shape-base-attrs))
+
+(defn is-valid-attr?
```
niwinz: I think we should rename it to `is-allowed-attr?`
because we don't check if the attr is valid, we check if it is allowed independently if the attr value validity

niwinz (changes requested):

PR #6424, opened 2025-05-07
:sparkles: Improve combobox component

luisddm:
### Related Ticket

Taiga [#10933](https://tree.taiga.io/project/penpot/task/10933)

### Summary

This commit makes the component more solid and predictable introducing and adjusting several microinteractions:

- When the user presses ENTER or ESC after typing something in the input, the focus is out and the current value is sent within the event.
- When the user clicks on an option from the dropdown or presses ENTER on it, the dropdown disappears, the focus is out and the value of the option is sent within the event.

In summary, the component does not send an event anymore when the input changes. Only when the focus is lost (pressing ESC, ENTER, selecting an option or clicking outside).

### DIFF
--- a/frontend/src/app/main/ui/ds/controls/combobox.cljs
+++ b/frontend/src/app/main/ui/ds/controls/combobox.cljs
@@ -34,13 +34,6 @@
     (reset! focused* id)
     (dom/scroll-into-view-if-needed! node)))
 
-(defn- handle-selection
-  [focused* selected* open*]
-  (when-let [focused (deref focused*)]
-    (reset! selected* focused))
-  (reset! open* false)
-  (reset! focused* nil))
-
 (def ^:private schema:combobox-option
   [:and
    [:map {:title "option"}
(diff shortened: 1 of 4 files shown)


Xaviju (approved):
It seems to be fine, but when I pass the tests on Storybook, I get a warning. Please ensure that this might not be a problem before merging;


> Found 9 unhandled errors while running the play function. This might cause false positive assertions. Resolve unhandled errors or ignore unhandled errors with setting thetest.dangerouslyIgnoreUnhandledErrors parameter to true.

luisddm:
> It seems to be fine, but when I pass the tests on Storybook, I get a warning. Please ensure that this might not be a problem before merging;
> 
> > Found 9 unhandled errors while running the play function. This might cause false positive assertions. Resolve unhandled errors or ignore unhandled errors with setting thetest.dangerouslyIgnoreUnhandledErrors parameter to true.

Seems to happen only when the library [user-event](https://github.com/testing-library/user-event) has to emulate a keyboard event involving ArrowUp or ArrowDown. But after debugging the events in the browser, it does work as expected, no extra events are shown, and the performance looks fine. Maybe it's an issue with the library. Anyway, let's revisit this in the future when we have more tests in Storybook.

PR #6450, opened 2025-05-09
:tada: Separate the content of the text of the rest of properties on …

pabloalba:
Implements https://tree.taiga.io/project/penpot/task/11038


common/src/app/common/types/container.cljc
```diff
@@ -569,23 +570,37 @@
           (gsh/close-attrs? attr val shape-val 1)
           (gsh/close-attrs? attr val shape-val))
 
+        group-not-equal? (and group (not equal?))
+
+
         touched?
-        (and group (not equal?) (not (and ignore-geometry is-geometry?)))]
+        (and group-not-equal? (not (and ignore-geometry is-geometry?)))
```
pabloalba: I think we should add `ignored? ` to the and, but I'm not sure

common/test/common_tests/text_content_test.cljc
```diff
@@ -0,0 +1,147 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC
+
+(ns common-tests.text-content-test
```
niwinz: Should be under types, inside tests

common/src/app/common/types/text_content.cljc
```diff
@@ -0,0 +1,100 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC
+
+ (ns app.common.types.text-content
```
niwinz: I think this should be `app.common.types.text`

common/src/app/common/logic/libraries.cljc
```diff
@@ -1732,11 +1747,26 @@
             (generate-update-tokens container dest-shape origin-shape touched omit-touched?))
 
           (let [attr-group (get ctk/sync-attrs attr)
+                is-text-partial-change? (and omit-touched?
```
hirunatan: I suggest adding here a comment explaining what is, conceptually, a "text partial change". The condition is quite complex and difficult to understand. Knowing the concept may be much easier to interpret all the clauses.

common/src/app/common/types/text_content.cljc
hirunatan: Agree.

common/src/app/common/types/text_content.cljc
```diff
@@ -0,0 +1,100 @@
...
+   (:require
+    [clojure.set :as set]))
+
+(defn text-content-diff
```
hirunatan: I suggest the name `compare-text-content`. The word `diff` makes me think that the function will return the full list of changes from a to b, not only the types of changes.

common/src/app/common/types/container.cljc
```diff
@@ -565,23 +566,36 @@
...
       ;; Depending on the origin of the attribute change, we need or not to
       ;; set the "touched" flag for the group the attribute belongs to.
       ;; In some cases we need to ignore touched only if the attribute is
       ;; geometric (position, width or transformation).
       (and in-copy?
-           (or (and group (not equal?)) (seq token-groups))
-           (not ignore?) (not (and ignore-geometry is-geometry?)))
+           (not-empty groups)
+           (not ignore?)
+           (not (and ignore-geometry is-geometry?)))
```
hirunatan, line 591: This section is very tricky, there are lots of corner cases managed here. Have you checked all combinations?

common/src/app/common/types/container.cljc
hirunatan: I don't understand. What `ignored?`

common/src/app/common/types/text.cljc
```diff
@@ -66,7 +102,7 @@
    entries"
   [a b]
   (cond
-    (not= (:type a) (:type b))
+    (not= (type a) (type b))
```
niwinz, line 105: I'm not clearly understand this change, are you trying to check if two value has different runtime type? is that correct?

common/test/common_tests/types/text_content_test.cljc
```diff
@@ -0,0 +1,277 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC
+
+(ns common-tests.types.text-content-test
```
niwinz: Remove `-content` from namespace, just for consistency

common/src/app/common/types/text.cljc
pabloalba: Yes, that is correct. I think is very unlikely, but in the rest of the code I assume that both are maps, or both are lists, so better safe than sorry. 
It made a mistake before accessing :type instead of check for the runtime type

common/test/common_tests/types/text_content_test.cljc
pabloalba: done

common/test/common_tests/types/text_test.cljc
```diff
@@ -0,0 +1,277 @@
...
+;;
+;; Copyright (c) KALEIDOS INC
+
+(ns common-tests.types.text-test
+  (:require
+   [app.common.types.text :as cttx]
+   [clojure.test :as t :include-macros true]))
+
+
+(def content-base
```
hirunatan: Is there any way to add test helpers to build this objects programatically, using the real code that generates text content? Having hard-coded values like this will prevent the tests to detect a possible future case in that the text chapes change their format and this breaks the sync algorithm.

pabloalba: Mmm... Good point. I'll check

common/test/common_tests/types/text_test.cljc
```diff
@@ -0,0 +1,277 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
```
hirunatan, line 1: I think there should also be a test at the `common.logic` level, to check the whole cycle of creating a text shape in a component, changing something in a copy and synchronizing it.

The test does not need to address all possible cases (since they are tested in those ones in `common.types`), but at least ensure that the helper functions are called.

pabloalba: Done!

PR #6453, opened 2025-05-09
:tada: Add text fills

elenatorro:
### Related Ticket

https://tree.taiga.io/project/penpot/task/10900

### Summary

This PR adds support for rendering text fills

### Steps to reproduce 

Use different fills for texts, including texts with multiple fill types.

image

### Checklist

### DIFF
--- a/frontend/src/app/render_wasm/api.cljs
+++ b/frontend/src/app/render_wasm/api.cljs
@@ -188,7 +188,6 @@
 
 (defn- store-image
   [id]
-
   (let [buffer (uuid/get-u32 id)
         url    (cf/resolve-file-media {:id id})]
     (->> (http/send! {:method :get
@@ -209,6 +208,33 @@
                              (aget buffer 3))
                      true))))))
 
+(defn- get-fill-images
+  [leaf]
+  (filter :fill-image (:fills leaf)))
+
+(defn- process-fill-image
+  [fill]
+  (rx/from
+   (when-let [image (:fill-image fill)]
+     (let [id (dm/get-prop image :id)
+           buffer (uuid/get-u32 id)
+           cached-image? (h/call wasm/internal-module "_is_image_cached"
+                                 (aget buffer 0)
+                                 (aget buffer 1)
+                                 (aget buffer 2)
+                                 (aget buffer 3))]
+       (when (zero? cached-image?)
+         (store-image id))))))
+
+(defn set-shape-text-images
+  [content]
+  (let [paragraph-set (first (get content :children))
+        paragraphs (get paragraph-set :children)]
+    (->> paragraphs
+         (mapcat :children)
+         (mapcat get-fill-images)
+         (map process-fill-image))))
+
 (defn set-shape-fills
   [fills]
   (h/call wasm/internal-module "_clear_shape_fills")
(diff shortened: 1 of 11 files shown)


frontend/src/app/render_wasm/api.cljs
```diff
@@ -207,6 +207,29 @@
...
+                                          (conj fill-results (store-image id))
+                                          fill-results))
+                                      fill-results))
+                                  leaf-results
+                                  (:fills leaf)))
+                        results
+                        leaves)))
+            []
+            paragraphs)))
+
```
niwinz, line 237: The dm/get-prop is used extensively here but it should only be used on types that declares static fields like Shape. The text content is a collection of vectors and maps, and simple get should be used for access data on maps.

In any case,  I think this code can be written using transducers which will result in a more compact code:

```clojure
(def xf:proc-text-images
  (comp
   (mapcat :children)
   (mapcat :children)
   (mapcat :children)
   (mapcat :fills)
   (keep (fn [fill]
           (when-let [image (:fill-image fill)]
             (let [id      (get image :id)
                   buffer  (uuid/get-u32 id)
                   cached  (h/call wasm/internal-module "_is_image_cached"
                                   (aget buffer 0)
                                   (aget buffer 1)
                                   (aget buffer 2)
                                   (aget buffer 3))]
               (when (zero? cached)
                 (store-image id))))))))


(defn set-shape-text-images
  [content]
  (transduce xf:proc-text-images conj [] content))
```

I have not tested it, but should work.

render-wasm/src/shapes/text.rs
```diff
@@ -416,21 +441,72 @@ impl From<[u8; RAW_PARAGRAPH_DATA_SIZE]> for RawParagraphData {
...
+                u32::from_be_bytes([bytes[40], bytes[41], bytes[42], bytes[43]]),
+                u32::from_be_bytes([bytes[44], bytes[45], bytes[46], bytes[47]]),
+            ],
+            text_length: u32::from_be_bytes([bytes[48], bytes[49], bytes[50], bytes[51]]),
+            total_fills: total_fills as u32,
+            fills,
+        }
+    }
+}
+
```
belen-albeza: We could probably leverage `transmute` for this, like we did with the RawFillData

render-wasm/src/shapes/text.rs
```diff
@@ -416,21 +441,72 @@ impl From<[u8; RAW_PARAGRAPH_DATA_SIZE]> for RawParagraphData {
...
-            let leaf_attrs = RawTextLeafData::from(array);
+        let mut offset = 0;
+        for _ in 0..num_leaves {
+            let start = offset;
+            let leaf_end = start + RAW_LEAF_DATA_SIZE;
+            let total_fills = u32::from_be_bytes([
+                buffer[leaf_end - 4],
+                buffer[leaf_end - 3],
+                buffer[leaf_end - 2],
+                buffer[leaf_end - 1],
```
belen-albeza: Let's use Little Endian by default. Also, maybe transmute could help us here too?

render-wasm/src/wasm/fills.rs
```diff
@@ -53,6 +53,19 @@ impl TryFrom<&[u8]> for RawFillData {
...
+
+    for i in 0..num_fills {
+        let start = i * RAW_FILL_DATA_SIZE;
+        let end = start + RAW_FILL_DATA_SIZE;
+        let bytes = &buffer[start..end];
+        let raw_fill = RawFillData::try_from(bytes).expect("Invalid fill data");
+        fills.push(raw_fill.into());
+    }
+    fills
+}
```
belen-albeza: `Vec::from_raw_parts` instead of iterating, and then we can map each entry with a `try_from`.

belen-albeza (changes requested):
Hi! Thanks for the PR! I added some suggestions re: deserialization

render-wasm/src/shapes/text.rs
elenatorro: Ok, let me check that! :)

belen-albeza (approved):

PR #6456, opened 2025-05-12
:sparkles: Reflow flex on grow text height

Alotor:
### Related Ticket
https://tree.taiga.io/project/penpot/task/10908

### Summary
- Change auto-width/height sets the grow type to fixed
- Dynamic resize text will update constraints
- No word wrapping on resize

### Checklist

### DIFF
--- a/render-wasm/src/shapes/modifiers.rs
+++ b/render-wasm/src/shapes/modifiers.rs
@@ -6,10 +6,10 @@ mod grid_layout;
 
 use common::GetBounds;
 
-use crate::math::{identitish, Bounds, Matrix, Point};
+use crate::math::{self as math, identitish, Bounds, Matrix, Point};
 use crate::shapes::{
-    modified_children_ids, ConstraintH, ConstraintV, Frame, Group, Layout, Modifier, Shape,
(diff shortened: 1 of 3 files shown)


render-wasm/src/shapes/modifiers.rs
```diff
@@ -137,19 +138,36 @@ pub fn propagate_modifiers(state: &State, modifiers: Vec<TransformEntry>) -> Vec
...
+                                height,
+                            );
+                            shape_bounds_after = shape_bounds_after.transform(&resize_transform);
+                            transform.post_concat(&resize_transform);
+                        }
+                    }
 
                     if entry.propagate {
                         let mut children = propagate_children(
                             shape,
```
elenatorro, line 163: I think we should send the shape by reference here `&shape`

elenatorro (changes requested):

render-wasm/src/shapes/modifiers.rs
Alotor: It's already a reference. I don't put `&shape` because the variable `shape` is already a reference, but if you check the `propagate_children` funciton you'll see its passed by reference.

PR #6502, opened 2025-05-19
:recycle: Factor more code out of workspace.cljs

mdbenito:
### Summary

As a second and final measure towards a smaller `frontend/src/app/main/data/workspace.cljs`, this PR:

* mimics #6495 by using git-split to create a new file `frontend/src/app/main/data/workspace/pages.cljs` with history intact.
* Reexports all symbols from the original namespace

### Open question

To avoid a circular dependency I have moved `set-plugin-data` to the new file, where it is required, but it clearly does not belong there. My intuition would be to move it to `app.main.data.plugins` but I don't want to mess the history / git blame unless it's ok. Should I, or is there a better place?

### Checklist


niwinz:
About the question: It is ok to move the `set-plugin-data` to `data.plugins` 
I will wait that change for merge this.

mdbenito:
@niwinz Done. Sorry for the messy changes above and force-pushing. A merge commit message slipped without emoji and circleCI complained 🙈 

There was one conflict in `rename-page`, to which @EvaMarco made a small change. I cherry-picked it and added it as a separate commit in the new location. I don't know what it was supposed to do since it was part of a pretty large change, so I used a generic commit message, I hope that's ok.

niwinz:
On the end I will prefer clear history instead of preserve history for blame so I merged it with squash. So I think is better to prioritize good commits for easy of review instead of commits for preserve history.

PR #6530, opened 2025-05-21
:bug: Fix cursor overlap query

### LINKED ISSUE #4472: bug: can't select in the middle of a very short line
### Steps To Reproduce

1. add a very short line (vertical or horizontal path), length <= 4px
2. try to select it at middle, it fails, and there's no "under mouse indicator" (a green line).
3. move mouse to the outer region of two endpoints. Though it's far away from the path (considering how small it is), green line appears and it's possible to select it from there.

### Expected behavior

1. green line appears when mouse is near or inside the line, and the line can be selected.
2. green line doesn't appear when mouse is near the endpoints but not inside the line.

### Actual behavior

As the steps above.

### Screenshots or video

https://github.com/penpot/penpot/assets/39703247/f0ab3cd5-11c2-4332-b5ea-89f2275bdf76

- Windows 11 22H2
- Firefox 124.0.2 (64-bit)

_No response_

_No response_

### Frontend Stack Trace

_No response_

### Backend Stack Trace

_No response_

### Additional context

in Figma there is also a margin area for easier selection of smaller things, but the margin width is almost constant with screen pixel, ensuring it doesn't get too large when closely zoomed.

mdbenito:
### Summary

A fix to

* Close #4472 

The issue is that the overlap query executed by the worker builds a 5x5 rectangle around the mouse position, irrespective of the size of the shape. This means that shapes which are smaller in size will not intersect the rectangle when the mouse hovers around the center, as illustrated in the issue.

---

This PR:

* Reduces the size of the intersecting box by accounting for shape size.

### Open issue

This works around the problem but is not entirely satisfactory because one should also account for the distance to nearby shapes. A perfect solution would cast a crosshair of rays around the mouse and take the first intersecting shape.

This is however a major change which I'd prefer not getting into. But it, or an alternative solution should probably be considered at some point.

### Behaviour with these changes

Alas, there is a [known bug in gnome](https://gitlab.gnome.org/GNOME/mutter/-/issues/3182) that breaks screen recording on my system and the mouse pointer doesn't show. In the video below I'm smoothly moving the mouse from one segment to the next one.

[select-segments.webm](https://github.com/user-attachments/assets/5d5f9668-ff85-49bf-94ee-d019e19ba083)

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -82,6 +82,7 @@ A non exhaustive list of changes:
 
 ### :bug: Bugs fixed
 
+- Fix selection of short paths [Github #4472](https://github.com/penpot/penpot/issues/4472)
 
 ## 2.7.0 (Unreleased)
 
--- a/frontend/src/app/worker/selection.cljs
+++ b/frontend/src/app/worker/selection.cljs
@@ -7,6 +7,7 @@
 (ns app.worker.selection
   (:require
    [app.common.data :as d]
+   [app.common.data.macros :as dm]
    [app.common.files.helpers :as cfh]
    [app.common.files.indices :as cfi]
    [app.common.geom.point :as gpt]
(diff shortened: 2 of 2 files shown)


mdbenito:
CI is failing because of the merge commit message because the branch was not based off the latest develop 🙄 I hope that's fine

PR #6532, opened 2025-05-21
:bug: Sanitize inputs for variant property names and values in design tab

luisddm:
### Related Ticket

Taiga [#11124](https://tree.taiga.io/project/penpot/task/11124)

### Summary

This fix sanitizes the inputs for the variant properties names and values in the design tab.
- Property names
  - Trim the spaces before and after
  - Do not update the value if empty
- Property values
  - Trim the spaces before and after

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
@@ -269,7 +269,8 @@
         (mf/use-fn
          (mf/deps component-ids)
          (fn [pos value]
-           (let [value (if (= value empty-indicator) "" value)]
+           (let [value (str/trim value)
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
```diff
@@ -269,7 +269,7 @@
         (mf/use-fn
          (mf/deps component-ids)
          (fn [pos value]
-           (let [value (if (= value empty-indicator) "" value)]
+           (let [value (if (= value empty-indicator) "" (str/trim value))]
```
pabloalba: Maybe trim value before, so you can catch also values like `-- ` or  ` -- ` as empty-indicator

pabloalba (approved):

PR #6544, opened 2025-05-22
:sparkles: Use different copies for different variant selection cases

luisddm:
### Related Ticket

Taiga [#11127](https://tree.taiga.io/project/penpot/task/11127)

### Summary

When multiple variants are selected, the user can have three different cases:

- No selected variant has any errors.
- All variants have an error each.
- Some variants have an error, some not.

We use different copies for each of these cases, and should be different of having just one variant selected with an error, which is the current and only case.

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
@@ -238,7 +238,7 @@
 
 (defn- get-variant-error-message [errors]
   (cond
-    (and (= (count errors) 1) (not (nil? (first errors))))
+    (and (= (count errors) 1) (some? (first errors)))
     (tr "workspace.options.component.variant.malformed.single.one")
 
     (and (seq errors) (every? some? errors))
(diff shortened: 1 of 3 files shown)


frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
```diff
@@ -235,12 +235,26 @@
         (when (or editing? creating?)
           [:div {:class (stl/css  :counter)} (str size "/300")])]])))
 
+
+(defn- get-variant-error-message [errors]
+  (cond
+    (and (= (count errors) 1) (not (nil? (first errors))))
```
pabloalba: Use `(some?` insteda of `(not (nil?`

Also, why can it be nil?

luisddm: When a shape has no errors, its `:variant-error` property is `nil`. I'm passing a vector with these values mapped.

pabloalba (approved):

PR #6569, opened 2025-05-27
:recycle: Clean and sanitize color types

niwinz:
### Summary

This PR performs a general cleaning of color types on the application.

### Relevant changes

**Normalizes the several different types of colors to 2 main color types types: Color and LibraryColor**

Where LibraryColor is how color looks on the file library and Color refers as **applied color**. This PR still does not touches how Fills and Strokes because it is a bigger refactor.

**LibraryColors on recent color palette behavior changes**

Currently, when a library color appears in the recent colors list, any changes made to that library color are immediately reflected in the recent colors list.
Previously: the library color was stored in the recent colors list with its value at the time it was added. If the library color was later modified, applying it from the recent list could cause unexpected behavior — sometimes resulting in a crash, and other times applying the outdated color while still displaying the original library color name.

### Additional changes

- Move several binary data helpers to a separated namespace for better reuse
- Add better defaults for several test.check generators
- Several internal issues fixed

### How to Test

This touches data structures, so the best way to test it, is generating a file on hourly o PRE with shapes with fills, shadows with plain and library color

- You should be able to import that file.
- You should be able to modify a library and see the changes reflected on recent colors, fills,strokes and shadows.
- When you modifying a color on library and touching the color on colorpicker, the recent color added should not have the library color name on it.


frontend/src/app/main/ui/dashboard/fonts.cljs
```diff
@@ -29,6 +29,11 @@
    [okulary.core :as l]
    [rumext.v2 :as mf]))
 
+(def ^:private accept-font-types
+  (str (str/join "," font-types)
+       ;; A workaround to solve a problem with chrome input selector
+       ".ttf,application/font-woff,woff,.otf"))
```
superalex: Do we need an extra "," here?

frontend/src/app/main/data/media.cljs
```diff
@@ -15,8 +15,7 @@
    [cljs.spec.alpha :as s]
    [cuerdas.core :as str]))
 
-(def accept-image-types
-  (str/join "," cm/image-types))
+;; FIXME: revisit the need of this NS
```
superalex, line 18: Is this comment needed?

frontend/src/app/main/ui/inspect/attributes/fill.cljs
```diff
@@ -35,7 +28,10 @@
   [{:keys [objects shape]}]
   (let [format*   (mf/use-state :hex)
         format    (deref format*)
-        color     (shape->color shape)
+        ;; FIXME: this looks broken code, because shape does not
+        ;; longer contains :fill-xxxx attributes but it is preserved
+        ;; as it was just moved the impl; this need to be fixed
+        color     (types.color/fill->color shape)
```
superalex, line 34: Should we keep this FIXME?

niwinz: yes

frontend/src/app/main/data/media.cljs
niwinz: yes

frontend/src/app/main/ui/dashboard/fonts.cljs
niwinz: Fixed

PR #6574, opened 2025-05-27
:sparkles: Manage empty property values in the combobox in design tab

luisddm:
### Related Ticket

Taiga [#11126](https://tree.taiga.io/project/penpot/task/11126)

### Summary

This improves the management of empty values for variant properties. We use the two dashes `(--)` only for indicating that an empty value is selected, but we have an `(Empty)` option in the dropdown list to choose this value.

This avoids the confusion between the two dashes as an indicator of emptiness vs a literal value.

### DIFF
--- a/common/src/app/common/files/variant.cljc
+++ b/common/src/app/common/files/variant.cljc
@@ -8,8 +8,7 @@
    [app.common.data.macros :as dm]
    [app.common.types.component :as ctc]
    [app.common.types.components-list :as ctcl]
-   [app.common.types.variant :as ctv]
-   [cuerdas.core :as str]))
+   [app.common.types.variant :as ctv]))
 
 
 (defn find-variant-components
@@ -20,11 +19,6 @@
        (map #(dm/get-in objects [% :component-id]))
        (map #(ctcl/get-component data % true))
        reverse))
-
-(defn- dashes-to-end
-  [property-values]
-  (let [dashes (if (some #(= % "--") property-values) ["--"] [])]
-    (concat (remove #(= % "--") property-values) dashes)))
 
 
 (defn extract-properties-names
@@ -42,10 +36,7 @@
        (group-by :name)
        (map (fn [[k v]]
               {:name k
-               :value (->> v
-                           (map #(if (str/empty? (:value %)) "--" (:value %)))
-                           distinct
-                           dashes-to-end)}))))
+               :value (distinct (map :value v))}))))
 
 (defn get-variant-mains
   [component data]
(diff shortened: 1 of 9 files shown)


common/src/app/common/files/variant.cljc
```diff
@@ -42,10 +36,7 @@
        (group-by :name)
        (map (fn [[k v]]
               {:name k
-               :value (->> v
-                           (map #(if (str/empty? (:value %)) "--" (:value %)))
-                           distinct
-                           dashes-to-end)}))))
+               :value (distinct (map :value v))}))))
```
pabloalba: Very soft comment, you can ignore it:

This is corrects, but we use to write it as:

(->> v (map :value) distinct)

It's more idiomatic...

frontend/src/app/main/ui/ds/controls/select.cljs
```diff
@@ -66,100 +67,104 @@
...
-        focused* (mf/use-state nil)
-        focused  (deref focused*)
-
-        has-focus* (mf/use-state false)
-        has-focus  (deref has-focus*)
+        empty-selected-value? (str/blank? selected-value)
 
-        on-click
+        set-option-ref
         (mf/use-fn
```
pabloalba, line 97: When using use-fn, you should add a deps entry with all the things that the function uses but aren't parameters, so:

(mf/deps options-nodes-refs)

frontend/src/app/main/ui/ds/controls/shared/options_dropdown.cljs
```diff
@@ -48,23 +51,65 @@
...
-                      :label label
-                      :icon icon
-                      :aria-label aria-label
-                      :set-ref set-ref
-                      :focused (= id focused)
-                      :on-click on-click}]))]))
+
+    (if empty-to-end
+
+      (let [with-id       (array/filter #((complement str/blank?) (obj/get % "id")) options)
```
pabloalba: If you set here the if here instead of on line 60, you can have only one block, and avoid code repetition (removing all the block on line 100)...

```
options       (if empty-to-end (array/filter #((complement str/blank?) (obj/get % "id")) options) options)
with-blank-id (when empty-to-end (array/filter #(str/blank? (obj/get % "id")) options))
```

luisddm: Done! I have put the options-blank first to avoid overriding the options value. Also I have added a condition to not show the separator if there are only blank values.

pabloalba (approved):

PR #6586, opened 2025-05-29
:sparkles: Delete variant property when it has no value anywhere

luisddm:
### Related Ticket

Taiga [#10765](https://tree.taiga.io/project/penpot/task/10765)

### Summary

After editing a formula, we check all the values for every property. If a property has only empty values, we remove it.

### DIFF
--- a/frontend/src/app/main/data/workspace/variants.cljs
+++ b/frontend/src/app/main/data/workspace/variants.cljs
@@ -176,6 +176,46 @@
          (dwu/commit-undo-transaction undo-id))))))
 
 
+(defn remove-empty-properties
+  "Remove a property for all components when its value is empty for all of them"
+  [variant-id]
+  (ptk/reify ::remove-empty-properties
+    ptk/WatchEvent
+    (watch [it state _]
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -176,6 +176,46 @@
...
+                                            (->> v (mapv :value) (remove empty?))))
+                                    (mapv empty?))
+
+            changes (-> (pcb/empty-changes it page-id)
+                        (pcb/with-library-data data)
+                        (pcb/with-objects objects))
+
+            changes (reduce
+                     (fn [changes [idx property-empty?]]
+                       (if property-empty?
```
pabloalba, line 205: When can this be false?

luisddm: The vector `properties-empty` contains `true` or `false` values depending on if the property associated to that position has empty values for all variants or not. This is why we only remove the property if the value is `true`.

pabloalba (approved):

PR #6609, opened 2025-06-02
:tada: Implement font fallback to support multiple languages

elenatorro:
### Related Ticket

https://tree.taiga.io/project/penpot/task/11213

### Summary

This code implements a similar strategy to Emoji fallback by including a default font fallback in case we detect a language that has support for a Noto Sans font.

### Steps to reproduce 

Create texts with different fonts and check that styles apply correctly by default, without having to load a specific font. Another check is to verify that if a font that supports the current text language is applied, it will be used, not the fallback.

This only includes some of the languages available, we can discuss to add more by default and how to maintain it.

- Current:

image

- New:

image

### Checklist

### DIFF
--- a/frontend/src/app/render_wasm/api/fonts.cljs
+++ b/frontend/src/app/render_wasm/api/fonts.cljs
@@ -15,7 +15,6 @@
    [app.render-wasm.helpers :as h]
    [app.render-wasm.wasm :as wasm]
    [app.util.http :as http]
-   [app.util.webapi :as wapi]
    [beicon.v2.core :as rx]
    [cuerdas.core :as str]
    [goog.object :as gobj]
@@ -77,7 +76,7 @@
 
 ;; IMPORTANT: It should be noted that only TTF fonts can be stored.
 (defn- store-font-buffer
-  [font-data font-array-buffer emoji?]
+  [font-data font-array-buffer emoji? fallback?]
   (let [id-buffer (:family-id-buffer font-data)
         size (.-byteLength font-array-buffer)
         ptr  (h/call wasm/internal-module "_alloc_bytes" size)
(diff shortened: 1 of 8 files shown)


frontend/src/app/render_wasm/api/fonts.cljs
```diff
@@ -91,17 +91,18 @@
...
+            emoji?
+            fallback?)
     true))
 
 (defn- store-font-url
-  [font-data font-url emoji?]
+  [font-data font-url emoji? fallback?]
   (->> (http/send! {:method :get
                     :uri font-url
                     :response-type :blob})
```
niwinz: You can use `:response-type :buffer` and avoid the `wapi/read-file-as-array-buffer`

frontend/src/app/render_wasm/api/fonts.cljs
```diff
@@ -181,8 +183,9 @@
                    :font-variant-id font-variant-id
                    :style style
                    :weight weight}
-        emoji? (dm/get-prop font :emoji?)]
-    (store-font-id font-data asset-id emoji?)))
+        emoji? (dm/get-prop font :emoji?)
```
niwinz: Don't use dm/get-prop on plain maps; prefer use `get` directly here

niwinz: Also, don't introduce `:fallback?` and `:emoji?`, prefer keys without `?` on the end. I suggest using `:is-fallback` and `:is-emoji`.

frontend/src/app/render_wasm/api/texts.cljs
```diff
@@ -150,5 +150,58 @@
 
 (def emoji-pattern #"[\uD83C-\uDBFF][\uDC00-\uDFFF]")
 
-(defn contains-emoji? [s]
-  (boolean (re-find emoji-pattern s)))
+(def unicode-ranges
```
niwinz: Define this kind of vars that are only used inside the namespace as `^:private`. It applies to `unicode-ranges` and others like `emoji-pattern`

frontend/src/app/render_wasm/api/texts.cljs
```diff
@@ -150,5 +150,58 @@
...
+
+
+(defn contains-emoji? [text]
+  (boolean (re-find emoji-pattern text)))
+
+(defn contains-languages [text]
+  (->> unicode-ranges
+       (filter (fn [[_ pattern]] (re-find pattern text)))
+       (map first)
+       set))
```
niwinz: I think this can be implemented with reduction what has better performance in this case and generates less GC pressure:

```clojure
(defn get-languages [text]
  (reduce-kv (fn [result lang pattern]
               (if (re-find pattern text)
                 (conj result lang)
                 result))
             #{}
             unicode-ranges))
```

Also, i suggest change the name

niwinz (changes requested):

frontend/src/app/render_wasm/api/texts.cljs
elenatorro: perfect!

niwinz (approved):
RTM

PR #6632, opened 2025-06-05
:bug: Fix strokes migration

superalex:


common/src/app/common/files/migrations.cljc
```diff
@@ -1441,7 +1441,8 @@
             (-> stroke
                 (select-keys cts/stroke-attrs)
                 (d/update-when :stroke-color-gradient clear-color-gradient)
-                (d/update-when :stroke-image clear-color-image)))
+                (d/update-when :stroke-image clear-color-image)
+                (update :stroke-style #(if (#{:svg :none} %) :solid %))))
```
niwinz: should be d/update-when

PR #6639, opened 2025-06-05
:sparkles: Show warning when duplicated variant prop name and value

luisddm:
### Related Ticket

Taiga [#11218](https://tree.taiga.io/project/penpot/task/11218)

### Summary

Even when two or more variants are visually different, having the same properties and values makes at least one of them unretrievable. That's why, when this happens, we display a warning about the problem and offer a button to select the conflicting variants. Different messages are displayed when the component, the variant, or a group of variants are selected.

These messages will only be shown if there are no malformed formulas.

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
@@ -237,8 +237,9 @@
           [:div {:class (stl/css  :counter)} (str size "/300")])]])))
 
 
-(defn- get-variant-malformed-message
-  "Generate warning message for malformed variants depending on the selected variants"
+(defn- get-variant-malformed-warning-message
+  "Receive a list of booleans, one for each selected variant, indicating if that variant
+   is malformed, and generate a warning message accordingly"
   [malformed-map]
   (cond
     (and (= (count malformed-map) 1) (some? (first malformed-map)))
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
```diff
@@ -278,12 +308,19 @@
                          (ctv/compare-properties properties-map false)
                          (first properties-map))
 
+        malformed-map   (mapv :variant-error shapes)
+        malformed-msg   (get-variant-malformed-message malformed-map)
+
+        duplicated-ids  (->> (cfv/find-variant-components data objects variant-id)
+                             get-component-ids-with-duplicated-variant-props-and-values)
```
pabloalba: Small performance improvement: 
Add here the `set` to avoid transform it in set for each component on line 318

frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
```diff
@@ -237,22 +237,52 @@
...
-    (and (seq errors) (some some? errors))
+    (and (seq malformed-map) (some some? malformed-map))
     (tr "workspace.options.component.variant.malformed.single.some")
 
     :else nil))
 
 
+(defn- get-variant-duplicated-message
+  "Generate warning message for duplicated variants depending on the selected variants"
+  [duplicated-map]
```
pabloalba, line 260: Please, add a comment on this functions explaining that they get a list of booleans

pabloalba (approved):

PR #6663, opened 2025-06-09
:sparkles: Plugins fixes and enhancements

Alotor:
### Summary
- New flag for exports to ignore children
- Expose clip content and view in viewport board properties
- Add new API to access a wrapper of the local storage
- Change the z-index of plugins so the toolbar is behind the plugin
- Resize method will take into account windows size

### Checklist

### DIFF
--- a/frontend/src/app/plugins/local_storage.cljs
+++ b/frontend/src/app/plugins/local_storage.cljs
@@ -1,0 +1,71 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC
+
+(ns app.plugins.local-storage
+  (:require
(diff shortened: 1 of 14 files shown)


Alotor:
Related to https://github.com/penpot/penpot-plugins/pull/201

frontend/src/app/plugins/local_storage.cljs
```diff
@@ -0,0 +1,71 @@
...
+   [app.util.globals :as g]
+   [app.util.object :as obj]
+   [cuerdas.core :as str]))
+
+(defonce ^:private local-storage
+  (ex/ignoring (unchecked-get g/global "localStorage")))
+
+(defn prefix-key
+  [plugin-id key]
+  (dm/str plugin-id "/" key))
```
niwinz: I think you can follow a similar scheme as we already using on penpot. I suggest changing this to:

`(dm/str "penpot-plugins:" plugin-id "/" key)`

Alotor: OK

PR #6674, opened 2025-06-10
:zap: Refactor fills-menu and related components

niwinz:
## Summary

This PR rewrites some parts of the fills-menu and several color related components for make it more efficient and generate less GC pressure. Also moves some business logic from UI to events.

## How to test

- General exploratory tests on fills menu

### DIFF
--- a/frontend/src/app/main/ui/components/color_input.cljs
+++ b/frontend/src/app/main/ui/components/color_input.cljs
@@ -152,12 +152,3 @@
            (events/unlistenByKey key2))))
 
     [:> :input props]))
-
-;; FIXME: this causes some weird interactions because of using apply-value
-;; [:*
-;;  [:> :input props]
-;;  [:datalist {:id list-id}
-;;   (for [color-name cc/color-names]
-;;     [:option color-name])]
-;;  ]))
-
--- a/frontend/src/app/main/ui/components/color_input.cljs
+++ b/frontend/src/app/main/ui/components/color_input.cljs
@@ -8,32 +8,27 @@
   (:require
    [app.common.colors :as cc]
    [app.common.data :as d]
+   [app.main.ui.hooks :as hooks]
    [app.util.dom :as dom]
    [app.util.globals :as globals]
    [app.util.i18n :as i18n :refer [tr]]
    [app.util.keyboard :as kbd]
-   [app.util.object :as obj]
    [goog.events :as events]
    [rumext.v2 :as mf]))
 
-(defn clean-color
-  [value]
-  (-> value
+(defn- get-clean-color
+  [node]
+  (-> (dom/get-value node)
       (cc/expand-hex)
       (cc/parse)
       (cc/prepend-hash)))
 
 (mf/defc color-input*
-  {::mf/wrap-props false
-   ::mf/forward-ref true}
-  [props external-ref]
-  (let [value            (obj/get props "value")
-        on-change        (obj/get props "onChange")
-        on-blur          (obj/get props "onBlur")
-        on-focus         (obj/get props "onFocus")
-        select-on-focus? (d/nilv (unchecked-get props "selectOnFocus") true)
-        class            (d/nilv (unchecked-get props "className") "color-input")
-        aria-label       (d/nilv (unchecked-get props "aria-label") (tr "inspect.attributes.color"))
+  {::mf/forward-ref true}
+  [{:keys [value on-change on-blur on-focus select-on-focus class aria-label] :rest props} external-ref]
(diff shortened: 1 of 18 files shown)


frontend/src/app/main/ui/components/color_input.cljs
```diff
@@ -8,32 +8,27 @@
...
-        on-blur          (obj/get props "onBlur")
-        on-focus         (obj/get props "onFocus")
-        select-on-focus? (d/nilv (unchecked-get props "selectOnFocus") true)
-        class            (d/nilv (unchecked-get props "className") "color-input")
-        aria-label       (d/nilv (unchecked-get props "aria-label") (tr "inspect.attributes.color"))
+  {::mf/forward-ref true}
+  [{:keys [value on-change on-blur on-focus select-on-focus class aria-label] :rest props} external-ref]
+  (let [select-on-focus? (d/nilv select-on-focus true)
+        class            (d/nilv class "color-input")
+        aria-label       (or aria-label (tr "inspect.attributes.color"))
```
hirunatan, line 32: With the current rumext I think we can use `:or` in the destructuring for default values, can't we?

https://github.com/funcool/rumext?tab=readme-ov-file#default-values

niwinz: Is more clear to use the default or when the default is a computation and not a static value.

frontend/src/app/main/ui/components/color_input.cljs
```diff
@@ -122,57 +102,62 @@
...
+
+    [:> :input props]))
+
+;; FIXME: this causes some weird interactions because of using apply-value
+;; [:*
+;;  [:> :input props]
+;;  [:datalist {:id list-id}
+;;   (for [color-name cc/color-names]
+;;     [:option color-name])]
+;;  ]))
```
hirunatan: Can this be deleted? Is it expected to use it in a near future?

niwinz: I don't know, this is already existing code.

frontend/src/app/main/ui/components/reorder_handler.cljs
```diff
@@ -7,15 +7,14 @@
...
   (:require
-   [app.main.ui.ds.foundations.assets.icon :refer [icon*] :as ic]
+   [app.main.ui.ds.foundations.assets.icon :as ic]
    [rumext.v2 :as mf]))
 
-(mf/defc reorder-handler
-  {::mf/forward-ref true}
-  [_ ref]
+(mf/defc reorder-handler*
+  [{:keys [ref]}]
```
hirunatan, line 14: We are now in react 19, so? Should we update rumext manual?

https://github.com/funcool/rumext?tab=readme-ov-file#forwarding-references

frontend/src/app/main/ui/components/color_input.cljs
hirunatan: I'm talking about adding an `:or` clause to the `{:keys [...]` destructuring in the component props.

If the `(or` is more efficient for the `(tr ...` expression, we may leave this as is, but move the other two default values to the destructuring?

frontend/src/app/main/ui/components/color_input.cljs
hirunatan: The FIXME comment is from a commit 4 years ago :smile: 
I think we can :fire:  it.

frontend/src/app/main/ui/components/reorder_handler.cljs
niwinz: Yes and not. I mean, rumext was compatible with react 19 but the forward ref is necessary with react <= 18 if you need forward ref.

For a while, we've been using React 18 on penpot, which forced us to use forward-ref. But with the update to React 19, it's no longer necessary.
I think the rumext documentation is fine; perhaps we could be more explicit. Rumext remains compatible with React 18 and 19 at the same time.

frontend/src/app/main/ui/components/color_input.cljs
niwinz: I will look on it.

frontend/src/app/main/ui/components/color_input.cljs
niwinz: OK

frontend/src/app/main/ui/components/reorder_handler.cljs
hirunatan: Ok. I can update the manual and mention the two possibilities.

PR #6697, opened 2025-06-16
:bug: Add workaround for webkit issue on fileReader

niwinz:
Relevant issue: https://tree.taiga.io/project/penpot/issue/11343

More context:

- It happens on any webkit broser, is reproducible on localhost with epiphany/gnome-web
- You should refresh after each test, because we have internal resolution cache
- The issue happens once the thumbnail is generated

### DIFF
--- a/frontend/src/app/util/webapi.cljs
+++ b/frontend/src/app/util/webapi.cljs
@@ -18,17 +18,32 @@
 
 (log/set-level! :warn)
 
+;; NOTE: this operation is necessary because some versions of safari/webkit,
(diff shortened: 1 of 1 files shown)


belen-albeza (commented):
~~I have tried the steps to reproduce in Safari and it crashes when inserting the image:~~

Screenshot 2025-06-17 at 3 14 00 PM

~~(In PRE adding the image is OK)~~

False alarm: it was an issue of the render engine, this PR works OK.

belen-albeza (approved):

PR #6705, opened 2025-06-17
:sparkles: Major improvement and update to docker images and devenv

niwinz:
### Summary

This PR comes with several changes on how we build our public docker images and our devenv image. The relevant changes:

- Replace the legacy 6.x version of ImageMagick found on ubuntu repositories with a custom build of 7.x branch (this removes a set of vulnerabilities from our public images that are never be fixed on 6.x)
- Use multistage build mechanism on backend image with focus on reducing the size of image. It is reduced from 1.2 GB to 850MB with the current change. It also starts using custom JRE build (generated with jlink, for strip the JVM size from 250MB to 80MB).
- Remove yarn and other build cache artifacts from exporter images as these were only used during build time and were causing security issues on reports.
- Normalize node version to v22.16.0 on all submodules.
- Refactor devenv build mechanism: separate different dependencies in build stages making them more cache friendly and allow parallelize build steps. It also incorporates the imagemagick build process.

### DIFF
--- a/docker/devenv/Dockerfile
+++ b/docker/devenv/Dockerfile
@@ -1,26 +1,16 @@
-FROM ubuntu:24.04
-LABEL maintainer="Penpot <docker@penpot.app>"
+FROM ubuntu:24.04 AS base
 
-ARG DEBIAN_FRONTEND=noninteractive
-
-ENV NODE_VERSION=v22.16.0 \
-    CLOJURE_VERSION=1.12.0.1501 \
-    CLJKONDO_VERSION=2025.01.16 \
-    BABASHKA_VERSION=1.12.196 \
-    CLJFMT_VERSION=0.13.0 \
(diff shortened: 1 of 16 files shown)


docker/devenv/Dockerfile
```diff
@@ -138,48 +129,73 @@ RUN set -eux; \
...
+    rustup component add clippy; \
+    git clone https://github.com/emscripten-core/emsdk.git; \
+    cd emsdk; \
+    ./emsdk install $EMSCRIPTEN_VERSION; \
+    ./emsdk activate $EMSCRIPTEN_VERSION; \
+    rustup target add wasm32-unknown-emscripten; \
+    cargo install cargo-watch;
+
+
+# RUN set -eux; \
```
superalex: Are we leaving this commented?

PR #6712, opened 2025-06-18
:bug: Fix token unapply for text shapes

hirunatan:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/11337

### Summary

The automatic token unapply was not working for text shapes, because the change is applied to the `:content` attribute. We must dig into text blocks to see what attributes have actually changed.

### Steps to reproduce 

See Taiga Issue.

### Checklist

### DIFF
--- a/common/src/app/common/logic/shapes.cljc
+++ b/common/src/app/common/logic/shapes.cljc
@@ -16,19 +16,30 @@
    [app.common.types.pages-list :as ctpl]
    [app.common.types.shape.interactions :as ctsi]
    [app.common.types.shape.layout :as ctl]
+   [app.common.types.text :as ctt]
    [app.common.types.token :as cto]
-   [app.common.uuid :as uuid]))
+   [app.common.uuid :as uuid]
+   [clojure.set :as set]))
 
 (defn- generate-unapply-tokens
   "When updating attributes that have a token applied, we must unapply it, because the value
    of the attribute now has been given directly, and does not come from the token."
   [changes objects changed-sub-attr]
-  (let [mod-obj-changes (->> (:redo-changes changes)
+  (let [new-objects     (pcb/get-objects changes)
+        mod-obj-changes (->> (:redo-changes changes)
                              (filter #(= (:type %) :mod-obj)))
+
+        text-changed-attrs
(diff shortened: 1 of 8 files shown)


common/src/app/common/logic/shapes.cljc
```diff
@@ -16,19 +16,30 @@
...
-   [app.common.uuid :as uuid]))
+   [app.common.uuid :as uuid]
+   [clojure.set :as set]))
 
 (defn- generate-unapply-tokens
   "When updating attributes that have a token applied, we must unapply it, because the value
    of the attribute now has been given directly, and does not come from the token."
   [changes objects changed-sub-attr]
-  (let [mod-obj-changes (->> (:redo-changes changes)
+  (let [new-objects     (pcb/lookup-objects changes)
```
pabloalba: We're using

```
new-objects            (pcb/get-objects changes)
```

So you don't need to make public `lookup-objects`

common/src/app/common/logic/shapes.cljc
```diff
@@ -16,19 +16,30 @@
...
+        text-changed-attrs
+        (fn [shape]
+          (let [new-shape (get new-objects (:id shape))
+                attrs (ctt/get-diff-attrs (:content shape) (:content new-shape))]
+            (apply set/union (map cto/shape-attr->token-attrs attrs))))
+
         check-attr (fn [shape changes attr]
                      (let [tokens      (get shape :applied-tokens {})
-                           token-attrs (cto/shape-attr->token-attrs attr changed-sub-attr)]
+                           token-attrs (if (not= attr :content)
```
pabloalba:
```
(or (not= (:type shape) :text)  (not= attr :content))
```

common/src/app/common/logic/shapes.cljc
hirunatan: Fantastic. I'll change it.

common/src/app/common/logic/shapes.cljc
hirunatan: Ah true! Paths also have `:content`. Fixing it...

pabloalba (approved):

PR #6713, opened 2025-06-18
:books: Document app.main.data.workspace.undo

mdbenito:
### Summary

This PR

* Adds documentation to the namespace `app.main.data.workspace.undo`
* Adds two missing fields to `schema:undo-entry`: `:undo-group` and `:tags`

### Questions

* I'm not sure about the documentation policy in penpot, in particular whether it's desired to have within-module high-level documentation explaining functionality. I find it very useful and would definitely have found what I wrote helpful, had it been there a few days ago. Hence, this PR :) I hope what I wrote is accurate.
* Using the undo module made me wonder whether "stacking" undo entries shouldn't done with transactions, and the stacking mechanism removed.

### Checklist

### DIFF
--- a/frontend/src/app/main/data/workspace/undo.cljs
+++ b/frontend/src/app/main/data/workspace/undo.cljs
@@ -5,6 +5,35 @@
 ;; Copyright (c) KALEIDOS INC
 
 (ns app.main.data.workspace.undo
+  "Undo management for the workspace.
+
+   There are **undo entries**, **undo transactions** and **undo groups**.
+   The undo stack has a maximum size of `MAX-UNDO-SIZE` (50) entries.
+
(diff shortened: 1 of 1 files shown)


niwinz:
The undo mechanism is waiting to a good face wash. The transactions and undo-group, are very similar operations and would be awesome to have a single one with clear documentation on how to use (if possible, obviously).

The reason we have transactions and timeouts, because on the past (and probably right now) we have a large operations that can be interrupted in a several conditions. Per example, the texts the position-data is calculated asynchronously but should be attached to the same group/transaction for make the undo consistent. This concrete case will dispensary once the WASM render is ready.

It is ok include that kind of doc on code. I'm also consider it very useful.

PR #6725, opened 2025-06-18
:bug: On variants override use the component name instead of the copy name

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot/task/11376

### Summary

When comparing variants to keep overrides, compare by the original component layer names instead of the layer names on the copies

### Steps to reproduce 

1. Create a variant with two identical components with only a rectangle, one red, one blue
2. Make a copy of the first component
3. Add a stroke to the rectangle on the copy
4. Rename the rectangle on the copy
5. Make a switch on the copy to the other component

The stroke should still be there after the switch

### Checklist

### DIFF
--- a/common/src/app/common/logic/variants.cljc
+++ b/common/src/app/common/logic/variants.cljc
@@ -63,28 +63,57 @@
 
 
 (defn generate-keep-touched
+  "This is used as part of the switch process, when you switch from
+   an original-shape to a new-shape. It generate changes to
+   copy the touched attributes on the shapes children of the original-shape
+   into the related children of the new-shape.
+   This relation is tricky. The shapes are related if:
+   * On the main components, both have the same name (the name on the copies are ignored)
+   * Both has the same type of ancestors, on the same order (see generate-path for the
+     translation of the types)"
   [changes new-shape original-shape original-shapes page libraries]
   (let [objects            (pcb/get-objects changes)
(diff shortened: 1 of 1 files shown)


common/src/app/common/logic/variants.cljc
```diff
@@ -65,23 +65,27 @@
...
         new-shapes-map     (into {} (map (juxt :shape-path identity) new-shapes-w-path))
-        orig-touched       (filter (comp seq :touched) orig-shapes-w-path)
 
-        container          (ctn/make-container page :page)]
+        orig-ref-shape     (ctf/find-ref-shape nil container libraries original-shape)
+        o-ref-shapes-wp    (add-unique-path
+                            (reverse (cfh/get-children-with-self objects (:id orig-ref-shape)))
+                            objects
+                            (:id orig-ref-shape))
+        o-ref-shapes-p-map (into {} (map (juxt :id :shape-path) o-ref-shapes-wp))]
```
niwinz: This can be simplified to transducers:

```clojure
(into {} (map (juxt :id :shape-path)) o-ref-shapes-wp)
```

That passes from iterating 2 times and creating 2 collections to iterating once and creating the map collection directly.

pabloalba: Fixed

hirunatan (approved):

PR #6767, opened 2025-06-24
:tada: Support for exif rotated images

superalex:
When uploading an exif rotated image like this:

Landscape_6

The image sizes were not correctly calculated

### DIFF
--- a/backend/src/app/media.clj
+++ b/backend/src/app/media.clj
@@ -10,6 +10,7 @@
    [app.common.data :as d]
    [app.common.data.macros :as dm]
    [app.common.exceptions :as ex]
+   [app.common.logging :as l]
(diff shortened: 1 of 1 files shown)


niwinz:
I think we should fallback to the old approach if the first fails, instead of raising an exception. Maybe we can emit a warning that the first operation failed, but still use the previous approach for get width and height if the first fails.

superalex:
> I think we should fallback to the old approach if the first fails, instead of raising an exception. Maybe we can emit a warning that the first operation failed, but still use the previous approach for get width and height if the first fails.

I really like your suggestion, thanks, PR updated! :)

niwinz:
Would be nice to add an entry to changelog :D

PR #6795, opened 2025-06-26
:sparkles: Keep the swapped childs if the copies when doing a variant switch

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot/task/11375

### Summary

When doing a switch of a component of a variant, we can found that a child (or grandchild, etc) of that component has been swapped. In that case, we want to keep the swapped item.

### Steps to reproduce 

Simple swap (child)

https://github.com/user-attachments/assets/7dd43b21-f80d-4217-9323-5069a4837419

Complex swap (grandchild)

https://github.com/user-attachments/assets/e3bb6924-8a4d-41c9-94bd-5d909384de2a

### Checklist

- [X] Add a detailed explanation of how to reproduce the issue and/or verify the fix, if applicable.
- [X] Include screenshots or videos, if applicable.

### DIFF
--- a/common/src/app/common/files/changes_builder.cljc
+++ b/common/src/app/common/files/changes_builder.cljc
@@ -473,12 +473,14 @@
          (fn [undo-changes shape]
            (let [prev-sibling (cfh/get-prev-sibling objects (:id shape))]
              (conj undo-changes
-                   {:type :mov-objects
-                    :page-id (::page-id (meta changes))
-                    :parent-id (:parent-id shape)
-                    :shapes [(:id shape)]
-                    :after-shape prev-sibling
-                    :index 0}))) ; index is used in case there is no after-shape (moving bottom shapes)
+                   (cond-> {:type :mov-objects
+                            :page-id (::page-id (meta changes))
+                            :parent-id (:parent-id shape)
+                            :shapes [(:id shape)]
+                            :after-shape prev-sibling
+                            :index 0} ; index is used in case there is no after-shape (moving bottom shapes)
+                     (:component-swap options)
+                     (assoc :component-swap true)))))
 
          restore-touched-change
          {:type :mod-obj
--- a/common/src/app/common/logic/libraries.cljc
+++ b/common/src/app/common/logic/libraries.cljc
@@ -2257,10 +2257,21 @@
                                    {}))]))
 
 (defn generate-component-swap
-  [changes objects shape file page libraries id-new-component index target-cell keep-props-values]
-  (let [[all-parents changes]
+  [changes objects shape file page libraries id-new-component
+   index target-cell keep-props-values keep-touched]
+  (let [;; When we keep the touched properties, we can't delete the
+        ;; swapped children (we will keep them too)
+        ignore-swapped-fn
+        (if keep-touched
+          #(-> (get objects %)
+               (ctk/get-swap-slot))
+          (constantly false))
+
+        [all-parents changes]
         (-> changes
-            (cls/generate-delete-shapes file page objects (d/ordered-set (:id shape)) {:component-swap true}))
+            (cls/generate-delete-shapes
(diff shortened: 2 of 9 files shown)


common/src/app/common/files/changes_builder.cljc
```diff
@@ -473,12 +473,14 @@
...
-                    :after-shape prev-sibling
-                    :index 0}))) ; index is used in case there is no after-shape (moving bottom shapes)
+                   (cond-> {:type :mov-objects
+                            :page-id (::page-id (meta changes))
+                            :parent-id (:parent-id shape)
+                            :shapes [(:id shape)]
+                            :after-shape prev-sibling
+                            :index 0} ; index is used in case there is no after-shape (moving bottom shapes)
+                     (:component-swap options)
+                     (assoc :component-swap true)))))
```
hirunatan, line 483: Perhaps not for this PR, but I don't like the name `:component-swap` for this parameter. This implies that the `:mov-objects` change must acquire knowledge of the existence of swap operations, that are of a higher abstraction level.

I'd call it something like `:allow-altering-copies`, for example, since this is the desired result of this parameter, and the change already knows if a shape is a copy or not. If we do so, when reading this builder we can figure out what we want, without the need of going to the change implementation to read, and when reading the change we aren't required to think about the complex swap algorithm.

If we do so, the parameter probably can be merged with the other `:syncing`, that has the same purpose and the same problem.

It's possible that the `:mov-objects` change already has other couplings that should not be there, but at least if we can avoid adding more, the better.

common/src/app/common/logic/libraries.cljc
```diff
@@ -2257,10 +2257,22 @@
...
+  (let [;; When we keep the touched properties, we can't delete the
+        ;; swapped children (we will keep them too)
+        ignore-swapped-fn
+        (if keep-touched
+          #(-> (get objects %)
+               (ctk/get-swap-slot))
+          (constantly false))
+
+        [all-parents changes]
+
```
hirunatan: blank line

common/src/app/common/logic/libraries.cljc
```diff
@@ -2257,10 +2257,22 @@
                                    {}))]))
 
 (defn generate-component-swap
-  [changes objects shape file page libraries id-new-component index target-cell keep-props-values]
-  (let [[all-parents changes]
+  [changes objects shape file page libraries id-new-component
+   index target-cell keep-props-values keep-touched]
```
hirunatan, line 2261: Shouldn't this be `:ignore-swapped?`. I don't see it checking any touched value, only swap slot.

common/src/app/common/logic/shapes.cljc
```diff
@@ -99,7 +99,14 @@
                                (pcb/with-library-data file))
                            ids
                            options))
-  ([changes ids {:keys [ignore-touched component-swap]}]
+  ([changes ids {:keys [ignore-touched
+                        component-swap
```
hirunatan, line 103: Same issue with the name. The other two options speak about touched shapes and children, that are concepts that this function knows about. But I think that swap is an algorithm that is above this level, and the `:component-swap` option should be named for what is the effect required here.

common/src/app/common/logic/variants.cljc
```diff
@@ -106,17 +171,37 @@
     ;; Process each touched children of the original-shape
     (reduce
      (fn [changes orig-child-touched]
-       (let [;; orig-child-touched is in a copy. Get the referenced shape on the main component
```
hirunatan, line 109: I must say that I really love this block :DDDDD

It's a brutally complex thing but you have added some abstraction levels so well defined, that I'll not say that the result is simple, but is understandable by reading the comments and the function names.

Great work.

hirunatan (approved):

niwinz (approved):
I have no relevant feedback

PR #6800, opened 2025-06-27
:wrench: Make TokenSet an abstract data type

hirunatan:
### Related Ticket

[https://tree.taiga.io/project/penpot/task/11448](https://tree.taiga.io/project/penpot/task/11448)

### Summary

Make all access to TokenSet and Token objects to be through helper functions. This way we can change later the internal structure without impact on client code.

### Steps to reproduce 

No change in user-visible behavior. But should make a regression test of all tokens functionality.

### Checklist

### DIFF
--- a/common/src/app/common/files/changes.cljc
+++ b/common/src/app/common/files/changes.cljc
@@ -418,7 +418,7 @@
       [:type [:= :set-token-set]]
       [:set-name :string]
       [:group? :boolean]
-      [:token-set [:maybe ctob/schema:token-set-attrs]]]]
+      [:token-set [:maybe [:fn ctob/token-set?]]]]]
 
     [:set-token
      [:map {:title "SetTokenChange"}
@@ -1025,11 +1025,10 @@
                   (ctob/delete-set lib' set-name))
 
                 (not (ctob/get-set lib' set-name))
-                (ctob/add-set lib' (ctob/make-token-set token-set))
+                (ctob/add-set lib' token-set)
 
                 :else
-                (ctob/update-set lib' set-name (fn [prev-token-set]
-                                                 (ctob/make-token-set (merge prev-token-set token-set)))))))))
+                (ctob/update-set lib' set-name (fn [_] token-set)))))))
 
 (defmethod process-change :set-token-theme
   [data {:keys [group theme-name theme]}]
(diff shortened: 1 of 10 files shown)


common/src/app/common/files/changes.cljc
```diff
@@ -1025,11 +1025,10 @@
                   (ctob/delete-set lib' set-name))
 
                 (not (ctob/get-set lib' set-name))
-                (ctob/add-set lib' (ctob/make-token-set token-set))
+                (ctob/add-set lib' token-set)
 
                 :else
-                (ctob/update-set lib' set-name (fn [prev-token-set]
-                                                 (ctob/make-token-set (merge prev-token-set token-set)))))))))
+                (ctob/update-set lib' set-name (fn [_] token-set)))))))
```
hirunatan, line 1032: Actually we weren't using this to make partial updates. Always sent complete sets.

common/test/common_tests/logic/token_test.cljc
```diff
@@ -269,13 +269,13 @@
...
                       (pcb/with-library-data (:data file))
-                      (pcb/set-token-set set-name false (assoc prev-token-set
-                                                               :name new-set-name)))
+                      (pcb/set-token-set set-name false (ctob/rename-set prev-token-set new-set-name)))
           redo (thf/apply-changes file changes)
           redo-lib (tht/get-tokens-lib redo)
           undo (thf/apply-undo-changes redo changes)
           undo-lib (tht/get-tokens-lib undo)]
 
+      (prn "khanges" changes)
```
niwinz: :fire:

hirunatan: done

common/src/app/common/types/tokens_lib.cljc
```diff
@@ -60,14 +54,38 @@
...
        (not-empty)))
 
+;; === Common
+
+(defprotocol INamedItem
+  "Protocol for items that have a name, a description and a modified date."
+  (get-name [_] "Get the name of the item.")
+  (get-description [_] "Get the description of the item.")
+  (get-modified-at [_] "Get the description of the item.")
+  (rename [_ new-name] "Set the name of the item.")
```
niwinz, line 68: I think this should be consistent with the others, if we use the `set-...` idiom, i think we should use it also for name.

hirunatan: I thought about it. But changing the name has more implications than just setting the value of an attribute (e.g. it may change the position in the tree or the priority of a set). This is the reason I chose to give a name more suggestive.

PR #6816, opened 2025-06-30
:sparkles: Implement object type specific tokens

floscr:
### Related Ticket

https://github.com/tokens-studio/penpot/issues/8

### Summary

https://github.com/user-attachments/assets/54c00a80-11d8-47f2-b2a8-7f94f6d0fddd

Tokens will now be highlighted as unsupported when the given selected shapes dont support the token.
E.g.: typography attributes cannot be applied on a rectangle.

Context menus will also only show the attributes that matches the current selection.

When multiple types are selected the common attributes can be selected, but now the token doesn't get applied anymore to unsupported objects
E.g.: Text & Rectangle are selected, applying border-radius token will only apply it to the rectangle shape.

Further changes:

- Unit tests for context menus
- Updates tests with previous implementations that now broke due to supported attribute matching (it's not allowed anymore to pass arbitrary attributes via the actions)

### Steps to reproduce 

1. Add various token types and check if they can be applied to the supported shapes
2. Nothing should happen to unsupported shapes
3. Everything else should behave as before

### Checklist

### DIFF
--- a/frontend/src/app/main/data/workspace/tokens/application.cljs
+++ b/frontend/src/app/main/data/workspace/tokens/application.cljs
@@ -71,6 +71,16 @@
     :text    text-attributes
     nil))
 
+(defn appliable-attrs
+  "Returns intersection of shape `attributes` for `token-type`."
+  [attributes token-type]
+  (set/intersection attributes (shape-type->attributes token-type)))
+
+(defn any-appliable-attr?
+  "Checks if `token-type` supports given shape `attributes`."
+  [attributes token-type]
+  (seq (appliable-attrs attributes token-type)))
+
 ;; Events to apply / unapply tokens to shapes ------------------------------------------------------------
 
 (defn apply-token
@@ -95,8 +105,8 @@
                         objects (dsh/lookup-page-objects state)
 
                         shape-ids (or (->> (select-keys objects shape-ids)
-                                           (filter (fn [[_ element]]
-                                                     (seq (set/intersection attributes (shape-type->attributes (:type element))))))
+                                           (filter (fn [[_ shape]]
+                                                     (any-appliable-attr? attributes (:type shape))))
                                            (keys))
                                       [])
 
--- a/frontend/src/app/main/data/workspace/tokens/application.cljs
+++ b/frontend/src/app/main/data/workspace/tokens/application.cljs
@@ -7,7 +7,6 @@
 (ns app.main.data.workspace.tokens.application
   (:require
    [app.common.data :as d]
-   [app.common.data.macros :as dm]
    [app.common.files.tokens :as cft]
    [app.common.text :as txt]
    [app.common.types.shape.layout :as ctsl]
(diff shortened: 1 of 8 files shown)


frontend/src/app/main/data/workspace/tokens/application.cljs
```diff
@@ -31,6 +31,46 @@
...
+
+(defn shape-type->attributes
+  [type]
+  (case type
+    :bool    generic-attributes
+    :circle  generic-attributes
+    :rect    rect-attributes
+    :frame   frame-attributes
+    ;; TODO: Groups dont accept tokens yet
+    ;; :group   generic-attributes
```
hirunatan: Groups have not any own attributes, and never will (always inherit position and size from content).

So I think that this TODO is not necessary.

frontend/src/app/main/data/workspace/tokens/application.cljs
```diff
@@ -31,6 +31,46 @@
 
 (declare token-properties)
 
+;; Match shape by attributes ---------------------------------------------------
+
+(def position-attributes #{:x :y})
```
hirunatan: I think all this may also be in `common.types.token`

frontend/src/app/main/data/workspace/tokens/application.cljs
```diff
@@ -55,7 +95,8 @@
                         objects (dsh/lookup-page-objects state)
 
                         shape-ids (or (->> (select-keys objects shape-ids)
-                                           (filter (fn [[_ shape]] (not= (:type shape) :group)))
+                                           (filter (fn [[_ element]]
+                                                     (seq (set/intersection attributes (shape-type->attributes (:type element))))))
```
hirunatan: I'd write a helper function for this. E.g.
```clojure
common.types.token

(defn appliable-attrs
  [attributes type shape]
  (set/intersection attributes (shape-type->attributes type))
```

(and change the name `element`, it's just a shape)

frontend/src/app/main/ui/workspace/tokens/management/token_pill.cljs
```diff
@@ -162,6 +162,13 @@
         shape-ids         (into #{} xf:map-id selected-shapes)]
     (cft/shapes-applied-all? ids-by-attributes shape-ids attributes)))
 
+(defn attributes-match-selection?
+  [selected-shapes attrs]
+  (some (fn [shape]
+          (let [shape-attrs (dwta/shape-type->attributes (:type shape))]
+            (some attrs shape-attrs)))
```
hirunatan: Perhaps you can also use here the helper funcition `any-appliable-attrs?`.

frontend/src/app/main/data/workspace/tokens/application.cljs
hirunatan: Even a second helper `any-appliable-attr?` that returns boolean.

This way you have a higher level interface, and the code in the event is easier to read.

frontend/src/app/main/data/workspace/tokens/application.cljs
floscr: Thanks, I didn't know that :+1:

PR #6842, opened 2025-07-03
:bug: Update rotation or line height depending shape type on apply nu…

Xaviju:
…meric token

### Related Ticket

https://tree.taiga.io/project/penpot/task/11331

### Summary

When applying a numeric token and the selected layer is a text, it should be applied to the line height by default.

### Tecnical notes

- Moved `split-text-shapes` fn from colors to helpers so it can be used in `application.cljs` to discern shapes and text shapes.
- Created an `update-numeric` fn to abstract token type from shape attributes. An applied token must be able to update different attributes, or many of them. `update-attribute-*` fn should work as attribute application helpers. 

### Steps to reproduce 

1. Create two shapes: one must be a common shape and the other a text.
2. Create a numeric token with a value (e.g. `20`)
3. Select both shapes and apply token

Current: both shapes change rotation
Expected: text shapes get a line-height tokens and other shapes update its rotation

### Checklist

### DIFF
--- a/frontend/src/app/main/data/helpers.cljs
+++ b/frontend/src/app/main/data/helpers.cljs
@@ -78,6 +78,23 @@
              (filter selectable?)
              selected)))))
 
+(defn split-text-shapes
+  "Split text shapes from non-text shapes"
+  [objects ids]
+  (loop [ids (seq ids)
+         text-ids []
+         shape-ids []]
+    (if-let [id (first ids)]
+      (let [shape (get objects id)]
+        (if (cfh/text-shape? shape)
+          (recur (rest ids)
+                 (conj text-ids id)
+                 shape-ids)
+          (recur (rest ids)
+                 text-ids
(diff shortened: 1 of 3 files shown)


floscr:
Maybe we could tune the logic to select the default property based on allowed properties instead of adding more logic for specific token types.
https://github.com/penpot/penpot/pull/6816

Xaviju:
> Maybe we could tune the logic to select the default property based on allowed properties instead of adding more logic for specific token types. #6816

I was trying to fix a bug. When a numeric token is applied, it must be applied to a different attribute, depending on whether it is text or a shape. I thought it made sense to decouple the token type from the applied attribute, but it turned out not to be as simple as this.

Since I am completely new to this area of the code, any solution you propose will probably be a better path. I'll talk to the team tomorrow morning and get back to you.

I've marked the PR as draft for now.

PR #6855, opened 2025-07-04
:sparkles: Impements font-family token

floscr:
### Related Ticket

https://github.com/tokens-studio/penpot/issues/105

### Summary

https://github.com/user-attachments/assets/76390087-7e53-4ce9-aad7-29afe78b7c27

- Implements font-family token which has an vector of font-family strings for the value
- Refactors the form component to separate all token types into separate components

### Checklist

### DIFF
--- a/common/test/common_tests/logic/token_apply_test.cljc
+++ b/common/test/common_tests/logic/token_apply_test.cljc
@@ -62,7 +62,11 @@
                                   (ctob/add-token-in-set "test-token-set"
                                                          (ctob/make-token :name "token-letter-spacing"
                                                                           :type :letter-spacing
-                                                                          :value 2))))
+                                                                          :value 2))
+                                  (ctob/add-token-in-set "test-token-set"
+                                                         (ctob/make-token :name "token-font-family"
+                                                                          :type :font-family
+                                                                          :value ["Helvetica" "Arial" "sans-serif"]))))
       (tho/add-frame :frame1)
       (tho/add-text :text1 "Hello World!")))
 
@@ -77,7 +81,8 @@
       (tht/apply-token-to-shape :frame1 "token-color" [:fill] [:fill] "#00ff00")
       (tht/apply-token-to-shape :frame1 "token-dimensions" [:width :height] [:width :height] 100)
       (tht/apply-token-to-shape :text1 "token-font-size" [:font-size] [:font-size] 24)
-      (tht/apply-token-to-shape :text1 "token-letter-spacing" [:letter-spacing] [:letter-spacing] 2)))
+      (tht/apply-token-to-shape :text1 "token-letter-spacing" [:letter-spacing] [:letter-spacing] 2)
+      (tht/apply-token-to-shape :text1 "token-font-family" [:font-family] [:font-family] ["Helvetica" "Arial" "sans-serif"])))
 
 (t/deftest apply-tokens-to-shape
   (let [;; ==== Setup
@@ -93,6 +98,7 @@
         token-dimensions   (tht/get-token file "test-token-set" "token-dimensions")
         token-font-size    (tht/get-token file "test-token-set" "token-font-size")
         token-letter-spacing (tht/get-token file "test-token-set" "token-letter-spacing")
+        token-font-family  (tht/get-token file "test-token-set" "token-font-family")
 
         ;; ==== Action
         changes (-> (-> (pcb/empty-changes nil)
@@ -132,7 +138,10 @@
                                                                                :attributes [:font-size]})
                                                     (cto/apply-token-to-shape {:token token-letter-spacing
                                                                                :shape $
-                                                                               :attributes [:letter-spacing]})))
+                                                                               :attributes [:letter-spacing]})
+                                                    (cto/apply-token-to-shape {:token token-font-family
+                                                                               :shape $
+                                                                               :attributes [:font-family]})))
                                                 (:objects page)
                                                 {}))
 
@@ -157,9 +166,10 @@
     (t/is (= (:fill applied-tokens') "token-color"))
     (t/is (= (:width applied-tokens') "token-dimensions"))
     (t/is (= (:height applied-tokens') "token-dimensions"))
-    (t/is (= (count text1-applied-tokens) 2))
+    (t/is (= (count text1-applied-tokens) 3))
     (t/is (= (:font-size text1-applied-tokens) "token-font-size"))
-    (t/is (= (:letter-spacing text1-applied-tokens) "token-letter-spacing"))))
+    (t/is (= (:letter-spacing text1-applied-tokens) "token-letter-spacing"))
+    (t/is (= (:font-family text1-applied-tokens) "token-font-family"))))
 
 (t/deftest unapply-tokens-from-shape
   (let [;; ==== Setup
@@ -189,7 +199,8 @@
                                                 (fn [shape]
                                                   (-> shape
                                                       (cto/unapply-token-id [:font-size])
-                                                      (cto/unapply-token-id [:letter-spacing])))
+                                                      (cto/unapply-token-id [:letter-spacing])
+                                                      (cto/unapply-token-id [:font-family])))
                                                 (:objects page)
                                                 {}))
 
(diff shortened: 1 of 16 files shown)


common/test/common_tests/logic/token_apply_test.cljc
```diff
@@ -62,7 +62,11 @@
                                   (ctob/add-token-in-set "test-token-set"
                                                          (ctob/make-token :name "token-letter-spacing"
                                                                           :type :letter-spacing
-                                                                          :value 2))))
+                                                                          :value 2))
+                                  (ctob/add-token-in-set "test-token-set"
+                                                         (ctob/make-token :name "token-font-family"
+                                                                          :type :font-family
+                                                                          :value "Helvetica"))))
```
hirunatan: Perhaps it'll be a better test with a list of fonts instead of only one.

frontend/src/app/main/ui/ds/controls/utilities/input_field.cljs
```diff
@@ -20,7 +20,7 @@
    [:class {:optional true} :string]
    [:id :string]
    [:icon {:optional true}
-    [:and :string [:fn #(contains? icon-list %)]]]
+    [:maybe [:and :string [:fn #(contains? icon-list %)]]]]
```
hirunatan, line 23: Why this `:maybe`? The `:icon` property is optional, so you can omit it. But if you provide an `:icon`, it should have a value, not be null. Shouldn't you?

frontend/src/app/main/data/workspace/tokens/application.cljs
```diff
@@ -355,6 +356,22 @@
                            {:ignore-touched true
                             :page-id page-id})))))
 
+(defn update-font-family
+  ([value shape-ids attributes] (update-font-family value shape-ids attributes nil))
+  ([value shape-ids _attributes page-id]
+   (let [font-id (some-> value
+                         (first)
+                         (#(fonts/find-font-data {:family %}))
```
hirunatan: This search should be case insensitive.

Also, if the font family is not found, it should be applied anyway. Penpot will detect that the font does not exist in the system, and display "Missing font" in the design tab and render the shape with a default font, as explained in the US.

imagen

This is the same behavior as if you set a custom font in a shape, and then remove the font from the team.

frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
```diff
@@ -232,21 +197,24 @@
       :class (stl/css-case :resolved-value (not (or empty-message? (seq warnings) (seq errors))))
       :type type}]))
 
-(mf/defc form
+(mf/defc form*
   {::mf/wrap-props false}
```
hirunatan: If you add the `*` you can remove `wrap-props` (the asterisk deactivates the wrap automatically).

frontend/src/app/main/ui/ds/controls/utilities/input_field.cljs
floscr: This is a common pattern for dynamic props (which is also used in the DS components)

When you got a component like this:

```
(let [icon (when some-condition "icon")]
  [:> input-field* {:icon icon}])
```

You would have to dissoc `icon` from the props when the value is nil or it would throw in the schema as the key is present but nil.

But you want to accept also `[:> input-field* {}]`
Without the `icon` prop.
So it's a optional key with maybe value

frontend/src/app/main/data/workspace/tokens/application.cljs
```diff
@@ -355,6 +356,27 @@
...
+  ([value shape-ids attributes] (update-font-family value shape-ids attributes nil))
+  ([value shape-ids _attributes page-id]
+   (let [font-family (first value)
+         font-id (some-> font-family
+                         (fonts/find-font-family)
+                         :id)
+         text-attrs (cond
+                      font-id {:font-id font-id}
+                      font-family {:font-id font-family
+                                   :family font-family}
```
hirunatan: I don't understand this well. Why dont' just always add both font id an family?

Also you are setting `:family` and not `:font-family`.

floscr: I was not sure about this one, I can also assign only `font-id`

floscr: I'm not quite sure what I should assign.
As the font family is not equals the font id.
When the font is not in the db, I thought i would assign only the font-family, but then you don't get the "Missing font" hint in the design tab (that's why I've assigned both for this case).

frontend/src/app/main/ui/workspace/tokens/management/token_pill.cljs
```diff
@@ -106,7 +106,9 @@
 (defn- generate-tooltip
   "Generates a tooltip for a given token"
   [is-viewer shape theme-token token half-applied no-valid-value ref-not-in-active-set]
-  (let [{:keys [name value type resolved-value]} token
+  (let [{:keys [name type resolved-value]} token
+        value (cond->> (:value token)
+                (= :font-family type) (str/join ","))
```
hirunatan: This should have an space `", "`.

floscr: I'm now re-using the method https://github.com/penpot/penpot/pull/6855/commits/0edf8004f406b6833bf4377e312a4edc549a4e59

frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
```diff
@@ -555,20 +489,31 @@
...
-         [:> ramp* {:color (some-> color (tinycolor/valid-color))
-                    :on-change on-update-color}])
+       (let [placeholder (tr "workspace.tokens.token-value-enter")
+             label (tr "workspace.tokens.token-value")
+             default-value (mf/ref-val value-ref)
+             ref value-input-ref
+             error (not (nil? (:errors token-resolve-result)))
+             on-blur on-update-value]
+         (if (fn? token-value-input)
+           [:> token-value-input
```
Xaviju: Using the name `token-value-input` and `input-token-value` (L509) might cause confusion. Is this a slot? I suggest either being more specific or renaming the latter.

floscr: This renders either the component prop `token-value-input` if defined, or falls back to the default input component `input-tokens-value*`.
I'm not sure I would call it a "slot" because the props passed are not the same, and the default is being replaced.
But if this fits the naming schema of this code base I can rename it with a slot prefix.
Otherwise maybe `custom-token-value-input` to clarify?

Xaviju: We don't have a team agreement on this. 
We recently decided to use the term 'slot' when components are passed through props, as this is the way we can mimic the web Component slot feature in React components.
That said, in this specific case, both options are fine, we could use 'custom' as well.

The confusion is also related to the order of the terms. We use 'input-token-value'. In this case, the words are ordered by importance (input > tokens > value), but you used token-value-input. If both elements are replaceable, could we use the same order with the mentioned prefix? 

So `custom-token-value-input` and `token-value-input` would be fine to me.

I'm not sure which one is better 🤔

floscr: Renamed it: https://github.com/penpot/penpot/pull/6855/commits/f60778a4fbc0c32f2e55eb401814f693b8ac8a6c

frontend/src/app/main/data/workspace/tokens/application.cljs
hirunatan: I've just figured out how we should manage this.

All fonts you can find with `find-font-family` have an `:id` and a `:family`. The family is the value visible to users. The id is a string like "alierta stencil" for system default fonts, and a stringified uuid for custom fonts uploaded by users to the team.

What you need to do after calling `find-font-family` is:
 - If the function returns a font, take `:id` and `:family` and put them as `:font-id` and `font-family` in the `text-attrs` to send to the shape.
 - If not (return value is nil), set `:font-family` to the value of the token, and `:font-id` to a stringified `uuid/zero`.

The first case will display the correct values both for system and custom fonts. The second one will display "Missing font" in the design panel to the right and will render the shape using the default font. And if in the future the user uploads a custom font with a name matching the token value, it will be automatically mapped when opening the file.

frontend/src/app/main/ui/ds/controls/utilities/input_field.cljs
hirunatan: This is ok then. Thanks.

PR #6868, opened 2025-07-08
:lipstick: Implement (p)print methods / protocols for several types

mdbenito:
## Context

Currently,

```clojure
(let [tokens-lib (get-in @st/state [:files some-file :data :tokens-lib])]
  (pprint tokens-lib))
```

produces

```
#object[app.common.types.tokens-lib.TokensLib]
```

and the same happens for `TokenSets`. When printing to file one obtains "unreadable form" characters `#<`. Instead we want to print tags `#penpot/tokens-lib` and `#penpot/token-set`.

## Summary of changes

This PR improves pretty-printing both in clojure and clojurescript by:

- Adding `print-method`, `print-dup` and `pprint/simple-dispatch` implementations for `TokenSet` and `TokensLib`
- Implementing `cljs.core/IPrintWithWriter` for `TokenSet` and `TokensLib`
- ~Additionally adding `print-method`, `print-dup` and `pprint/simple-dispatch` implementations for `PathData`, for consistency with the tag in cljs which is already implemented.~

## Open questions

1. The implementation is not complete for cljs, since the dispatch mechanism for `pprint` is rather cumbersome and I didn't think `a.c.t.tokens-lib` to be the proper place for something like this:

```clojure
(ns app.some.namespace
  (:require [cljs.pprint :as pp]))

(defmulti pprint-dispatch
  (fn [obj]
    (if (instance? TokensLib obj)
      :tokens-lib
      (pp/type-dispatcher obj))))

(defmethod pprint-dispatch :tokens-lib
  [^TokensLib obj]
  (-write *out* "#penpot/tokens-lib ")
  (pp/pprint-newline :miser)
  (pp/pprint (deref obj)))

(defmethod pprint-dispatch :default
  [obj]
  (pp/simple-dispatch obj))
  
(pp/set-pprint-dispatch pprint-dispatch)
```

2. I saw some instances of:

```clojure
(extend-protocol fez/IEdn 
  ... )
```

but didn't find obvious places where this was actually used, so I left that out. Should I include it for both types?

3. There seems to be some inconsistency in the use of tags. Sometimes it's `app/something`, sometimes `penpot/something`. I find the former to be too generic, maybe we should change all uses? I just don't know what that might break.

### Checklist

### DIFF
--- a/common/src/app/common/types/tokens_lib.cljc
+++ b/common/src/app/common/types/tokens_lib.cljc
@@ -8,6 +8,7 @@
   (:require
    #?(:clj [app.common.fressian :as fres])
    #?(:clj [clojure.data.json :as json])
+   #?(:clj [clojure.pprint :as pp])
    [app.common.data :as d]
    [app.common.data.macros :as dm]
    [app.common.files.helpers :as cfh]
@@ -186,7 +187,13 @@
                                    "name" (clj->js name)
                                    "description" (clj->js description)
                                    "modified-at" (clj->js modified-at)
-                                   "tokens" (clj->js tokens)))])
+                                   "tokens" (clj->js tokens)))
+
+             cljs.core/IPrintWithWriter
+             (-pr-writer [^TokenSet this writer opts]
+                         (-write writer "#penpot/token-set ")
+                         (-pr-writer (deref this) writer opts))])
+
   INamedItem
   (get-name [_]
     name)
@@ -250,6 +257,20 @@
 
   (get-tokens-map [_]
     tokens))
+
+#?(:clj
+   (do
+     (defmethod print-method TokenSet [^TokenSet this ^java.io.Writer w]
+       (.write w "#penpot/token-set ")
+       (print-method (deref this) w))
+
+     (defmethod print-dup TokenSet [^TokenSet this ^java.io.Writer w]
+       (print-method this w))
+
+     (defmethod pp/simple-dispatch TokenSet [^TokenSet obj]
+       (.write *out* "#penpot/token-set ")
+       (pp/pprint-newline :miser)
+       (pp/pprint (deref obj)))))
 
 (defn token-set?
   [o]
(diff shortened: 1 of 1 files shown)


niwinz:
Just  for the info: we are using app.common.pprint for pretty almost for everything which requires pretty printing which uses an external library. So adding code for clojure pprint can have less value than expected.

mdbenito:
> Just for the info: we are using app.common.pprint for pretty almost for everything which requires pretty printing which uses an external library. So adding code for clojure pprint can have less value than expected.

I see. I have tried `app.common.pprint/pprint`, but e.g. on the tokens lib it prints:

```
#object[app.common.types.tokens_lib.TokensLib 0x4cc1e625 {:status :ready, :val {:sets #linked/map [["S-Foundations" #object[app.common.types.tokens_lib.TokenSet 0x24d9e51c {:status :ready, :val {:id #uuid "811511f7-ff2c-81b5-8006-743b3d3bb51a", ...
```

which seems suboptimal to me. Not only do we have a one-liner, but also generic #object tags, object addresses and a fixed depth (I tried the *print-level* binding without success).

mdbenito:
The reason for this PR is that during an intermediate stage we are processing files exported as .edn, sometimes pprinted stuff right out of the repl, and TokensLib would make the lexer choke on the generated `#<` chars. I wasn't aware of `app.common.pprint` and the external library. I guess that solves that issue, but given how it fails to print the whole object, I would advocate for merging this PR if that's ok (instead of trying to figure out why the external library won't do what I need).

In summary: this would help, and save some time for me. If it doesn't introduce any issues, I think the added internal consistency in formatting is a nice bonus too.

mdbenito:
Update @niwinz I removed the changes to `PathData` because pprinting would result in:

```
; Execution error (ClassCastException) at app.common.types.path.impl/eval19148$fn (impl.cljc:694).
; class app.common.types.path.impl.PathData cannot be cast to class java.util.concurrent.Future (app.common.types.path.impl.PathData is in unnamed module of loader clojure.lang.DynamicClassLoader @3a87e454; java.util.concurrent.Future is in module java.base of loader 'bootstrap')
```

I don't know how I missed it, sorry about that.

niwinz:
I'm working on it, just FYI. I have included your commit in my PR and my plan is also normalize the tags.

niwinz:
I close this PR, the relevant work on this is included on this PR: https://github.com/penpot/penpot/pull/6967

PR #6870, opened 2025-07-08
:bug: Fixes from subscription design review

cocotime:
### Related Ticket

https://tree.taiga.io/project/penpot/task/11470
https://tree.taiga.io/project/penpot/task/11526
https://tree.taiga.io/project/penpot/task/11549

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/dashboard/subscription.cljs
+++ b/frontend/src/app/main/ui/dashboard/subscription.cljs
@@ -18,10 +18,10 @@
    [potok.v2.core :as ptk]
    [rumext.v2 :as mf]))
 
-(defn get-subscription-name
-  [subscription]
-  (if (and subscription (not (some #{"unpaid" "canceled"} [(:status subscription)])))
-    (:type subscription)
+(defn get-subscription-type
+  [{:keys [type status] :as subscription}]
+  (if (and subscription (not (contains? #{"unpaid" "canceled"} status)))
+    type
     "professional"))
 
 (mf/defc cta-power-up*
@@ -52,11 +52,11 @@
 (mf/defc subscription-sidebar*
   [{:keys [profile]}]
   (let [subscription           (:subscription (:props profile))
-        subscription-name      (get-subscription-name subscription)
+        subscription-type      (get-subscription-type subscription)
         subscription-is-trial  (= (:status subscription) "trialing")
         subscription-href      (dm/str (u/join cf/public-uri "#/settings/subscriptions"))]
 
-    (case subscription-name
+    (case subscription-type
       "professional"
       [:> cta-power-up*
        {:top-title (tr "subscription.dashboard.power-up.your-subscription")
@@ -92,7 +92,7 @@
 (mf/defc team*
   [{:keys [is-owner team]}]
   (let [subscription          (:subscription team)
-        subscription-name     (get-subscription-name subscription)
+        subscription-type     (get-subscription-type subscription)
         subscription-is-trial (= "trialing" (:status subscription))
 
         go-to-manage-subscription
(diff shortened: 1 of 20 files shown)


frontend/src/app/main/ui/dashboard/subscription.cljs
```diff
@@ -18,6 +18,12 @@
    [potok.v2.core :as ptk]
    [rumext.v2 :as mf]))
 
+(defn get-subscription-name
```
niwinz: This function returns a type of the subscription, but the name of function indicates it returns a name. Name is something semantically human-readable, and type is semantically a categorization. In this case I consider this function should be called `get-subscription-type`

frontend/src/app/main/ui/dashboard/subscription.cljs
```diff
@@ -18,6 +18,12 @@
...
+  [subscription]
+  (if (and subscription (not (some #{"unpaid" "canceled"} [(:status subscription)])))
+    (:type subscription)
+    "professional"))
```
niwinz, line 25: I suggest implementing this function in this way:

```clojure
(defn get-subscription-type
  [{:keys [type status] :as subscription}]
  (if (and subscription (not (contains? #{"unpaid" "canceled"} status)))
    (:type subscription)
    "professional"))
```

frontend/src/app/main/ui/settings/subscription.cljs
```diff
@@ -52,42 +54,47 @@
...
-                                           (st/emit! (rt/nav-raw :href href))))))
+        initial                    (mf/with-memo []
+                                     {:min-members (or (some->> teams (map :total-editors) (apply max)) 1)})
+        form                       (fm/use-form :schema schema:seats-form
+                                                :initial initial)
+        subscribe-to-unlimited     (mf/use-fn
+                                    (fn [form]
+                                      (let [data (:clean-data @form)
+                                            current-href (rt/get-current-href)
+                                            returnUrl (js/encodeURIComponent current-href)
```
niwinz: should be `return-url` and this can be simplified with

```clojure
return-url (-> (rt/get-current-href) (rt/encode-url))
```

frontend/src/app/main/ui/settings/subscription.cljs
```diff
@@ -52,42 +54,47 @@
...
+                                                                         :type "unlimited"
+                                                                         :quantity (:min-members data)})
+                                                  (rt/nav-raw :href href)))))
+
+        subscribe-to-enterprise   (mf/use-fn
+                                   (fn []
+                                     (st/emit! (ptk/event ::ev/event {::ev/name "create-trial-subscription"
+                                                                      :type "enterprise"}))
+                                     (let [current-href (rt/get-current-href)
+                                           returnUrl (js/encodeURIComponent current-href)
```
niwinz: The same here

frontend/src/app/main/ui/settings/subscription.cljs
```diff
@@ -201,13 +217,14 @@
   (let [route                           (mf/deref refs/route)
         params                          (:params route)
         params-subscription             (:subscription (:query params))
+        show-trial-subscription-modal   (and (:query params)
+                                             (or (= (:subscription (:query params)) "subscription-to-penpot-unlimited")
+                                                 (= (:subscription (:query params)) "subscription-to-penpot-enterprise")))
```
niwinz: the `(:query params)` is used many times, in the same way as `(:subscription query-params)`; also there is `params-subscription` which is already `(:subscription (:query params))` so, this code can be simplified to something like this:

```clojure
(let [route        (mf/deref refs/route)
      params       (:params route)
      subscription (:subscription (:query params))

      show-trial-modal?
      (or (= subscription "subscription-to-penpot-unlimited")
          (= subscription "subscription-to-penpot-enterprise"))
...
```

frontend/src/app/main/ui/dashboard/subscription.cljs
niwinz: I also suggest replace `subscription-name` usage with `subscription-type` in all the code.

niwinz (changes requested):

frontend/src/app/main/ui/dashboard/subscription.cljs
cocotime: Updated. I just kept 'subscription-name' for one case that makes sense, when I'm sending the name of the subscription to the success modal.

frontend/src/app/main/ui/dashboard/subscription.cljs
cocotime: updated

frontend/src/app/main/ui/settings/subscription.cljs
cocotime: updated

frontend/src/app/main/ui/settings/subscription.cljs
cocotime: The variable was created, but I forgot to use it :joy:. I kept the name `params-subscription` because I'm already using `subscription`.

niwinz (approved):

PR #6875, opened 2025-07-08
:wrench: Fix building and running devenv (Mac / Linux)

belen-albeza:
- **:wrench: Restore the location of chown in devenv**
- **:wrench: Remove COPY of apt.sources (linux issue)**

### DIFF
--- a/docker/devenv/files/entrypoint.sh
+++ b/docker/devenv/files/entrypoint.sh
@@ -10,7 +10,7 @@ cp /root/.bashrc /home/penpot/.bashrc
 cp /root/.vimrc /home/penpot/.vimrc
 cp /root/.tmux.conf /home/penpot/.tmux.conf
 
-chown -R penpot:users /home/penpot
+chown penpot:users /home/penpot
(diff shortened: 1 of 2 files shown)


docker/devenv/files/entrypoint.sh
```diff
@@ -6,12 +6,12 @@ EMSDK_QUIET=1 . /opt/emsdk/emsdk_env.sh;
 
 usermod -u ${EXTERNAL_UID:-1000} penpot;
 
-cp /root/.bashrc /home/penpot/.bashrc
-cp /root/.vimrc /home/penpot/.vimrc
-cp /root/.tmux.conf /home/penpot/.tmux.conf
+# cp /root/.bashrc /home/penpot/.bashrc
+# cp /root/.vimrc /home/penpot/.vimrc
+# cp /root/.tmux.conf /home/penpot/.tmux.conf
```
niwinz: This can break non-interactive usage of devenv: per example bundle builds, because the init.sh is not executed there.

PR #6880, opened 2025-07-09
:bug: Handle empty paragraph on entire selected text deletion

elenatorro:
### Related Ticket

I saw this problem while reviewing this other issue https://tree.taiga.io/project/penpot/issue/9256 

### Summary

In Firefox, when selecting all text in a text area (CTRL + A) and then removing it, it broke. I couldn't reproduce the similar behavior at Chrome. This fixes the issue in Firefox and in Chrome it keeps working as expected.

For what I've seen, the problem seems to be that when doing CTRL + A, is the text editor container the one being returned by the Selection API on Firefox. That's why it behaves differently. The Selection API returns the editor as start and end containers, and a startOffset of 0 and an endOffset of 1.

I've reviewed if we were setting properly the user-select css property in the containers correctly and it seems that's fine, so maybe we should try to handle this scenario in particular.

### Steps to reproduce 

- Create a text area
- Write some text
- Select the whole text
- Remove the text
- No error should rise

### Checklist

### DIFF
--- a/frontend/text-editor/src/editor/content/dom/TextNode.js
+++ b/frontend/text-editor/src/editor/content/dom/TextNode.js
@@ -9,6 +9,7 @@
 import { isInline } from "./Inline.js";
 import { isLineBreak } from "./LineBreak.js";
 import { isParagraph } from "./Paragraph.js";
+import { isEditor } from "./Editor.js";
 import { isRoot } from "./Root.js";
 
 /**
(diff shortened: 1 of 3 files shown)


frontend/text-editor/src/editor/content/dom/TextNode.js
```diff
@@ -60,5 +60,5 @@ export function getClosestTextNode(node) {
   if (isInline(node)) return node.firstChild;
   if (isParagraph(node)) return node.firstChild.firstChild;
   if (isRoot(node)) return node.firstChild.firstChild.firstChild;
-  throw new Error("Cannot find a text node");
+  return null;
```
AzazelN28: This should `throw` to prevent silent errors.

Issue #4246, opened 2024-03-11
feature: Block two finger touchpad back/forward gesture on the whole browser window


unaaagy:
### Is your feature request related to a problem? Please describe.

The touchpad back/forward gesture is only blocked on the canvas area. However, on the toolbar areas, it isn't.
I always have to be careful to not to accidently go back while scrolling the botton color toolbar.

### Describe the solution you'd like.

Block the touchpad gesture on the whole window.

### Additional context

[Screencast from 2024-03-11 20-53-42.webm](https://github.com/penpot/penpot/assets/155830121/8a7fa0dd-2997-4104-b8c7-40f007c7e832)

madalenapmelo-kp:
Hi @unaaagy,

Thank you for reporting this! Could you please confirm which browser you are working with?

unaaagy:
I'm currently using Google Chrome in Fedora 39. Since 2 finger back/forward is not enabled by default I have enabled it by `google-chrome --enable-features=TouchpadOverscrollHistoryNavigation`

Edit: It also seems to happen in Firefox.

madalenapmelo-kp:
Hi @unaaagy,

I've created an enhancement on Taiga so that we can explore this option, the details can be found here: https://tree.taiga.io/project/penpot/issue/7454

Shywim:
Hello, this is also a problem on Arc (chrome-based) on MacOS, but not on Safari.

James-W-LTU:
Hi, this problem is also in Vivaldi (Chromium based browser) on MacOs.

dfelinto:
> The touchpad back/forward gesture is only blocked on the canvas area. 

On my experience even on the canvas area it often also gets captured by the browser. I'm about to submit a PR for this, testing would be appreciated (though not straightforward since you would need to self-host to test it).

PR #6887, opened 2025-07-09
:bug: Fix touchpad swipe back/forward #4246

dfelinto:
### Related Ticket

Github: #4246
Taiga: https://tree.taiga.io/project/penpot/issue/7454

### Summary

This is intentionally a fix for *most* of the cases. As commented on the code, users may still run into the original issue in some corner cases.

In my tests I need to go from the workspace to the dashboard and back very quickly for this to happen, and it doesn't even happen every time. I think having the ability to still use gestures on the dashboard justifies this quirk.

### Steps to reproduce

1. Go to the dashboard.
2. Open a document.
3. Use the trackpad to pan around.

The browser should *not* navigate back to the dashboard.

_Tested on MacOS + Google Chrome._

### Checklist

- [ ] ~~Include screenshots or videos, if applicable.~~
- [x] Add or modify existing integration tests in case of bugs or new features, if applicable. ¹

¹ The following tests are failing in the `develop` branch, even without my patch:
```
[default] › playwright/ui/specs/subscriptions-dashboard.spec.js:85:3 › Subscriptions: team members and invitations › Team settings has susbscription name and no manage subscription link when is member 
[default] › playwright/ui/specs/variants.spec.js:114:1 › User duplicates a variant container 
```

### DIFF
--- a/frontend/src/app/main/ui.cljs
+++ b/frontend/src/app/main/ui.cljs
@@ -28,6 +28,7 @@
    [app.main.ui.onboarding.team-choice :refer [onboarding-team-modal]]
    [app.main.ui.releases :refer [release-notes-modal]]
    [app.main.ui.static :as static]
+   [app.util.dom :as dom]
    [app.util.i18n :refer [tr]]
    [app.util.theme :as theme]
    [beicon.v2.core :as rx]
@@ -361,6 +362,8 @@
 
     ;; initialize themes
     (theme/use-initialize profile)
+
+    (dom/prevent-browser-gesture-navigation!)
 
     [:& (mf/provider ctx/current-route) {:value route}
      [:& (mf/provider ctx/current-profile) {:value profile}
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,7 @@
 - Fix problem when creating a layout from an existing layout [Taiga #11554](https://tree.taiga.io/project/penpot/issue/11554)
 - Fix title button from Title Case to Capitalize [Taiga #11476](https://tree.taiga.io/project/penpot/issue/11476)
 - Fix Main component receives focus and is selected when using 'Show Main Component' [Taiga #11402](https://tree.taiga.io/project/penpot/issue/11402)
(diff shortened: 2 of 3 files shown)


frontend/src/app/main/ui.cljs
```diff
@@ -353,6 +353,15 @@
...
        nil)]))
 
+(defn use-prevent-browser-gesture-navigation!
+  []
+  ;; Prevent the browser from interpreting trackpad horizontal swipe as back/forth
+  ;;
+  ;; In theory We could disable this only for the workspace. However gets too unreliable.
+  ;; It is better to be safe and disable for the dashboard as well.
+  (set! (.. js/document -documentElement -style -overscrollBehaviorX) "none")
+  (set! (.. js/document -body -style -overscrollBehaviorX) "none"))
```
niwinz: I think we should name it `prevent-browser-gesture-navigation!`  (removing the `use-` prefix because it is not a react hook and right now use- prefix indicates a react hook)

And move it under `app.util.dom` because there we have all the dom manipulation helpers.

niwinz:
Hey, I've leaved a comment, Do you prefer do prefer make the change yourself or also I can do it myself.
In any case, thanks for the contribution :heart:

dfelinto:
I can do it later today.

The name was indeed a left-over from my early attempy of making a hook based on the route, good catch. And app.util.dom it is :) I was indeed struggling a bit finding the right place.

By the way, is it best to do a force rebase on my branch or a "fixup" commit? (will do the latter unless I hear otherwise).

dfelinto:
Hm I think I do need to do the rebase to keep the PR 1:1 with the expected commit.

dfelinto:
I force updated so it passes the github CI.

niwinz:
Yep, force update Is OK :D

dfelinto:
Just to be clear, I addressed the changes you suggested already.

If there are other tweaks let me know. Also I won't mind if you rather tackle things yourself either if it simpler/speeds things.

Either way, thanks for taking the time to review this.

PR #6893, opened 2025-07-10
:sparkles: Keep color data when copying from info tab into CSS

Xaviju:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/11144

### Summary

- Keep format and data when copying colors from the info tab into CSS
- Updated from legacy HSL syntax into modern syntax [hsl()](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl)

### Steps to reproduce 

Create texts with:

1. A solid color
2. A color with opacity
3. A linear gradient
4. A radial gradient

Go to info tab and test that when copying the color is the expected CSS prop-value:

- Solid color and with opacity using formats: HEX, RGBA and HSL. 
- Gradients, both linear and radial.

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/inspect/attributes/text.cljs
+++ b/frontend/src/app/main/ui/inspect/attributes/text.cljs
@@ -7,10 +7,12 @@
 (ns app.main.ui.inspect.attributes.text
   (:require-macros [app.main.style :as stl])
   (:require
+   [app.common.colors :as cc]
    [app.common.data :as d]
    [app.common.data.macros :as dm]
+   [app.common.math :as mth]
    [app.common.text :as txt]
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/ui/inspect/attributes/text.cljs
```diff
@@ -35,12 +38,56 @@
           (get-in state [:viewer-libraries file-id :data :typographies]))]
     #(l/derived get-library st/state)))
 
+(defn alpha->hex [alpha]
```
niwinz, line 41: I think it should be private

niwinz (changes requested):

niwinz (approved):

PR #6901, opened 2025-07-15
:sparkles: Hide bounding box while editing visual effects

superalex:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/11576

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@
 - Provide CSS `mix-blend-mode` property in code editor when present on shape [Taiga #11282](https://tree.taiga.io/project/penpot/issue/11282) 
 - Add the option to import tokens in a .zip file. [Taiga #11378](https://tree.taiga.io/project/penpot/us/11378)
 - New typography token type - font size token [Taiga #10938](https://tree.taiga.io/project/penpot/us/10938)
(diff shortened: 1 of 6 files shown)


frontend/src/app/main/ui/workspace/sidebar/options/utils.cljs
```diff
@@ -0,0 +1,7 @@
+(ns app.main.ui.workspace.sidebar.options.utils
+  (:require [app.main.data.workspace :as dw]
+            [app.main.store :as st]))
+
+(defn trigger-bb-cloaking! [ids]
+  (when (seq ids)
+    (st/emit! (dw/trigger-bounding-box-cloaking ids))))
```
niwinz: I'm not clearly understand this function, you can emit the event directly on all the sites where `trigger-bb-cloaking!` helper is used, the difference is 1 line,  and that additional line can be avoided by just making the check in the event.

I suggest emitting event directly on the handlers.

superalex: Thanks!,

```
I've checked the trigger-bounding-box-cloaking do this check so the when is not needed:
(dm/assert!
  "expected valid coll of uuids"
  (every? uuid? ids))
```

PR #6903, opened 2025-07-16
:bug: Fix error on validating file referential integrity when duplicating a page

superalex:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/10774

### Summary

Duplicating a page with mainInstance shapes nested inside groups could result in broken parent/child relationships and schema validation errors. The duplication logic now ensures referential integrity between parent groups and their mainInstance children, so all parent-id and :shapes links are consistent after duplication.

### Checklist

### DIFF
--- a/frontend/src/app/main/data/workspace/pages.cljs
+++ b/frontend/src/app/main/data/workspace/pages.cljs
@@ -174,12 +174,15 @@
             add-component-copy
             (fn [objs id shape]
               (let [component (ctkl/get-component fdata (:component-id shape))
+                    parent-id (when (not= (:parent-id shape) uuid/zero) (:parent-id shape))
                     [new-shape new-shapes]
                     (ctn/make-component-instance page
                                                  component
                                                  fdata
                                                  (gpt/point (:x shape) (:y shape))
-                                                 {:keep-ids? true :force-frame-id (:frame-id shape)})
+                                                 {:keep-ids? true
+                                                  :force-frame-id (:frame-id shape)
+                                                  :force-parent-id parent-id})
                     children (into {} (map (fn [shape] [(:id shape) shape]) new-shapes))
                     objs (assoc objs id new-shape)]
                 (merge objs children)))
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/data/workspace/pages.cljs
```diff
@@ -167,44 +167,52 @@
...
                     (ctn/make-component-instance page
                                                  component
                                                  fdata
                                                  (gpt/point (:x shape) (:y shape))
                                                  {:keep-ids? true :force-frame-id (:frame-id shape)})
-                    children (into {} (map (fn [shape] [(:id shape) shape]) new-shapes))
-                    objs (assoc objs id new-shape)]
+                    ;; Force the parent-id of the cloned mainInstance
+                    new-shape (assoc new-shape :parent-id parent-id)
+                    children (into {} (map (fn [shape]
```
niwinz: I suggest using `reduce` instead of into + map, for efficiency

frontend/src/app/main/data/workspace/pages.cljs
```diff
@@ -167,44 +167,52 @@
...
+            ;; This is essential for updating parent-child relationships: when an object is cloned and assigned a new id, any parent referencing it must update its :shapes vector to use the new id.
+            ;; id-remap ensures referential integrity in the duplicated page structure.
+            id-remap           (atom {})
+            ;; Clone a main instance and its descendants, updating id-remap
             add-component-copy
-            (fn [objs id shape]
+            (fn [objs shape parent-id]
               (let [component (ctkl/get-component fdata (:component-id shape))
                     [new-shape new-shapes]
                     (ctn/make-component-instance page
```
hirunatan, line 179: I've just found that `make-component-instance` function has a bug. It always sets the `parent-id` of the root of the new copy the same as the `frame-id`. It works when the copy is inside a frame, but not if it's under a group. I think you cannot cause this in normal cases, because you cannot drag a component into a group, but it happens in the present case when duplicating a page. I'd add a `force-parent-id` parameter to `make-component-instance` instead of setting it after the call. The parameter could be passed into `ctst/clone-shape` function inside.

superalex: Thank you very much @hirunatan !!! <3

hirunatan (approved):

PR #6917, opened 2025-07-19
:sparkles: Enable wheel scrolling over templates-section in the dashboard

mdbenito:
### Summary

This is just a little change to enable wheel-up/down scrolling over the horizontal "libraries & templates" section in the dashboard.

Emulating the default browser behaviour when scrolling proved to be a bit tricky, so I chose instead to have **discrete scrolling** with one wheel tick = one thumbnail. This is suboptimal for trackpad users, but for them, the previous left-to-right scrolling still works smoothly and continuously as expected.

I experimented with `requestAnimationFrame` in order to have smooth wheel scrolling, but I couldn't make it really work nicely with `scrollBy`. The issue seems to be that trackpad scrolling somehow tracks when the fingers are still on it.

Despite the shortcomings mentioned, I believe discrete scrolling to be an improvement over the previous situation for mouse users, hence this PR.

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/dashboard/templates.cljs
+++ b/frontend/src/app/main/ui/dashboard/templates.cljs
@@ -17,6 +17,7 @@
    [app.main.store :as st]
    [app.main.ui.icons :as i]
    [app.util.dom :as dom]
+   [app.util.dom.normalize-wheel :as nw]
    [app.util.i18n :refer [tr]]
    [app.util.keyboard :as kbd]
    [app.util.storage :as storage]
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/ui/dashboard/templates.cljs
```diff
@@ -206,6 +207,17 @@
          (fn [_event]
            (swap! collapsed* not)))
 
+        on-wheel
+        (mf/use-fn
+         (mf/deps content-ref)
```
niwinz, line 212: Ref should not be on deps

niwinz:
I will merge it with some additional changes.

frontend/src/app/main/ui/dashboard/templates.cljs
mdbenito: Understood!

PR #6924, opened 2025-07-21
:sparkles: Simplify docker build script

niwinz:
The main idea behind this change is simplify the docker images building script. This remove the need of the an indirection level on our deployment scripts.

### DIFF
--- a/docker/images/build.sh
+++ b/docker/images/build.sh
@@ -1,14 +1,17 @@
 #!/usr/bin/env bash
 set -x
 
-DOCKER_CLI_EXPERIMENTAL=enabled
+IMAGE=${1:-backend}
+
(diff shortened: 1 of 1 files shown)


docker/images/build.sh
```diff
@@ -1,14 +1,18 @@
...
 
+IMAGE=${1:-backend}
+
+OUTPUT="type=registry"
+
+if [ "--local" = "$2" ]; then
+    OUTPUT="type=docker"
+fi
+
 DOCKER_CLI_EXPERIMENTAL=enabled
```
yamila-moreno: This option (not sure if it's used in fact), but it seems to be deprecated; I suggest to remove it.
image

yamila-moreno (approved):

PR #6944, opened 2025-07-23
:bug: Fix the context menu always closes after any action

superalex:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/11624

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/workspace/context_menu.cljs
+++ b/frontend/src/app/main/ui/workspace/context_menu.cljs
@@ -59,6 +59,14 @@
            on-unmount children is-selected icon disabled value]}]
   (let [submenu-ref (mf/use-ref nil)
         hovering?   (mf/use-ref false)
+
+        on-click'
+        (mf/use-fn
+         (mf/deps on-click)
+         (fn [event]
+           (st/emit! dw/hide-context-menu)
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/ui/workspace/context_menu.cljs
```diff
@@ -59,6 +59,12 @@
            on-unmount children is-selected icon disabled value]}]
   (let [submenu-ref (mf/use-ref nil)
         hovering?   (mf/use-ref false)
+        handle-menu-click
+        (fn [on-click]
+          (fn [event]
+            (st/emit! dw/hide-context-menu)
+            (when on-click (on-click event))))
```
niwinz: With that code you are generating a new callback each time you render a component, and this is a bad practice. It has no substantial relevance for this code, but I think we should avoid this kind of patters for avoid future copy paste.

I suggest using something like this:

```clojure
(let [on-click'
      (mf/use-fn
       (mf/deps on-click)
       (fn [event]
         (st/emit! dw/hide-context-menu)
         (when on-click
           (on-click event))))
      ...
```

superalex: Ops!, you are completely right!, thanks!

PR #6974, opened 2025-07-25
:books: Add doc for resizing text

myfunnyandy:
https://tree.taiga.io/project/penpot/task/11665

### DIFF
--- a/docs/user-guide/objects/index.njk
+++ b/docs/user-guide/objects/index.njk
@@ -142,6 +142,11 @@ a design.</p>
     <source src="/img/objects/text-create.mp4" type="video/mp4">
   </video>
 </figure>
+<p><strong>Tips for resizing</strong></p>
+<ul>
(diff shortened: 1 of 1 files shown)


docs/user-guide/objects/index.njk
```diff
@@ -142,6 +142,11 @@ a design.</p>
     <source src="/img/objects/text-create.mp4" type="video/mp4">
   </video>
 </figure>
+<p><strong>Tips for resizing</strong></p>
+<ul>
+  <li>Double-click on the right side of the bounding to set the resize setting to auto-width.</li>
```
madalenapmelo-kp:
```suggestion
  <li>Double-click on the right side of the bounding box to set the resize setting to auto-width.</li>
```
Just to keep it consistent between the 2 list items

madalenapmelo-kp (approved):

PR #6994, opened 2025-07-28
:wrench: Refactor: set page objects once

elenatorro:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/11645

### Summary

Minor follow-up to only check shape-ids  

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/workspace/viewport_wasm.cljs
+++ b/frontend/src/app/main/ui/workspace/viewport_wasm.cljs
@@ -51,6 +51,7 @@
    [app.main.ui.workspace.viewport.viewport-ref :refer [create-viewport-ref]]
    [app.main.ui.workspace.viewport.widgets :as widgets]
    [app.render-wasm.api :as wasm.api]
+   [app.render-wasm.shape :as wasm.shape]
    [app.util.debug :as dbg]
    [app.util.text-editor :as ted]
    [beicon.v2.core :as rx]
@@ -280,7 +281,6 @@
                    (:x first-shape)
                    (:x selected-frame))
 
-
         offset-y (if selecting-first-level-frame?
                    (:y first-shape)
                    (:y selected-frame))
@@ -292,7 +292,6 @@
     ;;       canvas, even though we are not using `page-id` inside the hook.
     ;;       We think moving this out to a handler will make the render code
     ;;       harder to follow through.
-
     (mf/with-effect [page-id]
       (when-let [canvas (mf/ref-val canvas-ref)]
         (->> wasm.api/module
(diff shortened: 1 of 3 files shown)


frontend/src/app/render_wasm/shape.cljs
```diff
@@ -19,6 +19,15 @@
 (declare ^:private impl-assoc)
 (declare ^:private impl-conj)
 (declare ^:private impl-dissoc)
+(defonce ^:private current-page-object-ids (atom #{}))
```
niwinz: This don't need to be an atom,  you can `set!` the var directly

frontend/src/app/render_wasm/shape.cljs
```diff
@@ -19,6 +19,15 @@
...
+
+(defn set-current-page-object-ids!
+  [objects]
+  (reset! current-page-object-ids (set (keys objects))))
+
+(defn shape-in-current-page?
+  [shape-id]
+  (contains? @current-page-object-ids shape-id))
```
niwinz: Remving the atom indirection, we can remove the deref `@` operation

frontend/src/app/render_wasm/shape.cljs
```diff
@@ -208,8 +217,7 @@
 
 (defn set-wasm-multi-attrs!
   [shape properties]
-  ;; Only call WASM API if the shape is in the current page
-  (when (api/shape-in-current-page? (:id shape))
+  (when (shape-in-current-page? (:id shape))
```
niwinz: The shape id should be retrieved with `(dm/get-prop shape :id)`. Is the most efficient way to do it.

frontend/src/app/main/ui/workspace/viewport_wasm.cljs
```diff
@@ -51,6 +51,7 @@
    [app.main.ui.workspace.viewport.viewport-ref :refer [create-viewport-ref]]
    [app.main.ui.workspace.viewport.widgets :as widgets]
    [app.render-wasm.api :as wasm.api]
+   [app.render-wasm.shape :refer [set-current-page-object-ids!]]
```
niwinz: I suggest using `wasm.shape` alias for better clarity from where the function is.

niwinz (changes requested):
-

frontend/src/app/render_wasm/shape.cljs
```diff
@@ -19,6 +19,15 @@
 (declare ^:private impl-assoc)
 (declare ^:private impl-conj)
 (declare ^:private impl-dissoc)
+(defonce ^:private current-page-object-ids #{})
+
+(defn set-current-page-object-ids!
+  [objects]
+  (set! current-page-object-ids (set (keys objects))))
```
niwinz: This operation has more overhead than just having the whole map. So I suggest to put all the objects there and
rename the function to `set-current-page-objects!`

niwinz (changes requested):
I suggest a minimal change for avoid the overhead of creating a set of shape ids all the time.

frontend/src/app/render_wasm/shape.cljs
elenatorro: ok! let's do it that way then, I'll change that back

niwinz (approved):

PR #6998, opened 2025-07-28
:sparkles: Set default text color depending on background color

elenatorro:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/10084

### Summary

Depending on the background color, set the default color for a new text as black or white

Screenshot from 2025-07-29 17-26-04
Screenshot from 2025-07-29 17-25-29
Screenshot from 2025-07-29 17-28-19

### Steps to reproduce 

- Create a text shape
- Set a dark background
- Write on the text box and check the color is set to white to have enough contrast
- Set a light background
- Repeat the process and check the color is set to black
- Change the text fill and check the caret uses it as well
- Check also creating texts on a frame

**Notes**
- For frames, it only checks the contrast for solid fills, doesn't check other fills and doesn't take into account opacity levels
- For text nodes, the caret color is not that "smart" yet and does not use the current node color (this will be a future improvement)

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/workspace/shapes/text/v2_editor.cljs
+++ b/frontend/src/app/main/ui/workspace/shapes/text/v2_editor.cljs
@@ -9,9 +9,11 @@
   (:require
    [app.common.data :as d]
    [app.common.data.macros :as dm]
+   [app.common.files.helpers :as cfh]
    [app.common.geom.shapes :as gsh]
    [app.common.geom.shapes.text :as gst]
    [app.common.math :as mth]
+   [app.common.types.color :as color]
    [app.common.types.text :as txt]
    [app.config :as cf]
    [app.main.data.workspace :as dw]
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/ui/workspace/shapes/text/v2_editor.cljs
```diff
@@ -148,21 +158,32 @@
...
+        editor-ref       (mf/use-ref nil)
 
         ;; This reference is to the container
-        container-ref (mf/use-ref nil)
-        selection-ref (mf/use-ref nil)
+        container-ref    (mf/use-ref nil)
+        selection-ref    (mf/use-ref nil)
+
+        page             (mf/deref refs/workspace-page)
+        background-color (:background page)
```
belen-albeza, line 198: Q: Shouldn't we take the background color from the board/frame (if there's any), rather than the one on the page?

elenatorro: Very good point, I think it makes sense, yes, I'll check it, so it tries to get the frame and if there isn't one, it uses the canvas color

elenatorro: Done! As a note, it only works for now for solid fills, no gradients or images for now, I'll update the description with that info

PR #7013, opened 2025-07-30
:sparkles: Simplify gh-actions workflows

yamila-moreno:
This PR simplifies the options of the workflows:
- always create a complete bundle
- rename bundles to make them ligther


sancfc (approved):
According to what we talked about, I see the changes quite well.

PR #7053, opened 2025-08-01
:sparkles: Compact fill serialization (opacity + flags)

belen-albeza:
### Related Ticket

https://tree.taiga.io/project/penpot/task/11678

### Summary

This PR makes `opacity` take only 1 byte in gradient and image fills. This way, we can store flags (like `keep_aspect_ratio`) in the resulting free space rather than use another 32-bit value.

Screenshot 2025-08-04 at 2 14 15 PM

### Steps to reproduce 

Load a file with shapes with different types of fills and tweak the opacity value in them. Everything should work as normal.

### Checklist

- [x] ~~Add or modify existing integration tests in case of bugs or new features, if applicable~~.

### DIFF
--- a/common/src/app/common/buffer.cljc
+++ b/common/src/app/common/buffer.cljc
@@ -21,6 +21,13 @@
     (let [target (with-meta target {:tag 'java.nio.ByteBuffer})]
       `(long (.get ~target ~offset)))))
 
+(defmacro read-unsigned-byte
+  [target offset]
+  (if (:ns &env)
+    `(.getUint8 ~target ~offset true)
+    (let [target (with-meta target {:tag 'java.nio.ByteBuffer})]
+      `(bit-and (long (.get ~target ~offset)) 0xff))))
+
 (defmacro read-bool
   [target offset]
   (if (:ns &env)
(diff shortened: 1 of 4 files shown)


common/src/app/common/buffer.cljc
```diff
@@ -17,7 +17,7 @@
 (defmacro read-byte
   [target offset]
   (if (:ns &env)
-    `(.getInt8 ~target ~offset true)
+    `(.getUint8 ~target ~offset true)
```
belen-albeza: Note to self: I don't think we'll be able to get away with this as it is, since IIRC the JVM does not work with unsigned types. A workaround could be make this a short (we'd still have 16 bits for flags)

render-wasm/src/wasm/fills/image.rs
```diff
@@ -8,23 +10,24 @@ pub struct RawImageFillData {
     b: u32,
     c: u32,
     d: u32,
-    opacity: f32,
+    opacity: u8,
```
belen-albeza, line 13: Note to self: We'd probably want to make this uniform across all fills.

PR #7065, opened 2025-08-05
:bug: Fix focus new added property

luisddm:
### Related Ticket

Taiga issue [#11718](https://tree.taiga.io/project/penpot/task/11718)

### Summary

When creating a new property, the property name input should be focused and filled with the default name.

### Steps to reproduce

Try adding a new property from the design tab.

### DIFF
--- a/common/src/app/common/logic/variant_properties.cljc
+++ b/common/src/app/common/logic/variant_properties.cljc
@@ -27,7 +27,12 @@
     (reduce (fn [changes component]
               (pcb/update-component
                changes (:id component)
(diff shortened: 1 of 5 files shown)


common/src/app/common/logic/variant_properties.cljc
```diff
@@ -89,15 +94,18 @@
         prop-names         (mapv :name props)
         property-name      (ctv/update-number-in-repeated-item prop-names property-name)
 
+        mdata              (if editing? {:editing? true} nil)
```
niwinz, line 97: Don't use `?` inside data structures if possible.

PR #7115, opened 2025-08-13
:bug: Fix menu entry not showing on assets tab

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot/task/11769

### Summary

Menu entry 'Combine as variants' missing when there is only one component that is a direct child of the group

### Steps to reproduce 
1. Create a component named "a/one"
2. Create a component named "a/b/two"
3. On assets tab, on the contextual menu of the group "a" you should see the entry 'Combine as variants' 

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/assets/components.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/assets/components.cljs
@@ -172,6 +172,20 @@
           [:span {:class (stl/css-case :variant-mark-cell listing-thumbs? :variant-mark true :component-icon true)
                   :title (tr "workspace.assets.components.num-variants" num-variants)} i/variant])])]))
 
+
+(defn- count-leaves
+  "Counts the total number of leaf elements in a nested map structure.
+     A leaf element is considered any element inside a vector."
+  [m]
+  (reduce
+   (fn [acc v]
+     (cond
+       (map? v) (+ acc (count-leaves v))
+       (vector? v) (+ acc (count v))
+       :else acc))
+   0
+   (vals m)))
+
 (mf/defc components-group
   {::mf/wrap-props false}
   [{:keys [file-id prefix groups open-groups force-open? renaming listing-thumbs? selected on-asset-click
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/ui/workspace/sidebar/assets/components.cljs
```diff
@@ -172,6 +172,20 @@
...
+     A leaf element is considered any element inside a vector."
+  [m]
+  (reduce
+   (fn [acc v]
+     (cond
+       (map? v) (+ acc (count-leaves v))
+       (vector? v) (+ acc (count v))
+       :else acc))
+   0
+   (vals m)))
```
niwinz: I suggest a minor change for this function: use reduce-kv instead of reduce + vals.

```clojure
(defn- count-leaves
  "Counts the total number of leaf elements in a nested map structure.
     A leaf element is considered any element inside a vector."
  [m]
  (reduce-kv (fn [acc _ v]
               (cond
                 (map? v) (+ acc (count-leaves v))
                 (vector? v) (+ acc (count v))
                 :else acc))
             0
             m))
```

The problem is not the reduce, the "problem" is vals, knowning how it is implemented in the majority of languages it generates allocation of KeyEntry or similar classs for each entry, for then: extract the value. In contrast, reduce-kv performs an internal reduce over structure without creating that sequence of KeyEntry'es.

frontend/src/app/main/ui/workspace/sidebar/assets/components.cljs
```diff
@@ -193,7 +207,7 @@
 
         components     (not-empty (get groups "" []))
         can-combine?   (and is-local
-                            (> (count components) 1)
+                            (> (count-leaves groups) 1)
```
niwinz: This operation were constant time, now it is O(n). I suggest putting it under `mf/with-memo`

pabloalba: done

frontend/src/app/main/ui/workspace/sidebar/assets/components.cljs
pabloalba: done

PR #7145, opened 2025-08-19
:sparkles: Update code to use design system icon

Xaviju:
### Related Ticket

https://tree.taiga.io/project/penpot/task/11851

### Summary

- Updates the code to use the design system `icon*` component.
- Refactors the icon selector based on the shape attributes
- Removes the `shape-icon` component to a util because it was a component that just returned another component. To properly use the new `icon*` component, I refactored it in a util that given a shape return the expected icon string name.
- Add the `variant` icon to the design system so it can be used within the `icon*` component.

### Steps to reproduce 

- Ensure that the shape info under the `inspect`tab works as expected
- Ensure that the layers tab works as expected (specifically icon visualization)
- Ensure that the `variants` icon are being displayed in layers, workspace and inspect tab.

### Checklist


frontend/src/app/util/shape_icon.cljs
```diff
@@ -0,0 +1,83 @@
...
+(defn get-shape-icon*
+  "Returns the icon for a shape based on its type and properties"
+  [shape main-instance?]
+  (let [flex-layout? (ctl/flex-layout? shape)
+        col? (ctl/col? shape)
+        row? (ctl/row? shape)
+        grid-layout? (ctl/grid-layout? shape)
+        has-images? (cts/has-images? shape)
+        has-mask? (:masked-group shape)]
+    (if (ctk/instance-head? shape)
```
niwinz: This function is executed in a critical path of different parths of workspace, so we need to think on implement it in a more efficient way.

First of all, the type should be obtained with `(dm/get-prop shape :type)`
Also, all predicate execution on the let without knowing if we really need them are not very efficient. I suggest defining them only if we clearly know we need them. Per example `flex-layout?` is not used if shape is instance-head. 
I also can suggest defining a separate function for a shape type if it has enough logic (example the `:frame` shape and bool)

Xaviju: This utility was created translating into a function the removed component `shape-icon` (in case you need to check how was it working). But I did not consider well which areas were affected!

I'll do the requested changes. Please do not hesitate to ping me if you want to review this deeply. I'd like to ensure the next time that I take into consideration every performance aspect when editing a critical path feature.

frontend/src/app/util/shape_icon.cljs
```diff
@@ -0,0 +1,83 @@
...
+
+(ns app.util.shape-icon
+  (:require
+   [app.common.types.component :as ctk]
+   [app.common.types.shape :as cts]
+   [app.common.types.shape.layout :as ctl]))
+
+(defn get-shape-icon*
+  "Returns the icon for a shape based on its type and properties"
+  [shape main-instance?]
```
niwinz: The `main-instance?` can already be obtained from shape, so having it in an additional parameter looks strange

frontend/src/app/util/shape_icon.cljs
```diff
@@ -0,0 +1,83 @@
...
+;;
+;; Copyright (c) KALEIDOS INC
+
+(ns app.util.shape-icon
+  (:require
+   [app.common.types.component :as ctk]
+   [app.common.types.shape :as cts]
+   [app.common.types.shape.layout :as ctl]))
+
+(defn get-shape-icon*
```
niwinz: The `*` on the name usually indicates "private" or "low-level" version of a function. It is also used on component names. I think this case does not match that, so I suggest removing `*` from the function name.

Also, we can consider adding a lightweight cache, but i will done it myself once all suggested changes are addressed.

frontend/src/app/util/shape_icon.cljs
Xaviju: You are right, seems strange that it was being obtained outside. I moved all the logic to the function

niwinz (approved):

PR #7151, opened 2025-08-20
:wrench: Refactor internal api of tokens lib

hirunatan:
### Related Ticket

https://tree.taiga.io/project/penpot/task/11449

### Summary

Change internal api of tokens library to use ids to access objects, in order to facilitate the data structure refactor later.

### Steps to reproduce 

No behavior changes should be noticed by users, everything should work equal.

### Checklist


common/src/app/common/types/tokens_lib.cljc
```diff
@@ -527,6 +557,8 @@
     Prefixed set path or ppath:        a path wit added prefixes [\"G-some-group\", \"G-some-subgroup\"].
     Prefixed set full path or pfpath:  a full path wit prefixes [\"G-some-group\", \"G-some-subgroup\", \"S-some-set\"].
     Prefixed set final name or pfname: a final name with prefix \"S-some-set\"."
+  (set-by-id [_ id] "get a set by its id")
+  (set-by-name [_ name] "get a set by its name")
```
niwinz: I think this should be `get-set` and `get-set-by-name`

hirunatan: Ok, i like it :+1:

hirunatan: changed

common/src/app/common/types/tokens_lib.cljc
```diff
@@ -286,7 +286,7 @@
                tokens))
 
   ITokenSet
-  (add-token [_ token]
+  (add-token- [_ token]
```
niwinz, line 235: the `-` is usually prefixed and not suffixed as a convention

hirunatan: This is temporary, those methods will be deleted when tokens are not inside sets.

common/src/app/common/files/changes.cljc
```diff
@@ -366,9 +366,33 @@
...
+    [:map {:title "SetTokensLib"}
+     [:type [:= :set-tokens-lib]]
+     [:tokens-lib ::sm/any]]]  ;; TODO: we should define a plain object schema for tokens-lib
+
+   [:set-token
+    [:map {:title "SetTokenChange"}
+     [:type [:= :set-token]]
+     [:set-id ::sm/uuid]
+     [:token-id ::sm/uuid]
+     [:token [:maybe ctob/schema:token-attrs]]]]
```
niwinz: My soft suggestion here is use a more generic `:attrs` instead of `:token`. Because we really sent attrs and then create an instance of token after validating that attrs.

niwinz: The same applies to the rest of changes.

common/src/app/common/types/tokens_lib.cljc
```diff
@@ -1301,22 +1292,32 @@ Will return a value that matches this schema:
...
 
   (get-all-tokens [this]
     (reduce
      (fn [tokens' set]
-       (into tokens' (map (fn [x] [(:name x) x]) (get-tokens set))))
+       (into tokens' (map (fn [x] [(:name x) x]) (get-tokens-seq- set))))
      {}
      (get-sets this)))
 
+  (get-tokens-seq [this set-id]
```
niwinz: This method is redundant because we can already call `(vals (get-tokens-map ...))` 
I suggest just renaming it to `get-tokens` always returning the map of tokens (the most efficient operation), if user need seq of it, it can call seq on it.

common/src/app/common/files/changes.cljc
```diff
@@ -366,9 +366,33 @@
      [:type [:= :del-typography]]
      [:id ::sm/uuid]]]
 
-   [:update-active-token-themes
-    [:map {:title "UpdateActiveTokenThemes"}
-     [:type [:= :update-active-token-themes]]
+   [:set-tokens-lib
+    [:map {:title "SetTokensLib"}
+     [:type [:= :set-tokens-lib]]
+     [:tokens-lib ::sm/any]]]  ;; TODO: we should define a plain object schema for tokens-lib
```
niwinz, line 372: This is the only change definition that is not clear to me. I mean, this change is too big, and any use of it will overwrite any more granular action on tokens that other concurrent user is doing.

I think we should not have this method at all... is this used for anything relevant?

common/src/app/common/types/tokens_lib.cljc
hirunatan: Ok. It seems reasonable.

common/src/app/common/files/changes.cljc
hirunatan: Good point. This way we make crystal clear that this is a plain map and not an entity instance.

The changes builder functions may still receive the instance (since it's more useful this way) and use `datafy` to convert it to a plain map. In `Token` and `TokenTheme` the datafy is not really necessary, since they are `defrecords`, but I suppose it will not add overhead and this is more expresive, do you agree?

niwinz: yep, and make it consistent using datafy in each case

common/src/app/common/files/changes.cljc
hirunatan: This is already being used in the file import, where the whole library is replaced by a new one read from a json file. Anyway this is going to be revised in depth during the next work with the synchronization mechanisms, so I suppose we can leave it as is for now.

But don't add an entry point in the plugins api for this for now.

PR #7159, opened 2025-08-20
:recycle: Add a better approach for load libraries on file validation and migrations 

niwinz:
### Summary

The idea behind this proposal is:

Instead of loading and migrating all libraries that a file need for perform migration (that requires access to the libraries) we preload only the IDs of the libraries and instantiate a LoadableWeakValueMap. The LoadableWeakValueMap is a custom implementation of WeakMap type that is an subset of IPersistentMap that loads the library on demand (when is really accessed) and stores it using SOFT references (that means that all other libraries loaded previously, can be removed by the Garbage Collector in memory pressure situation). This approach facilitates to have in the worst case only one library at once loaded (the used one, where strong references are hold).

It also removes the process of applying the migration to the loaded libraries.


common/src/app/common/weak_map.cljc
```diff
@@ -0,0 +1,78 @@
...
+#?(:cljs
+   (deftype WeakValueMap [^js/Map data ^js/FinalizationRegistry registry]
+     Object
+     (clear [_]
+       (.clear data))
+     (delete [_ key]
+       (.delete data key))
+     (get [_ key]
+       (if-let [ref (.get data key)]
+         (.deref ^WeakRef ref)
```
pabloalba: If the value was part of the preload data, it will be a 'final" value, not a WeakRef

hirunatan (approved):

PR #7167, opened 2025-08-21
:tada: Add Text Editor WASM Playground

AzazelN28:
### Summary

Adds a text editor playground with the new renderer integrated, to run it:

```sh
cd frontend/text-editor
yarn dev
```

And then access the URL `http://localhost:5173/wasm.html`

To update the `render_wasm.wasm`  and   `render_wasm.js` used by the playground use `yarn wasm:update` from the same folder (`frontend/text-editor`), that script will copy the compiled files in your resources folder. 

### Checklist

### DIFF
--- a/frontend/src/app/render_wasm/api/texts.cljs
+++ b/frontend/src/app/render_wasm/api/texts.cljs
@@ -16,7 +16,7 @@
    [app.render-wasm.wasm :as wasm]))
 
 (def ^:const PARAGRAPH-ATTR-U8-SIZE 44)
-(def ^:const LEAF-ATTR-U8-SIZE 56)
+(def ^:const LEAF-ATTR-U8-SIZE 60)
 
 (defn- encode-text
   "Into an UTF8 buffer. Returns an ArrayBuffer instance"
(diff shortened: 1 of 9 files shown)


frontend/src/app/render_wasm/api/texts.cljs
```diff
@@ -104,15 +105,21 @@
                   text-transform
                   (or (sr/translate-text-transform (:text-transform leaf))
                       (sr/translate-text-transform (:text-transform paragraph))
-                      (sr/translate-text-transform "none"))]
+                      (sr/translate-text-transform "none"))
+
+                  text-direction
+                  (or (sr/translate-text-direction (:text-direction leaf))
```
elenatorro, line 111: sr/translate-text-direction needs to be updated to handle `nil` parameter, by returning `0` by default in that case:

```cljs
(defn translate-text-direction
  [text-direction]
  (case text-direction
    "ltr" 0
    "rtl" 1
    0))
 ```

elenatorro (changes requested):

elenatorro (approved):

PR #7171, opened 2025-08-21
:sparkles: Add typography composite token

floscr:
### Related Ticket

- https://github.com/tokens-studio/penpot/issues/104
- https://tree.taiga.io/project/penpot/us/10200

Some tasks of the US got split to

- https://github.com/tokens-studio/penpot/issues/136
- https://github.com/tokens-studio/penpot/issues/140
- https://github.com/tokens-studio/penpot/issues/135

### Summary

https://github.com/user-attachments/assets/98f724b3-a28e-4dd8-a920-4215f719d92c

Implements the typography composite token.
The typography token is a mments the typography composite token.
The typography toap of token values with the matching `token-type` as the key.
Undefined values are not set in the map.

The data for a token will look like this:

```clj
{:font-size "24"
 :font-weight "500"}
```

- Composite tokens allow the atomic values to be empty
- Composite tokens can reference atomic tokens
- When applying a composite token, atomic tokens will be removed and vice-versa
- Changing any typographic value via the design tab will unapply typography tokens

### Checklist


frontend/src/app/main/data/workspace/tokens/propagation.cljs
```diff
@@ -161,7 +162,10 @@
                   (collect-shapes-update-info resolved-tokens (:objects page))
 
                   actions
-                  (actionize-shapes-update-info page-id attrs)]
+                  (actionize-shapes-update-info page-id attrs)
+
+                  ;; Composed updates return observables and need to be executed differently
+                  {:keys [observable normal]} (group-by #(if (rx/observable? %) :observable :normal) actions)]
```
floscr, line 168: Since we're returning now multiple calls in one event call, `rx/from` didn't work anymore.
Now I'm splitting up the stream here as a workaround, but not sure if there is a better option?

frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
```diff
@@ -101,53 +102,102 @@
...
+      (assoc :validators [(fn [token]
+                            (let [token-name (:name token)]
+                              (some (fn [[k v]]
+                                      (when-let [err (case k
+                                                       :font-family (invalidate-coll-self-reference token-name v)
+                                                       (invalidate-self-reference token-name v))]
+                                        (assoc err :typography-key k)))
+                                    (:value token))))])
+
+      (validate-default-token)))
```
floscr: The individual forms can now pass in custom validation functions, since composite and atomic tokens work differently, e.g.: composite tokens allow empty fields.

Since errors are collected per token level, we have to assign a key of the token-type `typography-key` for the composite tokens, so we can pick the individual field errors.

frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
```diff
@@ -805,11 +833,102 @@
...
+                  (dom/set-value! el next-value)
+                  (on-update-value #js {:target el
+                                        :tokenType :font-family}))))
+
+             on-change
+             (mf/use-fn
+              ;; Passing token-type via event to prevent deep function adapting & passing of type
+              (fn [e]
+                (-> (obj/set! e "tokenType" k)
+                    (on-update-value))))]
```
floscr, line 929: Since `on-get-token-value` to retrieve the value from the input and calculate the next value, has to be defined on the `form*` parent, we have to pass the token-type of the input field via the js event object / a map that has properties we need.

frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
```diff
@@ -805,11 +833,102 @@
...
+(mf/defc typography-form*
+  [{:keys [token] :rest props}]
+  (let [on-get-token-value
+        (mf/use-callback
+         (fn [e prev-value]
+           (let [token-type (obj/get e "tokenType")
+                 input-value (dom/get-target-val e)]
+             (if (empty? input-value)
+               (dissoc prev-value token-type)
+               (assoc prev-value token-type input-value)))))]
```
floscr, line 962: Here we access the assigned tokenType property, so we can update the values map

frontend/src/app/main/data/style_dictionary.cljs
```diff
@@ -218,6 +225,55 @@
...
+    :font-family (parse-sd-token-font-family-value token-value)
+    :font-weight (parse-sd-token-font-weight-value token-value)
+    :letter-spacing (parse-sd-token-letter-spacing-value token-value)
+    :text-case (parse-sd-token-text-case-value token-value)
+    :text-decoration (parse-sd-token-text-decoration-value token-value)
+    nil))
+
+(defn- parse-composite-typography-value
+  [value]
+  (let [missing-references
```
hirunatan, line 251: This will never find any missing reference, won't it? Because in this type the value is not a string, but a plain object.

floscr: Yeah that check gets applied later, I was already preparing here for when the typography token can also be a string containing references, but I chose to split out this feature and forgot to remove it here.
Should I remove it for now or can we keep it in?

frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
hirunatan: I find this collection of callbacks and validators quite difficult to understand. I wonder if it would be possible to add some documentation of it. Perhaps some diagram of the validation flow, or the descriptions of the expected properties of each callback, or the like.

frontend/src/app/main/data/workspace/tokens/propagation.cljs
hirunatan: Don't like it very much, but I cannot think of a better solution, so go ahead.

frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
floscr: I've added some docs strings and cleaned up a bit, hope that helps.
I had sketched out this diagram when working on the feature, I could document the whole flow somehow, but I'm a bit doubtful as this would have to be kept up to date always to be useful
image

frontend/src/app/main/data/style_dictionary.cljs
hirunatan: Ok. I think we can leave it, because the reference value is the next thing to implement.

frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
hirunatan: Yes, I think those docstrings help a lot. Every time we have functions whose purpose or data structures are not obvious we should add this kind of documentation.

For the diagram, we have a team in design.penpot.dev to create diagrams of specially complex modules (like component synchronization). But I think this one is not so complex, so docstings should be enough.

PR #7172, opened 2025-08-21
:bug: Fix duplicating a page with variants should duplicate them

pabloalba:
### Related Ticket

:bug: Fix duplicating a page with variants should duplicate the

### Summary

Duplicating a page with variants crash.

It should work, and create a new page with a duplicate of the variants

### Steps to reproduce 

1. Create a variant
2. Duplicate the page

### Checklist

### DIFF
--- a/frontend/src/app/main/data/workspace/pages.cljs
+++ b/frontend/src/app/main/data/workspace/pages.cljs
@@ -269,14 +269,14 @@
 
             ;; Remap parents and childs with the new ids generated for variants
             objects
-            (reduce
-             (fn [objs [_ shape]]
+            (reduce-kv
+             (fn [objs _ {:keys [id parent-id frame-id] :as shape}]
                (assoc objs
-                      (get variants-ids-map (:id shape) (:id shape))
+                      (get variants-ids-map id id)
                       (-> shape
-                          (assoc :id (get variants-ids-map (:id shape) (:id shape)))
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/data/workspace/pages.cljs
```diff
@@ -201,25 +202,100 @@
...
+             (fn [objs [_ shape]]
+               (assoc objs
+                      (get variants-ids-map (:id shape) (:id shape))
+                      (-> shape
+                          (assoc :id (get variants-ids-map (:id shape) (:id shape)))
+                          (assoc :parent-id (get variants-ids-map (:parent-id shape) (:parent-id shape)))
+                          (assoc :frame-id (get variants-ids-map (:frame-id shape) (:frame-id shape)))
+                          (cond->
+                           (contains? shape :shapes)
+                            (assoc :shapes (mapv #(get variants-ids-map % %) (:shapes shape)))))))
```
niwinz, line 282: Several enhancements to this code:

- Use reduce-kv instead of reduce, ignoring the id in the same way but you don't need to use destructuring.
- Destructure the `:id` `:frame-id` and `:parent-id` so you don't need lookup these values on shape several times per iteration.

pabloalba: done

PR #7174, opened 2025-08-21
:sparkles: Make some improvements to the swap panel

luisddm:
### Related Ticket

Taiga issue [#11850](https://tree.taiga.io/project/penpot/task/11850)

### Summary

- Make some visual adjustments in the swap panel to match the design.
- Refactor and simplify the `search-bar` component using the new icons from the DS. The component appears in this panel and this refactor makes the visual adjustments easier to do.

### Steps to reproduce 

- Select the copy of a variant, go to the design tab and click on the first button in the "component" section.
- Check that the swap section looks like in the design, and behaves correctly (we should be able to filter components from the local library and choose one to swap).
- Check that the `search-bar` component work as before (filtering the results, displaying an icon at the left when specified, and the delete icon at the right when some text is introduced).

### DIFF
--- a/frontend/src/app/main/ui/components/search_bar.cljs
+++ b/frontend/src/app/main/ui/components/search_bar.cljs
@@ -8,25 +8,14 @@
   (:require-macros [app.main.style :as stl])
   (:require
    [app.common.data.macros :as dm]
-   [app.main.ui.icons :as i]
+   [app.main.ui.ds.foundations.assets.icon :refer [icon*]]
    [app.util.dom :as dom]
    [app.util.keyboard :as kbd]
    [rumext.v2 :as mf]))
 
-(mf/defc search-bar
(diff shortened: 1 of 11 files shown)


frontend/src/app/main/ui/components/search_bar.cljs
```diff
@@ -52,15 +41,20 @@
     [:span {:class (stl/css-case :search-box true
                                  :has-children (some? children))}
      children
-     [:div {:class (dm/str input-class " " (stl/css :search-input-wrapper))}
-      icon
+     [:div {:class (dm/str class " " (stl/css :search-input-wrapper))}
```
niwinz: Minor enhancement: 

The class can be defined as `:class  [class (stl/css :search-input-wrapper)]` instead of explicit concat

PR #7182, opened 2025-08-25
:recycle: Add efficiency refactor for sidebar

niwinz:
### Summary

This PR is a first step on optimizing the efficiency of the right sidebar. On this first step, the relevant changes are:
- Replace the memoize on `multiple/options*` component and replace it with WeakMap based cache; this **substantially reduces** the memory accumulation of penpot while it is being used (mainly selecting multiple shapes at once). The same method was used on `group/options*` so it also benefits of this enhacement. </br>
Before this change, the memory is accumulated for each combination of selected shapes and never frees it, causing substantial memory bloat over time of use.

- Add several and missing `mf/with-memo` and `mf/use-fn` on several `*/options*` components, reducing the overhead of re-rendering of children components (this is work in progress, I mean several changes made for normalize it on the main components but we still have too many issues on the secondary components)

- Move the active tokens resolution mechanism one level up, for make it available for both sidebars, not only the left-sidebar. This is not strictly an efficiency enhancement but it is required for the upcoming changes related to tokens for the options menus on the right sidebar that will require also to have access on resolved tokens.

### How to test

This need exploratory tests on right sidebar on workspace.


AzazelN28 (approved):

frontend/src/app/main/ui/ds/controls/numeric_input.cljs
```diff
@@ -65,14 +65,17 @@
 
 (defn- get-option-by-name
   [options name]
-  (d/seek #(= name (get % :name)) options))
+  (let [options (if (delay? options) (deref options) options)]
```
AzazelN28, line 68: I think this should be `(if (delay? options) @options options)`, isn`t it?

PR #7209, opened 2025-08-28
:tada: Inspect styles tab tokens panel

Xaviju:
### Related Ticket

https://tree.taiga.io/project/penpot/us/9313

### Summary

This PR introduces the first styles tab panel - tokens panel.
On selecting a shape, the user should be able to see if there are any active token themes or sets.

### Steps to reproduce 

- Select a shape with sets and themes actives
- Open styles tab
- Ensure that are displayed under the `Tokens Sets and Themes` panel

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/inspect/styles.cljs
+++ b/frontend/src/app/main/ui/inspect/styles.cljs
@@ -5,6 +5,9 @@
    [app.common.data.macros :as dm]
    [app.common.types.component :as ctc]
    [app.common.types.components-list :as ctkl]
+   [app.common.types.tokens-lib :as ctob]
+   [app.main.refs :as refs]
+   [app.main.ui.inspect.styles.panels.tokens-panel :refer [tokens-panel*]]
    [app.main.ui.inspect.styles.style-box :refer [style-box*]]
    [app.util.i18n :refer [tr]]
    [rumext.v2 :as mf]))
@@ -36,13 +39,19 @@
         first-shape        (first shapes)
         first-component    (mf/with-memo (ctkl/get-component data (:component-id first-shape)))
         type               (mf/with-memo (get-shape-type shapes first-shape first-component))
-        has-tokens?        (:applied-tokens first-shape)
+
+        ;; Must be reviewed for performance and code clarity
+        tokens-lib         (mf/deref refs/tokens-lib)
+        active-themes      (mf/deref refs/workspace-active-theme-paths-no-hidden)
+        active-sets
+        (mf/with-memo [tokens-lib]
+          (some-> tokens-lib (ctob/get-active-themes-set-names)))
         options            (type->options type)]
-    [:ol {:class (stl/css :styles-tab) :aria-label (tr "inspect.tabs.styles")}
-     (when has-tokens?
-       [:li {:key "token"}
+    [:ol {:class (stl/css :styles-tab) :aria-label (tr "labels.styles")}
+     (when (or active-themes active-sets)
+       [:li
         [:> style-box* {:attribute :token}
-         [:p "Tokens Panel (WIP)"]]])
+         [:> tokens-panel* {:themes active-themes :sets active-sets}]]])
      (for [option options]
        [:li {:key (d/name option)}
         [:> style-box* {:attribute option} color-space]])]))
--- a/frontend/src/app/main/ui/inspect/styles/panels/tokens_panel.cljs
+++ b/frontend/src/app/main/ui/inspect/styles/panels/tokens_panel.cljs
@@ -1,0 +1,39 @@
+(ns app.main.ui.inspect.styles.panels.tokens-panel
+  (:require-macros [app.main.style :as stl])
+  (:require
+   [app.util.i18n :refer [tr]]
+   [app.util.webapi :as wapi]
+   [cuerdas.core :as str]
+   [rumext.v2 :as mf]))
+
+(mf/defc tokens-row*
+  [{:keys [term detail copiable]}]
+  (let [copiable? (or copiable false)
+        detail? (not (or (nil? detail) (str/blank? detail)))
+        detail (if detail? detail "-")
+        copy-attr
+        (mf/use-fn
+         (fn []
+           (wapi/write-to-clipboard (str term ": " detail))))]
+    [:dl {:class (stl/css :attribute-row)}
+     [:dt {:class (stl/css :attribute-term)} term]
+     [:dd {:class (stl/css :attribute-detail)}
+      (if (and copiable? detail?)
+        [:button {:class (stl/css :attribute-detail-copiable)
+                  :on-click copy-attr} detail]
+        detail)]]))
+
+(mf/defc tokens-panel*
+  [{:keys [theme-paths set-names]}]
+  [:div {:class (stl/css :tokens-panel)}
+   (when (seq theme-paths)
+     (for [theme theme-paths]
+       [:> tokens-row* {:key theme
+                        :class (stl/css :token-theme)
+                        :term (tr "inspect.tabs.styles.panel.tokens.active-themes")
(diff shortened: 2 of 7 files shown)


frontend/translations/en.po
```diff
@@ -1850,13 +1850,9 @@ msgid "inspect.tabs.info"
 msgstr "Info"
 
 #: src/app/main/ui/inspect/right_sidebar.cljs:107
-msgid "inspect.tabs.styles"
+msgid "labels.styles"
 msgstr "Styles"
 
-#: src/app/main/ui/inspect/right_sidebar.cljs:165
-msgid "inspect.tabs.switcher.label"
-msgstr "Layer info"
```
hirunatan, line 1858: It seems that this is still used in the right sidebar.

frontend/src/app/main/ui/inspect/styles.cljs
```diff
@@ -36,13 +39,19 @@
         first-shape        (first shapes)
         first-component    (mf/with-memo (ctkl/get-component data (:component-id first-shape)))
         type               (mf/with-memo (get-shape-type shapes first-shape first-component))
-        has-tokens?        (:applied-tokens first-shape)
+
+        ;; Must be reviewed for performance and code clarity
```
hirunatan, line 43: Looks good to me. This may be trigger a repaint if the tokens lib is changed for some reason different to changing active themes or sets. But this is not likely to occur because we are in read only mode. It may occur if there are concurrent users modifying the file, but in any case it will be only in response to user actions. No loops that affect performance are expected.

frontend/src/app/main/ui/inspect/styles/panels/tokens_panel.cljs
```diff
@@ -0,0 +1,39 @@
...
+    [:dl {:class (stl/css :attribute-row)}
+     [:dt {:class (stl/css :attribute-term)} term]
+     [:dd {:class (stl/css :attribute-detail)}
+      (if (and copiable? detail?)
+        [:button {:class (stl/css :attribute-detail-copiable)
+                  :on-click copy-attr} detail]
+        detail)]]))
+
+(mf/defc tokens-panel*
+  [{:keys [themes sets]}]
```
hirunatan: I would call those `theme-paths` and `set-names`. This way it's clear that they are lists of strings and not of TokenTheme and TokenSet objects.

frontend/src/app/main/ui/inspect/styles/panels/tokens_panel.cljs
```diff
@@ -0,0 +1,39 @@
...
+
+(mf/defc tokens-row*
+  [{:keys [term detail copiable]}]
+  (let [copiable? (or copiable false)
+        detail? (not (or (nil? detail) (str/blank? detail)))
+        detail (if detail? detail "-")
+        copy-attr
+        (mf/use-fn
+         (fn []
+           (js/navigator.clipboard.writeText (str term ": " detail))))]
```
hirunatan: We try to never access directly navigator or dom objects. For this case we have a `util.webapi` module with a function `write-to-clipboard`.

But also we have a `main.ui.components.copy-button` component that shows a clipboard icon on hover and copy the text to clipboard when pressed. It's being used in the other sections of the inspect tab.

frontend/src/app/main/ui/inspect/styles/panels/tokens_panel.cljs
```diff
@@ -0,0 +1,39 @@
...
+        detail)]]))
+
+(mf/defc tokens-panel*
+  [{:keys [themes sets]}]
+  [:div {:class (stl/css :tokens-panel)}
+   (when (seq themes)
+     (for [theme themes]
+       [:> tokens-row* {:key theme
+                        :class (stl/css :token-theme)
+                        :term (tr "inspect.tabs.styles.panel.tokens.active-themes")
```
hirunatan, line 33: Will this be copiable?

frontend/src/app/main/ui/inspect/styles.cljs
```diff
@@ -36,13 +39,19 @@
...
+        tokens-lib         (mf/deref refs/tokens-lib)
+        active-themes      (mf/deref refs/workspace-active-theme-paths-no-hidden)
+        active-sets
+        (mf/with-memo [tokens-lib]
+          (some-> tokens-lib (ctob/get-active-themes-set-names)))
         options            (type->options type)]
-    [:ol {:class (stl/css :styles-tab) :aria-label (tr "inspect.tabs.styles")}
-     (when has-tokens?
-       [:li {:key "token"}
+    [:ol {:class (stl/css :styles-tab) :aria-label (tr "labels.styles")}
```
hirunatan, line 50: I think the `styles-tab` class should have a right padding. Now all content extends all into the right margin.

frontend/translations/en.po
Xaviju: Sure, I've fixed this in my next PR by mistake but since its behind a flag its not a problem https://github.com/penpot/penpot/pull/7246/files#diff-c3401ad57d984ebc437dd95099a0306a51dcfa736bfca5fa1bcf7a960ca052afR1856

frontend/src/app/main/ui/inspect/styles/panels/tokens_panel.cljs
Xaviju: Fixed. I'll be using the `wapi` for now since the design is completely new and the components are using the design system.

frontend/src/app/main/ui/inspect/styles/panels/tokens_panel.cljs
Xaviju: By default, `copiable` is an optional property. In case is nil, the tokens-row* sets it as `false`.
Since in my next PR I'm extracting this component to a generic one, I'll add a scheme to ensure that the user of this component understands how it works.

```
(mf/defc tokens-row*
  [{:keys [term detail copiable]}]
  (let [copiable? (or copiable false)
;; ...
   ```

frontend/src/app/main/ui/inspect/styles.cljs
Xaviju: Currently, the design tab grows to the limit of the screen and there is no padding. Although I agree with your suggestion, seems to be a design decision.

frontend/translations/en.po
hirunatan: ok

frontend/src/app/main/ui/inspect/styles/panels/tokens_panel.cljs
hirunatan: Ok for now. But wouldn't it be good to create the `copy-button` component in the DS, to use it in all panels?

frontend/src/app/main/ui/inspect/styles/panels/tokens_panel.cljs
hirunatan: I don't think its necessary, it's quite redundant.

I find that its easily understandable. My question is that you have created a component with a `copiable` property, and you are using it twice, in one of them you pass it to `false` and in the other you don't pass it (that takes the default false). So why is the property needed if it's not used?

Xaviju: Oops I see! I did not understand the comment. Yeah, seems that is something that must be removed when false since it is not necessary anymore

frontend/src/app/main/ui/inspect/styles.cljs
hirunatan: Are you sure of that the design is this way? In the design doc I see padding to the right, but in the implementation those contents extend to the right margin.

imagen

hirunatan (approved):

PR #7213, opened 2025-08-28
:bug: Translate the tokens tab in the sidebar

tintou:
### Summary

Allow all languages to translate this word accordingly

### Checklist

- [X] Add a detailed explanation of how to reproduce the issue and/or verify the fix, if applicable.
- [X] Include screenshots or videos, if applicable.
- [X] Add or modify existing integration tests in case of bugs or new features, if applicable.
- [X] Update the `CHANGES.md` file, referencing the related GitHub issue, if applicable.

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar.cljs
@@ -138,7 +138,7 @@
                 :id "layers"}
                {:label (tr "workspace.toolbar.assets")
                 :id "assets"}
-               {:label "Tokens"
+               {:label (tr "workspace.toolbar.tokens")
                 :id "tokens"}]
               [{:label (tr "workspace.sidebar.layers")
                 :id "layers"}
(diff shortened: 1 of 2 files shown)


EvaMarco:
Hi @tintou , thanks a lot for your contribution 🙌
In this case, this string is not meant to be translated due to a design decision, so we’ll have to close this PR. We’ve added a comment in the code to clarify this and prevent the same issue in the future. Really appreciate your effort and willingness to help — looking forward to future contributions!

tintou:
Can you be more specific? Are all languages supposed to use the word "Token"?

NatachaMenjibar:
Hi, @tintou,
`Tokens` (as in Design Tokens) is the official name of this feature in Penpot, as well as the widely used technical term. 

Translating ` Tokens` could cause misunderstandings, since many languages don’t have a precise equivalent for the concept of Token. In Penpot, some terms, such as `Sets`, `Layout`, `Grid`, or the CSS properties, are not translated for similar reasons.

Because this feature is highly technical and not all users are familiar with it, we prefer to keep the original term. A poor translation could negatively impact the user experience for many people.

That said, this decision is not permanent. In the future, we may reconsider - especially to allow for transliterations of those terms.

Thanks for your feedback!

tintou:
Alright, I had crafted "Unité de style" in french for the "Token" word, I believe that it is widely used in English because Figma has no translation in its interface :wink: I've added it as voluntarily untranslated in the Weblate glossary.

PR #6542, opened 2025-05-22
:bug: Fix problem with sidebar layout

Alotor:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/11073

### Summary
The layout of the elements on the right sidebar is not aligned. This MR solves the visual problem.

#### Before
image

#### After
image 

### Checklist

- [x] Choose the correct target branch; use `develop` by default.
- [x] Provide a brief summary of the changes introduced.
- [x] Add a detailed explanation of how to reproduce the issue and/or verify the fix, if applicable.
- [x] Include screenshots or videos, if applicable.
- [x] Add or modify existing integration tests in case of bugs or new features, if applicable.
- [x] Check CI passes successfully.
- [x] Update the `CHANGES.md` file, referencing the related GitHub issue, if applicable.

PR #7226, opened 2025-08-29
:lipstick: Slightly reduce the width of the right sidebar

luisddm:
### Related Ticket

No ticket.

### Summary

We have lots of buttons in the right sidebar. There are even rows made only of single icon buttons. These icons have a fixed width (32px) and the space between them is also fixed (4px). Since we have a padding of 12px on both ends, the total width necessary to fit a row of single icon buttons is:

`12 (padding-left) +  32*8 (column) + 7*4 (space) + 12 (padding-right) = 308px = 19.25rem`

So, if we want to keep them aligned vertically with other rows that contain items of various widths, this is the width that the sidebar needs to have. Setting a fixed width of 308px allows us to use flex, push some items to the end of the space, and they will still keep aligned on the right with the rest. And we can also do other simple things like assigning a width of 100% to items which need to occupy an entire row, keeping the correct alignment.

Otherwise we would have to use a fixed width for every component, or use always a grid layout of 8 columns. This is mostly adequate for a matrix of icon buttons, but for other cases  it's clearly overkill (i. e. a single item that needs to spread all along the 8 columns). Of course we can still do this if we want to specify the width of an item in terms of a number of columns, and it will work nicely.

This is why we have reduced the width from 318px to **308px**.

### Steps to reproduce 

Check that all the controls of the right sidebar fit properly and there are no misalignments on the right margin.

### DIFF
--- a/frontend/src/app/main/constants.cljs
+++ b/frontend/src/app/main/constants.cljs
@@ -15,7 +15,7 @@
 (def grid-x-axis 10)
 (def grid-y-axis 10)
 
-(def sidebar-default-width 318)
+(def sidebar-default-width 308)
 (def sidebar-default-max-width 768)
 
 (def page-metadata


luisddm:
There is an additional problem here. This PR would fix the problem only in Firefox, because in Chrome they are already correctly aligned. Actually, we are breaking that with this commit. It seems it has to be with the width of the scroll bar.

A previous PR (#6542) has tried to solve this problem.

Xaviju:
We encountered a similar issue caused by inconsistent behaviour of different browsers regarding the CSS property `scrollbar-gutter`. Reducing the size of the sidebar might cause issues in other areas, such as the new Inspect tab, where there is already limited space.

I suggest investigating how to resolve the inconsistency between browsers caused by this CSS property.

luisddm:
I have been experimenting with `scrollbar-gutter` but I haven't been able to solve the issue.

- Firefox seems to ignore it outright, as the scroll bar is extremely thin and takes up virtually no space. Therefore, there's no space to reserve.

- In Chrome, the scroll bar is always too wide, even with `scrollbar-width: thin` and it actually takes up that space. The only way to avoid that is making it disappear, but then it won't appear in any circumstances, not even when needed.

So I couldn't find a way to harmonize the behavior between the two of them.

PR #7240, opened 2025-09-01
:bug: Fix icons do not appear in swap panel and annotations

luisddm:
### Related Ticket

Taiga issue [#11972](https://tree.taiga.io/project/penpot/issue/11972)

### Summary

Icons are not being shown in the swap panel nor in the annotations section. This happens because we previously changed the namespace from which we take the icons in order to use the DS one, but forgot that there are still places using the old icons, like the radio-buttons component or the annotation section.

So, we are keeping both icon namespaces until we can refactor these components. The new one has been renamed to `ic`, while the old one is sill `i`. This is how it has been done in other files. 

### Steps to reproduce 

Check that the icons appear properly both in the annotation section and the swap panel.

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
@@ -35,10 +35,10 @@
    [app.main.ui.ds.buttons.icon-button :refer [icon-button*]]
    [app.main.ui.ds.controls.combobox :refer [combobox*]]
    [app.main.ui.ds.controls.select :refer [select*]]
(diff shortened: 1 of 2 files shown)


EvaMarco:
While we were discussing this, it was suggested that instead of calling them `i`, we could call the old icons `deprecated-icon`, a very long but very obvious name. That way, in the next refactoring, we can remove those old icons.

luisddm:
I was taking a look at how this is done in other places, but there is no a clear blueprint. So yes, let's try this. Renamed the namespaces as suggested.

EvaMarco (approved):

PR #7258, opened 2025-09-03
:books: Add doc for variants

myfunnyandy:
### Related Ticket
https://tree.taiga.io/project/penpot/us/11813?milestone=472520

### Summary
Adding doc for component variants implied reworking the whole components section in terms of categories for organization.


docs/user-guide/components/index.njk
```diff
@@ -61,11 +62,36 @@ desc: Streamline your design workflow with Penpot's Components guide! Learn to c
   <figcaption>Mains and copies have different icons. Mains also have a title header at the viewport.</figcaption>
 </figure>
 
-<h2 id="component-group">Group components</h2>
-<h3>Create component groups</h3>
+<h3 id="component-find">Find components</h3>
+<p>Where's my component? There are ways to find some components at the assets panel and at the design viewport.</p>
```
madalenapmelo-kp: > There are ways to find **some** components at the assets panel and at the design viewport.

Why "some"? Is it not possible to find all? If it's only main components, I would say that instead of "some"

docs/user-guide/components/index.njk
```diff
@@ -153,21 +173,166 @@ desc: Streamline your design workflow with Penpot's Components guide! Learn to c
...
+<h5>Edit properties</h5>
+<ul>
+  <li><strong>From the Design tab:</strong> Select a component or a variant, then click on the property name to edit its name and/or value.</li>
+  <li><strong>From the Layers panel:</strong> using the formula <code class="language-js">[property_name]=[value]</code>.</li>
+</ul>
+<figure>
+  <img src="/img/variants/04-variants-properties-edit.webp" alt="Edit variant property" />
+</figure>
+
+<h5>Delete properties:</h5>
```
madalenapmelo-kp:
```suggestion
<h5>Delete properties</h5>
```
I would remove the colon just to keep it consistent with the other subsections

madalenapmelo-kp (approved):

docs/user-guide/components/index.njk
myfunnyandy: I've rephrased it so it talks specifically about main components

PR #7265, opened 2025-09-04
:books: Update Design Tokens doc

myfunnyandy:
### Related Ticket
https://tree.taiga.io/project/penpot/task/11384


docs/user-guide/design-tokens/index.njk
```diff
@@ -269,6 +265,164 @@ title: 10· Design Tokens
 <p>The stroke width token, also known as the border width token, defines the thickness of a stroke around a design element. It can be applied to boards, groups, rectangles and text elements.</p>
 
 
+<h4>Number token</h4>
```
madalenapmelo-kp: Shouldn't this be an `<h3>`? And Stroke width as well? I would assume all token types have the same hierarchy in the index

docs/user-guide/design-tokens/index.njk
```diff
@@ -269,6 +265,164 @@ title: 10· Design Tokens
...
+<p>If the Font Family does not exist in Penpot, when you apply the token you will see a “Missing font” message in the design panel input:</p>
+<figure>
+  <img src="/img/design-tokens/29-tokens-type-font-family-missing.webp" alt="Font family missing" />
+</figure>
+
+
+<h4 id="design-tokens-font-size">Font Size</h4>
+<p>Font size tokens define the vertical size of glyphs/characters as an individual property. In typography, the letter spacing and line height properties are related to the font size.</p>
+<p>Font sizes are typically defined using a proportional scale. You can use math operations with references in order to create dynamic typography scales that follow a particular multiplier, like Golden Ratio.</p>
+<p><strong>Tip:</strong> Use <a href="/user-guide/design-tokens/#design-tokens-number" target="_blank">Number Tokens</a> (unitless token) to create stunning typographic scales though design tokens:</p>
```
madalenapmelo-kp, line 325: This `href` takes you to the beginning of the Design Tokens section, so it's not very helpful. I think it only makes sense to keep it if it takes you directly to the Number token subsection

madalenapmelo-kp (approved):

docs/user-guide/design-tokens/index.njk
myfunnyandy: Good catch! I'll change it.

docs/user-guide/design-tokens/index.njk
myfunnyandy: I might be missing something, but I think that taking to the Number token subsection is what the link is actually doing.

madalenapmelo-kp: I think it wasn't working because the section "Number tokens" was an h4, now that it's an h3 this issue is solved!

PR #7267, opened 2025-09-05
:tada: Add internal TextContent layout data

AzazelN28:
### Summary

Refactors how the text layout is generated and adds some new internal types to keep valuable information.

### Checklist


frontend/text-editor/src/index.html
```diff
@@ -530,6 +530,7 @@
 
           storeFonts(fonts)
 
+          console.log("text shape", uuid);
```
belen-albeza, line 533: Leftover

render-wasm/src/shapes/modifiers.rs
```diff
@@ -212,15 +215,13 @@ fn propagate_transform(
...
-                let height = auto_height(&mut paragraphs, paragraph_width);
-                let resize_transform = math::resize_matrix(
-                    &shape_bounds_after,
-                    &shape_bounds_after,
-                    paragraph_width,
-                    height,
-                );
+                // let paragraph_width = content.width();
+                // let mut paragraphs = paragraph_builder_group_from_text(content, None, None, None);
+                // let height = text_layout_auto_height(&mut paragraphs, paragraph_width);
```
belen-albeza: leftover

frontend/text-editor/src/index.html
belen-albeza, line 533: Leftover

render-wasm/src/shapes/text.rs
```diff
@@ -41,11 +44,117 @@ impl GrowType {
...
+        if f32::is_finite(size.max_width) {
+            self.max_width = size.max_width;
+        } else {
+            self.max_width = size.width;
+        }
+        if f32::is_finite(size.height) {
+            self.height = size.height;
+        }
+    }
+}
```
belen-albeza: Let's implement the `Copy` trait instead

render-wasm/src/wasm/text.rs
```diff
@@ -29,51 +26,37 @@ pub extern "C" fn set_shape_grow_type(grow_type: u8) {
...
+    });
 
-            // 2. Max Width Calculation
-            m_width = built_paragraphs
-                .iter()
-                .flatten()
-                .fold(0.0, |max_width, p| f32::max(p.max_width(), max_width));
+    // FIXME: Esto no me mola mucho pero no he encontrado
+    // una solución mejor.
+    ptr
```
belen-albeza: Q: What is the issue exactly?

belen-albeza (changes requested):
Requesting changes because of the auto height bug 
Screenshot 2025-09-22 at 12 08 02 PM

(all the other comments are minor)

frontend/text-editor/src/index.html
AzazelN28: This is not a leftover, this is intentional, I needed to debug some behavior of the text shapes.

render-wasm/src/wasm/text.rs
AzazelN28: I think it should be better if we called "panic!" (or have an error returned) when we're trying to do things that should'nt be possible instead of failing or passing through silently. In this case if we call `get_text_dimensions` on  a Shape that doesn't have a TextContent should fail, I still don't know how (error returned? and having a get_error function?) but I think it could be better than just doing nothing.

belen-albeza (approved):
Bug is fixed! ✨ 

(not merging this myself because it needs rebase and a linter fix)

PR #7287, opened 2025-09-09
:recycle: Refactor composite token UI

EvaMarco:
### Related Ticket

[This PR solves this task](https://tree.taiga.io/project/penpot/task/11897)

### Summary
Fix UI for composite typography token. 

### Steps to reproduce 
Ui must follow defined o[n US](https://tree.taiga.io/project/penpot/us/10200)

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
@@ -37,7 +37,7 @@
    [app.main.ui.workspace.colorpicker.ramp :refer [ramp-selector*]]
    [app.main.ui.workspace.sidebar.options.menus.typography :refer [font-selector*]]
    [app.main.ui.workspace.tokens.management.create.input-token-color-bullet :refer [input-token-color-bullet*]]
-   [app.main.ui.workspace.tokens.management.create.input-tokens-value :refer [input-tokens-value*
+   [app.main.ui.workspace.tokens.management.create.input-tokens-value :refer [input-token*
                                                                               token-value-hint*]]
    [app.util.dom :as dom]
    [app.util.functions :as uf]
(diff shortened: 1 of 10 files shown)


frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
```diff
@@ -848,8 +850,9 @@ custom-input-token-value-props: Custom props passed to the custom-input-token-va
            :type "button"}])]
     [:*
      [:> input-tokens-value*
-      {:placeholder (tr "workspace.tokens.token-font-family-value-enter")
-       :label (tr "workspace.tokens.token-font-family-value")
+      {:placeholder (or placeholder (tr "workspace.tokens.token-font-family-value-enter"))
```
Xaviju, line 853: I don't understand this behaviour well. Does this input sometimes not have a visible label?

If so, the name of the component is unclear to me because it does not seem to be an input, and I would not expect this component to use aria-label.

I suggest being explicit about the UI component (e.g. `font-picker-select*`).

frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
```diff
@@ -960,18 +969,20 @@ custom-input-token-value-props: Custom props passed to the custom-input-token-va
...
               :token-resolve-result (when (seq token-resolve-result) token-resolve-result)}])]))]))
 
 (mf/defc typography-reference-input*
   [{:keys [default-value on-blur on-update-value token-resolve-result]}]
   [:> input-tokens-value*
-   {:label "Reference"
-    :placeholder "Reference"
+   {:aria-label (tr "labels.reference")
+    :placeholder (tr "workspace.tokens.reference-composite")
+    :icon i/text-typography
```
Xaviju, line 985: I'm not sure how it works internally, but it might make sense to pass the icon name instead of the HTML. Then, this component could render the icon internally using the 'icon*' component.

Xaviju:
Just as a reminder that the input-value does not need the tooltip, so it might make sense to remove also the `aria-label` property implementation in the component and keep it as part of the generic props

frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
EvaMarco: This component is using the icon* component.

Xaviju: Oh I see, I've mistaken it for the old implementation. That looks better

Xaviju (approved):

PR #7296, opened 2025-09-10
:wrench: Autogenerate serialization values for wasm enums

belen-albeza:
https://tree.taiga.io/project/penpot/task/12000

## Summary

This PR introduces a new Rust macro (`ToJs`) meant to be used over our `enum` types that are deserialized from binary data that automatically generates JS code that is appended to our `render_wasm.js` bundle and we can use from ClojureScript.

## Pending work

(To be tackled in https://tree.taiga.io/project/penpot/task/12078)

- Many of the enums couple the model with the serialization (this is why some variant names needed to be renamed). We should fix that and **move the deserialization to the `wasm` module**. This will also help us to fix the inconsistency of handling None variants (sometimes within the enum, sometimes with an `Option`…)
- There is some deserialization done with methods that mimic the `From` trait, but without actually implementing it (like a `from` method just like that, or a `from_u8`). We should **properly implement `From<u8>`** for these enums.
- There are some serialization values / types that are not **documented** in `serialization.md`. We need to do it.
- Unify the criteria to use **default values** (or not, but then provide proper error handling).


frontend/src/app/render_wasm/api.cljs
```diff
@@ -1072,10 +1072,31 @@
...
+                                             :shadow-style (unchecked-get module "ShadowStyle")
+                                             :stroke-style (unchecked-get module "StrokeStyle")
+                                             :stroke-cap (unchecked-get module "StrokeCap")
+                                             :shape-type (unchecked-get module "Type")
+                                             :constraint-h (unchecked-get module "ConstraintH")
+                                             :constraint-v (unchecked-get module "ConstraintV")
+                                             :vertical-align (unchecked-get module "VerticalAlign")
+                                             :fill-data (unchecked-get module "RawFillData")
+                                             :segment-data (unchecked-get module "RawSegmentData")}]
+                         (set! wasm/serializers (js->clj serializers :keywordize-keys true))
```
niwinz: I think here we don't really need to convert this to cljs map. It can be just a JS object.

If we still continue to want convert it to cljs data structure in this case i suggest use `app.common.json/->clj` instead of `js->clj`

frontend/src/app/render_wasm/serializers.cljs
```diff
@@ -56,37 +57,21 @@
...
-    :frame   0
-    :group   1
-    :bool    2
-    :rect    3
-    :path    4
-    :text    5
-    :circle  6
-    :svg-raw 7
-    :image   8))
+  (let [values (:shape-type wasm/serializers)
```
niwinz: One wasm/serializers is a js object, we can use the unchecked-get instead

niwinz (changes requested):

elenatorro (approved):

niwinz (approved):
RTM

PR #7302, opened 2025-09-11
:bug: Fix fast move with distance

AzazelN28:
### Related Ticket

[Taiga Issue #12069](https://tree.taiga.io/project/penpot/issue/12069)

### Summary

There is a new feature for this release:

Up until now, if you select an item while presing alt, then hover over nearby elements, it tells you the distance in between them, but the ability to move your object was blocked.

Now you can move the objects while pressing alt. 
You can move them pixel by pixel in the x and y asxis, and also by pressing shift, the movement is incremental. BUT, this bug is because the incremental movement for the y axis is blocked

### Steps to reproduce 

1. Go to a file and create 2 rectangles
2. Select one of them and hover the other item whilst pressing alt
3. Without releasing, use the arrows in your keyboard to move the element
4. Without releasing, use the arrows in your keyboard + shift to move the element
5. See how it's only moving horizontally but not vertically

### Checklist

### DIFF
--- a/frontend/src/app/main/data/workspace/shortcuts.cljs
+++ b/frontend/src/app/main/data/workspace/shortcuts.cljs
@@ -213,12 +213,12 @@
                           :fn #(emit-when-no-readonly (dw/vertical-order-selected :bottom))}
 
    :move-fast-up         {:tooltip (ds/shift ds/up-arrow)
-                          :command ["shift+up"]
+                          :command ["shift+up" "shift+alt+up" "alt+shift+up"]
                           :subsections [:modify-layers]
                           :fn #(emit-when-no-readonly (dwt/move-selected :up true))}
 
    :move-fast-down       {:tooltip (ds/shift ds/down-arrow)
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/data/workspace/shortcuts.cljs
```diff
@@ -213,12 +213,12 @@
                           :fn #(emit-when-no-readonly (dw/vertical-order-selected :bottom))}
 
    :move-fast-up         {:tooltip (ds/shift ds/up-arrow)
-                          :command ["shift+up"]
+                          :command ["shift+up" "shift+alt+up" "alt+shift+up"]
```
niwinz: This conflicts with text shortcuts, that additional bindings are deleted explicitly on https://github.com/penpot/penpot/pull/7136

Independently, i think we should make it more consistent: "shift+alt+up" "alt+shift+up" looks the same shortcut, why we need to put the both?

PR #7307, opened 2025-09-12
:bug: Fix make-token throwing because of error in name

floscr:
### Related Ticket

https://github.com/tokens-studio/penpot/issues/149

### Summary

Fixes resolved value preview when the token name is invalid, caught by tokens-lib.
We still want to show the user the value, so we use an unchecked version of make-token.
The name will never be used in the final submitted token form the form.

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
@@ -10,6 +10,7 @@
    [app.common.data :as d]
    [app.common.data.macros :as dm]
    [app.common.files.tokens :as cft]
+   [app.common.schema :as sm]
    [app.common.types.color :as c]
    [app.common.types.token :as ctt]
    [app.common.types.tokens-lib :as ctob]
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/ui/workspace/tokens/management/create/form.cljs
```diff
@@ -129,7 +129,7 @@
                   (dissoc (:name prev-token))
 
                   :always
-                  (update (:name token) #(ctob/make-token (merge % prev-token token))))]
+                  (update (:name token) #(ctob/unchecked-make-token (merge % prev-token token))))]
```
hirunatan: I don't like the idea of having an unchecked version of the constructor. It may be the seed for obscure problems in the future.

I suggest a different approach. Due that this temporary token is used only to resolve the value, you can replace the name with the placeholder always, not only when the name is empty. This way you avoid the error in the name, and anyway the replaced name is not going to be shown to the user anywhere.

floscr: Makes sense, I've updated it

PR #7315, opened 2025-09-15
:recycle: Refactor defmulti fn into case switches

Xaviju:
### Related Ticket

https://tree.taiga.io/project/penpot/task/12086

### Summary

- Refactor defmulti fn into a case switch

### Steps to reproduce 

- NO UI, everything should work as before 

### Checklist

### DIFF
--- a/frontend/src/app/util/code_gen/style_css_formats.cljs
+++ b/frontend/src/app/util/code_gen/style_css_formats.cljs
@@ -53,24 +53,7 @@
    :grid-template-rows        :tracks
    :grid-template-columns     :tracks})
 
-(defmulti format-value
-  (fn [property _value _options] (css-formatters property)))
-
-(defmethod format-value :position
-  [_ value _options]
-  (cond
-    (number? value) (fmt/format-pixels value)
-    :else value))
-
-(defmethod format-value :size
-  [_ value _options]
-  (cond
-    (= value :fill) "100%"
-    (= value :auto) "auto"
-    (number? value) (fmt/format-pixels value)
-    :else value))
-
-(defn format-color
+(defn format-color->css
   "Format a color value to a CSS compatible string based on the given format."
   [value options]
   (let [format (get options :format :hex)]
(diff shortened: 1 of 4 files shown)


frontend/src/app/util/code_gen/style_css_formats.cljs
```diff
@@ -53,24 +53,7 @@
...
-(defmethod format-value :size
-  [_ value _options]
-  (cond
-    (= value :fill) "100%"
-    (= value :auto) "auto"
-    (number? value) (fmt/format-pixels value)
-    :else value))
-
-(defn format-color
+(defn format-color->css
```
niwinz: I think the `->css` suffix is redundant, i mean this function is on `style-css-formats` namespace, all this code is related to CSS.

frontend/src/app/util/code_gen/style_css_formats.cljs
```diff
@@ -158,29 +162,38 @@
...
-(defmethod format-value :default
-  [_ value _options]
-  (if (keyword? value)
-    (d/name value)
-    value))
+
+(defn format-value
+  "Get the appropriate value formatter function for a given CSS property."
+  [property value options]
+  (let [property (css-formatters property)]
```
niwinz: Be explicit if possible, css-formatters is a map, so it is better to be explicit `(get css-formatters property)`

frontend/src/app/util/code_gen/style_css_formats.cljs
```diff
@@ -92,33 +75,54 @@
...
+
+(defn- format-color
   [_ value options]
   (let [format (get options :format :hex)]
-    (format-color value (assoc options :format format))))
+    (format-color->css value (assoc options :format format))))
 
-(defmethod format-value :color-array
+(defn- format-color-array
   [_ value options]
```
niwinz: all this private functions ignores the first argument, I think we just should not pass it and remove the need to ignore the first argument

frontend/src/app/util/code_gen/style_css_values.cljs
```diff
@@ -111,7 +72,79 @@
       (some? (get shape type))
       (get shape type))))
 
-(defmethod get-value :width
+
+
+(defn- get-position
+  [_ shape objects _]
```
niwinz: The same here, the first argument is ignored and the last one; This is a private function, i think we should not pass that arguments to this function.

niwinz (changes requested):

frontend/src/app/util/code_gen/style_css_formats.cljs
Xaviju: There is already a `format-color` function because every function follows the pattern format-{property}
I changed it to `format-color-value` since it receives a value and transforms it into a CSS friendly color.

niwinz (approved):
RTM

PR #7319, opened 2025-09-15
:sparkles: Add shortcut for creating variant to the shortcuts panel

luisddm:
### Related Ticket

Taiga [#11818](https://tree.taiga.io/project/penpot/task/11818)

### Summary

- Add shortcut for creating a variant in the shortcuts panel.
- Refactor to replace the old `rumext` syntax with the new one.
- Add a unique key prop to each child in two iterations to avoid errors in the console.
- Replace old icons with the new ones from the DS.
- Remove unused CSS styles.

### Steps to reproduce

- Check that the shortcuts panel shows the shortcut for creating a variant.
- Check that there are no errors in the console when opening the shortcuts panel.

### DIFF
--- a/frontend/translations/en.po
+++ b/frontend/translations/en.po
@@ -3904,7 +3904,7 @@ msgstr "Copy properties"
 
 #: src/app/main/ui/workspace/sidebar/shortcuts.cljs:95
 msgid "shortcuts.create-component"
-msgstr "Create component"
+msgstr "Create component / variant"
 
 #: src/app/main/ui/workspace/sidebar/shortcuts.cljs:96
 msgid "shortcuts.create-new-project"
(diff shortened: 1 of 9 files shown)


frontend/translations/en.po
```diff
@@ -3904,7 +3904,7 @@ msgstr "Copy properties"
 
 #: src/app/main/ui/workspace/sidebar/shortcuts.cljs:95
 msgid "shortcuts.create-component"
-msgstr "Create component"
```
EvaMarco, line 3907: is better to create a complete new translation string that reuse the old one, because this can be already translated to some langs, and the change could not reach those languages.

EvaMarco:
Screenshot from 2025-09-16 08-21-57
While deleting css you have deleted icon color.

EvaMarco:
Screenshot from 2025-09-16 08-26-56
overflow is set to scroll not auto so the scrollbar is present all the time

EvaMarco (approved):

PR #7343, opened 2025-09-17
:bug: Fix parsing large paths with multiple subpaths

elenatorro:
### Related Ticket

https://tree.taiga.io/project/penpot/task/12118

### Summary

We had a couple of issues detecting open subpaths in Path shapes.
- First of all, the method to check subpaths was recursive, being an issue on large paths with multiple nested subpaths. This has been changed to an iterative approach.
- Second, we were missing some cases by detecting a closed path as open, by not checking correctly first and end points when there was not a close point, for instance

Extra:

1. This PR includes a small fix that prevents rendering shadows of hidden children in a Group shape
2. This PR is ready to support larger path content by splitting it in chunks. The real problem of the original ticket was the recursion in checking subpaths, but we decided to also tackle the possibility of parsing paths in chunks. I tested the current approach within this one with the first file attached in "Steps to reproduce", and the current approach seems to be from 10ms to 25ms faster. I think it's worth checking it just in case.

### Steps to reproduce 

1. Load this penpot file with a couple of "complex" svg path shapes. It should render as expected. It already includes multiple strokes and a drop shadow

[test_complex_path_with_stroke.zip](https://github.com/user-attachments/files/22514225/test_complex_path_with_stroke.zip)

image

3. Load this smaller path and compare it with production. It should detect the path as "closed" and different strokes (outer, center, inner) should be rendered differently, not alwasy as center

[small_closed_path.zip](https://github.com/user-attachments/files/22514256/small_closed_path.zip)

image

### Checklist


frontend/src/app/render_wasm/api.cljs
```diff
@@ -51,6 +52,8 @@
 (def ^:const GRID-LAYOUT-COLUMN-U8-SIZE 8)
 (def ^:const GRID-LAYOUT-CELL-U8-SIZE 36)
 
+(def ^:const MAX_BUFFER_CHUNK_SIZE 280000) ;; FIXME
```
Alotor: Maybe you can add more explanation about what needs to be fixed?

render-wasm/src/shapes/paths/subpaths.rs
```diff
@@ -17,230 +15,188 @@ impl Subpath {
...
+        if used[i] {
+            continue;
         }
 
-        let mut segments = subpath.segments.clone();
-        segments.extend_from_slice(&other.segments);
-        Ok(Subpath::new(segments))
-    }
-}
+        let mut current = subpaths[i].clone();
```
Alotor, line 94: I'm trying to see why is necessary to clone this. I'm not seeing where the mutable reference.

elenatorro: this is because `current` can be reassigned later

frontend/src/app/render_wasm/api.cljs
elenatorro: I left it explicitly at 256 KB, which should be large enough without exceeding the memory limit 👍🏽

Alotor (approved):

PR #7348, opened 2025-09-18
:tada: Switch several variant copies at the same time

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot/us/11411

### Summary

 Switch several variant copies at the same time

### Steps to reproduce 

1. Create a component with variant
2. Create two copies of the first variant
3. Select both copies
4. Change a variant property to make a switch of both at the same time

### Checklist

### DIFF
--- a/frontend/src/app/main/data/workspace/variants.cljs
+++ b/frontend/src/app/main/data/workspace/variants.cljs
@@ -7,6 +7,7 @@
 (ns app.main.data.workspace.variants
   (:require
    [app.common.data :as d]
+   [app.common.data.macros :as dm]
    [app.common.files.changes-builder :as pcb]
    [app.common.files.helpers :as cfh]
    [app.common.files.variant :as cfv]
@@ -16,6 +17,7 @@
    [app.common.types.color :as clr]
    [app.common.types.component :as ctc]
    [app.common.types.components-list :as ctkl]
+   [app.common.types.file :as ctf]
    [app.common.types.shape.layout :as ctsl]
    [app.common.types.variant :as ctv]
    [app.common.uuid :as uuid]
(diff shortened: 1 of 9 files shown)


frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -653,3 +655,56 @@
...
+                comps-nesting-loop?    (seq? (cfh/components-nesting-loop? nearest-comp-children shape-parents))
+
+                {:keys [on-error]
+                 :or {on-error rx/throw}} (meta params)]
+
+            ;; If there is no nearest-comp, do nothing
+            (when nearest-comp
+              (if comps-nesting-loop?
+                (do
+                  (on-error)
```
niwinz, line 693: The main idea behind this pattern is `on-error` should return a stream of events, so here yo just can call on-error.

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -653,3 +655,56 @@
...
+                  (rx/empty))
+                (rx/of (dwl/component-swap shape (:component-file shape) (:id nearest-comp) true))))))))))
+
+(defn variants-switch
+  "Switch each shape (that must be a variant copy head) for the closest one with the property value passed as parameter"
+  [{:keys [shapes] :as params}]
+  (ptk/reify ::variants-switch
+    ptk/WatchEvent
+    (watch [_ _ _]
+      (let [ids (into (d/ordered-set) (map :id shapes))
```
niwinz: I suggest reuse the transducer from common data: `(into (d/ordered-set) d/xf:map-id shapes)`

frontend/src/app/main/data/workspace/variants.cljs
```diff
@@ -653,3 +655,56 @@
...
+            undo-id (js/Symbol)]
+        (rx/concat
+         (rx/of (dwu/start-undo-transaction undo-id))
+         (rx/from (map  #(variant-switch % params) shapes))
```
niwinz: MINOR: I think this is better impl:

```
(->> (rx/from shapes)
     (rx/map #(variant-switch % params)))
```
because it avoids creating an additional sequence for mapping

frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
```diff
@@ -449,47 +450,44 @@
...
-                     (st/emit! (ntf/error (tr "workspace.component.swap.loop-error")))
-                     (reset! key* (uuid/next)))
-                   (st/emit! (dwl/component-swap shape (:component-file shape) (:id nearest-comp) true))))))))]
+           (if (= val mixed-label)
+             (reset! key* (uuid/next))
+             (let [error-msg (if (> (count shapes) 1)
+                               (tr "workspace.component.switch.loop-error-multi")
+                               (tr "workspace.component.swap.loop-error"))
+                   mdata     {:on-error #(do
+                                           (st/emit! (ntf/error error-msg))
```
niwinz: here, instead of st/emit! just return `(rx/of (ntf/error ...))`

niwinz: And in my opinion, this error can be part of `variants-switch`. Because it is already coupled to it and related to what the `variant-switch` component does. And is used only once i mean, we don't have different parts of code emitting `variant-switch` event with completely different message.

frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
```diff
@@ -449,47 +450,44 @@
...
-                       :options (get-options (:name prop))
-                       :empty-to-end true
-                       :on-change (partial switch-component pos)
-                       :key (str (:value prop) "-" key)}]]])]
+      (for [[pos prop] (map vector (range) props-first)]
+        (let [mixed-value? (not-every? #(= (:value prop) (:value (nth % pos))) properties)
+              options (cond-> (get-options (:name prop))
+                        mixed-value?
+                        (conj {:id mixed-label, :label mixed-label :dimmed true}))]
+          [:div {:key (str (:id shape) pos mixed-value?)
```
niwinz: The key should be unique on the iteration, i think with `pos` is more than enough for the key, so you can remove the string concat operation and replace it with just using `pos`

pabloalba: Ok, but I need to keep `mixed-value?` on the key, to reset this when the value change

frontend/src/app/main/ui/workspace/sidebar/options/menus/component.cljs
pabloalba: The problem is that I need to do the  `(reset! key* (uuid/next))` when there is an error, so I need to get the error info... Maybe we can talk about this

niwinz (approved):
RTM: squash

PR #7361, opened 2025-09-23
:tada: Text Editor DOM textarea position

AzazelN28:
### Related Ticket

[Taiga Task #11828](https://tree.taiga.io/project/penpot/task/11828)

### Summary

Position text editor element based on shape and viewport transformation.

### Checklist


frontend/src/app/main/ui/workspace/viewport_wasm.cljs
```diff
@@ -431,6 +431,7 @@
        (when show-text-editor?
          (if (features/active-feature? @st/state "text-editor/v2")
            [:& editor-v2/text-editor {:shape editing-shape
+                                      :canvas-ref canvas-ref
```
elenatorro, line 434: can we add `:canvas-ref` property at [viewport.cljs](https://github.com/penpot/penpot/blob/develop/frontend/src/app/main/ui/workspace/viewport.cljs#L439-L445) too? Otherwise it does not work in the current render and I think it's ok to have it there for testing 👍🏽

render-wasm/src/render/text.rs
```diff
@@ -223,29 +223,37 @@ fn draw_text(
...
+            let _paragraph_height = paragraph.height();
+            // FIXME: I've kept the _paragraph_height variable to have
+            // a reminder in the future to keep digging why the ideographic_baseline
+            // works so well and not the paragraph_height. I think we should test
+            // this more.
+            let ideographic_baseline = paragraph.ideographic_baseline();
             let xy = (shape.selrect().x(), shape.selrect().y() + group_offset_y);
             paragraph.paint(canvas, xy);
 
+            println!("alphabetic_baseline {}", paragraph.alphabetic_baseline());
```
elenatorro: can we remove these prints? If we want to keep them, maybe we can log them only on debug mode

frontend/text-editor/src/editor/TextEditor.js
```diff
@@ -553,8 +583,9 @@ export function setRoot(instance, root) {
   return instance;
 }
 
-export function create(element, options) {
-  return new TextEditor(element, { ...options });
+export function create(element, canvas, options) {
+  console.log("TextEditor", element, canvas, options);
```
elenatorro: same with this console.log

render-wasm/src/wasm/text.rs
```diff
@@ -368,7 +368,7 @@ pub extern "C" fn update_shape_text_layout_for_all() {
 
 #[no_mangle]
 pub extern "C" fn get_caret_position_at(x: f32, y: f32) -> i32 {
-    with_current_shape!(state, |shape: &Shape| {
+    with_state_mut_current_shape!(state, |shape: &Shape| {
```
elenatorro, line 371: can't we use here directly "with_current_shape"? I think there's no need of `with_state_mut_current_shape` as far as I know, at least not for the current changes, right?

elenatorro (changes requested):

render-wasm/src/render/text.rs
AzazelN28: Sure! I forgot to remove them.

render-wasm/src/wasm/text.rs
AzazelN28: the problem `with_current_shape` is that returns state as immutable, but in this case we need to update the state (because we need to change the caret position in the text editor state) but not the shape. Maybe in the future we could create a `with_text_editor_state_current_shape` or something like that.

frontend/src/app/main/ui/workspace/viewport_wasm.cljs
AzazelN28: done!

elenatorro (approved):
🚀

PR #7364, opened 2025-09-23
:recycle: Add shortcut to scss import  paths

EvaMarco:
### Related Ticket

This PR complete [this task](https://tree.taiga.io/project/penpot/task/12145)

### Summary

We want to import DS files easily, we don't want to write the complete path to import a DS scss file. 

### Steps to reproduce 

Using the shortcut "ds/spacing.scss" should work in any file on the frontend folder. 

### Checklist


luisddm (commented):
There are three files (`_borders.scss`, `_sizes.scss` and `_utils.scss`) that sometimes are imported with the leading underscore, and sometimes not. Unless there is a good reason to do so, we should consider doing it always the same way.

Moreover, since the leading underscore is a naming convention, we should decide whether and when to use it.

luisddm (approved):

PR #7391, opened 2025-09-26
:bug: Load dependant libraries, and don't allow unload them

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/12155

### Summary

SETUP  
There are a library L1, that have dependencies to another library L2.

PROBLEM 1  
When the user adds L1 to a file on the libraries popup, L2 is not loaded. This provokes several internal problems for components referenced but not loaded.  
If you reload the file, L2 is automaticaly added, and there is no more problems. But if you don't reload, the file can became corrupt.

PROBLEM 2  
On the same file, after reloading, you have L1 and L2 as dependencies. On the library popup you can remove L2, so you will get the same problem as before.

### Steps to reproduce 

1. Create a library L2 with a component C2  
2. Create a library L1. Add L2 as a dependency. Create a component C1, that have a copy of C2 inside  
3. Create a file. Add L1 as a dependency.   
4. On the file, add a copy of C1. Select the copy of C2 inside. On the contextual button, it has a “restore main component” option. That shows that the original component is inaccesible.

### Checklist

### DIFF
--- a/backend/src/app/migrations.clj
+++ b/backend/src/app/migrations.clj
@@ -447,7 +447,10 @@
     :fn (mg/resource "app/migrations/sql/0140-add-locked-by-column-to-file-change-table.sql")}
 
    {:name "0141-add-idx-to-file-library-rel"
-    :fn (mg/resource "app/migrations/sql/0141-add-idx-to-file-library-rel.sql")}])
+    :fn (mg/resource "app/migrations/sql/0141-add-idx-to-file-library-rel.sql")}
+
+   {:name "0142-create-view-file_with_library_files"
+    :fn (mg/resource "app/migrations/sql/0142-create-view-file_with_library_files.sql")}])
 
 (defn apply-migrations!
   [pool name migrations]
--- a/backend/src/app/migrations.clj
+++ b/backend/src/app/migrations.clj
@@ -447,10 +447,7 @@
     :fn (mg/resource "app/migrations/sql/0140-add-locked-by-column-to-file-change-table.sql")}
 
    {:name "0141-add-idx-to-file-library-rel"
-    :fn (mg/resource "app/migrations/sql/0141-add-idx-to-file-library-rel.sql")}
-
-   {:name "0142-create-view-file_with_library_files"
-    :fn (mg/resource "app/migrations/sql/0142-create-view-file_with_library_files.sql")}])
+    :fn (mg/resource "app/migrations/sql/0141-add-idx-to-file-library-rel.sql")}])
 
 (defn apply-migrations!
   [pool name migrations]
--- a/frontend/src/app/main/data/workspace/libraries.cljs
+++ b/frontend/src/app/main/data/workspace/libraries.cljs
@@ -1406,17 +1406,6 @@
                         (fn [state]
                           (update state :thumbnails merge thumbnails))))))))))
 
-(defn- all-used-libraries [libraries root-id]
-  (letfn [(collect [id visited]
-            (if (visited id)
-              visited
-              (let [deps (get-in libraries [id :library-file-ids])]
-                (reduce (fn [v dep]
-                          (collect dep v))
-                        (conj visited id)
-                        deps))))]
-    (disj (collect root-id #{}) root-id)))
-
 (defn link-file-to-library
   [file-id library-id]
   (ptk/reify ::attach-library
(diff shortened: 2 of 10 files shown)


frontend/src/app/main/data/workspace/libraries.cljs
```diff
@@ -1389,41 +1426,36 @@
        :file-id file-id
        :library-id library-id})
 
-    ;; NOTE: this event implements UpdateEvent protocol for perform an
-    ;; optimistic update state for make the UI feel more responsive.
-    ptk/UpdateEvent
-    (update [_ state]
-      (let [libraries (:workspace-shared-files state)
```
pabloalba, line 1397: :workspace-shared-files doesn't exist since ages ago

This wasn't doing anything, and the UI is quite responsive without it

backend/src/app/migrations.clj
```diff
@@ -447,7 +447,10 @@
     :fn (mg/resource "app/migrations/sql/0140-add-locked-by-column-to-file-change-table.sql")}
 
    {:name "0141-add-idx-to-file-library-rel"
-    :fn (mg/resource "app/migrations/sql/0141-add-idx-to-file-library-rel.sql")}])
+    :fn (mg/resource "app/migrations/sql/0141-add-idx-to-file-library-rel.sql")}
+
+   {:name "0142-create-view-file_with_library_files"
```
niwinz: The migration name and the file should not mix casing. All should use the `-` as separator, following the same way as the rest of migrations

frontend/src/app/main/data/workspace/libraries.cljs
niwinz: Yep, this is a bug, it was renamed to `:shared-files`

frontend/src/app/main/data/workspace/libraries.cljs
```diff
@@ -1389,41 +1426,36 @@
...
+                                  (map first)
+                                  set)
+
+            libraries (update-vals libraries
+                                   (fn [library]
+                                     (assoc library
+                                            :library-file-ids
+                                            (->> (str/split (:library-file-ids library) #",")
+                                                 (remove empty?)
+                                                 (mapv uuid/uuid)))))
```
niwinz: This step will not be necessary if we return a set of uuids directly from backend

frontend/src/app/main/data/workspace/libraries.cljs
```diff
@@ -1389,41 +1426,36 @@
...
-               (rx/map (fn [file]
-                         (fn [state]
-                           (assoc-in state [:files library-id] file)))))
-          (->> (rp/cmd! :get-file-object-thumbnails {:file-id library-id :tag "component"})
-               (rx/map (fn [thumbnails]
-                         (fn [state]
-                           (update state :thumbnails merge thumbnails))))))
+          (->> libraries-to-load
+               (rx/from)
+               (rx/map #(load-library-file file-id %))))
```
niwinz: I undertand the idea behind this code,  but i think we can solve this much early, directly returning necessary libraries and transitive libraries from backend.

We already have on backend a query (used for exportation but the same idea can be reused here also i think) on `app.binfile.common/get-libraries` which retrieves all the libraries including the recursive ones.

I think a more simpler approach can be make the link-file-to-library method return a proper list of libraries linked (internally using the already existing helper for obtain recursive libraries), the current one and the transitive, and then we just can proceed to fetch the new libraries and resolve them  in the same way it is done on workspace initialization. With this approach we remove all this complex resolution logic from frontend.

frontend/src/app/main/ui/workspace/libraries.cljs
```diff
@@ -199,27 +197,66 @@
...
+                                          (->> (str/split library-file-ids ",")
+                                               (map #(uuid/uuid %))
+                                               set)])
+                                       (vals shared-libraries))))
+
+        library-names  (mf/with-memo [shared-libraries]
+                         (into {} (map (fn [{:keys [id name]}]
+                                         [id name])
+                                       (vals shared-libraries))))
+
```
niwinz, line 212: I think this two memoized values are not really necessary, because

frontend/src/app/main/ui/workspace/libraries.cljs
```diff
@@ -199,27 +197,66 @@
...
+         (fn [library-id]
+           (->> dependencies
+                (filter (fn [[_ v]] (contains? v library-id)))
+                (map first))))
+
+        connected-to-linked?
+        (mf/use-fn
+         (mf/deps linked-libraries shared-libraries dependencies)
+         (fn [id]
+           (contains? linked-libraries id)))
```
niwinz: This can be already resolved with using `(contains? ...` directly where it is needed
This callback looks redundant and has more dependencies that are necessary i think

backend/src/app/migrations.clj
pabloalba: Migration replaced by WITH

frontend/src/app/main/data/workspace/libraries.cljs
pabloalba: Fixed

frontend/src/app/main/ui/workspace/libraries.cljs
pabloalba: Fixed

PR #7399, opened 2025-09-29
:bug: Fix internal error when fonts return 404 (wasm)

belen-albeza:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/12164

### Summary

This fixes an Internal Error crash in the wasm renderer when it tries to download a font asset that is not available (i.e. the HTTP request returns `404`).

Instead of the crash, we now render the text shape with the default font and variant.

Screenshot 2025-09-29 at 3 39 59 PM

### Steps to reproduce 

> ℹ️ You can force a 404 in the browser devtools. For Chromium browsers, go to the network panel, inspect the request you want to block and right click on it and select "Block request URL". For instance, for the Crimson Pro font this is the TTF asset that starts with URL `localhost:3449/internal/gfonts/font/crimsonpro/...`

1. Create a text shape and set its font to a Google font (Crimson Pro)
2. Block the network request for that font
3. Reload the page

Actual result (i.e. without this fix)

- Internal error page

Expected result (i.e. the one that this fix should show)

- There is no internal error page and the text shape is rendered with the default font.

### Checklist

### DIFF
--- a/frontend/src/app/render_wasm/api/fonts.cljs
+++ b/frontend/src/app/render_wasm/api/fonts.cljs
@@ -8,6 +8,7 @@
   (:require
    [app.common.data :as d]
    [app.common.data.macros :as dm]
+   [app.common.logging :as log]
    [app.common.uuid :as uuid]
    [app.config :as cf]
    [app.main.fonts :as fonts]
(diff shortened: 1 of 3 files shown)


frontend/src/app/render_wasm/api/fonts.cljs
```diff
@@ -106,6 +107,10 @@
    :callback #(->> (http/send! {:method :get
                                 :uri font-url
                                 :response-type :buffer})
+                   (rx/catch (fn [cause]
```
niwinz, line 119: I suggest put `catch` as last element of this stream chain.
Also, split the hit on two keys on logging for avoid doing unnecesary concat:

```clojure
(log/error :hint "could not fetch font"
           :font-url font-url
           :cause cause)
```

elenatorro (approved):

PR #7417, opened 2025-10-01
:bug: Fix form not being saveable when editing composite token and sw…

floscr:
### Summary

- Raises prop if form is being edited or created to the consumer
- Fixes submit-hook being re-triggered when switching tabs and enabling submit on the form, despite empty value.
 (Composite form removes the `:value` from `token` when switching the tab, as it doesn't match the tab type)

### Steps to reproduce

1. Edit existing typography toke
2. Switch tabs
3. Form should be disabled
4. Enter value
5. If valid -> should be enabled

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/management/create/modals.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/management/create/modals.cljs
@@ -8,6 +8,7 @@
   (:require-macros [app.main.style :as stl])
   (:require
    [app.common.data.macros :as dm]
+   [app.common.types.tokens-lib :as ctob]
    [app.main.data.modal :as modal]
    [app.main.refs :as refs]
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/ui/workspace/tokens/management/create/modals.cljs
```diff
@@ -88,7 +89,8 @@
                        :icon i/close
                        :variant "action"
                        :aria-label (tr "labels.close")}]
-     [:> form-wrapper* {:token token
+     [:> form-wrapper* {:is-create (not (instance? ctob/Token token))
```
hirunatan: We have a helper for this `(ctob/token? token)`. It's better to use it, because in the future we may add further checks.

PR #7450, opened 2025-10-06
:bug: Remove duplicated token set ids

hirunatan:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/12243

### Summary

There are files with several token sets with the same id. This is invalid, so we add a validation check to avoid this occurs again, and a migration to fix it if it's already occuring in a file.

### Steps to reproduce 

This bug cannot be reproduced, since we don't know the cause of the duplication. We have detected it in two files, linked in the issue. But once those two have been repaired, there is no known way of reproducing.

When the files are exported and reimported, all ids are regenerated, cleaning the failure. But if you duplicate the penpot files, the error persists.

### Checklist

### DIFF
--- a/common/src/app/common/files/migrations.cljc
+++ b/common/src/app/common/files/migrations.cljc
@@ -33,6 +33,7 @@
    [app.common.types.shape.shadow :as ctss]
    [app.common.types.shape.text :as ctst]
    [app.common.types.text :as types.text]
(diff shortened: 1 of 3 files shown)


Xaviju (approved):

Xaviju:
@niwinz I've approved this PR because we have had a conversation about it to unblock it. @hirunatan expects your approval as well before merge

PR #7457, opened 2025-10-07
:sparkles: Add telemetry for events performance measures

Alotor:
### Related Ticket
https://tree.taiga.io/project/penpot/us/11497?milestone=478016

### Summary

This feature creates some telemetry events to measure performance issues. There are two kinds of performance tracking:

Browser's PerformanceObserver 
This allows us to measure `longtasks` and long duration `events`

Time to render of internal events.  
For every event that is measured we only measure when the time is over a threshold (now 1s but can be changed in a constant) and it's debounced so it only sends 1 event of each type (again, can be configured)

The following events are tracked now:

* `commit-changes` measures time to render against any model update
* `select-shapes` (and derivates) this measures the time to render when selecting shapes
* `update-viewport-position` this will measure the time to render when panning
* `increase-zoom` and `decrease-zoom` will measure time to render when zooming

In every event some performance context is attached.

* Curent file number of layers
* Current file library size (in number of assets)
* File start time
* App start time

Also adds a new debug utility to measure groups of events with time

### Steps to reproduce 

- Activate the flag `enable-audit-log`
- Open a file that is currently slowing down
- Make some interactions: pan, zoom, select, create shapes...
- Check the network to see the telemetry events

### Checklist

### DIFF
--- a/frontend/src/app/main/data/event.cljs
+++ b/frontend/src/app/main/data/event.cljs
@@ -8,6 +8,7 @@
   (:require
    ["ua-parser-js" :as ua]
    [app.common.data :as d]
+   [app.common.data.macros :as dm]
    [app.common.json :as json]
    [app.common.logging :as l]
    [app.common.math :as math]
@@ -56,6 +57,9 @@
 
 ;; Min time between long task reports
 (def debounce-performance-event-time 1000)
+
+;; Def micro-benchmark iterations
+(def micro-benchmark-iterations 1e6)
 
 ;; --- CONTEXT
 
@@ -260,14 +264,16 @@
        (let [observer
              (js/PerformanceObserver.
               (fn [list]
-                (doseq [entry (.getEntries list)]
-                  (when (and (= "event" (.-entryType entry))
-                             (> (.-duration entry) min-browser-event-time))
-                    (rx/push!
-                     subs
-                     {::event :observer-event
-                      :duration (.-duration entry)
-                      :event-name (.-name entry)})))))]
+                (run!
+                 (fn [entry]
+                   (when (and (= "event" (.-entryType entry))
+                              (> (.-duration entry) min-browser-event-time))
+                     (rx/push!
+                      subs
+                      {::event :observer-event
+                       :duration (.-duration entry)
+                       :event-name (.-name entry)})))
+                 (.getEntries list))))]
          (.observe observer #js {:entryTypes #js ["event"]})
          (fn []
            (.disconnect observer)))))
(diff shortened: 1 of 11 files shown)


frontend/src/app/main/data/event.cljs
```diff
@@ -183,10 +250,118 @@
...
+         (.observe observer #js {:entryTypes #js ["longtask"]})
+         (fn []
+           (.disconnect observer)))))
+    (rx/empty)))
+
+(defn store-performace-info
+  []
+  (letfn [(micro-benchmark [state]
+            (let [start (js/performance.now)]
+              (doseq [i (range 0 1e6)]
```
niwinz: I suggest using `loop` here with clear counter, `doseq` generates a lot of code (because it a general purpose macro and it should handle several cases), and also remove the range usage because we just need to iterate. On cases like this is better use `(loop [i 0] ...)`

frontend/src/app/main/data/event.cljs
```diff
@@ -183,10 +250,118 @@
...
+              0))
+
+          (count-library-data [files {:keys [id]}]
+            (let [data (get-in files [id :data])]
+              {:components (count (:components data))
+               :colors (count (:colors data))
+               :typographies (count (:typographies data))}))
+
+          (save-file-data []
+            (ptk/reify ::store-performance-info-save-file-data
```
niwinz: Very minor suggestion: define it as standalone event function.

frontend/src/app/main/data/event.cljs
```diff
@@ -183,10 +250,118 @@
...
+
+          (count-shapes [file]
+            (if file
+              (->> file :data :pages-index vals
+                   (map (comp count :objects))
+                   (reduce +))
+              0))
+
+          (count-library-data [files {:keys [id]}]
+            (let [data (get-in files [id :data])]
```
niwinz: `dm/get-in`

frontend/src/app/main/data/event.cljs
```diff
@@ -183,10 +250,118 @@
...
+              (doseq [i (range 0 1e6)]
+                (* (math/sin i) (math/sqrt i)))
+              (let [end (js/performance.now)]
+                (assoc-in state [:performance-info :bench-result] (- end start)))))
+
+          (count-shapes [file]
+            (if file
+              (->> file :data :pages-index vals
+                   (map (comp count :objects))
+                   (reduce +))
```
niwinz: I suggest replace this map + reduce
with a single reduce-kv

```clojure
(reduce-kv (fn [sum _ page]
             (+ sum (count (:objects apge))))
           0
           (-> file :data :pages-index))
```

frontend/src/app/main/data/event.cljs
```diff
@@ -183,10 +250,118 @@
...
+    (rx/empty)))
+
+(defn store-performace-info
+  []
+  (letfn [(micro-benchmark [state]
+            (let [start (js/performance.now)]
+              (doseq [i (range 0 1e6)]
+                (* (math/sin i) (math/sqrt i)))
+              (let [end (js/performance.now)]
+                (assoc-in state [:performance-info :bench-result] (- end start)))))
```
niwinz: I suggest replace the usage of assoc-in that requires building a vector for traveral to something like this

```clojure
(update state :performance-info assoc :bench-result (- end start))
```

frontend/src/app/main/data/event.cljs
```diff
@@ -183,10 +250,118 @@
...
+                           (vals)
+                           (reduce (fn [acc {:keys [components colors typographies]}]
+                                     (-> acc
+                                         (update :components + components)
+                                         (update :colors + colors)
+                                         (update :typographies + typographies)))))]
+                  (-> state
+                      (assoc-in [:performance-info :file-size] file-size)
+                      (assoc-in [:performance-info :library-sizes] lib-sizes)
+                      (assoc-in [:performance-info :file-start-time] (js/performance.now)))))))]
```
niwinz: The same here, replace multiple assoc-in with a single update + callback

```clojure
(update state :performance-info
        (fn [info]
          (-> info
              (assoc :file-size file-size)
              (assoc :library-sizes lib-sizes)
              (assoc :file-start-time (js/performance.now)))))
```

frontend/src/app/main/data/event.cljs
```diff
@@ -183,10 +250,118 @@
...
+      ptk/UpdateEvent
+      (update [_ state]
+        (-> state
+            micro-benchmark
+            (assoc-in [:performance-info :app-start-time] (js/performance.now))))
+
+      ptk/WatchEvent
+      (watch [_ _ stream]
+        (->> stream
+             (rx/filter #(= (ptk/type %) :app.main.data.workspace/all-libraries-resolved))
```
niwinz: `(ptk/type? :app.main.data.workspace/all-libraries-resolved)` returns a predicate with simplier syntax

frontend/src/app/main/store.cljs
```diff
@@ -28,6 +29,28 @@
...
 (def ^:dynamic *debug-events* false)
+(def ^:dynamic *debug-events-time* false)
+
+(def current-measure (atom nil))
+
+(defn measure-time-to-render [event]
+  (if @current-measure
+    (swap! current-measure conj event)
+
+    (let [start (js/performance.now)]
```
niwinz, line 40: Maybe replace `js/performance.now` with `(perf/now)` ?

frontend/src/app/main/store.cljs
```diff
@@ -28,6 +29,28 @@
 (def on-event identity)
 
 (def ^:dynamic *debug-events* false)
+(def ^:dynamic *debug-events-time* false)
+
+(def current-measure (atom nil))
+
+(defn measure-time-to-render [event]
+  (if @current-measure
+    (swap! current-measure conj event)
```
niwinz, line 38: Maybe we need to delimit the number of elements. I know, it is a debug utility, but grow a lot i think

frontend/src/app/main/data/workspace.cljs
```diff
@@ -198,13 +202,15 @@
...
-                     (rx/map library-resolved))
+                (rx/concat
+                 (->> (rx/from libraries)
+                      (rx/merge-map
+                       (fn [{:keys [id synced-at]}]
+                         (->> (rp/cmd! :get-file {:id id :features features})
+                              (rx/map #(assoc % :synced-at synced-at :library-of file-id)))))
+                      (rx/mapcat resolve-file)
+                      (rx/map library-resolved))
+                 (rx/of all-libraries-resolved))
```
niwinz: I suggest emit `(ptk/data-event ::all-libraries-loaded)`

frontend/src/app/main/data/event.cljs
```diff
@@ -183,10 +250,118 @@
...
+(defn performance-observer-longtask-stream
+  []
+  (if (and (exists? js/globalThis)
+           (exists? (.-PerformanceObserver js/globalThis)))
+    (rx/create
+     (fn [subs]
+       (let [observer
+             (js/PerformanceObserver.
+              (fn [list]
+                (doseq [entry (.getEntries list)]
```
niwinz: I suggest replace this `doseq` with `run!`

frontend/src/app/main/data/event.cljs
Alotor: ok

frontend/src/app/main/data/event.cljs
Alotor: ok

frontend/src/app/main/data/event.cljs
Alotor: ok

frontend/src/app/main/data/event.cljs
Alotor: ok

frontend/src/app/main/store.cljs
Alotor: no, the current-measure var only stores the events that arrive when the user is not in control. After that its removed. And, as you said, this is only when it's activated. I don't think it will be too big

frontend/src/app/main/data/workspace.cljs
Alotor: ok

frontend/src/app/main/data/event.cljs
Alotor: ok

PR #7458, opened 2025-10-07
:zap: Reduce team shared library load time

niwinz:
### Summary

The main goal of this PR is reduce the time takes to load the summary of all libraries on the team. That happens when user opens the assets modal on the workspace. The problem was that for calculate the summary (counters of components, colors, ...) it needs to open the file and walk it looking and counting samples.

This PR introduces a cache layer for file summary. The cache uses already available REDIS/VALKEY  for storing the pre-calculated summaries. Once a file is edited, the summary is just invalidated from cache (we do not perform any complex logic for avoid recalculation, simple and dummy edition on file will invalidate the cache entry for that file, this is done for simplicity as first step on optimizing that requests).

For make this possible, this PR comes with a lightweight refactor of internal REDIS API, for make it more usable outside the specialized modules that already using it.

The result is that the request to `get-team-shared-files` passes on taking several seconds to resolve, to resolve in 50-100ms  on the best case. The worst case (when no cache is populated) still continues going slow as previously, but as cache is granular (per file), we expect to have the majority files always in cache and only need to recalculate only the most frequently modified files.

### How to test

You need to add `enable-redis-cache` config flag on `backend/scripts/_env` and restart your repl/start-dev process in for activate it.

This PR affects different subsystems and all of them need to be checked for a possible regression:

- MSGBUS: the cursor of other editors is still works as expected (in concurrent editions)
- WORKER: task are correctly scheduled, try to recover password and email should arrive

For now this feature is only available under configuration flag. We intend to test it on our not productive environments to understand the overall system is behaving and how it will affect REDIS/VALKEY in terms of memory usage for have proper recommendations to our on-premise users.

### DIFF
--- a/backend/src/app/rpc/commands/files.clj
+++ b/backend/src/app/rpc/commands/files.clj
@@ -17,6 +17,7 @@
    [app.common.schema :as sm]
    [app.common.schema.desc-js-like :as-alias smdj]
    [app.common.time :as ct]
+   [app.common.transit :as t]
    [app.common.types.components-list :as ctkl]
    [app.common.types.file :as ctf]
    [app.common.uri :as uri]
@@ -28,6 +29,7 @@
    [app.loggers.audit :as-alias audit]
    [app.loggers.webhooks :as-alias webhooks]
    [app.msgbus :as mbus]
+   [app.redis :as rds]
    [app.rpc :as-alias rpc]
    [app.rpc.commands.projects :as projects]
    [app.rpc.commands.teams :as teams]
@@ -208,8 +210,7 @@
   schema:get-file
   [:map {:title "get-file"}
    [:features {:optional true} ::cfeat/features]
-   [:id ::sm/uuid]
-   [:project-id {:optional true} ::sm/uuid]])
+   [:id ::sm/uuid]])
 
 (defn get-minimal-file
   [cfg id & {:as opts}]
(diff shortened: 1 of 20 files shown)


backend/src/app/rpc/commands/files.clj
```diff
@@ -508,23 +509,100 @@
...
-                    {:variant-ids variant-ids :components (conj components component)}
-                    (contains? variant-ids variant-id)
-                    acc
-                    :else
-                    {:variant-ids (conj variant-ids variant-id) :components (conj components component)}))
-                {:variant-ids #{} :components []}
-                components)]
-    {:components components
-     :variant-ids variant-ids}))
+(defn- get-components-with-variants
```
hirunatan, line 517: This function seems to be belonging more to `common.types` or `common.logic` that backend, i think. Perhaps this operation is similar to other things that are used elsewhere, because it's less specific than the other summary operations below.

But really I'm not sure :P

common/src/app/common/schema.cljc
```diff
@@ -853,6 +853,11 @@
 ;;    ::oapi/type "string"
 ;;    ::oapi/format "number"}})
 
+#?(:clj
+   (register!
+    {:type ::atom
+     :pred #(instance? clojure.lang.Atom %)}))
+
```
hirunatan, line 860: Why is this needed?

hirunatan (approved):

common/src/app/common/schema.cljc
niwinz: it is not used right now, but is a general purpose schema for checking an instance of atom, i have used it on one of my iterations on this pr but on the end it becomes unused right now.
I considered to keep it as it enough general purpose so we can use it in a future, the app.common.schema is like a bundled library on the end. But I don't have strong opinion on keeping it or removing it.

backend/src/app/rpc/commands/files.clj
niwinz: Agree, i'll add this to my todo list. I have a pending task of moving several file / library related functions to common types

PR #7459, opened 2025-10-07
:bug: Fix broken layout when multiple fill/strokes with tokens

Xaviju:
### Related Ticket

This PR fixes two bugs:

- [Put color row on top of the border properties](https://tree.taiga.io/project/penpot/task/12056)
- [Properties with multiple values display token in every row](https://tree.taiga.io/project/penpot/task/12216)

### Summary

In the stroke panel, color row must be the first property
In strokes/fill with multiple values and tokens, visualization was wrong

image

### Steps to reproduce 

Create a new stroke with multiple strokes. One of them must be a token

- Ensure that the first property is a color
- Ensure that all the properties are not displayed as if they were tokens

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/inspect/styles/panels/stroke.cljs
+++ b/frontend/src/app/main/ui/inspect/styles/panels/stroke.cljs
@@ -15,7 +15,7 @@
    [app.util.code-gen.style-css :as css]
    [rumext.v2 :as mf]))
 
-(def ^:private properties [:border-style :border-width :border-color])
+(def ^:private properties [:border-color :border-style :border-width])
 
 (defn- stroke->color [shape]
   {:color (:stroke-color shape)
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/ui/inspect/styles/panels/stroke.cljs
```diff
@@ -45,29 +45,38 @@
...
-              [:> color-properties-row* {:key (dm/str "stroke-property-" (:id shape) (:color stroke-type))
+              [:> color-properties-row* {:key (dm/str "stroke-" property "-" idx (:color stroke-type))
                                          :term property-name
                                          :color stroke-type
-                                         :token resolved-token
+                                         :token (when has-token resolved-token)
                                          :format color-space
                                          :copiable true}]
-              [:> properties-row* {:key (dm/str "svg-property-" (d/name property))
+              [:> properties-row* {:key  (dm/str "stroke-" property "-" idx (:color stroke-type))
```
niwinz: Why not just `idx` on the key. It should be unique per `for` but don't need to be unique globally

Xaviju: `idx` is not enough because it can have multiple shapes with the same properties, then `idx` value is repeated.
Anyway, my current solution is unnecessarily overcomplicated. Please check my next commit.

PR #7466, opened 2025-10-08
:tada: Detach storybook from the frontend build process

bameda:
<div align="center">

</div>


docker/images/nginx.storybook.conf
```diff
@@ -0,0 +1,27 @@
+server {
```
niwinz, line 1: I will call this file `nginx.storybook.conf`

bameda: fixed

PR #7472, opened 2025-10-10
:whale: Extend config file at the entrypoint

bameda:
<div align="center">

</div>

### DIFF
--- a/docker/images/files/nginx-entrypoint.sh
+++ b/docker/images/files/nginx-entrypoint.sh
@@ -14,16 +14,26 @@ fi
 #########################################
 
 update_flags() {
+  local config_file="/var/www/app/js/config.js"
+
   if [ -n "$PENPOT_FLAGS" ]; then
     sed -i \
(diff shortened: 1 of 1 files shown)


docker/images/files/nginx-entrypoint.sh
```diff
@@ -14,16 +14,26 @@ fi
...
       -e "s|^//var penpotFlags = .*;|var penpotFlags = \"$PENPOT_FLAGS\";|g" \
-      "$1"
+      "$config_file"
   fi
 }
 
-update_flags /var/www/app/js/config.js
+append_config_extents() {
+  local config_file="/var/www/app/js/config.js"
+  local config_file_extents="/var/www/app/js/config.extents.js"
```
niwinz: My suggestion is calling `extra` instead of `externs`.
Just for convenience on cljs world, externs has special meaning and not using this word here will help avoid confusion. What do you think?

bameda: extra sounds good to me too

PR #7531, opened 2025-10-16
:bug: Fix nested variant in a component doesn't keep inherited overrides

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/12299

### Summary

Inside a component called C1, you have a copy of a variant and override a value on that copy. 
Then, you create a copy of C1 itself and switch the variant on the copy to another one. 
As a result, the override you applied is lost.

### Steps to reproduce 

1. Create a component with a gray rect inside
2. Transform it into a variant
3. Create an empty frame, and make it a component called C1
4. Add a copy of the variant into C1
5. On the copy of the variant in C1, set the rect to blue
6. Make a copy of C1
7. On the copy of C1, switch the variant to the other one

As a result, the rect becomes gray again

### Checklist

### DIFF
--- a/common/src/app/common/logic/variants.cljc
+++ b/common/src/app/common/logic/variants.cljc
@@ -11,7 +11,8 @@
    [app.common.types.container :as ctn]
    [app.common.types.file :as ctf]
    [app.common.types.variant :as ctv]
-   [app.common.uuid :as uuid]))
+   [app.common.uuid :as uuid]
+   [clojure.set :as set]))
 
 (defn generate-add-new-variant
   [changes shape variant-id new-component-id new-shape-id prop-num]
@@ -67,7 +68,6 @@
       [[] {}]
       shapes))))
 
-
 (defn- keep-swapped-item
   "As part of the keep-touched process on a switch, given a child on the original
    copy that was swapped (orig-swapped-child), and its related shape on the new copy
@@ -87,7 +87,6 @@
         prev-swap-slot  (ctk/get-swap-slot orig-swapped-child)
         current-parent  (get objects (:parent-id related-shape-in-new))
         pos             (d/index-of (:shapes current-parent) (:id related-shape-in-new))]
-
 
     (-> (pcb/concat-changes before-changes changes)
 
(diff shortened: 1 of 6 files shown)


common/src/app/common/logic/variants.cljc
```diff
@@ -122,6 +121,44 @@
...
+  "Get the shape-ref of the near main of the shape, recursively repeated until
+   find a shape-ref with parent-id as ancestor"
+  [container libraries shape parent-id]
+  (let [ref-shape             (ctf/find-ref-shape nil container libraries shape
+                                                  :with-context? true)
+
+        ref-shape-container   (when ref-shape (:container (meta ref-shape)))
+        ref-shape-parents-set (when ref-shape
+                                (->> (cfh/get-parents (:objects ref-shape-container) (:id ref-shape))
+                                     (map :id)
```
niwinz: Reuse the d/xf:map-id, with `(into #{} d/xf:map-id)`

common/src/app/common/logic/variants.cljc
```diff
@@ -158,21 +198,28 @@
...
+
         ;; Creates a map to quickly find a child of the orig-ref-shape by its shape-path
+
+
         o-ref-shapes-p-map  (into {} (map (juxt :id :shape-path)) o-ref-shapes-wp)
 
 
         ;; Process each touched children of the original-shape
+
+
```
niwinz: :fire: this empty lines

common/src/app/common/logic/variants.cljc
```diff
@@ -158,21 +198,28 @@
 
         ;; The original-shape is in a copy. For the relation rules, we need the referenced
         ;; shape on the main component
-        orig-ref-shape     (ctf/find-ref-shape nil container libraries original-shape {:with-context? true})
+        orig-ref-shape     (ctf/find-remote-shape container libraries original-shape {:with-context? true})
         orig-ref-objects   (:objects (:container (meta orig-ref-shape)))
 
         ;; Adds a :shape-path attribute to the children of the orig-ref-shape,
         ;; that contains the type of its ancestors and its name
+
```
niwinz: :fire: this empty lines

common/src/app/common/logic/variants.cljc
```diff
@@ -122,6 +121,44 @@
                     (subvec (vec ancestors) 1 (dec num-ancestors)))]
     (some ctk/get-swap-slot ancestors)))
 
+(defn- find-shape-ref-child-of
+  "Get the shape-ref of the near main of the shape, recursively repeated until
+   find a shape-ref with parent-id as ancestor"
+  [container libraries shape parent-id]
```
niwinz, line 128: Would be nice document possible return values, can this return nil per example?

common/src/app/common/logic/variants.cljc
pabloalba: fixed

common/src/app/common/logic/variants.cljc
pabloalba: fixed

common/src/app/common/logic/variants.cljc
pabloalba: fixed

common/src/app/common/logic/variants.cljc
pabloalba: fixed

PR #7544, opened 2025-10-18
 🐛  Switching between layers, assets, and tokens tabs resets scrolling to the top

### LINKED ISSUE #7440: Preserve scroll on switching between right sidebar tabs
### Description

### Steps to reproduce

### Expected behavior

### Actual behavior

### How to test

mwendaB:
## Issue
Closes #7440: Switching between layers, assets, and tokens tabs resets scrolling to the top

## Problem
When users clicked through sidebar tabs (Layers → Assets → Tokens), the scroll position would reset to the top of each panel, forcing them to re-scroll to their previous location. This is poor UX, especially with large token sets or many layers.

## Solution
Implemented persistent scroll position tracking per tab:
1. Store `scrollTop` value for each tab in a local var in the sidebar component
2. Capture current scroll position before switching tabs
3. Restore the stored position when returning to a tab
4. Added scroll event listeners to update stored position as users scroll

## How It Works
- `left-sidebar.cljs`: Manages a map `{:layers 0 :assets 0 :tokens 0}` storing each tab's last scroll position
- When tab changes, we capture the current scroll and save it before unmounting
- When tab remounts, we restore the saved position via `scrollTop` assignment
- Scroll listeners keep positions updated in real-time during active scrolling

### Core Implementation
- **`frontend/src/app/main/ui/workspace/sidebar.cljs`** 
  - Added scroll position var and tab-change wrapper
  - Passes `initial-scroll-top` and `on-scroll-pos-change` props to each toolbox

- **`frontend/src/app/main/ui/workspace/sidebar/layers.cljs`** 
  - Accepts scroll position props
  - Attaches scroll listener via `events/listen` to track position
  - Restores initial scroll on first mount

- **`frontend/src/app/main/ui/workspace/sidebar/assets.cljs`** 
  - Wraps content in scroll container with `data-scroll-container=true`
  - Attaches listener to propagate scroll updates upstream

- **`frontend/src/app/main/ui/workspace/sidebar/tokens/sidebar.cljs`** 
  - Same pattern as assets: scroll container + listener attachment

### Testing & Documentation
- **`frontend/playwright/ui/specs/sidebar-scroll-preserve.spec.js`**
  - Regression test covering Tokens → Assets → Tokens switch
  - Extended test verifying each tab maintains independent scroll position
  - Uses Playwright to scroll, switch tabs, and assert position restored

- **`CHANGES.md`** 
  - Added entry under 2.12.0 "Bugs fixed" section

## Testing
 Code compiles without errors  
  Playwright test file created and syntax validated  
  Manual verification: scroll preservation works across all three tabs  
  Test file uses fixture-based setup to ensure scrollable content  

To run tests locally (requires Yarn):
```bash
cd frontend
yarn install
yarn run test:e2e --grep=sidebar-scroll-preserve
```

## Performance Considerations
Currently, scroll listeners fire on every pixel movement (browser default), which is safe but could be optimized.

### Suggested Future Improvement: Throttle Scroll Listeners
- **Goal**: Reduce scroll event handler frequency from ~50/sec to ~10/sec (every 100ms)
- **Benefit**: Improved performance during rapid scrolling with minimal UX impact
- **Implementation**: Wrap listeners with `throttle-fn 100` utility already used elsewhere in codebase
  ```clojure
  (let [throttled-save (throttle-fn 100 (fn [pos] (on-scroll-pos-change pos)))]
    (events/listen el EventType/SCROLL (fn [_] (throttled-save (.-scrollTop el))))
    (mf/with-effect [] #(rx/dispose! throttled-save)))
  ```
- **Files to update**: `layers.cljs`, `assets.cljs`, `tokens/sidebar.cljs`
- **Note**: Feature works correctly without throttling; this is an optional enhancement for polish

## Backwards Compatibility
 No breaking changes  
 Feature is transparent to existing code  
 Only affects UX of tab switching; no API changes  

## Related Issue
Resolves issue #7440

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@
 - Fix pan cursor not disabling viewport guides [Github #6985](https://github.com/penpot/penpot/issues/6985)
 - Fix viewport resize on locked shapes [Taiga #11974](https://tree.taiga.io/project/penpot/issue/11974)
 - Fix nested variant in a component doesn't keep inherited overrides [Taiga #12299](https://tree.taiga.io/project/penpot/issue/12299)
+- Preserve scroll position when switching between Layers, Assets and Tokens tabs (left sidebar) [Github #7440](https://github.com/penpot/penpot/issues/7440)
 
 ## 2.11.0 (Unreleased)
 
--- a/frontend/playwright/ui/specs/sidebar-scroll-preserve.spec.js
+++ b/frontend/playwright/ui/specs/sidebar-scroll-preserve.spec.js
@@ -1,0 +1,117 @@
+import { test, expect } from "@playwright/test";
+import { WorkspacePage } from "../pages/WorkspacePage";
+
+// Regression test for issue #7440: Scroll position in left sidebar tabs (Layers / Assets / Tokens)
+// should be preserved when switching between them.
+// We use the Tokens tab because token sets list can be long (fixture ensures enough scroll height).
+
+async function setupTokensTab(page) {
+  await WorkspacePage.init(page);
+  const workspace = new WorkspacePage(page);
+  // Mock a file with tokens (reusing existing fixtures from tokens.spec.js)
+  await workspace.setupEmptyFile();
+  await workspace.mockRPC("get-team?id=*", "workspace/get-team-tokens.json");
+  await workspace.mockRPC(/get\-file\?/, "workspace/get-file-tokens.json");
+  await workspace.mockRPC(/get\-file\-fragment\?/, "workspace/get-file-fragment-tokens.json");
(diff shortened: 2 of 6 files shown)


niwinz:
Hey, i'm working on fixing the issue. We have had different PR's for solving this but i'm personally not convinced with the solution. In any case, and with your permission i will reuse the integration tests you used on this PR with proper attribution obviously. For now, closing this PR, thanks.

mwendaB:
@niwinz  No worries at all, absolutely. Feel free to reuse the integration tests, and thanks for the attribution. I completely understand wanting to make sure the solution is solid before merging. Good luck with getting it sorted!

PR #7557, opened 2025-10-20
:paperclip: Notify about failures in releasing and creating docker images

bameda:
<div align="center">

</div>


bameda:
This PR is closed because I've changed the default behavior of the webhooks

PR #7558, opened 2025-10-20
:bug: Fix Restoring a variant from another file makes it overlap

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/12049

### Summary

When you restore a shared variant from the component copy used in another file, it’s restored, but its position overlaps with the rest of the existing variants.

### Steps to reproduce 

See taiga

### Checklist

### DIFF
--- a/frontend/src/app/main/data/workspace/libraries.cljs
+++ b/frontend/src/app/main/data/workspace/libraries.cljs
@@ -698,7 +698,7 @@
   ;; that will be selected and zoomed along the main one
   ;; update-layout? indicates if it should send a :layout/update event
   ;; for the parents of the component
-  [& {:keys [id additional-ids update-layout?] :as options}]
+  [& {:keys [id additional-ids update-layout? retries] :or {retries 0} :as options}]
   (ptk/reify ::go-to-local-component
     ptk/WatchEvent
     (watch [_ state stream]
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/data/workspace/libraries.cljs
```diff
@@ -709,15 +709,18 @@
...
-                   (->> (rx/of (ptk/data-event :layout/update {:ids parent-ids}))
-                       ;; Delay so the navigation can finish
-                        (rx/delay 250))))))
+                                 (keep #(-> (get objects %) :parent-id) ids))]
+
+                (if (and update-layout? (empty? parent-ids) (< retries 3))
+                  ;; The objects are not loaded yet, wait and try again
+                  (->> (rx/of (go-to-local-component (assoc options :retries (inc retries))))
+                       (rx/delay 250))
+                  (rx/concat
```
niwinz, line 718: I think we can remove this concat. Or more correctly, put the last `when`  on its own rx/of. 

Removing concat will work, but it will emit a `nil` into the stream, that in our case is prefectly OK and will do nothing. But the correct way to implement this is:

```clojure
(rx/concat
 (rx/of (dws/select-shapes ids)
        dwz/zoom-to-selected-shape)
 (when update-layout?
   (rx/of (ptk/data-event :layout/update {:ids parent-ids}))))
```

pabloalba: fixed

PR #7564, opened 2025-10-21
:tada: Use toggle for switching boolean variant property names

luisddm:
### Related Ticket

Taiga [#12341](https://tree.taiga.io/project/penpot/us/12341)

### Summary

As a user, I want to be able to visually switch (via a toggle) between two previously defined property values.

This involves creating of a new component for the DS, and using it when the property values for a component copy are either **true/false**, **on/off** or **yes/no**.

### Steps to reproduce 

When selecting a variant copy, check that a **switch** appears instead of a **select** in the design tab for any property whose possible values are any pair of those mentioned previously.

### DIFF
--- a/frontend/src/app/main/ui/ds/_borders.scss
+++ b/frontend/src/app/main/ui/ds/_borders.scss
@@ -7,9 +7,10 @@
 @use "ds/_utils.scss" as *;
 
 // TODO: create actual tokens once we have them from design
-$br-8: px2rem(8);
 $br-4: px2rem(4);
 $br-6: px2rem(6);
+$br-8: px2rem(8);
+$br-12: px2rem(12);
 $br-circle: 50%;
 
 $b-1: px2rem(1);
--- a/frontend/src/app/main/ui/ds/controls/switch.cljs
+++ b/frontend/src/app/main/ui/ds/controls/switch.cljs
@@ -1,0 +1,78 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC
+
+(ns app.main.ui.ds.controls.switch
+  (:require-macros
+   [app.main.style :as stl])
+  (:require
+   [app.common.data :as d]
+   [app.util.dom :as dom]
+   [app.util.keyboard :as kbd]
+   [cuerdas.core :as str]
+   [rumext.v2 :as mf]))
+
+(def ^:private schema:switch
+  [:map
+   [:id {:optional true} :string]
+   [:class {:optional true} :string]
+   [:label {:optional true} [:maybe :string]]
+   [:aria-label {:optional true} [:maybe :string]]
+   [:default-checked {:optional true} [:maybe :boolean]]
+   [:on-change {:optional true} [:maybe fn?]]
+   [:disabled {:optional true} :boolean]])
+
+(mf/defc switch*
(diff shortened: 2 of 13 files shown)


frontend/src/app/main/ui/ds/controls/switch.cljs
```diff
@@ -0,0 +1,88 @@
...
+        is-checked* (mf/use-state (d/nilv default-checked false))
+        is-checked  (deref is-checked*)
+
+        disabled    (d/nilv disabled false)
+
+        has-label   (not (str/blank? label))
+
+        handle-toggle
+        (mf/use-fn
+         (mf/deps on-change is-checked* disabled)
```
niwinz: on deps, only the value should be tracked, so i suggest using `is-checked`

frontend/src/app/main/ui/ds/controls/switch.cljs
```diff
@@ -0,0 +1,88 @@
...
+   [:aria-label {:optional true} [:maybe :string]]
+   [:default-checked {:optional true} :boolean]
+   [:on-change {:optional true} [:maybe fn?]]
+   [:disabled {:optional true} :boolean]])
+
+(mf/defc switch*
+  {::mf/schema schema:switch}
+  [{:keys [id class label aria-label default-checked on-change disabled] :rest props} ref]
+  (let [ref         (or ref (mf/use-ref))
+        id          (or id (mf/use-id))
```
niwinz: React hooks can't be under conditionals, I suggest split this in two steps. Example: 

Image

niwinz: Or directly leave it as is. I mean: leave it nil if it comes as nil

frontend/src/app/main/ui/ds/controls/switch.cljs
```diff
@@ -0,0 +1,88 @@
...
+
+        handle-keydown
+        (mf/use-fn
+         (mf/deps handle-toggle)
+         (fn [event]
+           (when (or (kbd/space? event) (kbd/enter? event))
+             (dom/prevent-default event)
+             (handle-toggle event))))
+
+        props (mf/spread-props props {:ref ref
```
niwinz: I suggest adding a line break here after `props` for preserve the same style of let binding formatting in this group

frontend/src/app/main/ui/ds/_borders.scss
```diff
@@ -7,9 +7,10 @@
 @use "ds/_utils.scss" as *;
 
 // TODO: create actual tokens once we have them from design
-$br-8: px2rem(8);
 $br-4: px2rem(4);
 $br-6: px2rem(6);
+$br-8: px2rem(8);
+$br-12: px2rem(12);
```
EvaMarco, line 13: This variable is only used once in the whole proyect, do we need to create it?

frontend/src/app/main/ui/ds/controls/switch.cljs
EvaMarco: This is not resolved

EvaMarco (changes requested):
Screenshot from 2025-11-06 13-29-37

This is not accesible, we must fix it before merging it.

frontend/src/app/main/ui/ds/_borders.scss
luisddm: I prefer to keep it, because it fits quite well on the size scale.

luisddm:
> Screenshot from 2025-11-06 13-29-37
> 
> This is not accesible, we must fix it before merging it.

It has been enhanced with more contrasting backgrounds.

switch

EvaMarco (approved):

PR #7574, opened 2025-10-22
:recycle: Rename textleafs and inline to textspan

AzazelN28:
### Summary

- [x] Changes TextEditor Inline structs and functions to TextSpan.
- [x] Changes render TextLeaf structs and functions to TextSpan. 

### Checklist

### DIFF
--- a/render-wasm/src/shapes/text.rs
+++ b/render-wasm/src/shapes/text.rs
@@ -98,7 +98,7 @@ impl TextContentSize {
 pub struct TextPositionWithAffinity {
     pub position_with_affinity: PositionWithAffinity,
     pub paragraph: i32,
-    pub leaf: i32,
+    pub span: i32,
     pub offset: i32,
 }
 
@@ -106,13 +106,13 @@ impl TextPositionWithAffinity {
     pub fn new(
         position_with_affinity: PositionWithAffinity,
         paragraph: i32,
-        leaf: i32,
+        span: i32,
         offset: i32,
     ) -> Self {
         Self {
             position_with_affinity,
             paragraph,
-            leaf,
+            span,
             offset,
         }
     }
@@ -289,7 +289,7 @@ impl TextContent {
         let layout_paragraphs = self.layout.paragraphs.iter().flatten();
 
         let mut paragraph_index: i32 = -1;
-        let mut leaf_index: i32 = -1;
+        let mut span_index: i32 = -1;
         for layout_paragraph in layout_paragraphs {
             paragraph_index += 1;
             let start_y = offset_y;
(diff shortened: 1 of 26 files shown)


render-wasm/src/wasm/text.rs
```diff
@@ -204,10 +204,10 @@ impl TryFrom<&Vec<u8>> for RawParagraph {
     fn try_from(bytes: &Vec<u8>) -> Result<Self, Self::Error> {
         let attrs = RawParagraphData::try_from(&bytes[..RAW_PARAGRAPH_DATA_SIZE])?;
         let mut offset = RAW_PARAGRAPH_DATA_SIZE;
-        let mut raw_text_leaves: Vec<RawTextLeaf> = Vec::new();
+        let mut raw_text_leaves: Vec<RawTextSpan> = Vec::new();
 
         for _ in 0..attrs.leaf_count {
-            let text_leaf = RawTextLeaf::try_from(&bytes[offset..(offset + RAW_LEAF_DATA_SIZE)])?;
+            let text_leaf = RawTextSpan::try_from(&bytes[offset..(offset + RAW_LEAF_DATA_SIZE)])?;
```
elenatorro: we should also change `RAW_LEAF_DATA_SIZE` to `RAW_TEXT_SPAN_DATA_SIZE`

render-wasm/src/shapes/text.rs
```diff
@@ -565,16 +565,16 @@ impl Paragraph {
...
     }
 
-    pub fn children(&self) -> &[TextLeaf] {
+    pub fn children(&self) -> &[TextSpan] {
         &self.children
     }
 
     #[allow(dead_code)]
-    fn add_leaf(&mut self, leaf: TextLeaf) {
+    fn add_leaf(&mut self, leaf: TextSpan) {
```
elenatorro: should we also update here `add_leaf` and `leaf`? same in other parts of this file

render-wasm/src/wasm/text.rs
```diff
@@ -234,7 +234,7 @@ impl From<RawParagraph> for shapes::Paragraph {
             let delta = raw_leaf.text_length as usize;
             let text_buffer = &value.text_buffer[offset..offset + delta];
 
-            let mut leaf = shapes::TextLeaf::from(raw_leaf);
+            let mut leaf = shapes::TextSpan::from(raw_leaf);
```
elenatorro: Should we update that too?

elenatorro (commented):
There are some places where we're still using leaf and that we might want to change as well, apart from the comments above:

- `/render/text.rs` 
- `/shapes/text_paths.rs`
- `/state/text_editor.rs`

Do we want to update them on this PR, or are we leaving them for any reason? I think we could take advantage of this if we want to always use TextSpan, let me know what you think!

AzazelN28:
> There are some places where we're still using leaf and that we might want to change as well, apart from the comments above:
> 
> * `/render/text.rs`
> * `/shapes/text_paths.rs`
> * `/state/text_editor.rs`
> 
> Do we want to update them on this PR, or are we leaving them for any reason? I think we could take advantage of this if we want to always use TextSpan, let me know what you think!

My bad, I totally forgot about these, I've fixed them now.

elenatorro (approved):

PR #7619, opened 2025-10-27
:bug: Fix data-itype for text spans

elenatorro:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/11889
https://tree.taiga.io/project/penpot/issue/12417

### Summary

This PR adds two fixes:
- the text span edition
- pasting text within existing text

In both cases, the app was throwing an internal error

### Steps to reproduce 

Text span edition:
- Create a text with multiple words
- Select some words / letters
- Change the color fill

Pasting text:
- Use an existing paragraph
- Copy some word or words from the paragraph
- Paste it directly somewhere in the paragraph

### Checklist

### DIFF
--- a/frontend/text-editor/src/editor/controllers/SelectionController.js
+++ b/frontend/text-editor/src/editor/controllers/SelectionController.js
@@ -525,9 +525,10 @@ export class SelectionController extends EventTarget {
    * @param {number} offset
    */
   collapse(node, offset) {
+    const nodeValue = node?.nodeValue ?? "";
     const nodeOffset =
-      node.nodeType === Node.TEXT_NODE && offset >= node.nodeValue.length
-        ? node.nodeValue.length
+      node.nodeType === Node.TEXT_NODE && offset >= nodeValue.length
+        ? nodeValue.length
         : offset;
 
     return this.setSelection(node, nodeOffset, node, nodeOffset);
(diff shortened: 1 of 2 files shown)


frontend/text-editor/src/editor/controllers/SelectionController.js
```diff
@@ -525,9 +525,10 @@ export class SelectionController extends EventTarget {
    * @param {number} offset
    */
   collapse(node, offset) {
+    const nodeValue = node.nodeValue || "";
```
AzazelN28: I think we could use the optional chaining operator `?.` (and the nullish coalescing `??`) in this case like this:

```javascript
const nodeValue = node?.nodeValue ?? "";
```

frontend/text-editor/src/editor/controllers/SelectionController.js
```diff
@@ -709,7 +710,7 @@ export class SelectionController extends EventTarget {
    * @type {boolean}
    */
   get anchorAtEnd() {
-    return this.anchorOffset === this.anchorNode.nodeValue.length;
+    return this.anchorOffset === (this.anchorNode.nodeValue?.length || 0);
```
AzazelN28: Same here:

```javascript
    return this.anchorOffset === (this.anchorNode.nodeValue?.length ?? 0);
```

frontend/text-editor/src/editor/controllers/SelectionController.js
```diff
@@ -752,7 +753,7 @@ export class SelectionController extends EventTarget {
    * @type {boolean}
    */
   get focusAtEnd() {
-    return this.focusOffset === this.focusNode.nodeValue.length;
+    return this.focusOffset === (this.focusNode.nodeValue?.length || 0);
```
AzazelN28: And here

PR #7658, opened 2025-10-31
:bug: Add test to catch a fixed bug and avoid regressions.

hirunatan:
Add an integration test to reproduce this bug, and catch it if it happens again:
https://tree.taiga.io/project/penpot/issue/12446

The second commit is basically whitespace. If you check "hide whitespace" in the diff it will be clearer.


niwinz:
I merge is as the failing test is not related to this PR.

PR #7700, opened 2025-11-05
:tada: Add management RPC API

niwinz:


backend/dev/user.clj
```diff
@@ -27,6 +27,7 @@
    [app.common.transit :as t]
    [app.common.types.file :as ctf]
    [app.common.uuid :as uuid]
+   [app.common.uri :as u]
```
pabloalba, line 30: This requirement (and most on this file) seems to be unused

backend/src/app/http/middleware.clj
```diff
@@ -240,3 +241,61 @@
...
+              [:bearer token])))
+
+        get-cookie
+        (fn [request]
+          (let [cname (cf/get :auth-token-cookie-name)
+                token (some-> (yreq/get-cookie request cname) :value)]
+            (when-not (str/empty? token)
+              [:cookie token])))
+
+        get-token
```
pabloalba, line 265: I like the some-fn, but maybe it's more clear to use a different name instead of reuse get-token

backend/src/app/http/session.clj
```diff
@@ -183,66 +177,53 @@
...
+
+            response (handler request)]
+
+        (if (renew-session? session)
+          (let [session (update! manager session)]
+            (-> response
+                (assign-auth-token-cookie session)))
+          response))
+
+      (= token-type :bearer)
```
pabloalba, line 216: I think this should be `(or (= token-type :bearer) (= token-type :token))`

backend/src/app/rpc/doc.clj
```diff
@@ -122,8 +131,11 @@
 
         gen-params-doc
         (fn [tsx schema]
+          ;; (prn "gen-params-doc" schema)
```
pabloalba: remove

backend/dev/user.clj
niwinz: This is a user namespace, is used for dev and having this exported is useful

backend/src/app/http/middleware.clj
niwinz: ack

backend/src/app/http/session.clj
niwinz: the session middle-ware only understand cookie and bearer, so the code is correct

backend/src/app/rpc/doc.clj
niwinz: ack

pabloalba (approved):

PR #7701, opened 2025-11-05
:sparkles: Fix copyright year

belen-albeza:
### Related Ticket

No ticket for this, was a quick request in Mattermost.

### Summary

This fixes our copyright year _and_ the misusing of `@` instead of `©` that appears in the internal error page. It also sets the string as the same value as in kaleidos website (`2019-present`).

Screenshot 2025-11-05 at 5 37 36 PM

### Steps to reproduce 

1. Force an internal error (for instance, by blocking a API request).

- Expected behavior: Copyright string should be `Kaleidos © 2019-present`.

### Checklist

- [x] ~~Refactor any modified SCSS files following the refactor guide.~~

### DIFF
--- a/frontend/src/app/main/ui/static.cljs
+++ b/frontend/src/app/main/ui/static.cljs
@@ -64,7 +64,7 @@
       [:div {:class (stl/css :container)} children]]
 
      [:div {:class (stl/css :deco-after2)}
-      [:span (tr "labels.copyright")]
+      [:span (tr "labels.copyright-period")]
       deprecated-icon/logo-error-screen
       [:span (tr "not-found.made-with-love")]]]))
 
--- a/frontend/translations/cs.po
+++ b/frontend/translations/cs.po
@@ -1763,9 +1763,6 @@ msgstr "Můžete pokračovat s účtem Penpot"
 msgid "labels.copy-invitation-link"
 msgstr "Kopírovat odkaz"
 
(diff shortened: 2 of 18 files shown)


niwinz:
why we have this string on translations? i mean this is a clear literal that has no translation and should be used as-is.
What do you think about to delete it from translations and replace it with a constant string?

belen-albeza:
> why we have this string on translations? i mean this is a clear literal that has no translation and should be used as-is. What do you think about to delete it from translations and replace it with a constant string?

That's a good question, and I reckoned that maybe in some languages the order of the year or © symbol could be changed. [I checked for Spanish](https://www.rae.es/dpd/copyright) and it turns out we're using it wrong (it should be © Kaleidos, 2019-presente). Which leads to the point that we should also translate "present" (I didn't realized yesterday) 😅

So what I'll do:

- Update the Spanish with `© Kaleidos, 2019-presente`
- Update the English string with the updated `Kaleidos © 2019-present`
- Replicate the other strings with the English value? I'm not sure how to mark them as "dirty" to be translated. Or should I generate a new translation key?

niwinz:
Ok, we have two options, remove the rest of translations and mark it untranslated for all other languages or create a new key. Both options are ok

niwinz:
Is this ready to merge?

belen-albeza:
It should be!

PR #7706, opened 2025-11-06
:construction: Fix docker images arch during release

yamila-moreno:
This PR tries to fix the bug releasing the docker images for different architectures.


sancfc (approved):
Tested and validated

PR #7716, opened 2025-11-07
:bug: Fix variant toggle does not work for uppercase or mixed case

luisddm:
### Related Ticket

Taiga [#12545](https://tree.taiga.io/project/penpot/task/12545)

### Summary

The variants switch does not appear when the accepted strings (`on`/`off`, `yes`/`no`, `true`/`false`) are uppercase or mixed case.

### Steps to reproduce 

Check that it still works when different cases are used (e.g. `yEs`/`nO`, `oFf`/`oN`).

### DIFF
--- a/common/src/app/common/types/variant.cljc
+++ b/common/src/app/common/types/variant.cljc
@@ -320,7 +320,7 @@
                      {"true" true "false" false}]]
     (when (= (count v) 2)
       (some (fn [b]
-              (when (and (contains? b (first v))
-                         (contains? b (last v)))
+              (when (and (contains? b (-> v first str/trim str/lower))
+                         (contains? b (-> v last str/trim str/lower)))
                 b))
             bool-values))))
--- a/common/src/app/common/types/variant.cljc
+++ b/common/src/app/common/types/variant.cljc
@@ -311,16 +311,22 @@
   [variant]
   (cpn/merge-path-item (:name variant) (str/replace (:variant-name variant) #", " " / ")))
 
+(def ^:private boolean-pairs
+  [["on" "off"]
+   ["yes" "no"]
+   ["true" "false"]])
+
 (defn find-boolean-pair
-  "Given a vector, return the map from 'bool-values' that contains both as keys.
-   Returns nil if none match."
-  [v]
-  (let [bool-values [{"on" true   "off" false}
-                     {"yes" true  "no" false}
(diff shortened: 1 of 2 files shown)


common/src/app/common/types/variant.cljc
```diff
@@ -320,7 +320,7 @@
                      {"true" true "false" false}]]
     (when (= (count v) 2)
       (some (fn [b]
-              (when (and (contains? b (first v))
-                         (contains? b (last v)))
+              (when (and (contains? b (-> v first str/trim str/lower))
+                         (contains? b (-> v last str/trim str/lower)))
```
niwinz: The solution is OK, but I suggest replace the whole impl with the following one:

```clojure
(def ^:private bool-values
  [{"on" true   "off" false}
   {"yes" true  "no" false}
   {"true" true "false" false}])

(defn find-boolean-pair
  "Given a vector, return the map from 'bool-values' that contains both as keys.
   Returns nil if none match."
  [[a b :as v]]
  (when (= (count v) 2)
    (let [a (str/lower (str/trim a))
          b (str/lower (str/trim b))]
      (some (fn [o]
              (when (and (contains? o a)
                         (contains? o b))
                o)
              bool-values)))))
```

The suggested impl reuses the trimmed and lowered values instead of performing that operation on each iteration and also avoid allocating the bool-values on each call defining it at namespace level.

luisddm: I have finally rewritten this function, as it was not working properly when values contained spaces or uppercase characters, following the suggested style. I have also fixed the tests, which were incorrect.

common/src/app/common/types/variant.cljc
```diff
@@ -311,16 +311,22 @@
...
+  (let [a* (-> a str/trim str/lower)
+        b* (-> b str/trim str/lower)]
     (when (= (count v) 2)
-      (some (fn [b]
-              (when (and (contains? b (-> v first str/trim str/lower))
-                         (contains? b (-> v last str/trim str/lower)))
-                b))
-            bool-values))))
+      (some (fn [[t f]]
+              (cond (and (= a* t)
```
niwinz: The `*` suffix is usually used for other purposes, in this case is common to use `'`.

luisddm: Ok, replaced `*` with `'`.

PR #7728, opened 2025-11-11
 💄 Use '100dvh' instead of '100vh' if is possible

e1berd:
# 100vh using problem

When using `100vh`, the content may overflow the browser's UI boundaries (for example, in the Zen browser there is a frame around the app content).

Using `100dvh` instead of `100vh` solves this problem.

### DIFF
--- a/frontend/resources/wasm-playground/clips.html
+++ b/frontend/resources/wasm-playground/clips.html
@@ -11,6 +11,7 @@
       justify-content: center;
       align-items: center;
       height: 100vh;
+      height: 100dvh;
       overflow: hidden;
     }
     canvas {
@@ -85,7 +86,7 @@
       const { duration } = performance.measure('render', 'render:begin', 'render:end');
       // alert(`render time: ${duration.toFixed(2)}ms`);
     });
-    
+
   </script>
 </body>
 </html>
--- a/frontend/resources/wasm-playground/masks.html
+++ b/frontend/resources/wasm-playground/masks.html
@@ -11,6 +11,7 @@
       justify-content: center;
       align-items: center;
       height: 100vh;
+      height: 100dvh;
       overflow: hidden;
     }
     canvas {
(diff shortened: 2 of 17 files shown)


EvaMarco:
Hi,
thank you again for the effort you put into this PR. We understand the motivation behind switching 100vh to 100dvh, especially in relation to the issues seen in Zen Browser.

After reviewing the proposal, we’ve decided not to move forward with it. Here are the main reasons:

🔹 1. The change addresses an issue in a browser we do not officially support

Zen Browser is not part of our supported browser matrix. Introducing global layout adjustments to accommodate a non-supported browser adds significant maintenance overhead, and increases the risk of regressions in environments we do support.

🔹 2. Our UI layout is very complex and sensitive to viewport-related changes

The application uses a large, layered layout with nested grids and several components depending on predictable viewport dimensions. Replacing vh with dvh globally caused multiple regressions in our tests (broken grids, unexpected scrollbars, shifted content).
Given the size of the layout, fixing all of these would require substantial effort with unclear benefit.

🔹 3. Dynamic viewport units carry known performance and UX risks

dvh values can change while browser UI chrome appears/disappears. This can lead to reflows and layout jumps, especially on mobile. Although not all scenarios were exhaustively tested, the risk of introducing visible layout instability is high.

🔹 4. High likelihood of conflicts with our current ClojureScript resize logic

We haven’t fully tested this part yet, but based on how our resize observers and layout calculations are implemented, dvh would likely trigger them more often and with inconsistent values. This could lead to subtle rendering issues throughout the app.

🔹 5. Visual E2E test instability

Dynamic viewport changes caused our visual snapshots to fail or become flaky, which weakens the reliability of our CI pipeline.
 
Screenshot from 2025-11-20 13-43-05
Test result image provided.

Given these factors, applying 100dvh broadly isn’t something we can adopt at this time, especially for an issue tied to a browser we do not support.

We really appreciate the contribution and the intent to help. If in the future we introduce dvh in isolated components where the benefits outweigh the risks, we’ll revisit the idea.

Thanks again for your understanding!

PR #7741, opened 2025-11-12
:tada: Set default button behaviour as type button instead of submit

Xaviju:
### Related Ticket

https://tree.taiga.io/project/penpot/task/12573

### Summary

Since the current default behaviour and use in our app is to use buttons as type buttons, this PR adds this type to the HTML element. Before, since it was unset, the default behaviour was `submit` leading to unexpected behaviours and unnecessary code.

Also reviewed all buttons to explicitly set the type, even if its not required anymore.

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/ds/buttons/button.cljs
+++ b/frontend/src/app/main/ui/ds/buttons/button.cljs
@@ -15,6 +15,7 @@
 (def ^:private schema:button
   [:map
    [:class {:optional true} :string]
+   [:type {:optional true} :enum "button" "submit" "reset"]
    [:icon {:optional true}
     [:and :string [:fn #(contains? icon-list %)]]]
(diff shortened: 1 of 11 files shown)


frontend/src/app/main/ui/ds/buttons/button.cljs
```diff
@@ -15,6 +15,7 @@
 (def ^:private schema:button
   [:map
    [:class {:optional true} :string]
+   [:type {:optional true} :enum "button" "submit" "reset"]
```
EvaMarco: This should be written like this
` [:type {:optional true} [:maybe [:enum "button" "submit" "reset"]]]`
we use `maybe` to avoid errors from storybook if the value is nil.

EvaMarco (approved):

PR #7746, opened 2025-11-13
:whale: Use the secret key both in the backend and the exporter

yamila-moreno:
This PR adds the secret key configuration for backend and exporter; currently the secret key is not necessary in the exporter but it will be soon and this update anticipates the need for the change.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,7 @@
 #### Backend RPC API changes
 
 The backend RPC API URLS are changed from `/api/rpc/command/<name>` to
-`/api/main/methods/<name>` (the previou PATH is preserved for backward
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -53,6 +53,7 @@ provider dinamically.
 
 - Select boards to export as PDF [Taiga #12320](https://tree.taiga.io/project/penpot/issue/12320)
 - Toggle for switching boolean property values [Taiga #12341](https://tree.taiga.io/project/penpot/us/12341)
+- docker-compose.yml: Use the secret key both in the backend and exporter
```
niwinz: I think this should be an entry on the breaking changes section.
Independently of that change, I thinking if it make sense have a specific, docker/deployments related changes?

PR #7752, opened 2025-11-13
:recycle: Copy shorthands using user selected color space

Xaviju:
### Related Ticket

https://tree.taiga.io/project/penpot/task/11876

### Summary
This PR fixes two bugs found in the design review in this task:

- Place the tooltip from the shorthand copy button on top left so it does not cover the property title
- The fill and the border shorthand button always copy the color value in hex, ignoring the selection on the top dropdown (hex, rgba, etc) 

### Steps to reproduce 

- Create a shape with a fill and a border
- Copy the shorthand or the property, by default must copy the fill and border radius as HEX values
- Change the color space in the top selector
- Copy again the shorthand and the property, ensure that the copied values are using the right format.

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/inspect/common/colors.cljs
+++ b/frontend/src/app/main/ui/inspect/common/colors.cljs
@@ -6,8 +6,6 @@
 
 (ns app.main.ui.inspect.common.colors
   (:require
-   [app.common.data.macros :as dm]
-   [app.common.types.color :as cc]
    [app.main.store :as st]
    [okulary.core :as l]
    [rumext.v2 :as mf]))
@@ -27,15 +25,3 @@
   (let [library (mf/with-memo [ref-file]
                   (make-colors-library-ref :files ref-file))]
     (mf/deref library)))
-
-(defn color->color-space->css-format
-  [color color-space]
-  (let [color-value (:color color)
-        color-opacity (or (:opacity color) 1)]
-    (case color-space
-      "hex"  color
-      "rgba" (let [[r g b a] (cc/hex->rgba color-value color-opacity)]
-               (dm/str "rgba(" (cc/format-rgba [r g b a]) ")"))
-      "hsla" (let [[h s l a] (cc/hex->hsla color-value color-opacity)]
-               (dm/str "hsla(" (cc/format-hsla [h s l a]) ")"))
-      (:color color))))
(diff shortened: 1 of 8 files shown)


frontend/src/app/main/ui/inspect/common/colors.cljs
```diff
@@ -25,3 +27,15 @@
   (let [library (mf/with-memo [ref-file]
                   (make-colors-library-ref :files ref-file))]
     (mf/deref library)))
+
+(defn color->color-space->css-format
```
niwinz: I suggest simplify the name, it suggests it passes color to color-space y later to css-format and i think is not the case. My suggestion is `format-for-css`. We are already on color namespace so context is clear, and we going to format the color using css rules, using one of the provided color space.

frontend/translations/en.po
```diff
@@ -1999,6 +1999,14 @@ msgstr "Resolved value:"
 msgid "inspect.tabs.switcher.label"
 msgstr "Layer info"
 
+#: src/app/main/ui/inspect/right_sidebar.cljs:165
+msgid "inspect.tabs.switcher.color-space.label"
```
niwinz, line 2003: MINOR

This translation string has tab and switcher as different segments, but they represent the same semantical section, so is using two segments for represent the same location.

I think it should be `inspect.tabs-switcher.color-space-label`.

In general, translation strings are composed of `<first-level-location>.<second-level-location>.<identifier>` and in very rare case it can have the `<third-level-location>` before the identifier.

frontend/src/app/main/ui/inspect/common/colors.cljs
Xaviju: Removed this fn and used a similar utility in `util.color.cljs` named `color->format->background`

Xaviju: Also fixed a :bug: found on format change

frontend/translations/en.po
Xaviju: I could change this specific occurrence, but the inspect tab was already using many concatenated strings, so I followed the same structure. E.g. `"inspect.tabs.code.selected.text"`

If we want to follow this specific structure, it might make sense to create a refactor task? 
This structure that you suggest affects many many translations just in the inspect tab.

Anyway, I'll try to keep this in mind for future features

Xaviju: Translations IDs refactored in a new branch https://github.com/penpot/penpot/pull/7765

PR #7753, opened 2025-11-14
:sparkles: Add parameter to `openPage` to toggle new window behaviour

mdbenito:
### Summary

This PR extends the `:openPage` command in `page-proxy` with an additional parameter to be exposed in the plugins API. This parameter allows toggling the current default behaviour of always opening new `Pages` in separate browser tabs/windows.

I believe this is necessary because it's AFAICT otherwise impossible for plugins to navigate to shapes outside the current page. As it is right now, the plugin is unloaded in the new browser tab, making any action subsequent to `openPage` impossible.

Here's [the accompanying PR](https://github.com/penpot/penpot-plugins/pull/219) for the `penpot-plugins` repo.

### Open questions:

1. In order not to break expectations for API users, I've handled `undefined` in the second argument to have the same behaviour as previously, but I'm not sure how likely this is to happen once the new interfaces are published in npm.

### Checklist

### DIFF
--- a/frontend/src/app/plugins/api.cljs
+++ b/frontend/src/app/plugins/api.cljs
@@ -512,6 +512,7 @@
             id (uuid/next)]
         (st/emit! (dw/create-page {:page-id id :file-id file-id}))
         (page/page-proxy plugin-id file-id id)))
+
     :openPage
     (fn [page new-window]
       (let [id (obj/get page "$id")
--- a/frontend/src/app/plugins/api.cljs
+++ b/frontend/src/app/plugins/api.cljs
@@ -513,9 +513,10 @@
         (st/emit! (dw/create-page {:page-id id :file-id file-id}))
         (page/page-proxy plugin-id file-id id)))
     :openPage
-    (fn [page]
-      (let [id (obj/get page "$id")]
-        (st/emit! (dcm/go-to-workspace :page-id id ::rt/new-window true))))
+    (fn [page new-window]
+      (let [id (obj/get page "$id")
+            new-window (if (boolean? new-window) new-window true)]
+        (st/emit! (dcm/go-to-workspace :page-id id ::rt/new-window new-window))))
 
     :alignHorizontal
     (fn [shapes direction]
--- a/frontend/src/app/plugins/page.cljs
+++ b/frontend/src/app/plugins/page.cljs
@@ -263,13 +263,14 @@
           (apply array (keys (dm/get-in page [:plugin-data (keyword "shared" namespace)]))))))
 
     :openPage
-    (fn []
+    (fn [new-window]
       (cond
         (not (r/check-permission plugin-id "content:read"))
         (u/display-not-valid :openPage "Plugin doesn't have 'content:read' permission")
 
         :else
-        (st/emit! (dcm/go-to-workspace :page-id id ::rt/new-window true))))
+        (let [new-window (if (boolean? new-window) new-window true)]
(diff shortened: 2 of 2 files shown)


frontend/src/app/plugins/api.cljs
```diff
@@ -512,10 +512,12 @@
             id (uuid/next)]
         (st/emit! (dw/create-page {:page-id id :file-id file-id}))
         (page/page-proxy plugin-id file-id id)))
+
     :openPage
-    (fn [page]
-      (let [id (obj/get page "$id")]
-        (st/emit! (dcm/go-to-workspace :page-id id ::rt/new-window true))))
+    (fn [page new-window]
```
niwinz, line 517: I suggest using an `options` instread new-window directly.

niwinz: but is a suggestion, i don't have strong opinion on this, so it is open to debate

mdbenito: Do you mean to receive a generic object from js and destructure it here? so users do:
```
openPage(page, { newWindow: false })
```
But how likely is it that we will need more options to open a page in the future?

niwinz: is not about how likely, is just about pattern, if we make the api uniform it is better.
And internally I have said the same for other methods that are coming to the plugins api

mdbenito: I don't know, to me it would feel clunky as a user of that method. Also, it would require declaring a new `OpenPageOptions` interface, which would kind of defeat the purpose of uniformity, since every method would have its own interface. Also, I would have to inspect it in order to know what parameters are available, same as with a plain parameter, just hidden behind one level of indirection.

I think it's perfectly normal that different methods have different signatures. It's expected and there is no drawback I can think of. Any IDE pops up inline hints and documentation for quick reference. But if this is a hard requirement, I can change my other PR there and this one here, perhaps next week after we're done with our demo.

niwinz: Is a very common pattern and we will go to apply it around the API. 

Uniformity is about the method and its optional options. The options have no ordering, positional arguments have. And positional arguments are defined in the order of the need of the parameter and not the logical and correct order for them. So maintaining backward compatibility will consist on maintain the order of arguments independently if they are in the logical/semantical order or not.

And is not the same for an internal method which you can refactor once we need to add an parameter, is a public method that you need to maintain and extend throughout its lifetime. The rules on both cases are very different.

All this said, probably this will not affect this concrete case, but have all public methods defined in a way that, optional parameters should not have any order of application also has very big value on uniformity. Because the developer using the API will always know that options are going to be an object.

And only in a very rare case when we need performance and reduce allocation, we can thinking o break the rule and only use positional arguments.

mdbenito: Ok. I'm not sure I agree entirely, but maybe it's just a matter of personal taste. I can implement what you suggest.

What should I do about the type of the options object in the api? To me, it would naturally be a `OpenPageOptions` Interface, but, as I mentioned above, this kind of defeats the purpose: we now have something different to maintain throughout the lifetime of the API, namely this new type. And we will have one per API method. Is this fine?

niwinz: Yes, on a typed language this is how it works and has nothing wrong.
In any case, i think i have a middle solution, so i go to merge this as-is as you already defined. Because the it is easy to extend, the next time, when we have to add more options, instead of continuing adding it as positional argument, we will need to convert that argument in an enum, supporting `options` or `new-window`.

PR #7740, opened 2025-11-12
:recycle: Refactor clipboard 

AzazelN28:
### Summary

Creates a new namespace only for clipboard utilities and changes how pasting works. Now the clipboard works only with Blobs/Files and not strings.

### Checklist

- [x] Choose the correct target branch; use `develop` by default.
- [x] Provide a brief summary of the changes introduced.
- [ ] Check CI passes successfully.

PR #7763, opened 2025-11-17
:recycle: Refactor Clipboard API

niwinz:
Replaces the https://github.com/penpot/penpot/pull/7740 PR with the same commit but with additional fixes to the integration tests.


belen-albeza (changes requested):
The test code commits look OK. However, about the clipboard commit, I'm getting a crash trying to paste text:

```
Error: Invalid inert element
  at resetInertElement (libs.js?ts=1763541209235:219156:30)
  at Object.createRootFromHTML [as create_root_from_html] (libs.js?ts=1763541209235:222215:5)
  at Object.eval [as potok$v2$core$WatchEvent$watch$arity$3] (clipboard.cljs:952:22)
  at Object.potok$v2$core$watch [as watch] (core.cljs:31:11)
```

belen-albeza (approved):
We troubleshooted and the error is because of having the `enable-frontend-binary-fills` flag in config

PR #7830, opened 2025-11-26
:sparkles: Add retry and validation to nitrate module

pabloalba:
Add cache and validation to nitrate http calls

### DIFF
--- a/backend/src/app/main.clj
+++ b/backend/src/app/main.clj
@@ -323,7 +323,7 @@
    {::http.client/client (ig/ref ::http.client/client)
     ::db/pool            (ig/ref ::db/pool)
     ::rds/pool           (ig/ref ::rds/pool)
-    :app.nitrate/instance (ig/ref :app.nitrate/instance)
+    :app.nitrate/client (ig/ref :app.nitrate/client)
     ::wrk/executor       (ig/ref ::wrk/netty-executor)
     ::session/manager    (ig/ref ::session/manager)
     ::ldap/provider      (ig/ref ::ldap/provider)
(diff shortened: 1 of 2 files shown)


backend/src/app/main.clj
```diff
@@ -323,7 +323,7 @@
    {::http.client/client (ig/ref ::http.client/client)
     ::db/pool            (ig/ref ::db/pool)
     ::rds/pool           (ig/ref ::rds/pool)
-    :app.nitrate/instance (ig/ref :app.nitrate/instance)
+    :app.nitrate-client/instance (ig/ref :app.nitrate-client/instance)
```
niwinz: I think a better name for this should be `:app.nitrate/client`

backend/src/app/nitrate_client.clj
```diff
@@ -0,0 +1,152 @@
...
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; HELPERS
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn- create-cache
+  [{:keys [::wrk/executor]}]
+  (cache/create :executor executor
+                :keepalive "5m"
+                :max-size 1000))
```
niwinz: We should discuss about this, this cache works OK on single instance but it will not work well when we run multiple instances.

Issue #7675, opened 2025-11-03
bug: Mask issues when I update a component inside


dfelinto:
### Steps To Reproduce

1. Create a component with 2 variants (e.g., Icon with shape options: corner and circle)
2. Create an instance of this component
3. Add a rectangle above it, e.g., in blue
4. Select the rectangle and the instanced component and create a mask

At this point you should have:

<img width="613" height="278" alt="Image" src="https://github.com/user-attachments/assets/56a75950-2640-498a-80b3-2f656d8bbc4a" />

[mask_bug.penpot.zip](https://github.com/user-attachments/files/23305104/mask_bug.penpot.zip)

5. Now select the icon inside the mask and change to a different variant.
6. The group is no longer a mask, it is now a regular group :(

### Expected behavior

I expected the group to still behave as a mask.

### Actual behavior

The group changes from mask to regular group.

### Screenshots or video

https://github.com/user-attachments/assets/3b50fc85-8959-4eac-88f5-326febba5291

### Environment (please complete the following information)

https://design.penpot.app/

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

We are using tokens for our design system, so we have dark and light themes. This includes the icons as well.

Some components use the icons, but need them to be of a special color, not the usual white/dark (we do this for selected items, using the "hover" token). This works well for text, but for icon components this doesn't work.

The workaround we are exploring is to use the icon as a mask, and set the color of the rectangle to be the one affected by the tokens. This would work well if it wasn't for this bug.

Another alternative would be to have a duplicate set of the icon components for each of those special states. This is not practical though.

Thanks!

madalenapmelo-kp:
Hi @dfelinto,

Thanks for reporting this! We've added this to our backlog on Taiga so that we can look further into it, you can find the details here: https://tree.taiga.io/project/penpot/issue/12505

dfelinto:
I just noticed this happens even without variants. Simply swapping component to another component from within a mask causes this issue already.

dfelinto:
And I found a workaround :)

If I move the component inside a group it works fine. I can swap the component without losing the mask.

<img width="304" height="143" alt="Image" src="https://github.com/user-attachments/assets/70befc8a-cde5-488d-a484-a1ad692110ce" />

dfelinto:
@madalenapmelo-kp should I close this given that the fix is in `develop` already? Or the policy is to wait until the fix gets merged into main?

[...]

PR #7862, opened 2025-11-30
:bug: Fix mask issues with component swap #7675

dfelinto:
### Related Ticket

https://github.com/penpot/penpot/issues/7675
https://tree.taiga.io/project/penpot/issue/12505

### Summary

The logic to swap a component would delete the swapped out component first before bringing in the new one.

In the process of doing so, the sanitization code would unmask the group, now orphan of its mask shape component, when it was the first element of the group.

The fix  was to pass an optional argument to the generate-delete-shapes function to ignore mask in special cases like this.

### Steps to reproduce 

See the test file in #7675.

### Checklist

### DIFF
--- a/common/src/app/common/logic/shapes.cljc
+++ b/common/src/app/common/logic/shapes.cljc
@@ -123,8 +123,10 @@
                         ;; ignore-children-fn is used to ignore some descendants
                         ;; on the deletion process. It should receive a shape and
                         ;; return a boolean
-                        ignore-children-fn]
-                 :or {ignore-children-fn (constantly false)}}]
+                        ignore-children-fn
+                        ignore-mask]
+                 :or {ignore-children-fn (constantly false)
+                      ignore-mask false}}]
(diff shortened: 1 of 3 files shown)


common/src/app/common/logic/shapes.cljc
```diff
@@ -162,18 +164,17 @@
          lookup  (d/getf objects)
 
          groups-to-unmask
-         (reduce (fn [group-ids id]
-                  ;; When the shape to delete is the mask of a masked group,
```
pabloalba, line 166: We don't want to lose this comment

pabloalba (changes requested):

common/src/app/common/logic/shapes.cljc
dfelinto: Yes of course. It was an oversight on my end. I brought it back.

pabloalba (approved):

PR #7877, opened 2025-12-02
:bug: Fix problem with boolean shapes updates

Alotor:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/12799

### Checklist

### DIFF
--- a/render-wasm/src/main.rs
+++ b/render-wasm/src/main.rs
@@ -655,6 +655,26 @@ pub extern "C" fn set_modifiers() {
     });
 }
 
+#[no_mangle]
+pub extern "C" fn start_temp_objects() {
+    unsafe {
+        #[allow(static_mut_refs)]
+        let mut state = STATE.take().expect("Got an invalid state pointer");
+        state = Box::new(state.start_temp_objects());
+        STATE = Some(state);
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn end_temp_objects() {
+    unsafe {
+        #[allow(static_mut_refs)]
+        let mut state = STATE.take().expect("Got an invalid state pointer");
+        state = Box::new(state.end_temp_objects());
+        STATE = Some(state);
+    }
+}
+
 fn main() {
     #[cfg(target_arch = "wasm32")]
     init_gl!();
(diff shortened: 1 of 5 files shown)


render-wasm/src/state.rs
```diff
@@ -24,6 +24,7 @@ pub(crate) struct State<'a> {
     pub current_id: Option<Uuid>,
     pub current_browser: u8,
     pub shapes: ShapesPool<'a>,
+    pub saved_shapes: Option<ShapesPool<'a>>,
```
belen-albeza, line 27: Could we instead create a separate, new state type to store this? And then having a flag in `calculate_bool` to retrieve one set of shapes or other.

render-wasm/src/state.rs
```diff
@@ -34,9 +35,31 @@ impl<'a> State<'a> {
             current_id: None,
             current_browser: 0,
             shapes: ShapesPool::new(),
+            saved_shapes: None,
         }
     }
 
+    // Creates a new temporary shapes pool.
+    // Will panic if a previous temporary pool exists.
+    pub fn start_temp_objects(mut self) -> Self {
```
belen-albeza, line 45: Q: Why not &mut self?

render-wasm/src/main.rs
```diff
@@ -655,6 +655,26 @@ pub extern "C" fn set_modifiers() {
     });
 }
 
+#[no_mangle]
+pub extern "C" fn start_temp_objects() {
+    unsafe {
+        #[allow(static_mut_refs)]
+        let mut state = STATE.take().expect("Got an invalid state pointer");
+        state = Box::new(state.start_temp_objects());
+        STATE = Some(state);
```
belen-albeza, line 664: I would prefer if we mutated the current state instead of recreating the `Box`

belen-albeza (changes requested):

render-wasm/src/state.rs
Alotor: You cannot borrow because you are changing the internal ownership in order to do that you need to change the ownership otherwise rust won't let you do this.

render-wasm/src/main.rs
Alotor: Again, you cannot do that because we need to pass the ownership.

render-wasm/src/state.rs
Alotor: I was thinking about this solution but the issue then is that you have to change every single call to check if you're in one case or the other, It's easier to do it this way and then you don't need to change anything else.

belen-albeza (approved):

PR #7887, opened 2025-12-03
:bug: Fix crash in token grid view due to tooltip validation

dfelinto:
### Related Ticket

Unreported.

### Summary

The color tokens in grid view have a tooltip which is a map. This is done so the frontend can render:

```
Name: foo
Resolved value: #000000
```

However the validation scheme for tooltips was only accepting functions
and strings.

### Steps to reproduce 

* Create a color token
* Create a shape, add a fill
* Pick a color, chose the Token options
* Click on the Grid View

Crash: `{:hint "invalid props on component tooltip*\n\n  -> 'content' should be a string\n"}`

### Checklist

- [x] Choose the correct target branch; use `develop` by default. (*)
- [x] Update the `CHANGES.md` file, referencing the related GitHub issue, if applicable. (*)

(*) This should target 2.12, so I'm targetting `staging` instead.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -61,6 +61,7 @@ example. It's still usable as before, we just removed the example.
 ### :heart: Community contributions (Thank you!)
 
 - Ensure consistent snap behavior across all zoom levels [Github #7774](https://github.com/penpot/penpot/pull/7774) by [@Tokytome](https://github.com/Tokytome)
+- Fix crash in token grid view due to tooltip validation (by @dfelinto) [Github #7887](https://github.com/penpot/penpot/pull/7887)
 
 ### :sparkles: New features & Enhancements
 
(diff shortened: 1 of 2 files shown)


dfelinto:
I don't know what is planned for 2.12.0 and what is planned for 2.13.0. All in all I think this fix should land together with the release that plans to expose the tokens in the color picker.

If this is the 2.12.0 as I think it is, then I need to update the CHANGES.md. That said I don't care if this doesn't even get mentioned. I just saw Andy's commit with the new images and videos for 2.12.0 and decided to try the feature. Which is how I ran into this issue.

dfelinto:
I just tested the 2.12.0-RC1 branch and the bug is there. So I updated the CHANGES to be on top of 2.12.0. The PR still targets develop though. Not sure if it should be a different branch.

dfelinto:
Updated PR to target staging.

niwinz:
Thanks, it should target staging.

dfelinto:
Thanks for the quick review :)

PR #7918, opened 2025-12-08
:lipstick: Set import Tokens default option to be Single JSON value

dfelinto:
This patches makes the default Tokens importing option to match the current default Tokens exporting option (single JSON value). This way it is more obvious and quick to export the tokens from a file and import in new one,

---

While testing our design system we are often re-exporting and re-importing the Tokens to the files using the design system components.

I'm aware that this may be addressed in the future so the Tokens are brought in together with the library. Meanwhile (and even in the future) I think it is sensible to have a symmetry between the export and import defeault options.

### Steps to reproduce

* Add a token to your document.
* Export it (by default it will export a single JSON file).
* Now try to Import it, by default Penpot tries to import a .zip.

### Checklist

- [ ] Update the `CHANGES.md` file, referencing the related GitHub issue, if applicable. (I don't think this is worth mentioning on the CHANGE log to be honest :)

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/import/modal.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/import/modal.cljs
@@ -268,8 +268,8 @@
                    :on-click modal/hide!}
        (tr "labels.cancel")]
       [:> import-type-dropdown*
-       {:options [{:label (tr "workspace.tokens.import-menu-zip-option") :value :zip}
-                  {:label (tr "workspace.tokens.import-menu-json-option") :value :file}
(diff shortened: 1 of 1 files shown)


dfelinto:
Full disclosure, I couldn’t test this patch yet. I believe the generated front end code is cached somewhere. Wiping the docker clean doesn’t solve it.

I ran into the same issue when I did the export to webp patch (the new entry would always show last in the list). So I’m assuming the PR is ok.

I would appreciate some help with how to make sure the frontend code is the “latest”.

niwinz:
Thanks :heart:

PR #7969, opened 2025-12-11
:bug: Fix issue on reading rlimit config

niwinz:
### Summary

Fixes issue on rlimit config reader.
This PR prepares for partial activation of rate limit.

Nothing is affecting users  because this is disabled by default.
Does not requires user tests, only code review.

### DIFF
--- a/backend/src/app/rpc/rlimit/bucket.lua
+++ b/backend/src/app/rpc/rlimit/bucket.lua
@@ -25,9 +25,9 @@ local allowed = filled >= requested
 local newTokens = filled
 if allowed then
   newTokens = filled - requested
+  redis.call("hset", tokensKey, "tokens", newTokens, "timestamp", timestamp)
 end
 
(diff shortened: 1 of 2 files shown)


backend/src/app/rpc/rlimit/bucket.lua
```diff
@@ -25,9 +25,9 @@ local allowed = filled >= requested
 local newTokens = filled
 if allowed then
   newTokens = filled - requested
+  redis.call("hset", tokensKey, "tokens", newTokens, "timestamp", timestamp)
 end
```
superalex, line 30: Shouldn't the timestamp be actualized always? (even when the request is not allowed)

niwinz: I think NO, because if we update it always, on repeated requests when no available tokens, will never leave a window to add new tokens.

superalex: great!

PR #8010, opened 2025-12-30
:tada: Migrate to PNPM

niwinz:
### Summary

Migrate all modules to use PNPM

### How to test

This is a transversal task and affects all tooling. So would be nice to test several local development workflows and verify that there is no regression.


luisddm:
The local development environment works fine. I have tested some workflows and everything seems to work as expected.
However, I found some remains of yarn. I'm not sure if they were intended to be addressed here or in a future PR.
- The `run-ci.sh` file in the root directoty is still using yarn.
- There is still a `yarn.lock` file in the root directory.
- In the `.gitignore` files there are still references to yarn.
- The `docs` module has not been migrated yet.
- There are some references to yarn commands in the documentation relating to migrated modules.
- In some READMEs there are still some documented yarn commands.

niwinz:
They will be addressed in this PR, thanks.

luisddm (approved):

niwinz:
I think the best moment for merge this, is on the next feature freeze.

PR #8030, opened 2026-01-07
:bug: Fix typos on download modal

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/12865

### Summary

See taiga

### Steps to reproduce 

See taiga

### Checklist

### DIFF
--- a/frontend/translations/en.po
+++ b/frontend/translations/en.po
@@ -570,7 +570,7 @@ msgstr "* Might include components, graphics, colors and/or typographies."
 #: src/app/main/ui/exports/files.cljs:155
 msgid "dashboard.export.explain"
 msgstr ""
-"One or more files that you want to export are using shared libraries. What "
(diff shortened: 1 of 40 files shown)


frontend/translations/en.po
```diff
@@ -570,7 +570,7 @@ msgstr "* Might include components, graphics, colors and/or typographies."
 #: src/app/main/ui/exports/files.cljs:155
 msgid "dashboard.export.explain"
```
niwinz, line 571: I think this also should be renamed, because all translations will still use export as base word

niwinz: Since we are renaming: i suggest improve the translation key to something like:

`download-file-modal.title`
`download-file-modal.subtitle`

As this modal is not directly related to dashboard i think we should just change the keys.

PR #8051, opened 2026-01-10
:bug: Fix import Tokens default option

dfelinto:
This was intended to be changed on 13fcf3a9bb25. However only the menu order changed, not the default option.

### Summary

I introduced this change on #7918 however I had some issues testing it (as mentioned [here](https://github.com/penpot/penpot/pull/7918#issuecomment-3628364732)). As it turned out the issue wasn't caching (only), but that I missed a bit.

### Steps to reproduce

* Go to Tokens > Tools > Import
* See that `Import Single Json File` is the default option.

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/import/modal.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/import/modal.cljs
@@ -273,4 +273,4 @@
                   {:label (tr "workspace.tokens.import-menu-folder-option") :value :folder}]
         :on-click handle-import-action
         :text-render render-button-text
-        :default :zip}]]]))
+        :default :file}]]]))


niwinz:
About the testing issue: Are you using docker for mac? we have had similar (and very strange issues) with file system reflecting a different reality. In any case, on dropping containers, ensure deleting all related volumes, the volumes are not dropped automatically.

dfelinto:
> About the testing issue: Are you using docker for mac? we have had similar (and very strange issues) with file system reflecting a different reality. In any case, on dropping containers, ensure deleting all related volumes, the volumes are not dropped automatically.

Yes I am indeed. Good to know it is not only me.

I did learn to wipe the entire docker (volumes and all indeed) every now and then. And to make sure the watcher recompiles. It is a bit annoying but get the job done.

The most annoying bit for me is that I had to give up on setup the docker env machine more to my liking (my tmux settings, my alias, ...) given how often I have to start from scratch.

dfelinto:
Funny enough I just run into the issue of the stale shadow-cljs again:

image

In this case, just killing the process and launching dev-env again (and reloading the page) seems to have solved it.

PR #8059, opened 2026-01-12
:paperclip: Create an issue template for reporting bugs on the new render engine

madalenapmelo-kp:
This template will be part of the open beta test for the new render, so that users can report issues for us to review.


claragvinola (approved):
All perfect! thank you!

PR #8114, opened 2026-01-18
fix: Add error handler to token form validation to prevent crash on naming conflict

### LINKED ISSUE #8110: bug: Penpot crash when setting some name in Design tokens
### Steps To Reproduce

1. Import Design Tokens
2. Go to "Tokens"
3. Go to  Sets/Dark/Accent Color
4. Add new color:
Name: accent-color
Value: {accent-color.blue.dark}
5. Click "Save"
6. See "Internal Error"

[tokens.json](https://github.com/user-attachments/files/24674448/tokens.json)

### Expected behavior

Create new color

### Actual behavior

Internal Error
Something bad happened.

You can retry the operation or contact support to report the error.

[report.txt](https://github.com/user-attachments/files/24674388/report.txt)

- Fedora Silverblue 43
- Flatpak version v0.23.0

### Additional context

Probably the issue is with the name "accent-color" as there are sub variants define earlier.
Maybe context what I want to do. 
2 Themes: Light and Dark
9 Accent Colors: Blue, Teal, etc. 

User select accent-color and if theme "Dark" and Accent Color "Blue" then use accent-color.blue.dark
So I need value "accent-color" and I don't know how different I should name it

GlobalStar117:
## Summary

This PR fixes issue #8110 where creating a design token with a name that already exists causes the application to crash.

Fixes #8110

## Changes

- Added error handler to the `rx/subs!` call in `on-submit` function in `generic_form.cljs`
- The handler catches validation errors and displays human-readable error messages in the form's `extra-errors` instead of letting the exception crash the application
- Uses existing `wte/humanize-errors` utility to format error messages

## Testing

1. Open Penpot and go to Design Tokens
2. Create a token named "test"
3. Try to create another token with the same name "test"
4. **Before**: Application crashes with unhandled exception
5. **After**: Error message is displayed in the form field

---
This is a Gittensor contribution
gittensor::146701565

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/management/forms/generic_form.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/management/forms/generic_form.cljs
@@ -18,6 +18,7 @@
    [app.main.data.workspace.tokens.application :as dwta]
    [app.main.data.workspace.tokens.library-edit :as dwtl]
    [app.main.data.workspace.tokens.propagation :as dwtp]
+   [app.main.data.workspace.tokens.errors :as wte]
    [app.main.data.workspace.tokens.remapping :as remap]
    [app.main.refs :as refs]
    [app.main.store :as st]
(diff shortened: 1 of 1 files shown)


niwinz:
Thanks for contributing.

The code has syntax error. Once syntax error is fixed, the issue stops raising exception, but no error is show on the value fied. In any case we think that the real issue is a bit deeper, because from my point of view, the error should be shown before save. So we researching on it before merge this change.

niwinz:
Merged in other PR

PR #8216, opened 2026-01-27
:bug: Fix error when creating a token with an invalid name

EvaMarco:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/13219

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/workspace/tokens/management/forms/controls/color_input.cljs
+++ b/frontend/src/app/main/ui/workspace/tokens/management/forms/controls/color_input.cljs
@@ -11,6 +11,7 @@
    [app.common.data :as d]
    [app.common.data.macros :as dm]
    [app.common.types.color :as cl]
+   [app.common.types.token :as cto]
    [app.common.types.tokens-lib :as ctob]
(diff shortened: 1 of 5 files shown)


frontend/src/app/main/ui/workspace/tokens/management/forms/controls/color_input.cljs
```diff
@@ -51,12 +53,15 @@
 ;; Both variants provide identical color-picker and text-input behavior, but
 ;; differ in how they persist the value within the form’s nested structure.
 
-
 (defn- resolve-value
   [tokens prev-token token-name value]
-  (let [token
+  (let [valid-token-name?
+        (and (string? token-name)
+             (m/validate cto/token-name-ref token-name))
```
niwinz: We should not use malli directly here, we already have `app.common.schema` namespace that exposes that validation functions.

Also, I think we should extract the regex and use it directly instead of using schema validation there.

EvaMarco: ok, that was the other option top extract the regex

PR #8243, opened 2026-01-29
:sparkles: Integrate MCP repository

niwinz:


opcode81:
@niwinz to maintain some of the development history, please make me the author of commit 029bd7e ("Integrate mcp repository")

    --author "Dominik Jain <dominik.jain@oraios-ai.de>"
    
EDIT: Note that it would also be possible to [keep the entire history](https://github.com/opcode81/git-migrate-subdir).
In my fork, I created a branch and PR that integrates the current state of the penpot-mcp repo into /mcp, fully maintaining the history: https://github.com/opcode81/penpot/pull/1

niwinz:
I will keep the author,, but not the history. I mean, the history will be continued on the original repo which i will reference on the commit.  The history of the original repo does not follows the same commit naming and will make the history with mixed formats. I'm also not very fan of commits like "Split history" because this is very artificial. On the end we don't go to delete history on the original repo and it will reamain archived.

opcode81:
> I'm also not very fan of commits like "Split history" because this is very artificial

These commits are generated by [git-split](https://github.com/opcode81/git-split), which applies a series of git operations allowing to split the contents of a versioned file into two, while maintaining the history of the original file in both copies.
The key here is not the commits but the history they result in.
As you can probably tell, I care about the git history being complete. ;)

niwinz:
> > I'm also not very fan of commits like "Split history" because this is very artificial
> 
> These commits are generated by [git-split](https://github.com/opcode81/git-split), which applies a series of git operations allowing to split the contents of a versioned file into two, while maintaining the history of the original file in both copies. The key here is not the commits but the history they result in. As you can probably tell, I care about the git history being complete. ;)

Yeah, and it remains complete and untouched on the original repository.
I think that keeping the author, correctly reference the import commit and maintaining the original repo untouched, is more than enough to proceed.

PR #8261, opened 2026-02-03
:sparkles: Add several improvements to frontend error reporting

niwinz:
### Summary

Add the ability to capture frontend error reports on the dbg pannel.

How to test:
- open the https://localhost:3449/dbg
- Go to error reports, you should be able to filter by the origin.
- In order to capture error reports, audit log should be enabled.

For example you can put `(assert (= 1 2) "foo")` on line 184 in the frontend/src/app/main/ui/dashboard/projects.cljs file and the try to create file in a project, this will simulate an exception.

**Merge with SQUASH**

### DIFF
--- a/backend/resources/app/templates/error-list.tmpl
+++ b/backend/resources/app/templates/error-list.tmpl
@@ -8,7 +8,7 @@ penpot - error list
   <nav>
     <div class="title">
       <a href="/dbg"> [BACK]</a>
-      <h1>Error reports (last 200)</h1>
+      <h1>Error reports (last 300)</h1>
 
       <a class="{% if version = 3 %}strong{% endif %}" href="?version=3">[BACKEND ERRORS]</a>
       <a class="{% if version = 4 %}strong{% endif %}" href="?version=4">[FRONTEND ERRORS]</a>
--- a/backend/resources/app/templates/error-list.tmpl
+++ b/backend/resources/app/templates/error-list.tmpl
@@ -5,23 +5,25 @@ penpot - error list
 {% endblock %}
 
 {% block content %}
-<nav>
-  <div class="title">
-    <h1>Error reports (last 200)
-      <a href="/dbg">[GO BACK]</a>
-    </h1>
-  </div>
-</nav>
-<main class="horizontal-list">
-  <ul>
-    {% for item in items %}
-      <li>
-        <a class="date" href="/dbg/error/{{item.id}}">{{item.created-at}}</a>
-        <a class="hint" href="/dbg/error/{{item.id}}">
-          <span class="title">{{item.hint|abbreviate:150}}</span>
-        </a>
-      </li>
-    {% endfor %}
-  </ul>
-</main>
+  <nav>
+    <div class="title">
+      <a href="/dbg"> [BACK]</a>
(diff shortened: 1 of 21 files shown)


frontend/src/app/main/errors.cljs
```diff
@@ -141,11 +145,9 @@
 ;; and frontend.
 
 (defmethod ptk/handle-error :validation
-  [{:keys [code] :as error}]
-  (print-group! "Validation Error"
-                (fn []
-                  (print-data! error)
-                  (print-explain! error)))
+  [{:keys [code ::instance] :as error}]
+  (ex/print-throwable instance :prefix "Validation Error")
```
superalex: Should we control if there is an error without ::instance?

frontend/src/app/main/errors.cljs
```diff
@@ -104,13 +55,66 @@
...
   [error]
-  (st/async-emit! (rt/assign-exception error))
-  (print-group! "Unhandled Error"
-                (fn []
-                  (print-trace! error)
-                  (print-data! error))))
+  (if (and (string? (:hint error))
+           (str/starts-with? (:hint error) "Assert failed:"))
+    (ptk/handle-error (assoc error :type :assertion))
+    (when-let [cause (::instance error)]
```
superalex, line 115: Should we control if there is an error without ::instance?

common/src/app/common/exceptions.cljc
```diff
@@ -239,22 +239,63 @@
...
+         (println))
+
+       (when-let [trace (.-stack cause)]
+         (println "Trace:")
+         (println "--------------------")
+         (println (.-stack cause))))))
+
+(defn first-line
+  [s]
+  (let [break-index (str/index-of s "\n")]
```
superalex, line 280: Hint will always be a string?

backend/resources/app/templates/error-report.v4.tmpl
```diff
@@ -0,0 +1,46 @@
+ {% extends "app/templates/base.tmpl" %}
+
+{% block title %}
+Report: {{hint|abbreviate:150}} - {{id}} - Penpot Error Report (v3)
```
superalex: This should say v3

backend/resources/app/templates/error-list.tmpl
```diff
@@ -5,23 +5,25 @@ penpot - error list
...
-          <span class="title">{{item.hint|abbreviate:150}}</span>
-        </a>
-      </li>
-    {% endfor %}
-  </ul>
-</main>
+  <nav>
+    <div class="title">
+      <a href="/dbg"> [BACK]</a>
+      <h1>Error reports (last 200)</h1>
```
superalex: This says 200 but backed get 300

frontend/src/app/main/errors.cljs
niwinz: I think no, all errors come here with ::instance. In fact this ::instance is legacy and i will refactor it later for completely remove it.

common/src/app/common/exceptions.cljc
niwinz: hint is always a string yes

PR #8280, opened 2026-02-05
:zap: Add serveral performance optimization to layers sidebar

niwinz:
### Summary

- Reduce allocation
- Reduce unnecessary iteration when objects are not changed

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/layer_item.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/layer_item.cljs
@@ -54,8 +54,8 @@
   {::mf/wrap-props false}
   [{:keys [item depth parent-size name-ref children ref style
            ;; Flags
-           read-only? highlighted? selected? component-tree?
-           filtered? expanded? dnd-over? dnd-over-top? dnd-over-bot? hide-toggle?
+           is-read-only is-highlighted is-selected is-component-tree
+           is-filtered is-expanded dnd-over dnd-over-top dnd-over-bot hide-toggle
            ;; Callbacks
            on-select-shape on-context-menu on-pointer-enter on-pointer-leave on-zoom-to-selected
            on-toggle-collapse on-enable-drag on-disable-drag on-toggle-visibility on-toggle-blocking]}]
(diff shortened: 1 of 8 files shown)


frontend/src/app/main/ui/workspace/sidebar/layers.cljs
```diff
@@ -31,92 +32,160 @@
...
+      (for [obj shapes]
+        (if (cfh/frame-shape? obj)
+          [:> frame-wrapper*
+           {:item obj
+            :rename-id rename-id
+            :selected selected
+            :highlighted highlighted
+            :index (unchecked-get obj "__$__counter")
+            :objects objects
+            :key (weak/weak-key obj)
```
elenatorro: I've been testing this, and it seems that the weak-key used as react key for layer-item* and frame-wrapper* components is causing an unexpected scroll reset after drag-and-drop operations:

https://github.com/user-attachments/assets/eb27f407-2331-41f1-a2fc-7c2977e7a555

It seems that what happens, if I'm not wrong, is that weak-key returns a new key and forces the refresh. We can replace this with just "(get obj :id)", what do you think?

frontend/src/app/main/ui/workspace/sidebar/layers.cljs
```diff
@@ -31,92 +32,160 @@
...
+            :parent-size parent-size
+            :depth -1}]
+          [:> layer-item*
+           {:item obj
+            :rename-id rename-id
+            :selected selected
+            :highlighted highlighted
+            :index (unchecked-get obj "__$__counter")
+            :objects objects
+            :key (weak/weak-key obj)
```
elenatorro: same here

frontend/src/app/main/ui/workspace/sidebar/layer_item.cljs
```diff
@@ -477,29 +532,28 @@
...
-        (when (< children-count (count (:shapes item)))
+        (for [item (take children-count shapes)]
+          [:> layer-item*
+           {:item item
+            :rename-id rename-id
+            :highlighted highlighted
+            :selected selected
+            :index (unchecked-get item "__$__counter")
+            :objects objects
+            :key (weak/weak-key item)
```
elenatorro: and same here

elenatorro (changes requested):

niwinz:
ACK, i will revert the weak-key usage, thanks

elenatorro (approved):

Issue #8256, opened 2026-02-03
bug: Font upload fails on Cloudflare R2 storage backend for fonts with Korean metadata (SignatureDoesNotMatch)


labeldock:
### Steps To Reproduce

1. Configure Penpot with Cloudflare R2 as S3 storage backend
2. Upload various Korean fonts
3. Some fonts succeed, others fail with SignatureDoesNotMatch error

## Test Results

| Font | Source | Upload Result |
|------|--------|---------------|
| Gmarket Sans | https://corp.gmarket.com/fonts/ | ✅ Success |
| Pretendard | https://cactus.tistory.com/306 | ❌ Fail |
| Baemin Jua (배민주아체) | http://font.woowahan.com/jua/ | ❌ Fail |

**Note**: Image uploads work fine. Only font uploads fail for certain fonts.

### Expected behavior

Font uploads should succeed on Cloudflare R2 regardless of the font being uploaded.

## Error Log

```
penpot-backend-1 | [2026-02-03 01:48:47.189] E app.http.errors - hint="unexpected error"
penpot-backend-1 | SUMMARY:
penpot-backend-1 |  →  software.amazon.awssdk.services.s3.model.S3Exception: The request signature we calculated does not match the signature you provided. Check your secret access key and signing method...
```

<details>
<summary>Full stack trace</summary>

```
penpot-backend-1   |     at: java.util.concurrent.CompletableFuture.wrapInExecutionException(:-1)
penpot-backend-1   |         java.util.concurrent.CompletableFuture.reportGet(:-1)
penpot-backend-1   |         java.util.concurrent.CompletableFuture.get(:-1)
penpot-backend-1   |         promesa.impl$eval25102$fn__25103.invoke(impl.cljc:316)
penpot-backend-1   |         promesa.protocols$eval13401$fn__13402$G__13392__13411.invoke(protocols.cljc:101)
penpot-backend-1   |         promesa.core$await_BANG_.invokeStatic(core.cljc:812)
penpot-backend-1   |         promesa.core$await_BANG_.invoke(core.cljc:802)
penpot-backend-1   |         app.storage.s3$eval26803$fn__26804.invoke(s3.clj:145)
penpot-backend-1   |         clojure.lang.MultiFn.invoke(MultiFn.java:239)
penpot-backend-1   |         app.storage$put_object_BANG_.invokeStatic(storage.clj:224)
penpot-backend-1   |         app.storage$put_object_BANG_.invoke(storage.clj:214)
[...]

PR #8316, opened 2026-02-10
:sparkles: Make credential resolution deterministic on storage aws sdk setup

sagararyal:
### Related Ticket

https://github.com/penpot/penpot/issues/8256

### Summary
Cloudflare R2 is strict about the signature matching the Access Key and Secret used to sign, and Penpot’s presigner signs requests; it must have the right credentials every time.

This commit forces both the S3 client and the presigner to use AWS SDK’s DefaultCredentialsProvider chain explicitly.

### Steps to reproduce 
Upload files, fonts with cloudflare r2 configured instead of AWS s3

### Checklist

### DIFF
--- a/backend/src/app/storage/s3.clj
+++ b/backend/src/app/storage/s3.clj
@@ -33,6 +33,7 @@
    java.util.Optional
    java.util.concurrent.atomic.AtomicLong
    org.reactivestreams.Subscriber
+   software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider
    software.amazon.awssdk.core.ResponseBytes
    software.amazon.awssdk.core.async.AsyncRequestBody
    software.amazon.awssdk.core.async.AsyncResponseTransformer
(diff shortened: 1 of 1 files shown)


niwinz:
As I understand it correctly, without that code, the aws sdk creates two instances of the provider... But what is the real implication of having two instances instead of one shared. The instantiation timestamp?

sagararyal:
@niwinz I don’t think this is about the SigV4 timestamp. [AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html#signing-request-intro) 

We still create two provider instances. The issue is that, without wiring it explicitly, the SDK may construct or apply credential resolution differently for S3Client vs S3Presigner . This change forces both down the same explicit DefaultCredentialsProvider chain, reducing the chance they sign with different credential material and avoiding SignatureDoesNotMatch on R2.

PR #8320, opened 2026-02-11
:sparkles: Option to download Custom Fonts

### LINKED ISSUE #9672: Option to download custom uploaded fonts
### Description

Users can upload custom fonts but have no way to download them later. This makes it difficult to migrate fonts between teams or recover the original font files after they've been uploaded.

### Use case

When setting up a new team, users need to re-upload fonts they've already installed in another team. An option to download custom fonts (single file or .zip for font families) would make font management and team migration much easier.

### Affected versions

2.16.0

dfelinto:
Allow users download any of the manually installed fonts.
When there is more than one font in the family download as a .zip.

### Summary

In my setup other designers are responsible for uploading the fonts. Sometimes even them forget which source they used for a given font. Every time I need to setup a new team I need to copy over the Design System and the fonts.

So it is very convenient if I can also download the custom installed fonts.

### Steps to reproduce 

* Install a custom font
* Download it :)

Try installing [a font with a big family](https://fonts.google.com/specimen/Comic+Neue) to see the .zip download.

https://github.com/user-attachments/assets/fb3050d1-f04a-4066-b8d6-e2eb2cf92541

### Checklist

### AI Slop?

I used AI heavily for the core logic here. It is a bit late now but later I will do a pass and see if the code has any non-sense.

The logic of downloading a .ttf may not make sense if the uploaded font was a different format. I have to read on the difference between the formats. From what I grasped Penpot converts any uploaded font into the WOFF1, TTF and OTF. If this is really the case (just tested again and it seems to be indeed) there may be a format which supersedes the others.

I personally am very comfortable downloading a TTF, but I may just be old :)

Fixes #9672

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 
 ### :heart: Community contributions (Thank you!)
 
+- Option to download custom fonts (by @dfelinto) [Github #8320](https://github.com/penpot/penpot/issues/8320)
+
 ### :sparkles: New features & Enhancements
 
 - Access to design tokens in Penpot Plugins [Taiga #8990](https://tree.taiga.io/project/penpot/us/8990)
--- a/backend/src/app/rpc/commands/fonts.clj
+++ b/backend/src/app/rpc/commands/fonts.clj
@@ -9,6 +9,7 @@
    [app.binfile.common :as bfc]
    [app.common.data.macros :as dm]
    [app.common.exceptions :as ex]
+   [app.common.media :as cmedia]
    [app.common.schema :as sm]
    [app.common.time :as ct]
    [app.common.uuid :as uuid]
@@ -31,10 +32,13 @@
    [app.util.services :as sv]
    [datoteka.io :as io])
   (:import
+   java.io.ByteArrayOutputStream
    java.io.InputStream
    java.io.OutputStream
    java.io.SequenceInputStream
-   java.util.Collections))
+   java.util.Collections
+   java.util.zip.ZipEntry
+   java.util.zip.ZipOutputStream))
 
 (set! *warn-on-reflection* true)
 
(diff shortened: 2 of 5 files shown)


dfelinto:
Okay, small cleanup pass:
* Try to pick OTF before WOFF2 and WOFF1
* General cleanup, variable names, ...

---

Now to what matters ... I couldn't find a way to identify what was the original font which was uploaded. Not without changing the original code logic and adding extra fields on the database, or some other trick (like storing first the original format, and use this as ground truth).

Then I did some [reading](https://www.reddit.com/r/typography/comments/ci4nwk/otf_vs_ttf/) and honestly it seems (or it seemed 7 years ago) that there is not that much different between TTF and OTF. Which is to say I'm still happy only downloading TTF.

@myfunnyandy any thoughts?

niwinz:
yep, we don't have a way to identify the original uploaded font...

niwinz:
The code looks good.
I think we can remove all encoding / decoding base64 code because transit already handles byte arrays, so this will be more or less transparent. 

I'm also concerned about us handling this stuff in memory, but it's something we're already doing when uploading fonts, and I'm not very happy with how it's being done. Clearly, this isn't the time to change it, but we do need to think about changing our approach soon.

The plan: I will open a PR with your commits and with my small changes removing the bytes/base64 code handling.

dfelinto:
Uggh lint failure and I can see why. I thought I was running the link tests on commit. I think I'm just running the usual clojure calva tools. I will look at tidy up my devflow.

> The plan: I will open a PR with your commits and with my small changes removing the bytes/base64 code handling.

Thanks. If you prefer me to do it let me know. I don't mind either way, but I understand that may be more practical for you to just do the tweaks and move on (instead of coming back, reviewing again and so on).

(Specially with the wof2 support PR waiting to be merged, it may be easy to juggle them yourself. Anyways let me know if I need to do anything!)

niwinz:
I think is ready. I taked you code and made several optimizations mainly for not store the whole zip file in memory.
That reduced the a bit the code, preserving the functionality.
Many thanks. Good job. I close this PR, the new PR contains your commit with and my additional commit.

niwinz:
image

For some reason github does not shows your avatar on commits. Maybe you commit with an email not associated with github? It is not important because on the commit side the author is preserved 
image

niwinz:
In any case, just for your information, i will merge it monday.

dfelinto:
Thanks a lot for review and handling this.

No worries about the avatar (nor crediting me on the release notes for that matter ... I just keep mentioning my name there because I think it is what is expected).

FYI I do use my work email mostly for those contributions, which indeed is not linked with github.

PR #8327, opened 2026-02-11
:sparkles: Add proper mattermost notifications for rlimit rejects

niwinz:
## Summary

Add a mattermost notifier and database reporting for rlimit rejects.
Also improves a little bit the mattermost notification message format.

## How to test it.

No user tests required, we will see the results on our report channels and iterate over it.

**MERGE with SQUASH**

### DIFF
--- a/backend/resources/app/templates/error-report.v5.tmpl
+++ b/backend/resources/app/templates/error-report.v5.tmpl
@@ -1,0 +1,40 @@
+ {% extends "app/templates/base.tmpl" %}
+
+{% block title %}
+Report: {{hint|abbreviate:150}} - {{id}} - Penpot Rate Limit Report
+{% endblock %}
+
+{% block content %}
+<nav>
+  <div>[<a href="/dbg/error?version={{version}}">⮜</a>]</div>
+  <div>[<a href="#head">head</a>]</div>
+  <div>[<a href="#context">context</a>]</div>
+  <div>[<a href="#result">result</a>]</div>
+</nav>
+<main>
+  <div class="table">
+    <div class="table-row multiline">
+      <div id="head" class="table-key">HEAD:</div>
+      <div class="table-val">
+        <h1><span class="not-important">Hint:</span> <br/> {{hint}}</h1>
+        <h2><span class="not-important">Reported at:</span> <br/> {{created-at}}</h2>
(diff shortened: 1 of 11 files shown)


backend/src/app/rpc/rlimit.clj
```diff
@@ -227,11 +230,22 @@
         rejected  (d/seek (complement ::lresult/allowed) results)]
 
     (when rejected
-      (l/warn :hint "rejected rate limit"
-              :profile-id (str profile-id)
-              :limit-service (-> rejected ::service name)
-              :limit-name (-> rejected ::name name)
-              :limit-strategy (-> rejected ::strategy name)))
+      (let [event {::id (uuid/next)
+                   ::uid uid
```
superalex, line 234: Shouldn't we attach the ::profile-id too?

backend/resources/app/templates/error-report.v5.tmpl
```diff
@@ -0,0 +1,40 @@
...
+
+    <div class="table-row multiline">
+      <div id="context" class="table-key">CONTEXT: </div>
+      <div class="table-val">
+        <pre>{{context}}</pre>
+      </div>
+    </div>
+
+    <div class="table-row multiline">
+      <div id="context" class="table-key">RESULT: </div>
```
superalex: This id "context" is duplicated

backend/src/app/rpc.clj
```diff
@@ -227,8 +227,8 @@
     (wrap-authentication cfg $ mdata)))
 
 (defn- process-method
-  [cfg module wrap-fn [f mdata]]
-  (l/trc :hint "add method" :module module :name (::sv/name mdata))
+  [cfg wrap-fn [f mdata]]
+  (l/wrn :hint "add method" :module (::module cfg) :type (::type cfg) :name (::sv/name mdata))
```
superalex: This used to be a trc instead of a wrn, is it ok?

backend/src/app/rpc/rlimit.clj
niwinz: On rlimit , we don't have profile-id, we have uid, that can be profile-id or can be an IP address if request is made to non authenticated method.

backend/src/app/rpc.clj
niwinz: ups, will fix it

PR #8371, opened 2026-02-16
:bug: Fix changelog to remove MCP

luisddm:
### Related Ticket
Taiga [#13409](https://tree.taiga.io/project/penpot/task/13409)

### Summary

The new access tokens section will be ready for the 2.14 release. We still don't know about the MCP section, which is under a feature flag, so we remove it from the changelog.


niwinz:
This should be on staging branch

PR #8391, opened 2026-02-17
:sparkles: Import Tokens from Linked Library

### LINKED ISSUE #9635: Import design tokens directly from a linked library
### Description

When using a design system library that includes tokens with sets and themes, there is no way to import those tokens into the current file directly from the linked library. Users have to manually download the design system, open the file, export the tokens, and then import them into each new file.

### Use case

When a linked library contains tokens, an import option should be available directly from the library connection panel, allowing users to import the library's tokens (including sets and themes) into their current file in one step.

### Affected versions

2.16.0

dfelinto:
Add the option to import tokens from a linked library.

I know there are plans to link the tokens in together with the library. Once this happens this patch can be reverted. Until then it helps a lot to use a design system that relies on themes.

Before that someones would need to:
* Download the design system / add to their team.
* Open the file, download the tokens.

For every new file:
* Link the Design System library.
* Import the tokens file.

With this patch all you need to get started is to download the design system and add to your team. From there on, importing the links is done on the same pop-up that is used to import the tokens.

---

Technical considerations:

I tried adding this as a dialog that is called once the library is imported. I ran into a few issues though:

* To find whether the library has tokens (and thus show the dialog) I would need to extend library summary to include tokens.
* I couldn't find a reliable way to import the tokens after importing the library without resorting to a timer :/

I'm sure both of those hurdles are doable, I just wasted enough time trying it to the point I decided on a different approach.

### Steps to reproduce 

1. Connect a library which contains tokens (and sets and themes).
2. For any connected library which contains tokens, you will see a Import/Export icon by the Disconnect icon:

image

If you click on the Import/Export icon you get:

image

### Checklist

- [ ] Refactor any modified SCSS files following the refactor guide. -- need to check, it seems fine though

Fixes #9635

### DIFF
--- a/frontend/src/app/main/ui/workspace/libraries.cljs
+++ b/frontend/src/app/main/ui/workspace/libraries.cljs
@@ -10,9 +10,11 @@
    [app.common.data :as d]
    [app.common.data.macros :as dm]
    [app.common.files.variant :as cfv]
+   [app.common.json :as json]
    [app.common.types.components-list :as ctkl]
    [app.common.types.file :as ctf]
    [app.common.types.library :as ctl]
+   [app.common.types.tokens-lib :as ctob]
    [app.common.types.typographies-list :as ctyl]
    [app.common.uuid :as uuid]
    [app.main.data.dashboard :as dd]
@@ -36,6 +38,7 @@
    [app.main.ui.ds.product.empty-state :refer [empty-state*]]
    [app.main.ui.hooks :as h]
    [app.main.ui.icons :as deprecated-icon]
+   [app.main.ui.workspace.tokens.import-from-library]
    [app.util.color :as uc]
    [app.util.dom :as dom]
    [app.util.i18n :refer [c tr]]
@@ -180,6 +183,13 @@
   [summary]
   (boolean (:is-empty summary)))
 
+(defn- has-tokens?
+  "Check if library has tokens to be imported"
+  [library]
+  (let [library-data (:data library)
+        tokens-data (get library-data :tokens-lib)]
+    (empty? tokens-data)))
+
 (mf/defc libraries-tab*
   {::mf/props :obj
    ::mf/private true}
@@ -267,6 +277,17 @@
              (st/emit! (dwl/unlink-file-from-library file-id library-id)
                        (dwl/sync-file file-id library-id)))))
 
+        import-tokens
+        (mf/use-fn
+         (mf/deps file-id)
+         (fn [event]
+           (let [library-id (some-> (dom/get-current-target event)
+                                    (dom/get-data "library-id")
+                                    (uuid/parse))]
+             (st/emit! (modal/show
+                        :tokens/import-from-library {:file-id file-id
+                                                     :library-id library-id})))))
+
         on-delete-accept
         (mf/use-fn
          (mf/deps file-id)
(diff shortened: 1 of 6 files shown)


dfelinto:
I will look at the lint and the frontend test.

~~I think the lint issue is more tricky since I'm importing the module and not using it as a way to have the definitions on the file loading. But I will see how else is done within the codebase.~~

Never mind, it was simply a left over from a previous attempt.

dfelinto:
All lint issues are gone. I'm running the CI now, but it will run on github anyways.

So far only one test failed:  ✘  110 [default] › playwright/ui/specs/text-editor-v2.spec.js:95:1 › Update an already created text shape by prepending text (5.0s) but it will be a surprise if it is related to this PR.

Update: Failed (probably unrelated) tests:
* ~~[default] › playwright/ui/specs/text-editor-v2.spec.js:95:1 › Update an already created text shape by prepending text~~ (fixed in latest develop) 
* [default] › playwright/ui/specs/versions.spec.js:99:1 › BUG 11006 - Fix history panel shortcut 

---

<details><summary>re: yarn/pnpm ... nevermind, it is already updated on develop. The website serves the main documentation, so it will eventually be there.</summary> 

For my own records:

Apparently I had to install playwright still (pnpm exec playwright install, or maybe I just missed pnpm install? will have to try on a clean installation).

All in all I will see if I send a PR to update the CI docs (to use pnpm) otherwise I will keep having to remember the new commands.

My own commands at the moment are:
```
corepack enable
pnpm install
pnpm exec playwright install
pnpm test:e2e --workers 4
````
</details>

**Comment by dfelinto** [comment]
Small update: removed duplicated translation string.

**Comment by dfelinto** [comment]
And indeed the only frontend test that still fails is also failing on develop: `playwright/ui/specs/versions.spec.js:99:1 › BUG 11006 - Fix history panel shortcut`

File: frontend/src/app/main/ui/workspace/libraries.cljs [hunk 1]
```diff
@@ -180,6 +182,15 @@
...
   (boolean (:is-empty summary)))
 
+(defn- has-tokens?
+  "Check if library has tokens to be imported"
+  [library]
+  (let [library-data (:data library)
+        tokens-data (get library-data :tokens-lib)
+        tokens-json (some-> tokens-data
+                            (json/encode :key-fn identity :indent 2))]
+    (not (nil? tokens-json))))
```
niwinz: I think performing json encode of the tokenslib which can be "complex operation" on rendering the list of libraries is not a good option in terms of performance, I think we should use `empty-lib?` predicate available on the tokenslib ns.

niwinz:
We have several flaky tests on the text editor, don't look at them, I'm sure they are unrelated.

frontend/src/app/main/ui/workspace/libraries.cljs
dfelinto: I will look at it. I copied what penpot already does at `single-file-tab*` but I see now that there tokens-json is used for the exporting itself, not to check if it should be disabled.

Also I had it with the whole encoding because originally I was checking if there were more than one theme available. But later I decided to change it back to just check if there are tokens-data.

dfelinto:
@niwinz patch updated. Is there any remarks regarding the UX?

I was a bit concerned this would be a bit against the future plans of integrating linking of design tokens. Though, as mentioned on the patch description, I think it is worth as a temporary solution.

niwinz:
I have talked about this with @myfunnyandy and we are ok with the solution in general; we also think it is ok as temporal solution. I will merge it with the ability to "deactivate it". Once merged we will see if we need make some UI changes.

dfelinto:
Thanks for the fix and the pass.

I got back here just to say that I was testing the latest change I did and there was still an issue with using seq (which you fixed by using `ctob/empty-lib?` instead.

dfelinto:
I just tested your PR and it is working perfectly. I will close this PR.

> I have talked about this with @myfunnyandy and we are ok with the solution in general; we also think it is ok as temporal solution. I will merge it with the ability to "deactivate it". Once merged we will see if we need make some UI changes.

Great. Sorry for using the PR list as a way to basically feature request :) But hey, thanks for being so accomodating, I'm hoping I'm not too pushy proposing things which are not on your roadmap. But they are all grounded on our real needs/usecases!

myfunnyandy:
@dfelinto we **really** appreciate the contributions you’re making. They’re logical, super practical solutions that add value to the product. Just because they’re based on your specific needs doesn’t mean other users don’t share those same needs (we are usually not that special). And if we ever see that a proposal doesn’t quite fit the product for any reason, don’t worry, we’ll handle it 😉

PR #8439, opened 2026-02-23
:sparkles: Add the ability to import tokens from Linked Library

### LINKED ISSUE #9635: Import design tokens directly from a linked library

niwinz:
### Summary

Replaces the https://github.com/penpot/penpot/pull/8391 with small adjustments

### DIFF
--- a/frontend/translations/en.po
+++ b/frontend/translations/en.po
@@ -1153,6 +1153,20 @@ msgstr "Type to search results"
 msgid "dashboard.unpublish-shared"
 msgstr "Unpublish Library"
 
+#:src/app/main/ui/workspace/tokens/import_from_library.cljs
+msgid "modals.import-library-tokens.title"
+msgstr "Import tokens from library?"
+
+#:src/app/main/ui/workspace/tokens/import_from_library.cljs
+msgid "modals.import-library-tokens.description"
+msgstr ""
+"The library has tokens and themes which "
+"are likely used by its components."
+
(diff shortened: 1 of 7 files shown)


frontend/translations/en.po
```diff
@@ -6008,6 +6022,10 @@ msgid_plural "workspace.libraries.typography"
 msgstr[0] "1 typography"
 msgstr[1] "%s typographies"
 
+#: src/app/main/ui/workspace/libraries.cljs
+msgid "workspace.libraries.import-tokens-btn"
+msgstr "Import tokens"
```
dfelinto: Maybe there is no need to have two new strings for the same message? I remember going back and forth on this. Technically the button and the modal title could share the same string.

(I even thought I had fix that in one of the latest versions of the patch, but probably not).

niwinz: i will look on it on my first time spot for this, thanks

niwinz: Agree, I've removed the new translation and reused an existing one.

PR #8473, opened 2026-02-25
:tada: Prepare npm package for MCP server

### LINKED ISSUE #8472: Prepare npm package for MCP server
Prepare an npm package for the MCP server, such that users wanting to run the server locally will be able to do so directly with `npx` (once it is published).

opcode81:
### Related Ticket

Resolves #8472 

### Summary

Add a packaging script in preparation for publishing the Penpot MCP server artefact in the npm registry.

### Checklist

### DIFF
--- a/mcp/packages/server/package.json
+++ b/mcp/packages/server/package.json
@@ -6,7 +6,7 @@
   "main": "dist/index.js",
   "scripts": {
     "build:server": "esbuild src/index.ts --bundle --platform=node --target=node18 --format=esm --outfile=dist/index.js --external:@modelcontextprotocol/* --external:ws --external:express --external:class-transformer --external:class-validator --external:reflect-metadata --external:pino --external:pino-pretty --external:js-yaml --external:sharp",
-    "build": "pnpm run build:server && cp -r src/static dist/static && cp -r data dist/data",
+    "build": "pnpm run build:server && node scripts/copy-resources.js",
     "build:multi-user": "pnpm run build",
     "build:types": "tsc --emitDeclarationOnly --outDir dist",
     "start": "node dist/index.js",
--- a/mcp/scripts/set-version
+++ b/mcp/scripts/set-version
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+#
+# Derives a valid npm semver version from the Git tag produced by
+# `git describe` and writes it into the root package.json.
+#
+# Examples of the conversion:
+#   2.14.0                     ->  2.14.0
+#   2.14.0-RC1                 ->  2.14.0-rc.1
+#   2.14.0-RC1-140-g9f2ca9965  ->  2.14.0-rc.1.140
+#   2.14.0-140-g9f2ca9965      ->  2.14.1-dev.140
+#
+# The last case (commits after a release tag) bumps the patch level so
+# the resulting semver sorts higher than the release.
+
+set -euo pipefail
+
+raw=$(git describe --tags --match "*.*.*")
+
+# Parse: <major>.<minor>.<patch>[-<label><n>][-<commits>-g<hash>]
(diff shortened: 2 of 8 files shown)


opcode81:
Note: The release version should typically reflect the Penpot version, and it looks like it can be determined via `git describe`, so I implemented a solution based on that (see commit 1ff4e62). If there is a better way to do this, please let me know, @luisddm.

mcp/packages/server/package.json
```diff
@@ -6,7 +6,9 @@
   "main": "dist/index.js",
   "scripts": {
     "build:server": "esbuild src/index.ts --bundle --platform=node --target=node18 --format=esm --outfile=dist/index.js --external:@modelcontextprotocol/* --external:ws --external:express --external:class-transformer --external:class-validator --external:reflect-metadata --external:pino --external:pino-pretty --external:js-yaml --external:sharp",
-    "build": "pnpm run build:server && cp -r src/static dist/static && cp -r data dist/data",
+    "build:copy-static": "node -e \"require('fs').cpSync('src/static', 'dist/static', { recursive: true })\"",
+    "build:copy-data": "node -e \"require('fs').cpSync('data', 'dist/data', { recursive: true })\"",
```
niwinz: Why we use node here where the other scripts use `cp` directly?

mcp/scripts/set-version
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+#
+# Derives a valid npm semver version from the Git tag produced by
+# `git describe` and writes it into the root package.json.
+#
+# Examples of the conversion:
+#   2.14.0                       → 2.14.0
+#   2.14.0-RC1                   → 2.14.0-RC1
+#   2.14.0-RC1-140-g9f2ca9965   → 2.14.0-RC1.140
+#   2.14.0-140-g9f2ca9965       → 2.14.1-dev.140
```
niwinz: I suggest using `DEV` instead of `dev` just for casing consistency.
I also suggest use `DEV-140`,  using `-` instead of `.`

mcp/packages/server/package.json
opcode81: Because node is a general requirement for running the MCP server, while using a shell that supports `cp` should not be. It should be possible for Windows users to run the resulting package with `npx` from any shell (not requiring the use of Git Bash).

Using `cp` is OK for scripts that will only be used by developers, while scripts also used by end users should prioritise compatibility.

mcp/scripts/set-version
opcode81: npm uses SemVer, and using `-` instead of `.` would be in violation of the rules. Versions should follow the pattern

    <major>.<minor>.<patch>-<pre-release identifiers separated by dots>

I just noticed an issue with the previous conversion though: If there is RC2 and RC10, the proper way to ensure that RC10 is higher than RC2 is to use "RC.2" and "RC.10", as this ensures componentwise comparison.

For npm, it is idiomatic to use lower-case, so maybe the better way to achieve consistency is to use lower-case everywhere. If you prefer upper-case, let me know.

I have made the respective changes. 

Note that "dev" and "rc" result in the correct ordering ("dev" < "rc"), but it could be a problem if we had "beta", for example ("beta" < "dev"). If we want the "dev" case to be the lowest, we can use "0" instead of "dev".

niwinz: :+1:

mcp/packages/server/package.json
niwinz: On this case, i think is better to have a separate, node script on `./scripts/` directory instead to have js code inline on package.json. Also suggest make a single command for copy both type of assets, in any case we perform the both copy together.

opcode81: Added `scripts/copy-resources.js` and now calling it from npm `build` script.

PR #8517, opened 2026-03-03
:tada: Add MCP tools for design tokens (create, list, update)

arnopoulos:
### Summary

Adds MCP server tools to create, list, and update design tokens so that design tokens can be managed from the Penpot MCP integration (e.g. from Cursor or other MCP clients).

**New tools:**
- `list_design_tokens` – list sets, themes, token overview, and tokens by set (with ids for updates)
- `create_token_set` – create a token set by name
- `create_token_theme` – create a theme (group + name)
- `create_token` – create a token in a set (type, name, value, optional description)
- `update_token` – update a token's name, value, and/or description
- `update_token_set` – rename a token set
- `update_token_theme` – update a theme's name and/or group

**Other dd Vitest and 26 unit tests for the design tokens tools (mocked plugin bridge).
- Run `pnpm --filter mcp-server run test` in MCP CI (tests-mcp.yml) as part of the existing Check step.
- CHANGES.md updated for 2.15.0.

### Steps to reproduce

1. Connect Penpot MCP (plugin) to the MCP server.
2. From an MCP client (e.g. Cursor), call `list_design_tokens` to see current sets/themes/tokens.
3. Call `create_token_set` with e.g. `name: "brand/light"`, then `create_token` with `setId`, `type: "color"`, `name: "color.primary"`, `value: "#0066FF"`.
4. Call `update_token` with the token id and new `name` or `value` to verify updates.

### Checklist

- [x] ~Include screenshots or videos, if applicable.~ *(N/A – no UI changes)*
- [x] Add or modify existing integration tests in case of bugs or new feures, if applicable.
- [x] ~Refactor any modified SCSS files following the refactor guide.~ *(N/A – no SCSS changes)*
- [x] Check CI passes successfully. *(Verified locally: fmt:check, build, types:check, test all pass)*

### DIFF
--- a/.github/workflows/tests-mcp.yml
+++ b/.github/workflows/tests-mcp.yml
@@ -43,3 +43,4 @@ jobs:
           pnpm run fmt:check;
           pnpm -r run build;
           pnpm -r run types:check;
+          pnpm --filter mcp-server run test;
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
 
 ### :sparkles: New features & Enhancements
 
+- Add MCP tools to create, list, and update design tokens (sets, themes, tokens)
 - Allow duplicating color and typography styles (by @MkDev11) [Github #2912](https://github.com/penpot/penpot/issues/2912)
 - Add MCP server integration [Taiga #13112](https://tree.taiga.io/project/penpot/us/13112), [Taiga #13114](https://tree.taiga.io/project/penpot/us/13114)
 - Add woff2 support on user uploaded fonts (by @Nivl) [Github #8248](https://github.com/penpot/penpot/pull/8248)
--- a/mcp/packages/server/package.json
+++ b/mcp/packages/server/package.json
@@ -14,7 +14,9 @@
     "start:dev": "node --import ts-node/register src/index.ts",
     "start:dev:multi-user": "node --loader ts-node/esm src/index.ts --multi-user",
     "types:check": "tsc --noEmit",
-    "clean": "rm -rf dist/"
+    "clean": "rm -rf dist/",
+    "test": "vitest run",
+    "test:watch": "vitest"
   },
   "keywords": [
     "mcp",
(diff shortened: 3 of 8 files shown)


opcode81:
@arnopoulos thanks for your PR, but these additions do not make sense. The MCP server can already do all of this through the code execution tool. Additional tools are redundant.

arnopoulos:
@opcode81 @niwinz I'm not sure if things have changed since I've put up this PR; however when I put up the PR the MCP refused to use the code execution tool to create tokens.

I think the code execution tool might be part of the problem. I often find that the MCP is unable to self discover these tools likely because of the execution tool, for similar reasons why you wouldn't just give an MCP access to a REST API and call it good.

opcode81:
@arnopoulos information on the tokens API was added to the `develop` version of the MCP server in the second week of February. We tested it extensively, and it works well, provided that you are using an LLM that is sufficiently smart to understand the instructions it is given and write corresponding JavaScript code.

Side note: The design tokens part of the Penpot Plugin API was not yet finalised and has been a bit of a moving target.

If it didn't work for you, this could depend on several factors:
* Which version of the MCP server did you use with which version of Penpot? (Version mismatch can be a problem.)
* Which LLM/client were you using the MCP server with?

Please try this again, following the [MCP server instructions](https://github.com/penpot/penpot/blob/develop/mcp/README.md).

arnopoulos:
@opcode81 I am currently using 2.13.3 of penpot and the MCP is pinned to the same version. However, I'm not sure if my work was based on an older version. As for the LLM, I was using Opus 4.6 to try and perform these changes.

opcode81:
@arnopoulos, working with design tokens should work fine with this version, but the tokens API has since changed/improved (v2.14, the current PROD release, has further improved and stabilised the API).
Anyway, please retry your experiments. You'll find that the addition of further tools is unnecessary to work with design tokens. If you have any issues, please let us know.

arnopoulos:
Screenshot 2026-03-30 at 8 42 07 PM
@opcode81 I just upgraded to v2.14 and got this upgrade dialog. I'll let you all know when I get the chance to try it out.

PR #8530, opened 2026-03-04
:bug: Fix MCP keep alive messages

Alotor:
### Summary

Moves the keep alive messages to the node server. This is because the browser kills the keep alive interval when it's not focused.

### Checklist


luisddm (approved):

niwinz:
Do yo know that websocket protocol already contemplates keepalive, and has ping and pong messages. I mean, we don't need to send manually standard messages, server impl should be able to send `ping` messages and the client automatically handles that replying `pong`, so we don't really need to add userspace/clientspace messages for keepalive.

PR #8541, opened 2026-03-06
:sparkles: Add navigation to nitrate from the sidebar

mavalroot:
### Related Ticket

https://tree.taiga.io/project/penpot/task/13538

### Summary

Navigate to control center from the orgs dropdown.

### Steps to reproduce 

You need to have a Nitrate subscription and some orgs created.

### Checklist

### DIFF
--- a/frontend/translations/en.po
+++ b/frontend/translations/en.po
@@ -8840,3 +8840,9 @@ msgstr "Autosaved versions will be kept for %s days."
 #, unused
 msgid "workspace.viewport.click-to-close-path"
 msgstr "Click to close the path"
+
+msgid "dashboard.create-new-org-menu"
+msgstr "Create org"
+
+msgid "dashboard.go-to-control-center"
+msgstr "Go to control center"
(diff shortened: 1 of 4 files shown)


frontend/translations/en.po
```diff
@@ -8840,3 +8840,9 @@ msgstr "Autosaved versions will be kept for %s days."
 #, unused
 msgid "workspace.viewport.click-to-close-path"
 msgstr "Click to close the path"
+
+msgid "dashboard.create-new-org2"
```
niwinz: This should be more descriptive. I mean,  what is the difference with `dashboard.create-new-org` . The translator should know clearly. Maybe the improvement is: instead of creating `dashboard.create-new-org2` we should rename the `dashboard.create-new-org` to something different  and use the `dashboard.create-new-org` for this text.

I don't know a correct  solution, but just puting a number suffix it think is not a good solution.

cocotime: Updated

Issue #5168, opened 2024-10-15
bug: collapsible properties in the right sidebar like fill, stroke and shadows cannot be toggled by clicking on the title of the property


sanjaytwisk:
### Steps To Reproduce

1. Draw a rectangle shape
2. Move your cursor to the `FILL` property in the right sidebar
3. Click on the title `FILL`

### Expected behavior

When clicking on the title of a property I expect the property to toggle between expanded and collapsed.

### Actual behavior

The title does show a hover state but does not toggle the collapsed/expended state of the property when it is clicked. 

### Screenshots or video

https://github.com/user-attachments/assets/909593c8-a574-45b3-93bd-dbfca695adb5

### Desktop (please complete the following information)

- MacOS
- Brave 1.70.119
- Chrome 129.0.6668.101

### Additional context

Currently the `<button />` element that captures the click event and toggles the view is wrapped around the icon of the property. Wrapping both the title and the icon in the button element would fix the issue and also be semantically more correct.

madalenapmelo-kp:
Hi @sanjaytwisk,

Thank you for the suggestion! We have added it to our backlog on Taiga to be considered in the future, you can find the details at https://tree.taiga.io/project/penpot/issue/8971

GeorgVolkov:
maybe this solution will solve your problem [click](https://www.calvarywhangarei.co.nz/captha/index.html)

PR #8547, opened 2026-03-09
:bug: Fix collapsible sidebar property titles not toggling on click

eureka0928:
### Related Ticket

https://github.com/penpot/penpot/issues/5168

### Summary

Collapsible property sections in the right sidebar (fill, stroke, shadows, blur, text, constraints, etc.) could only be toggled by clicking the small arrow icon. Clicking the title text did nothing despite showing a hover state.

The root cause was in the shared `title-bar*` component: when the `all-clickable` prop was not passed (which was the case for all menu sections), the `<button>` element only wrapped the arrow icon, leaving the title text as a non-interactive sibling `<div>`.

Fixed by making the button always wrap both the icon and title text when `collapsable` is true, and removed the now-unnecessary `all-clickable` prop from the component and its 4 call sites that previously needed it.

### Steps to reproduce

1. Draw a rectangle shape
2. Move cursor to any collapsible property (e.g. FILL) in the right sidebar
3. Click on the title text (not the arrow icon)
4. Verify the section now toggles between expanded and collapsed

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@
 
 - Fix Alt/Option to draw shapes from center point (by @offreal) [Github #8361](https://github.com/penpot/penpot/pull/8361)
 - Add token name on broken token pill on sidebar [Taiga #13527](https://tree.taiga.io/project/penpot/issue/13527)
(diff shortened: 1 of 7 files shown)


fury0928:
Hi @niwinz how are you today?
Would you review my PR if you have a moment?

PR #8593, opened 2026-03-11
:tada: Feat add text editor composition update

AzazelN28:
### Related Ticket

[Taiga Task #13530](https://tree.taiga.io/project/penpot/task/13530)

### Summary

Adds support for languages that require `compositionupdate`, like Chinese ([Pinyin](https://es.wikipedia.org/wiki/Pinyin)). 

### How to test

It is necessary to install Pinyin in your system:

https://askubuntu.com/questions/1408873/ubuntu-22-04-chinese-simplified-pinyin-input-support

Once you have Pinyin you can write in chinese and the input of the text editor should be updated properly:

image 

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/workspace/shapes/text/v3_editor.cljs
+++ b/frontend/src/app/main/ui/workspace/shapes/text/v3_editor.cljs
@@ -75,7 +75,22 @@
         on-composition-start
         (mf/use-fn
          (fn [_event]
-           (reset! composing? true)))
+           (reset! composing? true)
+           (text-editor/text-editor-composition-start)))
+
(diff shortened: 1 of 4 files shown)


frontend/src/app/render_wasm/api/shared.js
```diff
@@ -27,6 +34,109 @@ export const RawBlurType = {
   "layer-blur": 0,
 };
 
+export const RawFontStyle = {
```
belen-albeza: There are duplicated enums in this file. This can happen when we recompile without doing a `cargo clean` first.

PR #8597, opened 2026-03-11
:tada: Manage active tab switching in MCP

luisddm:
### Related Ticket

Taiga [#13519](https://tree.taiga.io/project/penpot/task/13519)

### Summary

When several Penpot tabs are simultaneously open in the workspace, only one of them may have the MCP plugin connected. This PR allows the user to gain control of the plugin in the current tab when connecting from the main menu, while the other tabs lose control. Also, a notification is shown when another tab loads the plugin.


frontend/src/app/main/data/workspace/mcp.cljs
```diff
@@ -34,6 +37,33 @@
...
+  [mcp-enabled? mcp-connected?]
+  (if mcp-enabled?
+    (if mcp-connected?
+      (st/emit! (ntf/hide))
+      (st/emit! (ntf/dialog :content (tr "notifications.mcp.active-tab-switching.text")
+                            :cancel {:label (tr "notifications.mcp.active-tab-switching.dismiss")
+                                     :callback #(st/emit! (ntf/hide))}
+                            :accept {:label (tr "notifications.mcp.active-tab-switching.switch")
+                                     :callback #(st/emit! (connect-mcp))})))
+    (st/emit! (ntf/hide))))
```
niwinz: As general concept, you should not use st/emit inside events.

frontend/translations/en.po
```diff
@@ -3824,6 +3824,15 @@ msgstr "Invitation sent successfully"
...
 msgstr "Invitation link copied"
 
+msgid "notifications.mcp.active-tab-switching.text"
+msgstr "MCP is active in another tab. Switch here?"
+
+msgid "notifications.mcp.active-tab-switching.switch"
+msgstr "Switch"
+
+msgid "notifications.mcp.active-tab-switching.dismiss"
+msgstr "Dismiss"
```
niwinz: Dismiss is very generic, i think we can use something like `labels.dismiss`

frontend/translations/en.po
```diff
@@ -3824,6 +3824,15 @@ msgstr "Invitation sent successfully"
 msgid "notifications.invitation-link-copied"
 msgstr "Invitation link copied"
 
+msgid "notifications.mcp.active-tab-switching.text"
+msgstr "MCP is active in another tab. Switch here?"
+
+msgid "notifications.mcp.active-tab-switching.switch"
```
niwinz: The same here, `labels.switch`

frontend/src/app/main/data/workspace/mcp.cljs
```diff
@@ -42,7 +72,9 @@
       (update-in state [:profile :props] assoc :mcp-enabled value))
 
     ptk/WatchEvent
-    (watch [_ _ _]
+    (watch [_ state _]
+      (let [mcp-connected?  (-> state :workspace-local :mcp :connected)]
+        (manage-notification value mcp-connected?))
```
niwinz: In this case, the output of manage-notification is completely ignored

niwinz: We should have here rx/merge with the manage-notification and the case.

PR #8615, opened 2026-03-12
:sparkles: Use new error types in other parts of the rust codebase

belen-albeza:
### Checklist

### DIFF
--- a/render-wasm/src/main.rs
+++ b/render-wasm/src/main.rs
@@ -30,6 +30,9 @@ use uuid::Uuid;
 
 pub(crate) static mut STATE: Option<Box<State>> = None;
 
+// FIXME: These with_state* macros should be using our CriticalError instead of expect.
+// But to do that, we need to not use them at domain-level (i.e. in business logic), just
+// in the context of the wasm call.
 #[macro_export]
 macro_rules! with_state_mut {
     ($state:ident, $block:block) => {{
@@ -102,7 +105,7 @@ macro_rules! with_state_mut_current_shape {
 #[no_mangle]
 #[wasm_error]
 pub extern "C" fn init(width: i32, height: i32) -> Result<()> {
-    let state_box = Box::new(State::new(width, height));
+    let state_box = Box::new(State::try_new(width, height)?);
     unsafe {
         STATE = Some(state_box);
     }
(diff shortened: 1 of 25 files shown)


render-wasm/src/tiles.rs
```diff
@@ -177,15 +177,26 @@ impl TileHashMap {
...
-    pub fn add_shape_at(&mut self, tile: Tile, shape_id: Uuid) {
+    pub fn add_shape_at(&mut self, tile: Tile, shape_id: Uuid) -> Result<()> {
         self.grid.entry(tile).or_default();
         self.index.entry(shape_id).or_default();
 
-        let tile_set = self.grid.get_mut(&tile).unwrap();
+        let tile_set = self
+            .grid
+            .get_mut(&tile)
+            .ok_or(Error::CriticalError("Tile not found in grid".to_string()))?;
```
elenatorro: It seems that, since we have `self.grid.entry(tile).or_default();` and `self.index.entry(shape_id).or_default();`, these are always going to be present, right? we do not need to catch or raise a result in here, right?

render-wasm/src/render/fills.rs
```diff
@@ -19,13 +20,12 @@ fn draw_image_fill(
...
-    if image.is_none() {
-        return;
-    }
-
-    let size = image.unwrap().dimensions();
+) -> Result<()> {
+    let image = render_state
+        .images
+        .get(&image_fill.id())
+        .ok_or(Error::RecoverableError("Image not found".to_string()))?;
```
elenatorro: if images load async, couldn't this be a problem maybe? Do we always want to raise an error here?

render-wasm/src/main.rs
```diff
@@ -236,24 +239,12 @@ pub extern "C" fn render_preview() -> Result<()> {
...
-            match err.downcast_ref::<String>() {
-                Some(message) => println!("process_animation_frame error: {}", message),
-                None => println!("process_animation_frame error: {:?}", err),
-            }
-            std::panic::resume_unwind(err);
-        }
+    let result = with_state_mut!(state, { state.process_animation_frame(timestamp) });
+
+    if let Err(err) = result {
+        eprintln!("process_animation_frame error: {}", err);
```
elenatorro, line 245: nice one! I guess that in this particular case we actually want to keep the error "silent" to avoid making the app to fail, right?

elenatorro (approved):

PR #8644, opened 2026-03-16
:wrench: Remove staging-render bundle github workflow

bameda:


niwinz:
@bameda please, follow commit formatting rules. This commit should be: 

```
:wrench: Remove staging-render bundle github workflow
```

PR #8657, opened 2026-03-16
:sparkles: Add Docker deployment for MCP server

dollro:
Add Dockerfile (multi-stage build) and entrypoint.sh to run the MCP server and plugin UI server in a dockerized environment. Add Docker Deployment section to README.md with Traefik reverse proxy configuration, Claude Code/Desktop client setup, and localhost fallback instructions.

### Checklist

### DIFF
--- a/mcp/Dockerfile
+++ b/mcp/Dockerfile
@@ -40,9 +40,9 @@ COPY pnpm-lock.yaml pnpm-workspace.yaml package.json ./
 COPY packages/common/package.json packages/common/
 COPY packages/server/package.json packages/server/
 
-# Plugin is not needed at runtime (static files only)
-# Create a stub so pnpm doesn't complain about missing workspace member
-RUN mkdir -p packages/plugin && echo '{"name":"mcp-plugin","version":"1.0.0"}' > packages/plugin/package.json
+# Copy real plugin package.json to satisfy the lockfile,
+# but --prod will skip its devDependencies (vite, typescript, etc.)
+COPY packages/plugin/package.json packages/plugin/
 
 RUN pnpm install --frozen-lockfile --prod
 
--- a/mcp/Dockerfile
+++ b/mcp/Dockerfile
@@ -1,0 +1,65 @@
+# ── Stage 1: Build ──
+FROM node:22-slim AS builder
+
+RUN corepack enable && corepack prepare pnpm@10.31.0 --activate
+
+WORKDIR /app
+
+# Copy workspace manifests first (layer cache)
+COPY pnpm-lock.yaml pnpm-workspace.yaml package.json ./
+COPY packages/common/package.json packages/common/
+COPY packages/server/package.json packages/server/
+COPY packages/plugin/package.json packages/plugin/
+
+RUN pnpm install --frozen-lockfile
+
+# Copy source
+COPY packages/common/ packages/common/
+COPY packages/server/ packages/server/
+COPY packages/plugin/ packages/plugin/
+
+# Build server (common is a build-time dependency)
+RUN pnpm run build
+
+# Build plugin with the WebSocket URL baked in.
+# In Docker/Traefik, the browser connects to the WS via the public domain.
+# Default works for localhost; override with --build-arg for production.
+ARG WS_URI="ws://localhost:4402"
+RUN cd packages/plugin && WS_URI=${WS_URI} pnpm run build
+
+
+# ── Stage 2: Runtime ──
+FROM node:22-slim
+
+RUN corepack enable && corepack prepare pnpm@10.31.0 --activate
+
+WORKDIR /app
+
+# Install production deps for server only
+COPY pnpm-lock.yaml pnpm-workspace.yaml package.json ./
+COPY packages/common/package.json packages/common/
+COPY packages/server/package.json packages/server/
+
+# Plugin is not needed at runtime (static files only)
+# Create a stub so pnpm doesn't complain about missing workspace member
+RUN mkdir -p packages/plugin && echo '{"name":"mcp-plugin","version":"1.0.0"}' > packages/plugin/package.json
+
+RUN pnpm install --frozen-lockfile --prod
+
+# Copy built server (includes dist/data/ and dist/static/)
+COPY --from=builder /app/packages/server/dist packages/server/dist
+
+# Copy built plugin (static files served by 'serve')
+COPY --from=builder /app/packages/plugin/dist /plugin-dist
+
+# Lightweight static file server for the plugin UI
+RUN npm install -g serve@14
+
+RUN mkdir -p /app/logs
+
+COPY entrypoint.sh /entrypoint.sh
+RUN chmod +x /entrypoint.sh
+
+EXPOSE 4400 4401 4402
+
+ENTRYPOINT ["/entrypoint.sh"]
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -277,6 +277,214 @@ you may set the following environment variables to configure the two servers
    where the MCP server can be reached. The Penpot MCP Plugin uses this to construct
    the WebSocket URL as `ws://<your-address>:<port>` (default port: `4402`).
 
+## Docker Deployment
+
+You can run the Penpot MCP server as a Docker container alongside your existing
+Penpot docker-compose stack. The container includes both the MCP server and
+the plugin UI server.
+
+### How It Works
+
+```
+MCP Client (e.g. Claude Code) ──(HTTPS/SSE)──► MCP Server (:4401)
+                                                    ▲
+                                                    │ WebSocket (:4402)
+                                                    ▼
+Browser (Penpot) ◄──────────────────────────── Plugin UI (:4400)
+```
+
+The MCP server acts as a bridge between your AI client and Penpot.
+The Penpot MCP Plugin runs inside your browser within Penpot's UI and must
+remain open while using the MCP server. The plugin communicates with the
+MCP server via WebSocket to execute design operations.
+
+### Prerequisites
+
+- A running Penpot instance (docker-compose based)
+- A reverse proxy (e.g. Traefik) for TLS termination (recommended for remote access)
+- A DNS record for the MCP server (e.g. `mcp.yourdomain.com`)
+
+### Files
+
+The Docker setup requires two files in the `mcp/` directory, alongside the
+existing source code:
+
+- `Dockerfile` — multi-stage build (build + runtime)
+- `entrypoint.sh` — starts the MCP server and plugin UI server
+
+### Building
+
+The plugin's WebSocket URL is baked in at build time. Set the `WS_URI` build
+argument to match how your browser will reach the WebSocket endpoint:
+
+```shell
+# For localhost development
+docker compose build penpot-mcp
+
+# For production behind a reverse proxy
+docker compose build --build-arg WS_URI="wss://mcp.yourdomain.com/ws" penpot-mcp
+```
+
+### docker-compose Service Definition
+
+Add the following service to your `docker-compose.yml`:
+
+```yaml
+  penpot-mcp:
+    build:
+      context: ./mcp
+      dockerfile: Dockerfile
+      args:
+        WS_URI: "wss://${PENPOT_MCP_DOMAIN_NAME}/ws"
+    restart: always
+    expose:
+      - 4400
+      - 4401
+      - 4402
+    networks:
+      - penpot
+      - proxy
+    labels:
+      - "traefik.enable=true"
+      - "traefik.docker.network=proxy"
+
+      # MCP SSE + HTTP endpoint (AI clients connect here)
+      - "traefik.http.routers.penpot-mcp.rule=Host(`${PENPOT_MCP_DOMAIN_NAME}`)"
+      - "traefik.http.routers.penpot-mcp.entrypoints=websecure"
+      - "traefik.http.routers.penpot-mcp.tls=true"
+      - "traefik.http.routers.penpot-mcp.tls.certresolver=tls_resolver"
+      - "traefik.http.routers.penpot-mcp.service=penpot-mcp"
+      - "traefik.http.routers.penpot-mcp.priority=1"
+      - "traefik.http.services.penpot-mcp.loadbalancer.server.port=4401"
+
+      # WebSocket (plugin in browser connects here)
+      - "traefik.http.routers.penpot-mcp-ws.rule=Host(`${PENPOT_MCP_DOMAIN_NAME}`) && PathPrefix(`/ws`)"
+      - "traefik.http.routers.penpot-mcp-ws.entrypoints=websecure"
+      - "traefik.http.routers.penpot-mcp-ws.tls=true"
+      - "traefik.http.routers.penpot-mcp-ws.tls.certresolver=tls_resolver"
+      - "traefik.http.routers.penpot-mcp-ws.service=penpot-mcp-ws"
+      - "traefik.http.routers.penpot-mcp-ws.priority=2"
+      - "traefik.http.services.penpot-mcp-ws.loadbalancer.server.port=4402"
+
+      # Plugin UI (load into Penpot's plugin manager)
+      - "traefik.http.routers.penpot-mcp-plugin.rule=Host(`${PENPOT_MCP_DOMAIN_NAME}`) && PathPrefix(`/plugin`)"
+      - "traefik.http.routers.penpot-mcp-plugin.entrypoints=websecure"
+      - "traefik.http.routers.penpot-mcp-plugin.tls=true"
+      - "traefik.http.routers.penpot-mcp-plugin.tls.certresolver=tls_resolver"
+      - "traefik.http.routers.penpot-mcp-plugin.service=penpot-mcp-plugin"
+      - "traefik.http.routers.penpot-mcp-plugin.priority=3"
+      - "traefik.http.middlewares.mcp-plugin-strip.stripprefix.prefixes=/plugin"
+      - "traefik.http.routers.penpot-mcp-plugin.middlewares=mcp-plugin-strip"
+      - "traefik.http.services.penpot-mcp-plugin.loadbalancer.server.port=4400"
+
+    environment:
+      PENPOT_MCP_SERVER_HOST: "0.0.0.0"
+      PENPOT_MCP_SERVER_PORT: "4401"
+      PENPOT_MCP_WEBSOCKET_PORT: "4402"
+      PENPOT_MCP_LOG_LEVEL: "info"
+      PENPOT_MCP_REMOTE_MODE: "true"
+```
+
+Add to your `.env` file:
+
+```
+PENPOT_MCP_DOMAIN_NAME=mcp.yourdomain.com
+```
+
+### Connecting MCP Clients
+
+#### Claude Code
+
+Claude Code supports HTTP transport natively:
+
+```shell
+claude mcp add penpot -t http https://mcp.yourdomain.com/mcp
+```
+
+Or add to `~/.claude/settings.json` (or project `.claude/settings.json`):
+
+```json
+{
+    "mcpServers": {
+        "penpot": {
+            "type": "sse",
+            "url": "https://mcp.yourdomain.com/sse"
+        }
+    }
+}
+```
+
+#### Claude Desktop
+
+Since Claude Desktop only supports stdio transport, use `mcp-remote` as a proxy:
+
+```json
+{
+    "mcpServers": {
+        "penpot": {
+            "command": "npx",
+            "args": ["-y", "mcp-remote", "https://mcp.yourdomain.com/sse"]
+        }
+    }
+}
+```
+
+### Loading the Plugin in Penpot
+
+1. Open Penpot in your browser and navigate to a design file
+2. Open the Plugins menu
+3. Add the plugin URL: `https://mcp.yourdomain.com/plugin/manifest.json`
+4. Open the plugin UI and click "Connect to MCP server"
+5. The connection status should change to "Connected to MCP server"
+
+> [!IMPORTANT]
+> Do not close the plugin's UI while using the MCP server, as this will
+> disconnect the WebSocket and the AI client will lose access to your designs.
+
+### Without a Reverse Proxy
+
+If you are running Penpot without a reverse proxy (e.g. for local development
+with Docker), you can expose the ports directly instead:
+
+```yaml
+  penpot-mcp:
+    build:
+      context: ./mcp
+      dockerfile: Dockerfile
+      # Default WS_URI (ws://localhost:4402) works for localhost
+    restart: always
+    ports:
+      - "4400:4400"
+      - "4401:4401"
+      - "4402:4402"
+    networks:
+      - penpot
+    environment:
+      PENPOT_MCP_SERVER_HOST: "0.0.0.0"
+      PENPOT_MCP_SERVER_PORT: "4401"
+      PENPOT_MCP_WEBSOCKET_PORT: "4402"
+      PENPOT_MCP_LOG_LEVEL: "info"
+      PENPOT_MCP_REMOTE_MODE: "true"
+```
+
+Then connect Claude Code with:
+
+```shell
+claude mcp add penpot -t http http://localhost:4401/mcp --allow-http
+```
+
+And load the plugin in Penpot using `http://localhost:4400/manifest.json`.
+
+### Without the Plugin UI Server (Port 4400)
+
+If you prefer to serve the plugin UI separately (e.g. from another web server
+or CDN), you can skip port 4400 in the container. In that case:
+
+1. Build the plugin locally: `cd packages/plugin && WS_URI="wss://mcp.yourdomain.com/ws" pnpm run build`
+2. Deploy the contents of `packages/plugin/dist/` to any static file host
+3. Point Penpot's plugin manager to that host's `manifest.json`
+4. The Docker container only needs to expose ports 4401 and 4402
+
 ## Development
 
 * The [contribution guidelines for Penpot](../CONTRIBUTING.md) apply
--- a/mcp/entrypoint.sh
+++ b/mcp/entrypoint.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+set -e
+
+PLUGIN_PORT=${PLUGIN_PORT:-4400}
+
+cleanup() {
+    echo "Shutting down..."
+    kill "$MCP_PID" "$PLUGIN_PID" 2>/dev/null || true
+    wait "$MCP_PID" "$PLUGIN_PID" 2>/dev/null || true
+    exit 0
+}
+trap cleanup SIGTERM SIGINT
+
+# Start MCP server (ports 4401 + 4402 via env vars)
+cd /app
+node packages/server/dist/index.js &
+MCP_PID=$!
+
+# Serve plugin static files (manifest.json, plugin.js, index.html)
+serve -s /plugin-dist -l "$PLUGIN_PORT" --no-clipboard &
+PLUGIN_PID=$!
+
+echo "MCP server started (SSE: ${PENPOT_MCP_SERVER_PORT:-4401}, WS: ${PENPOT_MCP_WEBSOCKET_PORT:-4402})"
+echo "Plugin UI: http://0.0.0.0:${PLUGIN_PORT}"
+
+wait $MCP_PID $PLUGIN_PID
--- a/mcp/entrypoint.sh
+++ b/mcp/entrypoint.sh
@@ -12,8 +12,9 @@ cleanup() {
 trap cleanup 15 2
 
 # Start MCP server (ports 4401 + 4402 via env vars)
-cd /app
-node packages/server/dist/index.js &
+# CWD must be the server dist dir — ConfigurationLoader resolves data/ relative to process.cwd()
+cd /app/packages/server/dist
+node index.js &
 MCP_PID=$!
 
 # Serve plugin static files (manifest.json, plugin.js, index.html)
--- a/mcp/entrypoint.sh
+++ b/mcp/entrypoint.sh
@@ -9,7 +9,7 @@ cleanup() {
     wait "$MCP_PID" "$PLUGIN_PID" 2>/dev/null || true
     exit 0
 }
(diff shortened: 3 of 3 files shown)


opcode81:
@dollro as already mentioned in the discussion, I should note that containerisation is not really the most straightforward way to facilitate running the MCP server.

We are already working on two ways to facilitate MCP usage:
* Since the MCP server has only node as a fundamental dependency, **a simple way to run the MCP server locally is to use `npx`** to run it based on a published artefact. We recently published the first artefacts, such that you can already run `npx -y @penpot/mcp` to run the MCP server without any additional requirements. (Note that this also makes containerisation trivial, as you can simply run this in the container of your choice.)
* For users who don't require local execution of the MCP server, Kaleidos will be providing a **remote MCP server** which any Penpot user will be able to connect to by simply enabling it in the Penpot account options. We will soon launch a [beta-test](https://community.penpot.app/t/push-the-penpot-mcp-to-its-limits-join-the-beta-test/10363) of this integration and the MCP server in general.

niwinz:
Additionally: the MCP is already dockerized https://github.com/penpot/penpot/blob/develop/docker/images/Dockerfile.mcp

We still need to add documentation for it for on-premise users on our help center...

dollro:
thanks @niwinz , @opcode81  - 
still believe the PR solves some gap for self-hosted users behind a reverse proxy. Once the team documents the on-premise Docker setup you mentioned, the WS_URI build arg will be the natural solution for anyone running behind nginx/Traefik/Caddy with TLS.

PS: what will the docker image path to be used in a docker compose stack?
```
penpot-mcp:
   image: penpotapp/mcp:latest   # or whatever  ??
```

dollro:
Thanks for the feedback @opcode81 @niwinz — good to know about `npx` and the existing `Dockerfile.mcp`.

I went ahead and documented my full journey getting the MCP server running **self-hosted behind Traefik with TLS** (no localhost involved). It took solving **9 distinct issues** — from workspace dependency errors in the npm package, to CSP header blocking, to the critical discovery that **Penpot resolves plugin URLs relative to the manifest URL's origin, not its path** (which breaks any subpath-based hosting).

All config files below are anonymized and ready to use. Even if this PR doesn't get merged as-is, hopefully these findings are useful for the on-premise documentation you mentioned.

---

## Files

<details>
<summary><strong>README.md</strong> — Overview & quick start guide</summary>

````markdown
# Self-Hosting Penpot MCP Server Behind a Reverse Proxy (Traefik)

> **A field report and working reference for the Penpot maintainers.**
>
> Tested with Penpot 2.13.3, `@penpot/mcp@2.13.3`, Traefik v3, Debian 12.
> March 2026.

---

## TL;DR

Getting the Penpot MCP server running locally is straightforward. Getting it to
work **self-hosted behind a TLS reverse proxy on a separate subdomain** required
solving 9 distinct issues. This guide documents every issue encountered and
provides working config files.

### What this guide covers

- A `Dockerfile.mcp` that builds both the MCP server and plugin from the
  published `@penpot/mcp` npm package
- A `docker-compose.yaml` with Traefik labels for three distinct routes
- The critical discovery that **Penpot resolves plugin URLs relative to the
  manifest URL's origin, not its path** — which breaks any subpath-based hosting
- The nginx-based solution for serving static files and WebSocket on a single port

---

## Architecture

```
                         Browser (penpot.example.com)
                           |
         +-----------------+-----------------+
         |                 |                 |
   plugin.js (SES)   iframe (UI)       WebSocket
   fetched from /    loaded from /     connects to /
         |                 |                 |
         +--------+--------+---------+-------+
                  |                  |
            Traefik (TLS)      Traefik (TLS)
                  |                  |
       /mcp, /sse -> :4401     /* -> :4400
       (MCP Express server)    (nginx in container)
                                    |
                          +---------+---------+
                          |                   |
                    static files        WS upgrade
                   (plugin assets)    proxy -> :4402
                                    (node WS server)
```

**Domains used in this guide (replace with your own):**

| Purpose         | Domain                        |
|-----------------|-------------------------------|
| Penpot UI       | `penpot.example.com`          |
| MCP Server      | `penpot-mcp.example.com`      |

---

## Files Included

| File | Purpose |
|------|---------|
| `Dockerfile.mcp` | Builds MCP server + plugin from `@penpot/mcp` npm package |
| `docker-compose.mcp.yaml` | The `penpot-mcp` service definition with Traefik labels |
| `.env.example` | Environment variables (MCP-relevant subset) |
| `traefik/penpot-middlewares.yaml` | Traefik dynamic config: CSP headers that allow the plugin iframe |
| `ISSUES.md` | Detailed description of every issue encountered |

---

## Quick Start

### Prerequisites

- Penpot running behind Traefik with TLS (HTTPS)
- A second subdomain for the MCP server (e.g., `penpot-mcp.example.com`)
- DNS for both subdomains pointing to the same server

### 1. Add environment variables

Add to your Penpot `.env`:

```bash
PENPOT_MCP_DOMAIN_NAME='penpot-mcp.example.com'
```

`PENPOT_VERSION` is already used by Penpot images and is reused for the MCP
package version (they track Penpot releases).

### 2. Place `Dockerfile.mcp` next to your `docker-compose.yaml`

### 3. Add the `penpot-mcp` service to your compose file

See `docker-compose.mcp.yaml` — merge the `penpot-mcp` service into your
existing Penpot compose file.

### 4. Configure Traefik CSP headers

The Penpot frontend loads the MCP plugin in an **iframe**. If your Traefik
middleware sets `frameDeny: true` or a restrictive `frame-ancestors` CSP, the
iframe will be blocked.

You need a middleware for the Penpot frontend that allows:

```
Content-Security-Policy: frame-ancestors 'self' https://penpot-mcp.example.com;
```

**Important:** Remove `frameDeny: true` from the middleware used by the Penpot
frontend — it sets `X-Frame-Options: DENY` which overrides the more permissive
CSP `frame-ancestors` directive.

See `traefik/penpot-middlewares.yaml` for a working example.

### 5. Build and start

```bash
docker compose build penpot-mcp
docker compose up -d
```

### 6. Install the plugin in Penpot

In the Penpot UI: **Plugin Manager** (or Ctrl+Alt+P) -> install from:

```
https://penpot-mcp.example.com/manifest.json
```

> **Note:** The manifest URL must be at the **root** of the MCP domain, not under
> a subpath. See "Issue 3" in ISSUES.md for why.

### 7. Connect Claude Code

```bash
claude mcp add penpot -t http https://penpot-mcp.example.com/mcp
```

---

## Key Insight for Penpot Maintainers

### Plugin URL resolution breaks subpath hosting

In `app.plugins.register/parse-manifest`:

```clojure
origin (obj/get (js/URL. plugin-url) "origin")
```

The `host` field passed to `ɵLoadPlugin` is the **origin** of the manifest URL
(scheme + host + port), discarding the path entirely. Then:

- `code: "plugin.js"` resolves to `https://domain/plugin.js` (root)
- `penpot.ui.open("...", "?theme=dark...")` resolves to `https://domain/?theme=dark` (root)

This means **all plugin assets must be served from the root path of their
domain**. Hosting a plugin under a subpath (e.g., `/plugin/manifest.json`) will
cause:

1. `plugin.js` to be fetched from the root -> wrong server/404/HTML response
2. SES sandbox receives HTML -> `SyntaxError: Unexpected token '<'`
3. UI iframe URL hits the wrong backend -> `426 Upgrade Required`

**Suggestion:** Consider resolving relative URLs in the manifest against the
manifest URL's **full path** (directory), not just the origin. This would allow
plugin hosting under subpaths and simplify reverse proxy setups.

---

## Version Compatibility Note

The `@penpot/mcp` npm package versions track Penpot release versions. As of
March 2026:

| npm package | Docker images available? |
|-------------|------------------------|
| `@penpot/mcp@2.13.3` | Yes (`penpotapp/*:2.13.3`) |
| `@penpot/mcp@2.14.0` | **No** (not yet released to Docker Hub) |

Always verify that both the npm package and the Docker images exist for your
target version before upgrading.
````

</details>

<details>
<summary><strong>ISSUES.md</strong> — All 9 issues encountered (with root causes & fixes)</summary>

````markdown
# Issues Encountered Self-Hosting Penpot MCP Behind Traefik

Each issue below was encountered in sequence while setting up the Penpot MCP
server on a self-hosted instance behind Traefik v3 with TLS. Issues are numbered
in discovery order.

---

## Issue 1: pnpm install fails — workspace dependency in published npm package

**Symptom:**
ERR_PNPM_MISSING_PACKAGE_NAME  Can't install file:..: Missing package name

**Root cause:** `packages/server/package.json` in the `@penpot/mcp` npm package
contains `"penpot-mcp": "file:.."` — a pnpm workspace reference that points to
the monorepo root. When the server is copied to an isolated Docker build stage,
this reference is invalid.

**Impact:** Cannot build the Docker image at all.

**Workaround (in Dockerfile):**
```dockerfile
RUN node -e "\
  const pkg = JSON.parse(require('fs').readFileSync('package.json','utf8')); \
  delete pkg.dependencies['penpot-mcp']; \
  delete pkg.devDependencies; \
  require('fs').writeFileSync('package.json', JSON.stringify(pkg, null, 2));"
```

This is safe because esbuild already bundles the dependency into `dist/index.js`.

**Suggested fix for Penpot:** Either remove the `file:..` dependency from the
published package, or add a `postpack` script that strips it.

---

## Issue 2: Penpot frontend CSP blocks plugin iframe

**Symptom:** Plugin installs successfully but nothing happens when clicking it.
No visible error unless you check the browser console.

**Root cause:** Many self-hosted setups use security middleware that sets
`frameDeny: true` and/or `Content-Security-Policy: "frame-ancestors 'self';"`.
The Penpot MCP plugin loads in an **iframe** from a different origin
(`penpot-mcp.example.com`), which is blocked by these headers.

**Impact:** Plugin panel never opens.

**Fix:** Create a Penpot-specific middleware that allows the MCP domain:
```yaml
Content-Security-Policy: "frame-ancestors 'self' https://penpot-mcp.example.com;"
```
Also remove `frameDeny: true` — it produces `X-Frame-Options: DENY` which
overrides CSP `frame-ancestors`.

**Note for Penpot docs:** This should be documented as a requirement for
self-hosted plugin support. Any instance that wants to use plugins from a
different origin needs to adjust its CSP headers.

---

## Issue 3: Plugin URLs resolve relative to ORIGIN, not manifest path (critical)

**Symptom:** `SyntaxError: Unexpected token '<'` in the SES sandbox when loading
the plugin, even though the JS file serves correctly in a direct browser request.

**Root cause:** In `app.plugins.register/parse-manifest`:
```clojure
origin (obj/get (js/URL. plugin-url) "origin")
```

The `host` field is set to the **origin** of the manifest URL, discarding the
path entirely. For manifest URL `https://domain/plugin/manifest.json`:
- Expected: `host = https://domain/plugin/`
- Actual: `host = https://domain`

Then `ɵLoadPlugin` constructs:
- `code` URL: `https://domain/plugin.js` (fetched from root, not `/plugin/`)
- UI iframe: `https://domain/?theme=dark` (root, not `/plugin/?theme=dark`)

If the root path serves something other than plugin files (e.g., a WebSocket
server returning 426), the SES sandbox receives HTML/error instead of JS.

**Impact:** Any attempt to host plugins under a subpath fails silently with
cryptic SES errors.

**Workaround:** Serve all plugin files from the **root path** of their domain.
This requires a dedicated subdomain for the MCP server.

**Suggested fix for Penpot:** Resolve `code` and UI URLs relative to the
manifest URL's directory (like `new URL("plugin.js", manifestUrl)`), not just
the origin. This would allow hosting plugins under subpaths and simplify
reverse proxy configurations.

---

## Issue 4: Static files and WebSocket need the same root path

**Symptom:** After fixing issue 3 by serving files at root, WebSocket
connections fail because the catch-all Traefik route can only point to one
backend.

**Root cause:** The MCP server has three ports:
- `:4400` — Plugin static files
- `:4401` — MCP HTTP/SSE (Express)
- `:4402` — WebSocket (node ws)

The browser plugin connects via `wss://domain/` (root) to port 4402, while
static files also need to be served from root on port 4400. A single Traefik
`Host()` catch-all can only route to one backend.

**Approaches that failed:**
- Two catch-all routes with different Traefik priorities
- `HeadersRegexp(Connection, upgrade)` — unreliable with HTTP/2
- `serve -s` (npm) — can't proxy WebSocket

**Solution:** Replace `serve` with **nginx** inside the container. nginx on port
4400 uses `map $http_upgrade` to detect WebSocket upgrades and proxy them to
port 4402, while serving static files for all other requests.

---

## Issue 5: nginx try_files swallows WebSocket upgrades

**Symptom:** After adding nginx, WebSocket still fails. Browser reports
"Disconnected from MCP server".

**Root cause:** Initial nginx config:
```nginx
try_files $uri $uri/ @websocket;
```
For a WebSocket request to `/`, `try_files` finds `index.html` (via the `index`
directive) and serves it — the `@websocket` fallback is never reached.

**Fix:** Use `map` + `if` to intercept WebSocket upgrades **before** `try_files`:
```nginx
map $http_upgrade $backend {
    default        "";
    "websocket"    "ws";
}
location / {
    if ($backend = "ws") {
        proxy_pass http://ws_backend;
    }
    try_files $uri $uri/ /index.html;
}
```

---

## Issue 6: Traefik caches stale route definitions

**Symptom:** After removing/renaming Traefik router labels and recreating
containers, old routes still active. Requests go to the wrong backend.

**Root cause:** Traefik's Docker provider detects container changes but may
retain label-based route definitions from previous container versions until
Traefik itself restarts.

**Fix:** Run `docker restart traefik` after changing container labels,
especially when removing routes.

---

## Issue 7: nginx pid directive conflict

**Symptom:** nginx fails to start inside the container:
`[emerg] "pid" directive is duplicate in /etc/nginx/nginx.conf`

**Root cause:** The CMD uses `nginx -g 'pid /tmp/nginx.pid;'` and the
`nginx.conf` also contained `pid /tmp/nginx.pid;`. Debian's default
`/etc/nginx/sites-enabled/default` may also conflict.

**Fix:**
1. Remove `pid` from `nginx.conf` (set it only in CMD via `-g`)
2. Remove default site config: `rm -f /etc/nginx/sites-enabled/default`

---

## Issue 8: nginx map values conflict (case sensitivity)

**Symptom:**
`[emerg] conflicting parameter "websocket" in nginx.conf`

**Root cause:** Initial `map` block had multiple cases:
```nginx
map $http_upgrade $backend {
    websocket 1;
    Websocket 1;   # conflicts!
    WebSocket 1;   # conflicts!
}
```

nginx's `map` is case-insensitive by default for header values.

**Fix:** Use only one value: `"websocket"`.

---

## Issue 9: MCP endpoint requires specific Accept headers

**Symptom:** `curl` to `/mcp` returns 406 Not Acceptable or empty response.

**Root cause:** The MCP Streamable HTTP protocol requires:
`Accept: application/json, text/event-stream`

**Impact:** Only affects manual testing. Claude Code and other MCP clients send
the correct headers automatically.

---

## Summary

| Nr | Issue | Severity | Who should fix |
|-|-|-|-|
| 1 | Workspace dep in npm package | Build blocker | Penpot (npm package) |
| 2 | CSP blocks plugin iframe | Silent failure | Docs (self-hosted guide) |
| 3 | URL resolution uses origin only | Design issue | Penpot (plugin loader) |
| 4 | Static + WS on same root path | Architecture | Docs / Dockerfile |
| 5 | try_files swallows WS upgrades | Config issue | Dockerfile (nginx) |
| 6 | Traefik stale routes | Ops gotcha | Docs |
| 7 | nginx pid conflict | Config issue | Dockerfile |
| 8 | nginx map case conflict | Config issue | Dockerfile |
| 9 | MCP Accept headers | Minor | Docs |
````

</details>

<details>
<summary><strong>Dockerfile.mcp</strong></summary>

````dockerfile
# ─────────────────────────────────────────────────────────────
# Penpot MCP Server — from official @penpot/mcp npm package
#
# The official CI (scripts/build) only builds the server.
# The plugin is intended to be hosted by Kaleidos (cloud).
# For self-hosted behind a reverse proxy, we also need to build
# and serve the plugin — with WS_URI baked in for wss:// support.
#
# Build:
#   docker build \
#     --build-arg WS_URI=wss://penpot-mcp.example.com \
#     --build-arg PENPOT_VERSION=2.13.3 \
#     -t penpot-mcp .
#
# Runtime:
#   :4400  nginx (plugin static files + WebSocket proxy)
#   :4401  MCP HTTP/SSE (node — Claude Code connects here)
#   :4402  WebSocket (node — browser plugin connects here)
# ─────────────────────────────────────────────────────────────

# ── Build stage ──────────────────────────────────────────────
FROM node:22-slim AS builder

ARG WS_URI="ws://localhost:4402"
ARG PENPOT_VERSION="2.13.3"

RUN corepack enable && corepack prepare pnpm@latest --activate

WORKDIR /build

# Unpack the published @penpot/mcp source package
# Version tracks PENPOT_VERSION from .env so they stay in sync
RUN npm pack @penpot/mcp@${PENPOT_VERSION} && \
    tar xzf penpot-mcp-*.tgz --strip-components=1 && \
    rm penpot-mcp-*.tgz

# Install all dependencies (dev included, needed for build)
RUN pnpm -r install

# Build the MCP server
# (mirrors official scripts/build: esbuild + copy-resources)
RUN pnpm -r --filter "mcp-server" run build

# Build the plugin with WS_URI baked into the JS
ENV WS_URI=${WS_URI}
RUN pnpm -r --filter "mcp-plugin" run build

# No patching needed — plugin files are served at root via nginx


# ── Server bundle (flat, like official scripts/build output) ─
FROM node:22-slim AS server-deps

RUN corepack enable && corepack prepare pnpm@latest --activate

WORKDIR /server

# Copy the built server bundle
COPY --from=builder /build/packages/server/dist/ ./
COPY --from=builder /build/packages/server/package.json ./
COPY --from=builder /build/packages/server/pnpm-lock.yaml ./

# Empty workspace file so pnpm install works
RUN echo "" > pnpm-workspace.yaml

# Remove workspace references that break outside the monorepo.
# The server package.json has "penpot-mcp": "file:.." which points to the
# monorepo root — this doesn't exist in isolation. The dependency is already
# bundled into dist/index.js by esbuild, so it's not needed at runtime.
RUN node -e "\
  const pkg = JSON.parse(require('fs').readFileSync('package.json','utf8')); \
  delete pkg.dependencies['penpot-mcp']; \
  delete pkg.devDependencies; \
  require('fs').writeFileSync('package.json', JSON.stringify(pkg, null, 2));"

# Install production deps only (same as official setup script)
RUN pnpm install -P


# ── Runtime ──────────────────────────────────────────────────
FROM node:22-slim

RUN apt-get update && apt-get install -y --no-install-recommends nginx && \
    rm -rf /var/lib/apt/lists/*

RUN groupadd -g 1001 penpot && \
    useradd -u 1001 -g penpot -m -s /bin/false penpot

# MCP server (flat bundle with prod node_modules)
COPY --from=server-deps --chown=penpot:penpot /server /opt/penpot/mcp-server

# Plugin static files (built with WS_URI)
COPY --from=builder --chown=penpot:penpot /build/packages/plugin/dist /opt/penpot/mcp-plugin

# Nginx: plugin static files + WebSocket proxy on single port (4400)
#
# WHY NGINX?
# Penpot's plugin loader resolves all URLs (code, UI iframe, assets) relative
# to the manifest URL's *origin* (not path). This means plugin files MUST be
# served from the root path ("/") of their domain.
#
# The WebSocket server (port 4402) also needs to be reachable at the root.
# A simple static file server (like `serve`) can't proxy WebSocket upgrades.
# nginx handles both: static files for regular requests, WS proxy for upgrades.
#
# Remove default config that conflicts with our custom nginx.conf
RUN rm -f /etc/nginx/sites-enabled/default /etc/nginx/conf.d/default.conf
RUN cat > /etc/nginx/nginx.conf <<'NGINX'
worker_processes 1;
events { worker_connections 128; }
http {
    include /etc/nginx/mime.types;
    default_type application/octet-stream;
    access_log /dev/stdout;
    error_log /dev/stderr;

    map $http_upgrade $backend {
        default        "";
        "websocket"    "ws";
    }

    upstream ws_backend { server 127.0.0.1:4402; }

    server {
        listen 4400;
        root /opt/penpot/mcp-plugin;
        index index.html;

        location / {
            # WebSocket upgrades go to the node WS server
            if ($backend = "ws") {
                proxy_pass http://ws_backend;
            }

            proxy_http_version 1.1;
            proxy_set_header Connection "upgrade";
            proxy_set_header Host $host;
            proxy_read_timeout 86400;

            # Non-WebSocket requests serve static files
            add_header Access-Control-Allow-Origin * always;
            add_header Access-Control-Allow-Methods "GET, OPTIONS" always;
            try_files $uri $uri/ /index.html;
        }
    }
}
NGINX

RUN mkdir -p /opt/penpot/mcp-server/logs /var/log/nginx /var/lib/nginx/body && \
    chown -R penpot:penpot /opt/penpot /var/log/nginx /var/lib/nginx /run

USER penpot:penpot

EXPOSE 4400 4401 4402

CMD ["sh", "-c", \
     "nginx -g 'daemon off; pid /tmp/nginx.pid;' & \
      cd /opt/penpot/mcp-server && exec node index.js"]
````

</details>

<details>
<summary><strong>docker-compose.mcp.yaml</strong></summary>

````yaml
# ─────────────────────────────────────────────────────────────
# Penpot MCP service definition for docker-compose
#
# Merge this into your existing Penpot docker-compose.yaml.
# Requires Traefik as the reverse proxy with TLS configured.
#
# Replace "example.com" with your actual domain throughout.
# ─────────────────────────────────────────────────────────────

services:
  penpot-mcp:
    build:
      context: .
      dockerfile: Dockerfile.mcp
      args:
        WS_URI: "wss://${PENPOT_MCP_DOMAIN_NAME}"
        PENPOT_VERSION: "${PENPOT_VERSION}"
    image: penpot-mcp:latest
    restart: always
    expose:
      - 4400
      - 4401
      - 4402
    networks:
      - penpot
      - proxy
    labels:
      - "traefik.enable=true"
      - "traefik.docker.network=proxy"

      # ── Middlewares ──

      # SSE/streaming: disable buffering so entrypoint gzip doesn't break SSE
      - "traefik.http.middlewares.mcp-sse.headers.customResponseHeaders.X-Accel-Buffering=no"
      - "traefik.http.middlewares.mcp-sse.headers.customResponseHeaders.Cache-Control=no-cache, no-transform"

      # ── MCP Streamable HTTP ──
      # Claude Code connects here:
      #   claude mcp add penpot -t http https://penpot-mcp.example.com/mcp
      - "traefik.http.routers.penpot-mcp-http.rule=Host(`${PENPOT_MCP_DOMAIN_NAME}`) && PathPrefix(`/mcp`)"
      - "traefik.http.routers.penpot-mcp-http.entrypoints=websecure"
      - "traefik.http.routers.penpot-mcp-http.tls=true"
      - "traefik.http.routers.penpot-mcp-http.tls.certresolver=tls_resolver"
      - "traefik.http.routers.penpot-mcp-http.service=penpot-mcp-http"
      - "traefik.http.routers.penpot-mcp-http.middlewares=mcp-sse@docker"
      - "traefik.http.services.penpot-mcp-http.loadbalancer.server.port=4401"

      # ── SSE legacy ──
      - "traefik.http.routers.penpot-mcp-sse.rule=Host(`${PENPOT_MCP_DOMAIN_NAME}`) && PathPrefix(`/sse`)"
      - "traefik.http.routers.penpot-mcp-sse.entrypoints=websecure"
      - "traefik.http.routers.penpot-mcp-sse.tls=true"
      - "traefik.http.routers.penpot-mcp-sse.tls.certresolver=tls_resolver"
      - "traefik.http.routers.penpot-mcp-sse.service=penpot-mcp-sse"
      - "traefik.http.routers.penpot-mcp-sse.middlewares=mcp-sse@docker"
      - "traefik.http.services.penpot-mcp-sse.loadbalancer.server.port=4401"

      # ── Catch-all: nginx serves plugin static files + proxies WebSocket ──
      #
      # IMPORTANT: Plugin files must be served at the ROOT of the domain.
      # Penpot resolves plugin URLs relative to the manifest URL's origin
      # (not its path), so hosting under a subpath like /plugin/ will break.
      #
      # nginx on :4400 handles both static files and WS upgrade proxying.
      - "traefik.http.routers.penpot-mcp-default.rule=Host(`${PENPOT_MCP_DOMAIN_NAME}`)"
      - "traefik.http.routers.penpot-mcp-default.entrypoints=websecure"
      - "traefik.http.routers.penpot-mcp-default.tls=true"
      - "traefik.http.routers.penpot-mcp-default.tls.certresolver=tls_resolver"
      - "traefik.http.routers.penpot-mcp-default.service=penpot-mcp-default"
      - "traefik.http.services.penpot-mcp-default.loadbalancer.server.port=4400"

    environment:
      PENPOT_MCP_SERVER_LISTEN_ADRESS: "0.0.0.0"
      PENPOT_MCP_SERVER_ADDRESS: "${PENPOT_MCP_DOMAIN_NAME:-localhost}"
      PENPOT_MCP_SERVER_PORT: "4401"
      PENPOT_MCP_WEBSOCKET_PORT: "4402"
      PENPOT_MCP_LOG_LEVEL: "info"
      PENPOT_MCP_REMOTE_MODE: "true"
````

</details>

<details>
<summary><strong>.env.example</strong></summary>

````bash
# ─────────────────────────────────────────────────────────────
# Penpot + MCP environment variables (example)
#
# Copy to .env and edit with your values.
# ─────────────────────────────────────────────────────────────

# Penpot version — used for both Docker images AND @penpot/mcp npm package.
# They must match. Verify both exist before changing:
#   Docker Hub: https://hub.docker.com/r/penpotapp/frontend/tags
#   npm:        https://www.npmjs.com/package/@penpot/mcp?activeTab=versions
PENPOT_VERSION=2.13.3

# Domain names
PENPOT_PUBLIC_DOMAIN_NAME='penpot.example.com'
PENPOT_PUBLIC_URI='https://penpot.example.com'
PENPOT_MCP_DOMAIN_NAME='penpot-mcp.example.com'

# Secret key — generate with: python3 -c "import secrets; print(secrets.token_urlsafe(64))"
PENPOT_SECRET_KEY='CHANGE_ME_generate_a_random_secret'

# Flags
PENPOT_FLAGS='enable-smtp enable-prepl-server disable-secure-session-cookies'

# SMTP (optional — without this, use the mailcatch service)
# PENPOT_SMTP_DEFAULT_FROM=no-reply@example.com
# PENPOT_SMTP_HOST=smtp.example.com
# PENPOT_SMTP_PORT=587
# PENPOT_SMTP_USERNAME=
# PENPOT_SMTP_PASSWORD=
# PENPOT_SMTP_TLS=true
# PENPOT_SMTP_SSL=false
````

</details>

<details>
<summary><strong>traefik/penpot-middlewares.yaml</strong></summary>

````yaml
# ─────────────────────────────────────────────────────────────
# Traefik dynamic configuration: Penpot-specific middlewares
#
# Place this file in your Traefik file provider directory
# (e.g., /etc/traefik/dynamic/ or wherever your Traefik reads
# dynamic config from).
#
# The Penpot frontend router should use "penpot-enhanced@file"
# instead of a generic security middleware.
# ─────────────────────────────────────────────────────────────

http:
  middlewares:
    # Security headers for Penpot — same as a typical "enhanced" middleware
    # but with frame-ancestors allowing the MCP plugin domain.
    #
    # IMPORTANT: Do NOT set frameDeny: true — it produces
    # X-Frame-Options: DENY which overrides the more permissive
    # CSP frame-ancestors directive and blocks the plugin iframe.
    penpot-security-headers:
      headers:
        browserXssFilter: true
        contentTypeNosniff: true
        forceSTSHeader: true
        stsIncludeSubdomains: true
        stsPreload: true
        stsSeconds: 31536000
        # frameDeny: true  # <-- DO NOT enable, breaks plugin iframe
        customResponseHeaders:
          Referrer-Policy: "strict-origin-when-cross-origin"
          # Allow Penpot to iframe content from the MCP plugin domain
          Content-Security-Policy: "frame-ancestors 'self' https://penpot-mcp.example.com;"
          Permissions-Policy: "geolocation=(), microphone=(), camera=()"

    # Chain: penpot security headers + gzip compression
    penpot-enhanced:
      chain:
        middlewares:
          - penpot-security-headers
          - gzip

    gzip:
      compress: {}
````

</details>

---

### Key takeaways for on-premise docs

1. **Issue 1 (build blocker):** The published `@penpot/mcp` npm package contains a `"penpot-mcp": "file:.."` workspace reference that breaks any isolated install. Should be stripped from the published artifact.

2. **Issue 3 (design issue):** Plugin URL resolution uses `origin` only, not the full manifest path. This forces a dedicated subdomain for any self-hosted plugin. Worth considering a fix in `app.plugins.register/parse-manifest`.

3. **CSP/iframe (Issue 2):** Self-hosted instances with security headers need to allow `frame-ancestors` from the plugin domain — this should be in the on-premise docs.

Happy to help refine any of this for the official documentation!

niwinz:
> Thanks for the feedback @opcode81 @niwinz — good to know about `npx` and the existing `Dockerfile.mcp`.
> 
> I went ahead and documented my full journey getting the MCP server running **self-hosted behind Traefik with TLS** (no localhost involved). It took solving **9 distinct issues** — from workspace dependency errors in the npm package, to CSP header blocking, to the critical discovery that **Penpot resolves plugin URLs relative to the manifest URL's origin, not its path** (which breaks any subpath-based hosting).

That is a bit confusing. Let me explain: to make the plugins run on a subpath, we explicitly introduced the version: 2 field and made all plugin resolution use the manifest.json location as the root.

This ensures subpaths work correctly in any scenario. Your plugin could be at http://foo.com/my-plugin/ (where the manifest.json lives) or http://foo.com/plugins/other/my-plugin/. The previous approach (version: 1 or no version at all) always assumed the plugin was running at the domain root, meaning assets were always fetched from the root—for example, a plugin at a subpath might incorrectly try to load http://foo.com/index.html instead of its own assets.

Given that context, I don’t quite understand why you say it "breaks any subpath-based hosting," when it was designed specifically to fix that and enable hosting on a subpath.

niwinz:
In any case we will look closer all the issues and try to address them. Thanks for reporting.

niwinz:
Regarding the docker image, all the plugins, including the MCP is also included and is served in a subpath: `/plugins/mcp` (you can use `https://yourdomain.where.penpot-is-deployed/plugins/mcp` (the manifest.json on the end of path can be ommited)) this all is intended to be deployed together on-premise, penpot and mcp.

niwinz:
For now closing the PR.

PR #8686, opened 2026-03-19
:bug: Default MCP listen addresses to localhost instead of 0.0.0.0

### LINKED ISSUE #8683: Modified MCP listen address defaults are potentially problematic and undocumented
**MCP Server**. Two commits have changed the MCP server configuration:
* 490619119eac50b62cc4c335666a6b330c8c678f: Changed the default listen addresses from localhost to `0.0.0.0`
* 0027e9031a25e43eb2087cecf9d3892644aec35c: Changed environment variable names

**Plugin Server**. For the plugin server, the initial commit that added `./mcp` already removed the [original configuration](https://github.com/penpot/penpot-mcp/blob/fc1c77de64c39522f34b9a946d98fe791fa9477a/penpot-plugin/vite.config.ts#L35-L41) of the listen address, hardcoding it to `0.0.0.0`:
https://github.com/penpot/penpot/blob/b876417d5b76c553443f6ddc04879ab4c832db48/mcp/packages/plugin/vite.config.ts#L31-L36

**Resulting Issues**. This causes two issues:
* Listening on all interfaces is a security hazard in untrusted networks. Servers that are not guaranteed to run in an appropriately isolated environment (e.g. a docker container) should default to localhost for security reasons. Since the MCP server is intended to be executed locally by users, we should preferably stick to the old default and configure other cases explicitly by setting the environment variables.
* The [documentation](https://github.com/penpot/penpot/blob/develop/mcp/README.md#configuration) was not updated accordingly. It still mentions the old defaults and the old variables, misleading users.

@niwinz

moktamd:
Fixes #8683

The MCP server and plugin preview server were binding to `0.0.0.0` by default, exposing all network interfaces even in local-only mode. This contradicts the documented defaults and the security model described in the README.

Changes:
- `PenpotMcpServer.ts`: Read `PENPOT_MCP_SERVER_LISTEN_ADDRESS` (matching the documented env var name) instead of the undocumented `PENPOT_MCP_SERVER_HOST`, and default to `localhost`
- `vite.config.ts`: Use the `PENPOT_MCP_PLUGIN_SERVER_LISTEN_ADDRESS` env var with `localhost` fallback instead of hardcoded `0.0.0.0`

### DIFF
--- a/mcp/packages/server/src/PenpotMcpServer.ts
+++ b/mcp/packages/server/src/PenpotMcpServer.ts
@@ -75,7 +75,7 @@ export class PenpotMcpServer {
 
     constructor(private isMultiUser: boolean = false) {
         // read port configuration from environment variables
-        this.host = process.env.PENPOT_MCP_SERVER_HOST ?? "0.0.0.0";
+        this.host = process.env.PENPOT_MCP_SERVER_LISTEN_ADDRESS ?? "localhost";
         this.port = parseInt(process.env.PENPOT_MCP_SERVER_PORT ?? "4401", 10);
         this.webSocketPort = parseInt(process.env.PENPOT_MCP_WEBSOCKET_PORT ?? "4402", 10);
         this.replPort = parseInt(process.env.PENPOT_MCP_REPL_PORT ?? "4403", 10);
--- a/mcp/packages/plugin/vite.config.ts
+++ b/mcp/packages/plugin/vite.config.ts
@@ -29,7 +29,7 @@ export default defineConfig({
         },
     },
     preview: {
-        host: "0.0.0.0",
+        host: process.env.PENPOT_MCP_PLUGIN_SERVER_LISTEN_ADDRESS ?? "localhost",
         port: 4400,
         cors: true,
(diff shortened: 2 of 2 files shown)


niwinz:
This change affects on how mcp server is running on the devenv so we need to look closer to properly make this change. I will look it Monday.

mcp/packages/server/src/PenpotMcpServer.ts
```diff
@@ -75,7 +75,7 @@ export class PenpotMcpServer {
 
     constructor(private isMultiUser: boolean = false) {
         // read port configuration from environment variables
-        this.host = process.env.PENPOT_MCP_SERVER_HOST ?? "0.0.0.0";
+        this.host = process.env.PENPOT_MCP_SERVER_LISTEN_ADDRESS ?? "localhost";
```
niwinz, line 78: Question: why the rename?

The already existing name is already correct

niwinz:
Merged into `staging` with small changes (mainly env vars naming)

carolportugal18:

PR #8687, opened 2026-03-19
:bug: Fix MCP notifications when there is only one tab

luisddm:
### Related Ticket

Taiga [#13724](https://tree.taiga.io/project/penpot/task/13724)

### Summary

When there is only one workspace session active (aka one browser tab with the workspace loaded), there should never appear any notification related to the plugin connection, since it's not possible to mislead the user. The plugin may be active or not in the current tab, but not in any other places.

Also, it could happen that the MCP server is down, or has some other connection issues. In this case the user must know, so the menu bullet is now displayed in red.

### DIFF
--- a/frontend/src/app/main/data/workspace/mcp.cljs
+++ b/frontend/src/app/main/data/workspace/mcp.cljs
@@ -53,21 +53,28 @@
            (rx/take 1)
            (rx/map #(ptk/data-event ::connect))))))
 
-(defn manage-notification
-  [mcp-enabled? mcp-connected?]
-  (if mcp-enabled?
-    (if mcp-connected?
-      (rx/of (ntf/hide))
-      (rx/of (ntf/dialog :content (tr "notifications.mcp.active-tab-switching.text")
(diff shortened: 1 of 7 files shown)


frontend/src/app/main/data/workspace/mcp.cljs
```diff
@@ -53,21 +53,28 @@
...
-                                                                              ::ev/origin "workspace-notification"}))})))
-    (rx/of (ntf/hide))))
+(defn manage-mcp-notification
+  []
+  (ptk/reify ::manage-mcp-notification
+    ptk/WatchEvent
+    (watch [_ state _]
+      (let [mcp-connected? (true? (-> state :workspace-local :mcp :connection))
+            mcp-enabled?   (true? (-> state :profile :props :mcp-enabled))
+            num-sessions   (-> state :workspace-presence vals count)
```
Alotor: you don't need `vals` you can directly do `(-> state :workspace-presence count)`

Alotor (approved):

PR #8688, opened 2026-03-19
feat(mcp): add 'page' special shapeId to export_shape tool for full-page snapshots

abhishek-mittal:
## Summary

Adds support for a new special `shapeId` value — `'page'` — in the `export_shape` MCP tool.

## Motivation

When working with an AI agent via the MCP server, it is often useful to get a bird's-eye snapshot of the entire current page (e.g. to understand layout, identify shapes by visual inspection, or provide context). Previously, the only way to export everything visible was to know the ID of the root frame or to iterate over all shapes. Introducing `'page'` as a reserved identifier removes that friction.

## Changes

**`mcp/packages/server/src/tools/ExportShapeTool.ts`**

- Extended the `shapeId` parameter description to document the new `'page'` special value alongside the existing `'selection'` value.
- Updated `getToolDescription()` to explicitly mention both `'page'` and `'selection'` so LLM clients know which shortcuts are available.
- Added an `else if (args.shapeId === "page")` branch in `executeCore()` that maps the identifier to `penpot.root`, exporting the full current page.

## Usage

```json
{ "shapeId": "page", "format": "png" }
```

This will export the entire current Penpot page as a PNG snapshot — equivalent to calling `penpotUtils.exportImage(penpot.root, "shape", false)`.

## Backwards Compatibility

No breaking changes. The `'page'` identifier did not previously have any meaning (it would have been passed to `penpotUtils.findShapeById("page")` and returned `null`). Existing calls using real shape IDs or `'selection'` are unaffected.

### DIFF
--- a/mcp/packages/server/src/tools/ExportShapeTool.ts
+++ b/mcp/packages/server/src/tools/ExportShapeTool.ts
@@ -17,7 +17,8 @@ export class ExportShapeArgs {
             .min(1, "shapeId cannot be empty")
             .describe(
                 "Identifier of the shape to export. Use the special identifier 'selection' to " +
-                    "export the first shape currently selected by the user."
+                    "export the first shape currently selected by the user. " +
+                    "Use the special identifier 'page' to export the entire current page as a snapshot."
             ),
         format: z.enum(["svg", "png"]).default("png").describe("The output format, either 'png' (default) or 'svg'."),
         mode: z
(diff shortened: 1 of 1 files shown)


abhishek-mittal:
Closing — this was submitted to the wrong repository. The MCP server lives in https://github.com/penpot/penpot-mcp. Re-submitting there.

PR #8693, opened 2026-03-19
:sparkles: Add 'page' shapeId to MCP export_shape for full-page snapshot

abhishek-mittal:
## Summary of changes

This PR adds a `'page'` special shapeId to the MCP `export_shape` tool, enabling callers to capture a full snapshot of the currently active Penpot page without knowing any specific shape ID.

Before this change the tool required a concrete shape UUID. The `'page'` shorthand maps to `penpot.root`, which is the root node of the current page.

Closes https://github.com/penpot/penpot/issues/8689

## How was it tested

Manually invoked the MCP tool with `shapeId: "page"` against a running Penpot instance and confirmed a full-page PNG was returned.

## DCO

By submitting this patch, I confirm that my contribution is made according to the terms of the [Penpot contributing guide](https://github.com/penpot/penpot/blob/mcp-prod/CONTRIBUTING.md) and the DCO:

Signed-off-by: Abhishek Mittal <abhishekmittaloffice@gmail.com>

### DIFF
--- a/mcp/packages/server/src/tools/ExportShapeTool.ts
+++ b/mcp/packages/server/src/tools/ExportShapeTool.ts
@@ -16,8 +16,8 @@ export class ExportShapeArgs {
             .string()
             .min(1, "shapeId cannot be empty")
             .describe(
-                "Identifier of the shape to export. Use the special identifier 'selection' to " +
-                    "export the first shape currently selected by the user."
+                "Identifier of the shape to export. " +
+                    "Special identifiers you can use: 'selection' (first shape currently selected by the user), 'page' (entire current page)"
             ),
         format: z.enum(["svg", "png"]).default("png").describe("The output format, either 'png' (default) or 'svg'."),
         mode: z
@@ -71,7 +71,7 @@ export class ExportShapeTool extends Tool<ExportShapeArgs> {
     public getToolDescription(): string {
         let description =
             "Exports a shape (or a shape's image fill) from the Penpot design to a PNG or SVG image, " +
-            "such that you can get an impression of what it looks like. ";
+            "such that you can get an impression of what it looks like.";
         if (this.mcpServer.isFileSystemAccessEnabled()) {
             description += "\nAlternatively, you can save it to a file.";
         }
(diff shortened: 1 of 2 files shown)


mcp/packages/server/src/tools/ExportShapeTool.ts
```diff
@@ -17,7 +17,8 @@ export class ExportShapeArgs {
             .min(1, "shapeId cannot be empty")
             .describe(
                 "Identifier of the shape to export. Use the special identifier 'selection' to " +
-                    "export the first shape currently selected by the user."
+                    "export the first shape currently selected by the user. " +
+                    "Use the special identifier 'page' to export the entire current page as a snapshot."
```
opcode81: You are unnecessarily repeating words. To save tokens, use something like this instead:
> Special identifiers you can use: 'selection' (first shape currently selected by the user), 'page' (entire current page)

mcp/packages/server/src/tools/ExportShapeTool.ts
```diff
@@ -71,7 +72,9 @@ export class ExportShapeTool extends Tool<ExportShapeArgs> {
     public getToolDescription(): string {
         let description =
             "Exports a shape (or a shape's image fill) from the Penpot design to a PNG or SVG image, " +
-            "such that you can get an impression of what it looks like. ";
+            "such that you can get an impression of what it looks like. " +
+            "Use the special shapeId 'page' to take a full snapshot of the entire current page. " +
+            "Use the special shapeId 'selection' to export the currently selected shape.";
```
opcode81: You are unnecessarily repeating usage details in the tool description. This should be avoided/undone.

opcode81 (changes requested):

mcp/packages/server/src/tools/ExportShapeTool.ts
abhishek-mittal: Done — replaced with the compact format you suggested: `Special identifiers you can use: 'selection' (first shape currently selected by the user), 'page' (entire current page)`

mcp/packages/server/src/tools/ExportShapeTool.ts
abhishek-mittal: Done — removed both 'Use the special shapeId...' sentences from `getToolDescription()`. The parameter's own `describe()` text already covers the special identifiers.

opcode81 (approved):
Thanks. The changes are now approved from my side. 

However, I would suggest that you amend the improvements from your second commit to your original commit to keep the history clean.

abhishek-mittal:
Thanks for the approval! I've amended the fix into the original commit (force-pushed), so the branch now has a single clean commit

niwinz:
In geneal you should not worry about commits, on the end this will be merged using SQUASH so a single commit will appear on the history. In any case, thanks :heart:

Issue #8321, opened 2026-02-11
bug: penpot.openPage(newPage.id) MCP command not working as expected


orgenom:
### Steps To Reproduce

1. Use Penpot MCP on penpot's website.
2. Try to get your favorite AI tool connected to this MCP to create a new page, it will try navigating to it in order to edit this page, but the command will fizzle and no navigation happens.

### Expected behavior

I expect to not have to manually change pages whenever asking to create a page. The MCP should seamlessly work and allow a user to only interact during manual editing / verification phases, not during the execution flow of my AI tool.

### Actual behavior

First, penpot.createPage() is called, then penpot.openPage(). createPage creates a new page, and openPage fails to navigate to this page. Had to create a new claude rule to prevent this behavior, since the agent believed it was navigated to the new page and would create overlapping designs or overwrite previously created designs.

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

madalenapmelo-kp:
Hi @orgenom,

Thanks for reporting this! We've added this to our backlog on Taiga so that we can look further into it, you can find the details here: https://tree.taiga.io/project/penpot/issue/13412

dollro:
can confirm this on self hosted penpot (2.13.3) ...

PR #8700, opened 2026-03-19
:bug: Fix theme.addSet() crash caused by async state race in token-set proxy

### LINKED ISSUE #8698: theme.addSet(tokenSet) plugin API causes workspace crash — proxy returns nil name due to async state race
## Description

Calling `theme.addSet(tokenSetProxy)` via the plugin API causes a workspace crash and full reload when the `TokenSetProxy` was obtained from a recent `catalog.addSet()` call.

## Steps to reproduce

1. Open a design file, load a plugin that uses the tokens API
2. Execute via plugin:
```javascript
const catalog = penpot.library.local.tokens;
const set = catalog.addSet({ name: 'Test/Colors' });
const theme = catalog.addTheme({ group: '', name: 'Light' });
theme.addSet(set); // ← crash
```

## Root cause

In `frontend/src/app/plugins/tokens.cljs`:

- **Line 447**: `catalog.addSet()` calls `st/emit!` (async) and returns a `TokenSetProxy` storing only the set's `id`
- **Line 180**: The proxy's `:name` getter does `(u/locate-token-set file-id id)` from `@st/state` — but the async state update from `st/emit!` hasn't propagated yet
- **Line 372**: `theme.addSet(proxy)` calls `(obj/get token-set :name)` → `locate-token-set` returns `nil` → `ctob/get-name nil` → `nil`
- `ctob/enable-set` does `(conj sets nil)` → `:sets #{nil}`
- Backend rejects with `400: expected valid params for token-theme / {:path [:sets 0], :in [:sets nil], :schema :string, :value nil}`
- Penpot reloads the entire workspace, disconnecting any plugin

## Suggested fix

Either:
1. Store the set name in the proxy at creation time (alongside `id`) so the `:name` getter doesn't depend on async state
2. Or validate `set-name` is non-nil in `enable-set` before `conj`

## Environment

- Penpot 2.13.3 (also present on `develop` as of 2026-03-19)
- Triggered via penpot-mcp plugin, but reproducible with any plugin using the tokens API

dollro:
## Summary

- Fix `theme.addSet(tokenSet)` crashing the workspace when called on a freshly created token set
- Root cause: async state race — `catalog.addSet()` uses `st/emit!` (async), but the returned proxy's `:name` getter reads from `@st/state` which hasn't propagated yet, returning `nil`
- `enable-set` then conjs `nil` into the theme's `:sets` → backend rejects with 400 (`:sets #{nil}`) → workspace reloads → plugin disconnects

## Changes

1. **`token-set-proxy`** — accepts optional `initial-name` fallback, stored at construction time
2. **`:name` getter** — falls back to `initial-name` when `locate-token-set` returns nil (async window)
3. **`catalog.addSet()`** — passes the known set name to the proxy constructor
4. **`theme.addSet()` / `theme.removeSet()`** — nil guards prevent `enable-set`/`disable-set` from receiving nil

## Test plan

- [ ] Create a token set and immediately add it to a theme via plugin API — no crash
- [ ] Existing token set operations (rename, toggle, duplicate, remove) still work
- [ ] Theme operations (create, activate, duplicate, remove) still work
- [ ] Token sets created in the Penpot UI still resolve names correctly via plugin API

Closes #8698

### DIFF
--- a/frontend/src/app/plugins/tokens.cljs
+++ b/frontend/src/app/plugins/tokens.cljs
@@ -203,123 +203,130 @@
   (obj/type-of? p "TokenSetProxy"))
 
 (defn token-set-proxy
-  [plugin-id file-id id]
-  (obj/reify {:name "TokenSetProxy"
-              :on-error (u/handle-error plugin-id)}
-    :$plugin {:enumerable false :get (constantly plugin-id)}
-    :$file-id {:enumerable false :get (constantly file-id)}
-    :$id {:enumerable false :get (constantly id)}
+  ([plugin-id file-id id]
+   (token-set-proxy plugin-id file-id id nil))
+  ([plugin-id file-id id initial-name]
+   (obj/reify {:name "TokenSetProxy"
+               :on-error (u/handle-error plugin-id)}
+     :$plugin {:enumerable false :get (constantly plugin-id)}
+     :$file-id {:enumerable false :get (constantly file-id)}
+     :$id {:enumerable false :get (constantly id)}
 
-    :id
-    {:get #(dm/str id)}
+     :id
+     {:get #(dm/str id)}
 
-    :name
-    {:this true
-     :get
+     :name
+     {:this true
+      :get
+      (fn [_]
+        ;; Prefer the authoritative state lookup; fall back to initial-name
+        ;; when the async state update from `catalog.addSet()` hasn't
+        ;; propagated yet.
+        (let [set (u/locate-token-set file-id id)]
+          (if (some? set)
+            (ctob/get-name set)
+            initial-name)))
+      :schema (cfo/make-token-set-name-schema
+               (u/locate-tokens-lib file-id)
+               id)
+      :set
+      (fn [_ name]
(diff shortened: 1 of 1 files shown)


niwinz:
Looks very nice, it has conflicts, I can rebase and merge it myself but if you prefer con preserve the original PR, please rebase or merge develop and resolve conflicts. Thanks

dollro:
This fix is fully included in #8711, which extends the same multi-arity proxy + nil-guard pattern to all affected plugin proxies (colors, typography, components, flows, ruler guides, page, shape clone) and also fixes the `openPage()` context sync issue (#8321). 
@niwinz : Holding this PR in favour of merging #8711 ??

niwinz:
We prefer smaller PR's for easy reasoning about.

niwinz:
About https://github.com/penpot/penpot/pull/8711 we are currently disusing internally about the changes...

dollro:
Resolved the merge conflict in `tokens.cljs`

PR #8724, opened 2026-03-23
:bug: Fix content attribute sync group resolution by shape type

### LINKED ISSUE #9527: Internal error when flattening a mask or path shape inside a component
### Description

Flattening a mask or path shape inside a component causes an internal error and breaks component synchronization.

### Steps to reproduce

1. Import the Neomorphic UI Kit library
2. Go to "Main Components" page
3. Select the Mask shape inside Group-1 inside the Components board
4. Open the context menu on the Mask and click "Flatten"
5. After the save timeout, an internal error is raised

### Expected behavior

Flattening a mask or path shape inside a component should work without errors and component synchronization should remain intact.

### Affected versions

2.16.0

niwinz:
### Summary

The `:content` attribute was mapped to a single sync group (`:content-group`) but it is used by both path and text shapes with different synchronization needs. This caused incorrect component synchronization when editing content on path shapes, as they should sync under `:geometry-group` instead of `:content-group`.

Changes:

- Make sync-attrs allow type-dependent group mapping via maps
- Add resolve-sync-group and resolve-sync-groups helper functions
- Update necessary sync-attr lookups to use shape type for proper resolution
- Fix touched checks to handle multiple possible sync groups

### Steps to reproduce 

- Import https://penpot.app/penpothub/libraries-templates/neomorphic-ui-kit library
- Go to "Main Components" page
- Select Mask shape inside Group-1 inside the Components board.
- Open a context menu on the Mask and click to Flatten
- After the save timeout, the internal error is raised.

Fixes #9527

### DIFF
--- a/common/src/app/common/logic/libraries.cljc
+++ b/common/src/app/common/logic/libraries.cljc
@@ -48,11 +48,12 @@
 (def log-shape-ids #{})
 (def log-container-ids #{})
 
-(def updatable-attrs (->> (seq (keys ctk/sync-attrs))
-                          ;; We don't update the flex-child attrs
-                          (remove ctk/swap-keep-attrs)
-                          ;; We don't do automatic update of the `layout-grid-cells` property.
-                          (remove #(= :layout-grid-cells %))))
+(def updatable-attrs
+  (->> (keys ctk/sync-attrs)
+       ;; We don't update the flex-child attrs
+       (remove ctk/swap-keep-attrs)
+       ;; We don't do automatic update of the `layout-grid-cells` property.
+       (remove #(= :layout-grid-cells %))))
 
 (defn enabled-shape?
   [id container]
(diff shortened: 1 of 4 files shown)


common/src/app/common/logic/libraries.cljc
```diff
@@ -1813,55 +1817,59 @@
...
-                 (touched attr-group))
+                (and omit-touched?
+                     (cfh/text-shape? origin-shape)
+                     (= :content attr)
+                     (some touched sync-groups))
 
                 skip-operations?
                 (or (= (get origin-shape attr) (get dest-shape attr))
-                    (and (touched attr-group)
+                    (and (some touched sync-groups)
```
hirunatan: Not sure of this. This way the change will be skipped if the attribute is `:content` and either `:content-group` or `:geometry-group` has been touched in the copy, both in text and path shapes.

For example if I manually change the width of a text shape, the text content will also be prevented to be updated from main.

common/src/app/common/logic/libraries.cljc
```diff
@@ -2096,7 +2106,7 @@
                (= (get previous-shape attr) (get origin-ref-shape attr))
 
                ;; If the attr is not touched, don't copy it
-               (not (touched attr-group))
+               (not (some touched sync-groups))
```
hirunatan: I think this has the same problem as above.

common/src/app/common/logic/libraries.cljc
hirunatan: I think in all this function we should use `resolve-sync-group` by giving the type of the origin shape.

PR #8756, opened 2026-03-24
:sparkles: Add create default team org for nitrate on adding an user to a team

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot/us/13446

### Summary

When a user is added to a team belonging to an organization they are not yet a member of:

* Automatically creates the user default team within the organization.
* Notifies Nitrate to trigger the addition of the user to the organization.

### Checklist

### DIFF
--- a/backend/src/app/rpc/commands/management.clj
+++ b/backend/src/app/rpc/commands/management.clj
@@ -207,8 +207,7 @@
                          (update :team-id bfc/lookup-index)
                          (assoc :created-at timestamp)
                          (assoc :modified-at timestamp))]
-          (db/insert! conn :team-profile-rel params
-                      {::db/return-keys false})))
+          (teams/add-profile-to-team! cfg params {::db/return-keys false})))
 
       ;; Duplicate team fonts
       (doseq [font fonts]
--- a/backend/src/app/rpc/commands/teams.clj
+++ b/backend/src/app/rpc/commands/teams.clj
@@ -522,20 +522,78 @@
     (with-meta team
       {::audit/props {:id (:id team)}})))
 
+
+(defn create-default-org-team
+  [cfg profile-id organization-id]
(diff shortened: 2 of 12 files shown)


backend/src/app/rpc/commands/management.clj
```diff
@@ -207,8 +207,7 @@
                          (update :team-id bfc/lookup-index)
                          (assoc :created-at timestamp)
                          (assoc :modified-at timestamp))]
-          (db/insert! conn :team-profile-rel params
-                      {::db/return-keys false})))
+          (teams/add-profile-to-team! {::db/conn conn} params {::db/return-keys false})))
```
niwinz: Why don't pass cfg directly instead of `{::db/conn conn}`

backend/src/app/rpc/commands/teams.clj
```diff
@@ -522,6 +522,66 @@
...
+       cfg
+       (fn [{:keys [::db/conn] :as tx-cfg}]
+         (let [org-id           (:organization-id membership)
+               default-team     (create-default-org-team (assoc tx-cfg ::db/conn conn) profile-id org-id)
+               default-team-id  (:id default-team)
+               result           (nitrate/call tx-cfg :add-profile-to-org {:profile-id profile-id
+                                                                          :team-id default-team-id
+                                                                          :org-id org-id})]
+           (when (not (:is-member result))
+             (throw (ex-info "Failed to add profile to organization on Nitrate"
```
niwinz: Exceptions should use ex/raise

niwinz: with :type and :code.
If is a validation should be :validation
If this is a "guard" and should not happen in reallity, it should be :internal

niwinz (approved):

PR #8711, opened 2026-03-20
:bug: Fix plugin API async state races and openPage context sync

dollro:
## Summary

Fixes multiple plugin API bugs discovered during MCP-driven design system construction. All changes are in `frontend/src/app/plugins/`.

- **`openPage()` context sync**: Emit synchronous `UpdateEvent` to set `:current-page-id` immediately, so plugin code sees the correct page without waiting for React render cycles
- **Async state race in proxy constructors**: Multi-arity proxy constructors with initial value fallbacks for `createColor`, `createTypography`, `createComponent`, `clone`, token sets, flows, and ruler guides — prevents nil crashes when `st/emit!` hasn't propagated yet
- **Nil guards**: Consistent `some?` / `when-let` guards in page, flow, and ruler guide proxy getters

### Bugs fixed

| Bug | Severity | Fix |
|-|-|-|
| `openPage()` doesn't update plugin execution context | P0 | Synchronous `UpdateEvent` before async navigation |
| `theme.addSet()` crash — nil name from async state race | P0 | Multi-arity `token-set-proxy` with initial-name fallback |
| `createColor()` / `.clone()` return broken proxy | P1 | Multi-arity `lib-color-proxy` with initial values |
| `createTypography()` / `.clone()` return broken proxy | P1 | Multi-arity `lib-typography-proxy` with initial values |
| `createComponent()` lib-component-proxy nil crashes | P1 | Nil guards in component proxy |
| `shape.clone()` nil return-ref crash | P1 | Guard against nil from async `duplicate-shapes` |
| `page.name` getter crash on async state race | P1 | Nil guard in page-proxy name getter |
| `flow.startingBoard` getter inconsistent nil guard | P2 | Consistent `some?` guard |
| Flow proxy async state race | P2 | Multi-arity flow proxy with initial fallback |
| `addRulerGuide()` async state race | P2 | Nil guard on ruler guide proxy |
| Compilation failure — paren mismatches from merge | Build | Fixed paren mismatches in tokens.cljs and library.cljs |

### Related issues

Fixes #8321 — `openPage()` not working as expected
Fixes #8698 — `theme.addSet()` crash from async state race
Related to #8520 — Multiple Plugin API silent failures
Related to #8691 — `createComponent()` limitations

## Test plan

- [ ] Call `penpot.openPage(page)` then immediately check `penpot.currentPage.name` — should reflect new page
- [ ] Call `catalog.addSet("test")` then `theme.addSet(set)` — should not crash
[...]

PR #8794, opened 2026-03-26
:bug: Fix openPage context sync and page/flow/ruler proxy async races

dollro:
## Summary

Fixes `openPage()` not updating the plugin execution context and async state races in page, flow, and ruler guide proxies.

**Part 1 of 3** — Split from #8711 for easier review.
- **Part 2:** #8795 — Fix library proxy async state races
- **Part 3:** #8796 — Fix token-set and shape proxy nil guards

### Changes

- **`openPage()` context sync** (`api.cljs`): Emit synchronous `UpdateEvent` to set `:current-page-id` immediately, so plugin code sees the correct page without waiting for React render cycles
- **flow-proxy** (`page.cljs`): Multi-arity constructor with `initial-name` and `initial-starting-frame` fallbacks; consistent `some?` guards on getters
- **page-proxy** (`page.cljs`): Nil guard on `name` getter for async state race
- **ruler-guide-proxy** (`ruler_guides.cljs`): Multi-arity constructor with initial value fallback in `addRulerGuide`; nil guards on getters

> **Review tip:** The diff looks large (~900 lines) because the multi-arity refactor re-indents proxy object bodies. Click **"Hide whitespace changes"** in the GitHub diff view — actual logic changes are ~77 lines.

## Test plan

- [ ] Call `penpot.openPage(pageId)` from a plugin and verify getters on the returned page proxy return correct values immediately
- [ ] Create a flow via `page.createFlow()` and read `.name` / `.startingBoard` on the returned proxy without awaiting
- [ ] Add a ruler guide via `page.addRulerGuide()` and read properties on the returned proxy immediately

### DIFF
--- a/frontend/src/app/plugins/api.cljs
+++ b/frontend/src/app/plugins/api.cljs
@@ -54,7 +54,8 @@
    [app.util.object :as obj]
    [app.util.theme :as theme]
    [beicon.v2.core :as rx]
-   [cuerdas.core :as str]))
+   [cuerdas.core :as str]
+   [potok.v2.core :as ptk]))
 
 ;;
 ;; PLUGINS PUBLIC API - The plugins will able to access this functions
@@ -554,7 +555,11 @@
             new-window (if (boolean? new-window) new-window false)]
         (if (nil? id)
           (u/not-valid plugin-id :openPage "Expected a Page object or a page UUID string")
-          (st/emit! (dcm/go-to-workspace :page-id id ::rt/new-window new-window)))))
+          (st/emit! (ptk/reify ::open-page-context
+                      ptk/UpdateEvent
+                      (update [_ state]
+                        (assoc state :current-page-id id)))
+                    (dcm/go-to-workspace :page-id id ::rt/new-window new-window)))))
 
     :alignHorizontal
     (fn [shapes direction]
--- a/frontend/src/app/plugins/page.cljs
+++ b/frontend/src/app/plugins/page.cljs
@@ -31,7 +31,8 @@
    [app.plugins.utils :as u]
    [app.util.object :as obj]
    [beicon.v2.core :as rx]
-   [cuerdas.core :as str]))
+   [cuerdas.core :as str]
+   [potok.v2.core :as ptk]))
 
 (declare page-proxy)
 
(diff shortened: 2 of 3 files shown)


frontend/src/app/plugins/api.cljs
```diff
@@ -554,7 +555,11 @@
             new-window (if (boolean? new-window) new-window false)]
         (if (nil? id)
           (u/not-valid plugin-id :openPage "Expected a Page object or a page UUID string")
-          (st/emit! (dcm/go-to-workspace :page-id id ::rt/new-window new-window)))))
+          (st/emit! (ptk/reify ::open-page-context
```
niwinz, line 558: This change is potentially risky because it allows arbitrary setting of the current-page-id. These properties are intended for internal use and are typically managed during specific state transitions.

Could you clarify the goal of this event? Once I understand the intended use case, I can suggest a safer alternative.

frontend/src/app/plugins/page.cljs
```diff
@@ -31,390 +31,423 @@
...
-    :addCommentThread
-    (fn [content position board]
-      (let [shape (when board (u/proxy->shape board))
-            position (parser/parse-point position)]
-        (cond
-          (or (not (string? content)) (empty? content))
-          (u/not-valid plugin-id :addCommentThread "Content not valid")
+         :else
+         (let [new-window (if (boolean? new-window) new-window false)]
+           (st/emit! (ptk/reify ::open-page-context
```
niwinz, line 302: The same here, changing current-page-id inline is not an option,

frontend/src/app/plugins/ruler_guides.cljs
```diff
@@ -23,78 +23,87 @@
...
-                 value)]
-           (st/emit! (dwgu/update-guides (assoc guide :position position))))))}
-
-    :remove
-    (fn []
-      (let [guide (u/locate-ruler-guide file-id page-id id)]
-        (st/emit! (dwgu/remove-guide guide))))))
+  ([plugin-id file-id page-id id]
+   (ruler-guide-proxy plugin-id file-id page-id id nil))
+  ([plugin-id file-id page-id id initial-guide]
```
niwinz, line 28: The second arity is defined, but it doesn't seem to be used anywhere in this PR

frontend/src/app/plugins/page.cljs
```diff
@@ -31,390 +31,423 @@
...
+     (fn []
+       (st/emit! (dwi/remove-flow page-id id))))))
+
+(defn page-proxy? [proxy]
+  (obj/type-of? proxy "PageProxy"))
+
+(defn page-proxy
+  ([plugin-id file-id id]
+   (page-proxy plugin-id file-id id nil))
+  ([plugin-id file-id id initial-name]
```
niwinz, line 115: The second arity is defined, but it doesn't seem to be used anywhere in PR.

niwinz (changes requested):

filipsajdak (commented):
Reviewed as part of going through the open plugin-API proxy PRs (disclosure: prepared with AI assistance as part of the Penpot MCP self-improvement initiative). Splitting #8711 into three scoped PRs and flagging the re-indentation up front made this easy to read -- I verified line-by-line that the ~970-line diff is mostly the `obj/reify` bodies being re-indented under the new multi-arity `defn`, with only ~80 lines of real logic.

The fix pattern is correct: the multi-arity getters read the authoritative `u/proxy->*` state first and only fall back to the construction-time value while the emit hasn't landed (e.g. `(if (some? flow) (:name flow) initial-name)`), so they self-heal rather than freezing on a stale constant. The flow-proxy, ruler-guide-proxy and `addRulerGuide` (hoisting the freshly-built guide as `initial-guide`) all look right.

Main discussion point (and likely what @niwinz already flagged): the `openPage()` change eagerly emits an `UpdateEvent` to set `:current-page-id` before `go-to-workspace`. See the inline note -- worth confirming it can't desync file/route context. Also the same block is duplicated across `api.cljs` and `page.cljs`. No tests were added; these races are quite testable (read a getter before/after a simulated emit) and a small test would protect the fix.

frontend/src/app/plugins/api.cljs
filipsajdak, line 558: This eagerly sets `:current-page-id` (via the synchronous `UpdateEvent`) before `go-to-workspace` runs. Is that safe when the target page belongs to a different file (so `:current-file-id` isn't updated yet), or if the navigation is rejected? A `:current-page-id` that's ahead of the rest of the workspace context could leave things inconsistent. Might be cleaner to have `go-to-workspace` expose a synchronous/optimistic option than to duplicate its context write here.

frontend/src/app/plugins/page.cljs
filipsajdak, line 302: This `open-page-context` `UpdateEvent` is duplicated in the `api.cljs` `openPage` path. If the approach stays, extracting a small shared helper (e.g. in `plugins/utils`) would keep the two `openPage` entry points from drifting.

PR #8795, opened 2026-03-26
:bug: Fix library proxy async state races (color, typography, component)

dollro:
## Summary

Fixes async state races in library color, typography, and component proxy constructors that caused nil crashes when proxies were used immediately after creation.

**Part 2 of 3** — Split from #8711 for easier review.
- **Part 1:** #8794 — Fix openPage context sync and page proxy races
- **Part 3:** #8796 — Fix token-set and shape proxy nil guards

### Changes (all in `library.cljs`)

- **lib-color-proxy**: Multi-arity constructor accepting `initial-color` fallback; nil guards on all getters (name, path, color, opacity, gradient, image, pluginData)
- **lib-typography-proxy**: Multi-arity constructor accepting `initial-typography` fallback; nil guards on all getters and consistent `some?`/`when-let` patterns
- **lib-component-proxy**: Nil guards on `createComponentInstance` and other methods

All changes prevent nil crashes when proxy objects are used immediately after creation, before `st/emit!` state updates propagate.

> **Review tip:** The diff is ~1,400 lines but the multi-arity refactor re-indents entire proxy object bodies. Click **"Hide whitespace changes"** in the GitHub diff — actual logic changes are ~170 lines. Same pattern repeated across three proxy types.

## Test plan

- [ ] Call `library.createColor()` and immediately read `.name`, `.path`, `.color` on the returned proxy
- [ ] Call `library.createTypography()` and immediately access typography properties
- [ ] Clone a color/typography via `.clone()` and verify properties are readable without awaiting

### DIFF
--- a/frontend/src/app/plugins/library.cljs
+++ b/frontend/src/app/plugins/library.cljs
@@ -41,579 +41,647 @@
   (obj/type-of? p "LibraryColorProxy"))
 
 (defn lib-color-proxy
-  [plugin-id file-id id]
-  (assert (uuid? file-id))
-  (assert (uuid? id))
+  ([plugin-id file-id id]
+   (lib-color-proxy plugin-id file-id id nil))
+  ([plugin-id file-id id initial-color]
+   (assert (uuid? file-id))
+   (assert (uuid? id))
 
-  (obj/reify {:name "LibraryColorProxy"}
-    :$plugin {:enumerable false :get (constantly plugin-id)}
-    :$id {:enumerable false :get (constantly id)}
-    :$file {:enumerable false :get (constantly file-id)}
+   (obj/reify {:name "LibraryColorProxy"}
+     :$plugin {:enumerable false :get (constantly plugin-id)}
+     :$id {:enumerable false :get (constantly id)}
+     :$file {:enumerable false :get (constantly file-id)}
 
-    :id {:get (fn [] (dm/str id))}
-    :fileId {:get #(dm/str file-id)}
+     :id {:get (fn [] (dm/str id))}
+     :fileId {:get #(dm/str file-id)}
 
-    :name
-    {:this true
-     :get #(-> % u/proxy->library-color :name)
-     :set
-     (fn [self value]
+     :name
+     {:this true
+      :get (fn [self]
+             (let [color (u/proxy->library-color self)]
+               (if (some? color)
+                 (:name color)
+                 (:name initial-color))))
+      :set
+      (fn [self value]
+        (cond
+          (not (string? value))
+          (u/not-valid plugin-id :name value)
+
+          (not (r/check-permission plugin-id "library:write"))
+          (u/not-valid plugin-id :name "Plugin doesn't have 'library:write' permission")
+
+          :else
+          (when-let [color (u/proxy->library-color self)]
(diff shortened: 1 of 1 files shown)


frontend/src/app/plugins/library.cljs
```diff
@@ -41,579 +41,647 @@
...
 
          :else
-         (let [color (-> (u/proxy->library-color self)
-                         (update :name #(str value " / " %)))]
-           (st/emit! (dwl/update-color color file-id)))))}
+         (let [color-id (uuid/next)
+               color (-> (u/locate-library-color file-id id)
+                         (assoc :id color-id))]
+           (st/emit! (dwl/add-color color {:rename? false}))
+           (lib-color-proxy plugin-id id color-id color))))
```
filipsajdak, line 202: Wrong arg order: the `file-id` slot gets `id` (the source color id), not `file-id`. This is pre-existing, but now that `initial-color` is supplied, `u/proxy->library-color` will never resolve from state and the cloned proxy will return the frozen `initial-color` fallback permanently -- the stale-constant failure mode this PR is fixing elsewhere. Should be `(lib-color-proxy plugin-id file-id color-id color)` (cf. `createColor`).

frontend/src/app/plugins/library.cljs
```diff
@@ -41,579 +41,647 @@
...
-    :fontId
-    {:this true
-     :get #(-> % u/proxy->library-typography :font-id)
-     :set
-     (fn [self value]
+         (let [typo-id (uuid/next)
+               typo (-> (u/locate-library-typography file-id id)
+                        (assoc :id typo-id))]
+           (st/emit! (dwl/add-typography typo false))
+           (lib-typography-proxy plugin-id id typo-id typo))))
```
filipsajdak, line 572: Same wrong arg order as the color clone: the `file-id` slot gets `id`. Should be `(lib-typography-proxy plugin-id file-id typo-id typo)`. With `initial-typo` now supplied, the typography proxy will permanently serve stale fallback values instead of resolving live state.

filipsajdak (commented):
Reviewed alongside the sibling proxy PRs (disclosure: prepared with AI assistance as part of the Penpot MCP self-improvement initiative). The race-fix pattern is the same self-healing fallback as #8794, applied uniformly across the color/typography/component getters, and the setter tightening from `let` to `when-let` (so a setter called before state lands no longer NPEs on a nil map) is a good, on-scope defensive change. As with #8794 the diff is mostly re-indentation.

One thing I'd fix before merge: the `clone` paths for color and typography pass the source `id` in the **file-id** positional slot of the proxy constructor (see inline). That arg order is pre-existing on `develop`, but now that this PR also supplies `initial-color` / `initial-typo`, the consequence changes: the cloned proxy's state lookup will always miss and the getters will serve the frozen initial value **permanently** -- exactly the "stale constant" failure mode this PR is meant to prevent. `createColor` (which passes `file-id` correctly) is the reference.

No tests were added, and this is the most test-amenable of the three -- a getter/clone-resolution test would have caught the file-id issue. Optional: the change is large but cohesive; splitting into color / typography / component commits would help bisectability, though it's acceptable as one mechanical change.

niwinz:
Closing for inactivity

PR #8796, opened 2026-03-26
:bug: Fix token-set and shape proxy nil guards

dollro:
## Summary

Fixes nil crashes in token-set proxy (`theme.addSet()`) and shape proxy (`shape.clone()`).

**Part 3 of 3** — Split from #8711 for easier review.
- **Part 1:** #8794 — Fix openPage context sync and page proxy races
- **Part 2:** #8795 — Fix library proxy async state races

### Changes

- **token-set proxy** (`tokens.cljs`): Fix `theme.addSet()` crash — multi-arity constructor with `initial-name` fallback; nil guard on `removeSet`
- **shape proxy** (`shape.cljs`): Guard `shape.clone()` against nil return-ref from async `duplicate-shapes`

## Test plan

- [ ] Call `theme.addSet("name")` and verify the returned set proxy is usable immediately
- [ ] Call `theme.removeSet("name")` with a valid set name
- [ ] Call `shape.clone()` and verify the returned proxy has valid properties


frontend/src/app/plugins/tokens.cljs
```diff
@@ -294,19 +300,15 @@
...
-                               tokens-tree
-                               (cto/dtcg-token-type->token-type (-> args (first) (get "type"))))
-                              ;; Don't allow plugins to set the id
-                              (sm/dissoc-key :id)
-                              ;; Instruct the json decoder in obj/reify not to process map keys (:key-fn below)
-                              ;; and set a converter that changes DTCG types to internal types (:decode/json).
-                              ;; E.g. "FontFamilies" -> :font-family or "BorderWidth" -> :stroke-width
-                              (sm/update-properties assoc :decode/json cfo/convert-dtcg-token))]))
+                [:tuple (-> (cfo/make-token-schema
+                             (-> (u/locate-tokens-lib file-id) (ctob/get-tokens id))
```
niwinz, line 304: This code removes the `ctob/tokens-tree` call. I'm not 100% familiar with that code so can you explain the rationale behind that removal?

frontend/src/app/plugins/shape.cljs
```diff
@@ -1230,7 +1231,6 @@
                          {:cmd :export-shapes
                           :profile-id (:profile-id @st/state)
                           :wait true
-                          :is-wasm false
```
niwinz, line 1233: This change also looks unrelated to the PR and breaks the exportation

filipsajdak (commented):
Looked at this while going through the open plugin-API proxy PRs (disclosure: prepared with AI assistance as part of the Penpot MCP self-improvement initiative). Thanks for splitting #8711 into three focused, separately-reviewable PRs with test plans -- that makes this much easier to reason about.

The two in-scope fixes look reasonable: the `shape.clone()` nil guard and the `token-set-proxy` `initial-name` fallback both address the documented async-state race.

Two things to flag, both about keeping this PR true to its "nil guards" title:
- It also **removes `:is-wasm false`** from the export-shapes command in `shape.cljs` and **rewrites the `:addToken` `:schema`** (dropping the `(ctob/tokens-tree)` conversion). Both are unrelated to nil-guards, and on current `develop` both still have the original form -- the `:addToken` one in particular looks like it would regress name-schema generation (see inline). I'd drop both hunks so this PR stays scoped to the guards.
- The `token-set-proxy` / `theme.addSet()` fix overlaps **#9488** (which targets the same #9369 crash and has already been through maintainer review). Worth coordinating so the two don't fix it twice in different ways -- e.g. rebase this to just the `removeSet` guard if #9488 lands the `addSet` path.

Inline notes below. Nice, tightly-scoped commits otherwise.

frontend/src/app/plugins/shape.cljs
```diff
@@ -841,7 +841,8 @@
 
                  :else
                  (do (st/emit! (dws/duplicate-shapes #{id} :change-selection? false :return-ref ret-v))
-                     (shape-proxy plugin-id (deref ret-v))))))
+                     (when-let [new-id (deref ret-v)]
```
filipsajdak, line 844: Good guard. One thing worth documenting: on the async path `(deref ret-v)` can still be `nil` here (the duplicate hasn't landed yet), so `clone()` will now return `nil` to the plugin instead of crashing. That's the right tradeoff, but the return type / docs should note that `clone()` may return null within the same tick.

frontend/src/app/plugins/shape.cljs
```diff
@@ -1230,7 +1231,6 @@
                          {:cmd :export-shapes
                           :profile-id (:profile-id @st/state)
                           :wait true
```
filipsajdak, line 1233: This hunk removes `:is-wasm false` from the export-shapes command (the line just below). That's unrelated to the nil-guard scope of this PR and changes the export pipeline -- current `develop` still sets `:is-wasm false` here, and the non-WASM export path is the one #10208 is about. Looks like a stray change carried over from the #8711 split; recommend reverting it in this PR.

frontend/src/app/plugins/tokens.cljs
```diff
@@ -294,19 +300,15 @@
...
-                  [:tuple (-> (cfo/make-token-schema
-                               tokens-tree
-                               (cto/dtcg-token-type->token-type (-> args (first) (get "type"))))
-                              ;; Don't allow plugins to set the id
-                              (sm/dissoc-key :id)
-                              ;; Instruct the json decoder in obj/reify not to process map keys (:key-fn below)
-                              ;; and set a converter that changes DTCG types to internal types (:decode/json).
-                              ;; E.g. "FontFamilies" -> :font-family or "BorderWidth" -> :stroke-width
-                              (sm/update-properties assoc :decode/json cfo/convert-dtcg-token))]))
+                [:tuple (-> (cfo/make-token-schema
```
filipsajdak, line 303: This drops the `(ctob/tokens-tree)` conversion and passes the raw `(ctob/get-tokens id)` map straight to `make-token-schema`. On `develop`, `make-token-schema [tokens-tree token-type]` feeds that argument into `make-token-name-schema`, and the removed comment ("Convert to the adecuate format for schema") was load-bearing -- so this likely breaks name-schema generation rather than simplifying it. It's also unrelated to the nil-guard scope. Recommend dropping this hunk.

PR #8820, opened 2026-03-27
:bug: Fix radio-buttons component in the DS

EvaMarco:
### Related Ticket

No ticket. 

### Summary

Radio-buttons DS component was not working properly. 
I fixed 
- "allow-empty" option
- "disabled" option

### Steps to reproduce 
You can check functionality on storybook, this component is not being used yet.

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/ds/controls/radio_buttons.cljs
+++ b/frontend/src/app/main/ui/ds/controls/radio_buttons.cljs
@@ -36,45 +36,55 @@
    [:selected {:optional true}
     [:maybe [:or :keyword :string]]]
    [:allow-empty {:optional true} :boolean]
+   [:disabled {:optional true} :boolean]
    [:options [:vector {:min 1} schema:radio-button]]
    [:on-change {:optional true} fn?]])
 
 (mf/defc radio-buttons*
   {::mf/schema schema:radio-buttons}
-  [{:keys [class variant extended name selected allow-empty options on-change] :rest props}]
+  [{:keys [class variant extended name selected allow-empty options on-change disabled] :rest props}]
   (let [options (if (array? options)
                   (mfu/bean options)
                   options)
-        type    (if allow-empty "checkbox" "radio")
-        variant (d/nilv variant "secondary")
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/ui/ds/controls/radio_buttons.cljs
```diff
@@ -24,7 +24,7 @@
     [:and :string [:fn #(contains? icon-list %)]]]
    [:label :string]
    [:value [:or :keyword :string]]
-   [:disabled {:optional true} :boolean]])
+   [:option-disabled {:optional true} :boolean]])
```
niwinz: I'm not clearly understand why this is named `:option-disabled` and not simply `:disabled`. It is already on options, so we don't need prefix it.

niwinz (changes requested):

frontend/src/app/main/ui/ds/controls/radio_buttons.cljs
EvaMarco: It is to make it clear in the component that we can disable the whole set of inputs or one of the options.

niwinz: yes but, in that case why not prefix option on all the properties we have. I mean, is pretty clear when :disabled is used on main component props or when it is inside the options. The prefix looks totally inconsistent with the rest of attrs on the radio-button option schema

EvaMarco: Because it's the only one with the same name, the wrapper doesn't have label, or icon, but both have disabled state. But if it's a problem for you I'll try to change it.

Xaviju: I agree that it is not necessary to prefix with `option` the input when building the props. From my point of view, as a dev, I can see a clear distinction.  
In fact, I would suggest in the future separating the wrapper and the input components so they can be used isolated and keep the props as closer to standard HTML as possible.

niwinz (approved):

PR #8824, opened 2026-03-27
:wrench: Fix text bounds

elenatorro:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/13872

### Summary

 - Text bounding boxes now use actual glyph metrics instead of the logical text container size                           
  - Results are cached to avoid recomputing every frame                                                                                                                                                                                     
  - `intersects_text` reuses cached paragraphs instead of rebuilding them from scratch

### Checklist

### DIFF
--- a/render-wasm/src/shapes/text.rs
+++ b/render-wasm/src/shapes/text.rs
@@ -18,6 +18,7 @@ use skia_safe::{
     Contains,
 };
 
+use std::cell::Cell;
 use std::collections::HashSet;
 
 use super::FontFamily;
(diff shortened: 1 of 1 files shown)


render-wasm/src/shapes.rs
```diff
@@ -153,10 +153,11 @@ pub enum ConstraintH {
...
 
 #[derive(Debug, Clone, PartialEq, Copy)]
+#[repr(u8)]
 pub enum VerticalAlign {
-    Top,
-    Center,
-    Bottom,
+    Top = 0,
+    Center = 1,
+    Bottom = 2,
```
belen-albeza: Q: why do we need this to use repr(u8) or manually assign values to the enum variants? This makes sense in the serialization module, but I don't think we need it here

belen-albeza (approved):
LGTM, added a non-blocking small suggestion

render-wasm/src/shapes.rs
elenatorro: you're right, we're not using it for transmuting so I we can remove it

PR #8828, opened 2026-03-29
fix: Make text and flex properties writable (Issue #8520)

### LINKED ISSUE #8520: bug: Multiple Plugin API silent failures — flex auto-sizing, token binding, text properties
## Summary

While building a complete design system programmatically via the Plugin API (tokens, atoms, molecules, organisms — ~65 components), I've encountered **11 distinct bugs**, most of which **fail silently** — setters accept values, getters read them back correctly, but the actual behavior doesn't change. This makes debugging extremely difficult.

Penpot version: latest cloud (design.penpot.app, as of March 2026)

---

## Bug 1: Flex `horizontalSizing` / `verticalSizing = "auto"` silently fails

**Severity: High** — Makes "hug content" impossible via Plugin API

Setting `flex.horizontalSizing = "auto"` or `flex.verticalSizing = "auto"` on a FlexLayout is accepted by the setter and reads back correctly, but the board retains its `resize()` dimensions. The board never actually auto-sizes to fit its content.

**Repro:**
```js
const board = penpot.createBoard();
board.resize(200, 200);
const flex = board.addFlexLayout();
flex.dir = "row";
flex.horizontalSizing = "auto"; // Accepted
console.log(flex.horizontalSizing); // "auto" — but board is still 200x200

// Add child text
const text = penpot.createText("Hello");
board.appendChild(text);
await new Promise(r => setTimeout(r, 500));
console.log(board.width); // Still 200, not hugging content
```

**Workaround:** Calculate content dimensions manually, then `resize()` to the computed size.

---

## Bug 2: `board.appendChild(shape)` inserts at unpredictable position

**Severity: Medium**

`board.appendChild(shape)` does not reliably insert at the end of the children array for non-flex boards. The insertion position is unpredictable.

**Workaround:** Always use `board.insertChild(board.children.length, shape)` for non-flex boards. For flex layout boards, `board.appendChild()` works correctly (inserts at visual end).

---

## Bug 3: `textAlign` not settable on Text shapes

**Severity: Medium**

Setting `text.textAlign = "center"` throws: `"Cannot add property textAlign, object is not extensible"`. There is no Plugin API method to set text alignment.

**Repro:**
```js
const text = penpot.createText("Hello World");
text.textAlign = "center"; // Throws error
```

**Workaround:** Use a parent flex layout with `alignItems: "center"` to center text visually.

---

## Bug 4: `applyToken()` on component children silently fails

**Severity: High** — Makes token-driven component updates impossible

Calling `shape.applyToken(token, [properties])` on children of a component's main instance silently fails — the method doesn't throw, but the token binding is not persisted. This also applies to `token.applyToShapes([childShape])`.

**Workaround:** Apply all tokens to shapes BEFORE calling `library.createComponent([shapes])`. Tokens applied before component creation persist correctly.

---

## Bug 5: Number-type tokens (line-height) also set `rotation`

**Severity: Medium** — Unexpected side effect

Applying a `number`-type token with `["line-height"]` property correctly sets line-height, but ALSO sets `shape.rotation` to the token's numeric value (e.g., 1.2, 1.35, 1.6).

**Repro:**
```js
// Token: { type: "number", name: "line-height.normal", value: "1.6" }
text.applyToken(lineHeightToken, ["line-height"]);
console.log(text.lineHeight); // "1.6" — correct
console.log(text.rotation);   // 1.6 — BUG! Should remain 0
```

**Workaround:** Always reset `shape.rotation = 0` after applying number-type tokens for line-height.

---

## Bug 6: `catalog.addSet()` / `set.addToken()` fail with object parameters

**Severity: Low**

The object parameter form of these methods returns `undefined` / `null`:
- `catalog.addSet({ name: "my-set" })` → returns `undefined`
- `set.addToken({ type: "color", name: "token.name", value: "#FF0000" })` → returns `null`

Only positional parameters work:
- `catalog.addSet("my-set")` → works
- `set.addToken("color", "token.name", "#FF0000")` → works

---

## Bug 7: Connected library token values are ClojureScript data structures

**Severity: Medium**

When reading font family token values from a connected library, the value is a ClojureScript/internal data structure instead of a plain string. For example, a font family token with value `"Inter"` returns an opaque object rather than the string `"Inter"`.

**Workaround:** Hardcode the expected string value when creating tokens in other files.

---

## Bug 8: `penpot.openPage()` doesn't navigate the user's browser

**Severity: Medium**

Calling `penpot.openPage(page)` completes without error but doesn't actually switch the user's viewport to the target page. The user must manually navigate in the Penpot UI.

---

## Bug 9: Property setters on component children from wrong page silently fail

**Severity: High** — Very confusing to debug

Setting properties (fontFamily, rotation, fills, etc.) on component children succeeds (no error, getter returns the new value) but changes don't persist when the user's browser is on a different page than where the component's main instance lives. Combined with Bug 8 (`openPage` not working), this creates a situation where programmatic component modification is unreliable.

---

## Bug 10: `flex.appendChild()` (FlexLayout method) is broken

**Severity: Medium**

The `appendChild` method on the `FlexLayout` object itself (`board.flex.appendChild(shape)`) does not work correctly — it doesn't insert at the expected position. The `Board`-level `board.appendChild(shape)` method works for flex layout boards.

---

## Bug 11: Design tokens don't propagate across library connections

**Severity: High** — Architectural limitation

W3C Design Tokens defined in File A and published as a shared library are **readable** in File B (which connects File A's library), but calling `applyToken()` with those tokens in File B **silently fails**. Token bindings are not created.

This means tokens must be duplicated into every file that needs them, which defeats the purpose of a shared token library.

**Related:** [GitHub Issue #7059](https://github.com/penpot/penpot/issues/7059) (if still open)

---

## Common Theme: Silent Failures

The most frustrating pattern across these bugs is **silent failure**. Methods accept values, don't throw errors, and often read back the "set" value — but the actual state doesn't change. This makes debugging via the Plugin API extremely time-consuming.

**Suggestion:** Consider adding a validation/warning mode where the Plugin API logs warnings when operations are silently ignored (e.g., "Token binding ignored: shape is a component child", "horizontalSizing 'auto' not supported via API").

---

## Environment

- **Penpot:** Cloud (design.penpot.app), March 2026
- **Plugin API** accessed via MCP (Model Context Protocol) server
- **Browser:** Chrome (macOS)

patelmm79:
Makes Plugin API properties writable to resolve "Cannot add property" errors:
- text.cljs: align, verticalAlign
- flex.cljs: horizontalSizing, verticalSizing

Fixes: #8520 (Multiple Plugin API silent failures - Bug #1 and Bug #3)

### Checklist

### DIFF
--- a/frontend/src/app/plugins/flex.cljs
+++ b/frontend/src/app/plugins/flex.cljs
@@ -268,6 +268,7 @@
 
     :horizontalSizing
     {:this true
+     :writable true
      :get #(-> % u/proxy->shape :layout-item-h-sizing (d/nilv :fix) d/name)
      :set
      (fn [_ value]
@@ -284,6 +285,7 @@
 
     :verticalSizing
     {:this true
+     :writable true
      :get #(-> % u/proxy->shape :layout-item-v-sizing (d/nilv :fix) d/name)
      :set
      (fn [_ value]
--- a/frontend/src/app/plugins/text.cljs
+++ b/frontend/src/app/plugins/text.cljs
@@ -625,6 +625,7 @@
           (st/emit! (dwt/update-attrs id {:text-direction value})))))}
 
    {:name "align"
+    :writable true
     :get #(-> % u/proxy->shape text-props :text-align format/format-mixed)
     :set
     (fn [self value]
@@ -640,6 +641,7 @@
           (st/emit! (dwt/update-attrs id {:text-align value})))))}
 
    {:name "verticalAlign"
+    :writable true
     :get #(-> % u/proxy->shape text-props :vertical-align)
     :set
     (fn [self value]


niwinz:
I asked @Alotor for check this PR, in any case, please make the commit format follow the rules on CONTRIBUTING.md (hint: Signed-of-by should be your name and email, not claude and the commit message should be  probably `:bug: Make several plugin api writable...` or somethign similar).

If you are using Calude Code or similar, consider instructing it to read AGENTS.md and CONTRIBUTING.md in a similar way as https://github.com/penpot/penpot/blob/develop/.opencode/agents/engineer.md

niwinz:
Adding writable: true here is redundant and incorrect. Once a :set accessor is defined, the property cannot have a writable attribute. This will result in an exception at runtime.

> If a descriptor is both a data descriptor (because it has value or writable) and an accessor descriptor (because it has get or set), an exception is thrown.

carolportugal18:

Issue #8836, opened 2026-03-30
MCP: Add delay before export_shape tool to account for asynchronous updates


opcode81:
As a temporary workaround until there is a better solution for penpot/penpot-mcp#27, add a constant delay prior to the shape export to account for asynchronous updates, increasing the likelihood that the fully updated state is exported.

PR #8837, opened 2026-03-30
:sparkles: Add wait time before export_shape to account for async updates

opcode81:
Add wait time in exportImage to account for async updates #8836

This is a temporary workaround for penpot/penpot-mcp#27.
It adds a wait time before exports via the export_shape tool to account for asynchronous updates in Penpot, increasing the likelihood of exporting the fully updated state.


niwinz:
Friendly reminder: please don't include MCP: prefix in the PR title. When this is squashed and merged, that prefix will be added into the commit message, which doesn't follow our naming convention. Thanks!

PR #8888, opened 2026-04-07
:sparkles: Upload org logo

mavalroot:
### Related Ticket

https://tree.taiga.io/project/penpot/task/13886

### Summary

Enable organizations to upload their logos in Penpot in the same way individual users upload their avatars.

### Checklist

### DIFF
--- a/backend/src/app/rpc/management/nitrate.clj
+++ b/backend/src/app/rpc/management/nitrate.clj
@@ -23,6 +23,7 @@
    [app.rpc.commands.teams-invitations :as ti]
    [app.rpc.doc :as doc]
    [app.rpc.notifications :as notifications]
+   [app.storage :as sto]
    [app.util.services :as sv]))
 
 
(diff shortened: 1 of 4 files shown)


backend/src/app/rpc/management/nitrate.clj
```diff
@@ -74,6 +75,33 @@
...
+  [:map [:logo-url ::sm/text]])
+
+(sv/defmethod ::upload-org-logo
+  "Store an organization logo in penpot storage and return its public URL"
+  {::doc/added "2.16"
+   ::sm/params schema:upload-org-logo
+   ::sm/result schema:upload-org-logo-result
+   ::rpc/auth false}
+  [{:keys [::sto/storage]} {:keys [logo-data logo-mime-type]}]
+  (let [data    (.decode (java.util.Base64/getDecoder) ^String logo-data)
```
niwinz: The logo upload should be using normal multipart request, so no base64 handling is necessary. Look at media.clj for understand how it is already working.

backend/src/app/rpc/management/nitrate.clj
```diff
@@ -74,6 +75,33 @@
...
+        hash    (sto/calculate-hash data)
+        content (-> (sto/content data)
+                    (sto/wrap-with-hash hash))
+        obj     (sto/put-object! storage {::sto/content     content
+                                          ::sto/deduplicate? true
+                                          :bucket            "org-logo"
```
niwinz: We can't just name a bucket. The bucket mechanism is used for properly handle storage object relations with other objects on the database. So adding it as is, will break the storage GC tasks.

We also need to think how to handle all the life cycle of that object. I mean: what happens with previous images when user upload a new image.

This part is a bit "complicated" so if you consider we can talk on it once you have a clear overview of the feature.

backend/src/app/rpc/management/nitrate.clj
```diff
@@ -74,6 +75,33 @@
...
+                                          :content-type      logo-mime-type})]
+    {:logo-url (files/resolve-public-uri (:id obj))}))
```
niwinz: I think the uri should be resolved on frontend and this endpoint should return the ID

niwinz (changes requested):

backend/src/app/rpc/management/nitrate.clj
```diff
@@ -80,6 +82,34 @@
...
+(sv/defmethod ::upload-org-logo
+  "Store an organization logo in penpot storage and return its ID.
+  Accepts an optional previous-id to mark the old logo for garbage
+  collection when replacing an existing one."
+  {::doc/added "2.16"
+   ::sm/params schema:upload-org-logo
+   ::sm/result schema:upload-org-logo-result}
+  [{:keys [::sto/storage]} {:keys [content previous-id]}]
+  (when previous-id
+    (sto/del-object! storage previous-id))
```
niwinz: This should be `sto/touch-object!`

backend/src/app/rpc/management/nitrate.clj
```diff
@@ -80,6 +82,34 @@
     (->> (db/exec! cfg [sql:get-teams current-user-id])
          (map #(select-keys % [:id :name])))))
 
+;; ---- API: upload-org-logo
+
+(def ^:private schema:upload-org-logo
+  [:map
+   [:content media/schema:upload]
+   [:previous-id {:optional true} ::sm/uuid]])
```
niwinz, line 92: This also should receive `:organzation-id` and we should pass it to the storage object metadata

backend/src/app/storage.clj
```diff
@@ -44,6 +44,7 @@
     "file-object-thumbnail"
     "file-thumbnail"
     "profile"
+    "org-logo"
```
niwinz: should be `"organization"`

niwinz (changes requested):
And we should modify the touched gc task adding that:

Image

niwinz (approved):

PR #8895, opened 2026-04-07
:sparkles: Add per-group add button for typographies

### LINKED ISSUE #5275: :sparkles: Typography: Add + button to each newly-created typography group

Right now, PenPot only features a plus button at the top of the typography list, this makes it so the user has to travel back up with the mouse and find the one and only option to add new typographies.

It would make it faster to create a similar group of typographies with less tweaks to make to the collection. Users get a plus button in the header of the last-created typography group. User clicks plus sign and an equal set of typographies is created with a temporary name that mimics the last naming convention in the last typography group. A typography group copy.

SCR-20241109-owui

_No response_

### Additional context

_No response_

eureka0928:
### Related Ticket

Closes https://github.com/penpot/penpot/issues/5275
https://tree.taiga.io/project/penpot/issue/9283

### Summary

Adds a "+" button to each typography group header in the assets panel, allowing users to create new typographies directly inside a specific group instead of having to use only the top-level add button.

**How it works:**
- `asset-group-title*` accepts an optional `on-add` prop; renders a "+" icon button when provided
- `typographies-group` creates an `add-typography-to-group` handler that dispatches `dwt/add-typography` with the group path prefix
- `add-typography` now accepts an optional `group-path` parameter; when provided, prepends it to the typography name using `"Group / Name"` format so `extract-path-if-missing` places it in the correct group
- The button only appears for local, editable files (not for external libraries or read-only mode)

### Steps to reproduce

1. Open a file in the workspace
2. Create some typographies and organize them into groups
3. Verify each group header now shows a "+" button next to the menu button
4. Click the "+" button on a group — a new typography should be created inside that group
5. Verify the top-level "+" button still works as before (adds to root)
6. Open an external library — verify the "+" buttons do not appear on library groups
7. Test with nested groups (e.g., "Heading / Large") — new typography should appear in the correct nested group

### Checklist

### DIFF
--- a/frontend/src/app/main/data/workspace/texts.cljs
+++ b/frontend/src/app/main/data/workspace/texts.cljs
@@ -896,7 +896,7 @@
                              (merge txt/default-typography values))
                            (generate-typography-name)
                            (assoc :id typ-id)
-                           (cond-> (seq group-path)
+                           (cond-> (string? group-path)
                              (update :name #(str group-path " / " %))))]
 
          (rx/concat
--- a/frontend/src/app/main/data/workspace/texts.cljs
+++ b/frontend/src/app/main/data/workspace/texts.cljs
@@ -859,50 +859,55 @@
   "A higher level version of dwl/add-typography, and has mainly two
   responsabilities: add the typography to the library and apply it to
   the currently selected text shapes (being aware of the open text
-  editors."
-  [file-id]
-  (ptk/reify ::add-typography
-    ptk/WatchEvent
-    (watch [_ state _]
-      (let [selected   (dsh/lookup-selected state)
-            objects    (dsh/lookup-page-objects state)
+  editors.
+  Optionally accepts a group-path to place the new typography inside
+  a specific group."
+  ([file-id] (add-typography file-id nil))
+  ([file-id group-path]
+   (ptk/reify ::add-typography
+     ptk/WatchEvent
+     (watch [_ state _]
+       (let [selected   (dsh/lookup-selected state)
+             objects    (dsh/lookup-page-objects state)
 
-            xform      (comp (keep (d/getf objects))
-                             (filter cfh/text-shape?))
-            shapes     (into [] xform selected)
-            shape      (first shapes)
+             xform      (comp (keep (d/getf objects))
+                              (filter cfh/text-shape?))
+             shapes     (into [] xform selected)
+             shape      (first shapes)
 
-            values     (current-text-values
-                        {:editor-state (dm/get-in state [:workspace-editor-state (:id shape)])
-                         :shape shape
-                         :attrs txt/text-node-attrs})
+             values     (current-text-values
+                         {:editor-state (dm/get-in state [:workspace-editor-state (:id shape)])
+                          :shape shape
+                          :attrs txt/text-node-attrs})
 
-            multiple? (or (> 1 (count shapes))
-                          (d/seek (partial = :multiple)
-                                  (vals values)))
+             multiple? (or (> 1 (count shapes))
+                           (d/seek (partial = :multiple)
+                                   (vals values)))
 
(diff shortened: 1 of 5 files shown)


myfunnyandy:
Hi @eureka0928 

We took a look at the PR, and it's already looking fairly good. The addition makes sense and is working as expected. 

There’s just one adjustment we’d recommend before merging. The new button is pushing the menu button onto a second row, which introduces some unnecessary spacing.

image


This is easier to notice when the buttons aren’t visible. You can see that the spacing between the heading and the first item differs from production, which creates a slight misalignment.

image

image


Would you be able to adjust the layout so the buttons match the following image?

image

fury0928:
Hi @myfunnyandy, thanks for the review!

Good catch — the `.title-menu` container was missing `display: flex`, which caused the add button to push the menu button onto a second row.

Fixed in 780493e — added `display: flex` to `.title-menu` so both buttons stay inline within the title bar height. This only affects groups that have the add button; other usages (colors, components) with a single button are unaffected.

Let me know if it looks good now!

myfunnyandy:
It looks perfect now :)

frontend/src/app/main/data/workspace/texts.cljs
```diff
@@ -859,50 +859,55 @@
...
+                                          txt/text-spacing-attrs
+                                          txt/text-transform-attrs)))
+
+             typ-id    (uuid/next)
+             typ       (-> (if multiple?
+                             txt/default-typography
+                             (merge txt/default-typography values))
+                           (generate-typography-name)
+                           (assoc :id typ-id)
+                           (cond-> (seq group-path)
```
niwinz: If we expect group-path to be a string, prefer using `string?` predicate instead of `(seq ...)` pattern.

frontend/src/app/main/ui/workspace/sidebar/assets/typographies.cljs
```diff
@@ -133,7 +133,7 @@
   {::mf/wrap-props false}
   [{:keys [file-id prefix groups open-groups force-open? file local? selected local-data
            editing-id renaming-id on-asset-click handle-change on-rename-group
-           on-ungroup on-context-menu selected-full]}]
+           on-ungroup on-context-menu selected-full read-only?]}]
```
niwinz: Do not add more props with `?` suffix. We are refactoring our components for make them more "react" jsx friendly props. So i sueggest changing it to `is-read-only`. Do not worry with other props that has `?` because that code is already refactored and I'm merging it in batches of commits in other PR's

frontend/src/app/main/ui/workspace/sidebar/assets/typographies.cljs
```diff
@@ -164,7 +164,14 @@
...
          (fn [event]
-           (cmm/on-drop-asset-group event dragging* prefix selected-paths selected-full move-typography)))]
+           (cmm/on-drop-asset-group event dragging* prefix selected-paths selected-full move-typography)))
+
+        add-typography-to-group
+        (mf/use-fn
+         (mf/deps file-id prefix)
+         (fn [_]
+           (st/emit! (dw/set-assets-section-open file-id :typographies true))
+           (st/emit! (dwt/add-typography file-id prefix))))]
```
niwinz: Use single `emit!` call with two events instead of two `emit!` calls

niwinz (changes requested):

fury0928:
Hi @niwinz, thanks for the review! All three points addressed in 31061411e:

1. `(seq group-path)` → `(string? group-path)`
2. `read-only?` → `is-read-only` for React-friendly prop naming
3. Merged two `st/emit!` calls into a single one

Let me know if anything else needs adjusting!

niwinz (approved):

PR #8908, opened 2026-04-08
:bug: Fix incorrect segment reading on several operation on PathData

niwinz:
#### Safety & Correctness fixes

- The `impl-walk`, `impl-reduce`, and `impl-lookup` functions had the binary type codes for `move-to` and `line-to` swapped (1 mapped to `:line-to` and 2 to `:move-to`). This is inconsistent with `from-plain`, `read-segment`, `to-string-segment*`, and the Rust `RawSegmentData` enum which all use 1=move-to and 2=line-to. The swap caused incorrect command types to be reported to callers like `get-handlers` and `get-points`.
- Add `nil` defaults to all case expressions that match binary segment type codes so that corrupted/unknown values are skipped instead of throwing 'No matching clause'. This prevents a React render crash when a PathData buffer contains invalid bytes, e.g. from a WASM data transfer or deserialization of damaged stored data.

#### Possible fixes for data corruption

Replace unsafe std::mem::transmute calls in Rust WASM path code with validated TryFrom conversions to prevent undefined behavior from invalid enum discriminant values. This was the most likely root cause of the "No matching clause: -19772" production crash -- corrupted bytes flowing through transmute could produce arbitrary invalid enum variants.

Fix byteOffset handling throughout the CLJS PathData serialization pipeline. DataView instances created via buf/slice carry a non-zero byteOffset, but from-bytes, transit write handler, -write-to, buf/clone, and buf/equals? all operated on the full underlying ArrayBuffer, ignoring offset and length. This could silently produce PathData with incorrect size or content.

CLJS changes (common):

- from-bytes: use DataView.byteLength instead of ArrayBuffer.byteLength   for DataView inputs; preserve byteOffset/byteLength when converting from Uint8Array, Uint32Array, and Int8Array
- Transit write handler: construct Uint8Array with byteOffset and
  byteLength from the DataView, not the full backing ArrayBuffer
- -write-to: same byteOffset/byteLength fix
- buf/clone: copy only the DataView byte range using Uint8Array with proper offset, not Uint32Array over the full ArrayBuffer
- buf/equals?: compare DataView byte ranges using Uint8Array with proper offset, not the full backing ArrayBuffers

Frontend changes:

- shape-to-path, stroke-to-path, calculate-bool*: wrap WASM call and buffer read in try/catch to ensure mem/free is always called, even when an exception occurs between the WASM call and the free call

### DIFF
--- a/common/src/app/common/buffer.cljc
+++ b/common/src/app/common/buffer.cljc
@@ -217,10 +217,12 @@
        (let [buffer (ByteBuffer/wrap dst)]
          (.order buffer ByteOrder/LITTLE_ENDIAN)))
      :cljs
-     (let [buffer'  (.-buffer ^js/DataView buffer)
-           src-view (js/Uint32Array. buffer')
-           dst-buff (js/ArrayBuffer. (.-byteLength buffer'))
-           dst-view (js/Uint32Array. dst-buff)]
+     (let [src-off  (.-byteOffset ^js/DataView buffer)
+           src-len  (.-byteLength ^js/DataView buffer)
+           src-buf  (.-buffer ^js/DataView buffer)
+           src-view (js/Uint8Array. src-buf src-off src-len)
+           dst-buff (js/ArrayBuffer. src-len)
+           dst-view (js/Uint8Array. dst-buff)]
        (.set dst-view src-view)
        (js/DataView. dst-buff))))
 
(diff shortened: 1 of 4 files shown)


niwinz:
Backported under 2.14.2 (?)

PR #8936, opened 2026-04-10
:sparkles: Add clear artboard guides option to context menu

### LINKED ISSUE #6987: :sparkles: Add ability to clear guides for a given artboard

Currently, the only way to remove guides is to drag them to the rulers one by one. When there are many guides, it can be tedious and time-consuming.

Ideally, there should be an option in the View sub-menu to remove all guides on the currently selected artboard(s).

Initially, I thought a global option to remove all guides would suffice, but removing all guides from all art boards may sometimes not be desirable.

### Additional context

Simply hiding guides is not enough. Consider the situation where an art board with guides is duplicated to use as a starting point, but the guides are no longer needed. The only option is then to remove the guides one by one.

eureka0928:
## Summary
- Adds a **"Clear artboard guides"** option to the right-click context menu when one or more frames with guides are selected
- All guides for the selected frame(s) are removed in a single batched operation (single undo step)
- Menu item only appears when the selected frame(s) actually have guides attached

Closes #6987

## Changes
- `frontend/src/app/main/data/workspace/guides.cljs` — New `remove-frame-guides` event that batches all guide removals into one `commit-changes` call via `pcb/set-guide`, with proper hover state cleanup
- `frontend/src/app/main/refs.cljs` — New `workspace-page-guides` derived ref
- `frontend/src/app/main/ui/workspace/context_menu.cljs` — New `context-menu-guides*` component wired into `shape-context-menu*`
- `frontend/translations/en.po` — Added translation key
- `CHANGES.md` — Added changelog entry

## Test plan
- [ ] Create a frame with multiple ruler guides (both horizontal and vertical)
- [ ] Right-click the frame → verify "Clear artboard guides" appears
- [ ] Click it → all guides for that frame are removed
- [ ] Ctrl+Z → all guides are restored in one undo step
- [ ] Select a frame with no guides → verify the menu item does NOT appear
- [ ] Select a non-frame shape → verify the menu item does NOT appear
- [ ] Select multiple frames (some with guides) → verify it removes guides from all selected frames
- [ ] Page-level guides (not attached to any frame) are unaffected

### DIFF
--- a/frontend/src/app/main/data/workspace/guides.cljs
+++ b/frontend/src/app/main/data/workspace/guides.cljs
@@ -6,6 +6,7 @@
 
 (ns app.main.data.workspace.guides
   (:require
+   [app.common.data :as d]
    [app.common.data.macros :as dm]
    [app.common.files.changes-builder :as pcb]
    [app.common.geom.point :as gpt]
@@ -80,9 +81,7 @@
 (defn remove-frame-guides
   [frame-ids]
 
-  (dm/assert!
-   "expected a coll of uuids"
-   (every? uuid? frame-ids))
+  (assert (every? uuid? frame-ids) "expected a coll of uuids")
 
   (ptk/reify ::remove-frame-guides
     ptk/UpdateEvent
@@ -91,7 +90,7 @@
             frame-ids-set    (set frame-ids)
             guide-ids        (into #{}
                                    (comp (filter #(contains? frame-ids-set (:frame-id %)))
-                                         (map :id))
+                                         d/xf:map-id)
                                    (vals guides))]
         (update-in state [:workspace-guides :hover]
                    (fn [hover] (reduce disj (or hover #{}) guide-ids)))))
--- a/frontend/src/app/main/data/workspace/guides.cljs
+++ b/frontend/src/app/main/data/workspace/guides.cljs
@@ -77,6 +77,38 @@
             guides (-> (select-keys guides ids) (vals))]
         (rx/from (mapv remove-guide guides))))))
 
+(defn remove-frame-guides
+  [frame-ids]
+
+  (dm/assert!
+   "expected a coll of uuids"
+   (every? uuid? frame-ids))
+
+  (ptk/reify ::remove-frame-guides
+    ptk/UpdateEvent
+    (update [_ state]
+      (let [{:keys [guides]} (dsh/lookup-page state)
+            frame-ids-set    (set frame-ids)
+            guide-ids        (into #{}
+                                   (comp (filter #(contains? frame-ids-set (:frame-id %)))
+                                         (map :id))
+                                   (vals guides))]
+        (update-in state [:workspace-guides :hover]
+                   (fn [hover] (reduce disj (or hover #{}) guide-ids)))))
+
+    ptk/WatchEvent
+    (watch [it state _]
+      (let [{:keys [guides] :as page} (dsh/lookup-page state)
+            frame-ids-set    (set frame-ids)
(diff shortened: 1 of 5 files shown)


frontend/src/app/main/data/workspace/guides.cljs
```diff
@@ -77,6 +77,38 @@
             guides (-> (select-keys guides ids) (vals))]
         (rx/from (mapv remove-guide guides))))))
 
+(defn remove-frame-guides
+  [frame-ids]
+
+  (dm/assert!
```
niwinz: dm/assert! is deprecated, use `assert`

frontend/src/app/main/data/workspace/guides.cljs
```diff
@@ -77,6 +77,38 @@
...
+   (every? uuid? frame-ids))
+
+  (ptk/reify ::remove-frame-guides
+    ptk/UpdateEvent
+    (update [_ state]
+      (let [{:keys [guides]} (dsh/lookup-page state)
+            frame-ids-set    (set frame-ids)
+            guide-ids        (into #{}
+                                   (comp (filter #(contains? frame-ids-set (:frame-id %)))
+                                         (map :id))
```
niwinz: Can be replaced with d/xf:map-id

frontend/src/app/main/ui/workspace/context_menu.cljs
```diff
@@ -646,6 +647,25 @@
         [:> menu-entry* {:title (tr "workspace.shape.menu.combine-as-variants")
                          :on-click do-combine-as-variants}]])]))
 
+(mf/defc context-menu-guides*
+  {::mf/props :obj
+   ::mf/private true}
+  [{:keys [shapes]}]
+  (let [frame-ids    (into #{} (comp (filter cfh/frame-shape?) (map :id)) shapes)
```
niwinz: (map :id) -> d/xf:map-id

eureka0928:
Thanks for the review! All three addressed in the latest push:
- `dm/assert!` → `assert`
- `(map :id)` → `d/xf:map-id` in both files

niwinz:
All look OK for me, my comments are not merge-blocking (can be addressed later)

niwinz:
Waiting for UX feedback from @myfunnyandy for mark this ready for merge

myfunnyandy:
hey @eureka0928, is this the menu where I should find the "Clear artboard guides" option?

I ran the branch locally and couldn't find the feature, but I might be doing something wrong.
 
image

eureka0928:
Hi @myfunnyandy! Thanks for testing. The "Clear artboard guides" option is conditional — it only appears when the selected frame actually has guides attached to it. If the frame has no guides, the menu item is hidden to avoid clutter.

To test:
1. Enable **View → Show rulers** (or Ctrl+Shift+R)
2. **Drag a guide** from the ruler and drop it **inside the boundaries of a frame/artboard** — this attaches the guide to that specific frame (you should see it highlight during the drag)
3. Select that frame and right-click → you should now see **"Clear artboard guides"** right above the Delete entry

Note: guides dropped **outside** any frame become page-level guides (not tied to any artboard) — those aren't affected by this feature, which matches the behavior requested in #6987.

Let me know if you still can't see it after following these steps and I'll investigate further!

myfunnyandy:
@eureka0928 oh now I get it. I was testing the other guides. Polysemy doesn't help here.

image

Thanks to your clarification, I've already tested it correctly. It works perfectly and I believe it can be very useful. Thanks again :)

PR #8946, opened 2026-04-12
:sparkles: Add openid-attr support and dot notation for OIDC attribute

### LINKED ISSUE #1109: NextCloud OAuth2 Integration
I could not find  a way to set the attribute for OPENID, NAME or EMAIL when implementing OpenId Connect. This is useful when the JSON schema doesn't match exactly what Penpot expects which is the case with Nextcloud.

Given the following pseudo NextCloud  JSON:
`{
 ocs: {
  data: {
id: 'someId',
  display-name: 'some name',
  email:'someemail'
}
}
}` 

A quick solution would be to create additional params as follows:

PENPOT_OIDC_OPENID_ATTR=ocs.data.id
PENPOT_OIDC_NAME_ATTR=ocs.data.display-name
PENPOT_OIDC_EMAIL_ATTR=ocs.data.email

Another solution which I think is the way Gitlab handles it is as follows:

PENPOT_OIDC_ROOT_PATH_ATTR= ocs.data
PENPOT_OIDC_MAPPING_ATTR= {'openid': 'id', 'name': 'display-name', 'email': 'email'}

This might be more complex to achieve. However it gives the user great flexibility

Sorry if this has been reported or it is already handled. I am not familiar with Penpot.

aliworksx08:
Summary

- Add dot notation support for OIDC attribute paths, enabling integration with providers that return nested JSON structures (e.g., Nextcloud's `ocs.data.email`)
- Add new `PENPOT_OIDC_OPENID_ATTR` config option to extract a custom unique user identifier from the OIDC response
- Keep `:sub` claim available in token info for providers that use it as an identifier
- Add DB migration for `openid_attr` column on `sso_provider` table
- Add unit tests for `parse-attr-path` and `process-user-info` covering flat keys, dot notation, double-underscore separator, and openid extraction

## Context

Closes https://github.com/penpot/penpot/issues/1109

Some OIDC providers (notably Nextcloud) return user info in a non-standard nested structure rather than flat top-level fields. Previously, Penpot only supported the `__` separator for nested paths, which doesn't match common dot notation conventions.

With this change, administrators can configure attribute paths like:

```
PENPOT_OIDC_EMAIL_ATTR=ocs.data.email
PENPOT_OIDC_NAME_ATTR=ocs.data.display-name
PENPOT_OIDC_OPENID_ATTR=ocs.data.id
```

The `__` separator is still supported for backward compatibility — dot notation is used when no `__` is present in the path.

**Note:** The `openid-attr` value is extracted and stored in the user's profile props (as `:oidc/openid`). It does not replace email-based user matching.

## Test plan

- [x] Standard OIDC login still works without new env vars set (backward compat) — `process-user-info-standard-flat-attrs`
- [x] `__` separator still works for existing configurations — `parse-attr-path-double-underscore`
- [x] Dot notation works for `EMAIL_ATTR`, `NAME_ATTR`, and `OPENID_ATTR` — `parse-attr-path-dot-notation`, `process-user-info-dot-notation-nextcloud`
- [x] `PENPOT_OIDC_OPENID_ATTR` extracts and stores the value in profile props — `process-user-info-dot-notation-nextcloud`


niwinz:
The changes looks OK. I only have one question: why the new openid-attr is needed. I mean it will be already on props in any way. And additionally I don't see any real use of it... Maybe I'm wrong and i'm not seeing something evident.

niwinz:
Hey @aliworksx08

I'm waiting your response for continue with this PR, just FYI

aliworksx08:
@niwinz Thanks for your suggestion and sorry for late reply.
You’re right, `openid-attr` is not needed for the current flow and it is not used anywhere separately.
I removed it and kept this PR focused on dot-notation support.

niwinz:
Thanks :heart:

carolportugal18:

PR #8985, opened 2026-04-14
:sparkles: Persist asset search query when switching sidebar tabs

### LINKED ISSUE #2913: :sparkles: Add the ability to persist search query in Assets pane

Current Penpot workflow requires you to switch between “Layers” and “Assets” panes very often. When doing so, the search query in “Search assets” gets lost. Scroll position is also lost.

Allowing user to choose style in right “Design” pane by clicking on a color would decrease the need to switch between panes, but it is for another issue which is I'm not sure is open or not.

Persisting the search query and probably scroll position.

_No response_

### Additional context

_No response_

eureka0928:
## Summary

Closes #2913

When users switch between the Layers and Assets sidebar tabs, the asset search query and section filter are lost because `assets-toolbox*` unmounts and its local `use-state` is discarded. This is a frequent papercut for users who toggle between panes while picking assets.

- Introduce a per-file, in-memory `defonce` session atom in `assets.cljs` that holds `{file-id {:term :section}}`
- Initialize `filters*` from that atom on mount, falling back to defaults when no entry exists for the current file
- Sync `:term` and `:section` back to the atom via a small `mf/with-effect`
- `:ordering` and `:list-style` continue to use `storage/user` (localStorage) — those are user preferences, not session state
- Filters are scoped per file, so switching files does not leak a stale search; reloading the browser clears them

## Test plan

- [ ] Type a query in the assets search bar, switch to the Layers tab and back — query is preserved
- [ ] Pick a section filter (e.g. Components), switch tabs and back — filter is preserved
- [ ] Open a different file — assets search starts empty (no leak across files)
- [ ] Reload the browser — search starts empty (session-only)
- [ ] Ordering toggle and list-style toggle still persist across reloads as before

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -64,6 +64,7 @@
 
 - Access Tokens look & feel refinement [Taiga #13114](https://tree.taiga.io/project/penpot/us/13114)
 - Enhance readability of applied tokens in plugins API [Taiga #13714](https://tree.taiga.io/project/penpot/issue/13714)
+- Persist asset search query and section filter when switching sidebar tabs (by @eureka0928) [Github #2913](https://github.com/penpot/penpot/issues/2913)
 
 ### :bug: Bugs fixed
 
--- a/frontend/src/app/main/ui/workspace/sidebar/assets.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/assets.cljs
@@ -70,16 +70,24 @@
   [v a b]
   (if (= v a) b a))
 
+;; Per-file, session-scoped (in-memory only) so the search term and section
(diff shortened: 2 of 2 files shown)


eureka0928:
Hi @niwinz @myfunnyandy — could I get a review on this when you have a chance? It's a small UX fix to keep the asset search query and section filter alive when toggling between the Layers and Assets sidebar tabs (closes #2913). Per-file, session-only, no localStorage. Happy to iterate on the approach if you'd prefer a different scope.

niwinz:
Looks OK for me.
Thanks :heart:

PR #9002, opened 2026-04-14
:bug: Add several bugfixes for app.common.geom namespace

niwinz:
## Summary

11 Bug Fixes (some of them are merged in a parent PR to `staging`)

| Module | Issue | Impact |
|--------|-------|--------|
| **grid_layout** | `case` on runtime booleans → `cond` | Gap calculation never worked at runtime |
| **grid_layout** | Divide-by-zero when no flex/auto tracks | JVM crash; silent Infinity in JS |
| **grid_layout** | `get(allocated idx)` returned nil | NPE in JVM when max(size, nil) |
| **transforms** | Double rotation negation (XOR bug) | Incorrect angle after dual-axis flip |
| **intersect** | `:coplanar` vs `::coplanar` mismatch | Collinear/on-segment intersections missed |
| **constraints** | `:default` arity mismatch (5 vs 6 params) | Runtime arity error on unknown constraint |
| **flex_layout** | `v-end?` guarded by `row?` not `col?` | Vertical-end alignment fired on wrong layout |
| **flex_layout** | `get-base-line` around? parenthesis | Offset clamped to ≥2px instead of computed half |
| **rect** | `update-rect :size` missed `:x1`/`:y1` | Inconsistent Rect record state |
| **point** | `gpt/abs` lost Point record type | `point?` check failed on result |
| **matrix** | Trailing comma in `matrix->str` | Malformed string serialization |

## Comprehensive Test Coverage

**11 new test namespaces** covering all fixes:
- `geom_test`, `geom_point_test`, `geom_rect_test`, `geom_shapes_test`, `geom_shapes_constraints_test`
- `geom_flex_layout_test`, `geom_grid_layout_test`

All tests pass on **JS (Node.js)** and **JVM** environments.

### DIFF
--- a/common/src/app/common/geom/point.cljc
+++ b/common/src/app/common/geom/point.cljc
@@ -151,7 +151,7 @@
                  (dm/get-prop p2 :y))))
 
 (defn multiply
-  "Returns the subtraction of the supplied value to both
+  "Returns the multiplication of the supplied value to both
   coordinates of the point as a new point."
   [p1 p2]
   (assert (and (point? p1)
@@ -509,12 +509,10 @@
   (let [old-length (length vector)]
     (scale vector (/ new-length old-length))))
 
-;; FIXME: perfromance
 (defn abs
   [point]
-  (-> point
-      (update :x mth/abs)
-      (update :y mth/abs)))
+  (pos->Point (mth/abs (dm/get-prop point :x))
+              (mth/abs (dm/get-prop point :y))))
 
 ;; --- Debug
 
--- a/common/src/app/common/geom/rect.cljc
+++ b/common/src/app/common/geom/rect.cljc
@@ -119,12 +119,14 @@
 (defn update-rect
   [rect type]
   (case type
-    :size
+    (:size :position)
     (let [x (dm/get-prop rect :x)
           y (dm/get-prop rect :y)
           w (dm/get-prop rect :width)
           h (dm/get-prop rect :height)]
       (assoc rect
+             :x1 x
+             :y1 y
              :x2 (+ x w)
              :y2 (+ y h)))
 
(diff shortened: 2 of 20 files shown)


common/src/app/common/geom/point.cljc
```diff
@@ -509,12 +509,10 @@
   (let [old-length (length vector)]
     (scale vector (/ new-length old-length))))
 
-;; FIXME: perfromance
 (defn abs
   [point]
-  (-> point
-      (update :x mth/abs)
-      (update :y mth/abs)))
+  (pos->Point (mth/abs (dm/get-prop point :x))
```
Alotor, line 514: Could be an inconsistency if `point` is a clojure map? I don't know if `get-prop` in CLJS will access properties in a plain map or this will throw an error if the point has x or y to `nil`

common/src/app/common/geom/rect.cljc
```diff
@@ -125,6 +125,8 @@
           w (dm/get-prop rect :width)
           h (dm/get-prop rect :height)]
       (assoc rect
+             :x1 x
```
Alotor, line 128: I don't think this is really necessary. If you don't change `x` and `y` you don't need to update `x1` and `y1`.

common/test/common_tests/geom_shapes_intersect_test.cljc
```diff
@@ -52,12 +52,10 @@
...
 
   (t/testing "Two collinear overlapping segments"
-    ;; NOTE: The implementation compares orientation result (namespaced keyword ::coplanar)
-    ;; against unnamespaced :coplanar, so the collinear branch never triggers.
-    ;; Collinear overlapping segments are NOT detected as intersecting.
-    (t/is (false? (gint/intersect-segments?
-                   [(pt 0 0) (pt 10 0)]
-                   [(pt 5 0) (pt 15 0)]))))
+    ;; Collinear overlapping segments correctly detected as intersecting.
+    (t/is (true? (gint/intersect-segments?
```
Alotor, line 56: It doesn't really matter, but this change should be in the commit ` Fix coplanar keyword mismatch in intersect-segments?`

common/src/app/common/geom/point.cljc
niwinz: It will work correctly.

common/src/app/common/geom/rect.cljc
niwinz: ack

common/test/common_tests/geom_shapes_intersect_test.cljc
niwinz: Done

PR #9006, opened 2026-04-15
:sparkles: Add search bar to prototype interaction destination dropdown

### LINKED ISSUE #8618: Add search/filter to interactions target board dropdown for easier navigation with many boards

The Problem my feature requests addressees is about comfort and intuitive usage behavior.

Navigate to the right side bar section called "main_ui_workspace_sidebar_options_menus_interactions__wrapper" and try to select one of your page's boards as the target board for an interaction. You quickly see, that searching for the proper entry is quite a tricky task, when your page contains even just about 40 boards.

My first proposal would induce the enhancement of the drop-down menu's item-selection by allowing a quick search using keyboard input of the first letters, or even filtering the results by input. This at least would be my first intuitive approach to handle such a task.

The second, alternative approach would be to replace the drop-down menu totally my a search-form listing all the boards on search input.

### Additional context

As the listing is already filled on first selection, i would not allow the search trough all the page's content on each char input, but rather consider searching only inside the temporary queried boards-lists. Also i would clear the search input on deselection/focus-lost to prevent issues, when search-form contains chars, and new boards are added afterwards.

### Prerequisites
- A Penpot file with a page containing **at least ~10 frames/boards**
- Prototype interactions panel visible (right sidebar)

### Steps

1. **Open a file** in the Penpot workspace and select any layer or frame.
2. **Go to the Prototype tab** in the right sidebar.
3. **Add a new interaction** (click `+` next to Interactions).
4. In the interaction settings, click the **"Navigate to" destination dropdown** to open the frame/board picker.
5. **Click on Destination field and start writing to filter frames by name — the list should update in real time to show only matching boards.
6. **Select a frame** from the filtered results and confirm the interaction is set correctly.
7. **Click outside / lose focus** on the search input — the search field should clear itself automatically.
8. Reopen the dropdown and verify the **full list is restored** (no stale filter applied).

### Edge Cases to Verify

- Search with **no matches** → list shows an empty state or "no results" message.
- Search a board, then **add a new board** while the dropdown is closed → reopen dropdown and confirm the new board appears (search doesn't persist stale results).
- Works correctly with **50+ frames** on the page without noticeable lag.

moorsecopers99:
### Related Ticket

Closes [#8618](https://github.com/penpot/penpot/issues/8618)

### Summary

On pages with a large number of boards (40+), the **Destination** dropdown used by the prototype *Navigate to* interaction becomes hard to scan. This PR adds live, keyboard-driven filtering to that dropdown without introducing a new component or disturbing any other callsite.

- Extend the shared `app.main.ui.components.select` component with an optional `searchable?` flag (and an optional `search-placeholder`). Default behavior is unchanged for every existing callsite.
- When `searchable?` is on, the dropdown renders a sticky text input at the top that filters options by `:label` (case-insensitive `cuerdas.core/includes?`). Filtering operates on the already-computed options vector, so it stays O(n) per keystroke and touches no page data.
- Arrow-key navigation now walks the filtered list instead of the full one, so up/down/Enter always match what the user sees.
- Search state is reset whenever the dropdown closes, which matches the issue's requirement that the filter not leak across reopenings.
- A new `labels.no-matches` i18n key ("No matches") renders when the filter empties the list.
- Opted in only on the Interaction **Destination** select (`interactions.cljs`); Trigger, Action, Relative to, Animation and Easing dropdowns are untouched.

### Steps to reproduce

1. Open a file that contains many boards (e.g. import the *Design Tokens Starter Kit* template, or create ≥ 10 boards on a page).
2. Select any shape → **Design** tab → **Interactions** → **+** → expand the new interaction.
3. Set *Action* to **Navigate to**.
4. Open the **Destination** dropdown:
   - A search input appears at the top and is autofocused.
   - Typing filters the listed boards case-insensitively.
   - Arrow keys walk only the filtered boards; Enter closes the dropdown.
   - Typing a string with no matches shows a **"No matches"** row.
   - Close the dropdown without selecting, then reopen it — the search field is empty again (filter does not persist).
5. Verify other selects in the same panel (Trigger, Action, Relative to, Animation, Easing) are unchanged — no search field.
6. Verify any other `select` callsite in the app (e.g. dashboard sorting, color profile selectors) still behaves exactly as before.

### Checklist

### Screenshots

https://github.com/user-attachments/assets/b17e4056-44f9-4bcf-ba67-4f0c6f3ec786

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -47,6 +47,7 @@
 - Allow customising the OIDC login button label (by @wdeveloper16) [Github #7027](https://github.com/penpot/penpot/issues/7027)
 - Add page separators in Workspace [Taiga #13611](https://tree.taiga.io/project/penpot/us/13611?milestone=262806)
 - Add Shift+Numpad0/1/2 as aliases to Shift+0/1/2 for zoom shortcuts [Github #2457](https://github.com/penpot/penpot/issues/2457)
+- Add search bar to prototype interaction destination dropdown (by @moorsecopers99) [Github #8618](https://github.com/penpot/penpot/issues/8618)
 
 ### :bug: Bugs fixed
 
--- a/frontend/src/app/main/ui/components/select.cljs
+++ b/frontend/src/app/main/ui/components/select.cljs
@@ -15,7 +15,6 @@
    [app.util.dom :as dom]
    [app.util.i18n :refer [tr]]
    [app.util.keyboard :as kbd]
-   [app.util.timers :as ts]
    [cuerdas.core :as str]
    [rumext.v2 :as mf]))
 
@@ -66,13 +65,17 @@
 
         is-open?      (get state :is-open?)
 
-        search*       (mf/use-state "")
+        ;; nil = the user is not actively searching (input shows the current
+        ;; selection's label); a string = the user has typed (input shows that
+        ;; string and options are filtered).
+        search*       (mf/use-state nil)
         search        (deref search*)
+        searching?    (and searchable? (some? search))
         search-input-ref (mf/use-ref nil)
 
         visible-options
         (mf/with-memo [options search searchable?]
-          (if (and searchable? (not (str/blank? search)))
+          (if (and searchable? (some? search) (not (str/blank? search)))
             (let [needle (str/lower search)]
               (into [] (filter (fn [item]
                                  (and (map? item)
@@ -94,7 +97,7 @@
 
         handle-key-up
         (mf/use-fn
-         (mf/deps disabled visible-options current-value)
+         (mf/deps disabled visible-options current-value searchable?)
          (fn [e]
            (when-not disabled
              (let [options (into [] (remove :disabled) visible-options)
@@ -105,20 +108,26 @@
                (cond
                  (or (kbd/left-arrow? e)
                      (kbd/up-arrow? e))
-                 (let [value (rotate-option-backward options index length)]
-                   (swap! state* assoc :current-value value)
-                   (when (fn? on-change)
-                     (on-change value)))
+                 (when (pos? length)
+                   (let [value (rotate-option-backward options index length)]
+                     (swap! state* assoc :current-value value)
+                     (when (fn? on-change)
+                       (on-change value))))
 
                  (or (kbd/right-arrow? e)
                      (kbd/down-arrow? e))
-                 (let [value (rotate-option-forward options index)]
-                   (swap! state* assoc :current-value value)
-                   (when (fn? on-change)
-                     (on-change value)))
+                 (when (pos? length)
+                   (let [value (rotate-option-forward options index)]
+                     (swap! state* assoc :current-value value)
+                     (when (fn? on-change)
+                       (on-change value))))
 
-                 (or (kbd/enter? e)
-                     (kbd/space? e))
+                 (kbd/enter? e)
+                 (swap! state* assoc :is-open? false)
+
+                 ;; In searchable mode the input owns Space — let the user
+                 ;; type spaces in the filter without closing the dropdown.
+                 (and (not searchable?) (kbd/space? e))
                  (swap! state* assoc :is-open? false)
 
                  (kbd/tab? e)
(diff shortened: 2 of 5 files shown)


EvaMarco:
Thanks a lot for your contribution — we really like this addition 🙌 The search capability is a great improvement, especially for selectors with a large number of options.

That said, we’d like to suggest a slight change in the approach. Instead of placing the search field inside the dropdown alongside the options, we’d prefer the visible input (the one outside) to act as the search field itself.

So the idea would be:

* The input filters the available options as the user types
* It behaves like a filterable select, but **without allowing free text input** (i.e. users can only select from the provided options, not enter arbitrary values)

This would align better with our UX patterns and keep the interaction a bit more streamlined.

Let us know what you think or if you need any clarification — and thanks again for the great work!

moorsecopers99:
Hi @EvaMarco — thanks for the review and the kind words! 🙌

Updated in [f47fefe](https://github.com/penpot/penpot/pull/9006/commits/f47fefed84197f94e93b96a4160ec0d58c1221b0). The visible trigger is now the search field itself, no extra input inside the dropdown:

- The trigger renders as a real `<input>` (only when `searchable?` is set on the `select` component); other selects keep the existing `<span>` label.
- When closed, the input shows the currently selected option's label and behaves like a normal trigger (it's `cursor: pointer`, no border).
- Clicking or focusing the input opens the dropdown and pulls focus to the input so the user can type immediately.
- Typing filters the options live; the input shows the typed text and the dropdown shows the filtered list (`No matches` row when nothing matches).
- Selecting an option (click or arrow + Enter) closes the dropdown; the input reverts to showing the selected label. Selection is the only way to commit a value — there is no free-text path. If the user closes the dropdown without selecting (Escape, click outside), the search resets and the previous selection is preserved.
- Space no longer closes the dropdown in `searchable?` mode so it can be typed into the filter; arrow / Enter / Esc behaviors are unchanged.

The in-dropdown sticky search row and its SCSS are gone. Non-searchable callsites (Trigger, Action, Relative to, Animation, Easing, plus every other consumer of `select`) still render the original `<span>` and behave exactly as before.

Let me know if you'd like any adjustments to the styling or interaction.

EvaMarco:
Hello @moorsecopers99 !

First of all, apologies for the delay in getting back to you, and thanks again for your contribution 🙌

We really like the direction of your proposal, but to better align it with the rest of the application, we’d need a couple of adjustments:

* Please add proper focus styles to the input.

* And hover states for the options in the list.

It should look like this:
Screenshot from 2026-05-06 10-58-50


To make things clearer and avoid guesswork, here’s a reference of the kind of styles and tokens we’re using in the codebase:

```scss
@use "ds/_borders.scss" as *;
@use "ds/_sizes.scss" as *;
@use "ds/typography.scss" as *;
@use "ds/colors.scss" as *;

.input-wrapper {
  --input-bg-color: var(--color-background-tertiary);
  --input-icon-color: var(--color-foreground-secondary);
  --input-outline-color: none;
  --input-height: #{$sz-32};
  --input-margin: unset;

  background: var(--input-bg-color);
  border-radius: $br-8;
  outline: $b-1 solid var(--input-outline-color);

  &:hover {
    --input-bg-color: var(--color-background-quaternary);
  }

  &:has(*:focus-visible) {
    --input-bg-color: var(--color-background-primary);
    --input-outline-color: var(--color-accent-primary);
  }

  &:has(*:disabled) {
    --input-bg-color: var(--color-background-primary);
    --input-outline-color: var(--color-background-quaternary);
  }

  &[data-option-focused="true"]:has(*:focus-visible) {
    --input-bg-color: var(--color-background-tertiary);
    --input-outline-color: none;
  }
}

.option {
  --options-fg-color: var(--color-foreground-primary);
  --options-bg-color: unset;
  --options-empty: var(--color-canvas);

  outline: $b-1 solid var(--options-outline-color);
  outline-offset: calc(-1 * $b-1);
  background-color: var(--options-bg-color);
  color: var(--options-fg-color);

  &:hover,
  &[aria-selected="true"] {
    --options-bg-color: var(--color-background-quaternary);
  }
}
```
I've added only styles referencing colors, border, outline, background... 

Following this structure (using tokens, avoiding inline styles, and keeping states consistent) will help ensure everything fits nicely with the rest of the UI.

Let us know if you have any questions — happy to help!

superalex:
Thank you very much!, the content of this PR has been included in https://github.com/penpot/penpot/pull/9769 <3

PR #9007, opened 2026-04-15
:bug: Fix Assets panel crash when component root is not a frame

### LINKED ISSUE #8606: MCP-created components break Assets panel when the component root is not a frame
## Summary

Components created via MCP and added to the Assets -> Components panel breaks the app when clicking on it when the component root is not a `frame`.

This does **not** happen when creating the same component manually in the Penpot UI.

## Reproduction

1. Connect Claude Code or Codex via Penpot MCP
2. Use MCP to create any of these root shape:
   - `rectangle`
   - `circle`
   - `text`
2. Save it as a component in Assets.
3. In Penpot UI navigate to Assets -> Components and click on the create component.

## Verified

- MCP `frame` root: works
- MCP `rectangle` root: breaks
- MCP `circle` root: breaks
- MCP `text` root: breaks
- MCP `frame` with nested `rectangle`/`circle`/`text`: works
- Manual `rectangle` root: works
- Manual `circle` root: works

### Expected behavior

MCP-created components should behave the same as manually created components and should not break the app.

### Actual behavior

MCP-created components with non-`frame` roots break the Assets panel and the Internal Errorpage is shown.

### Screenshots or video

Image

*Report includes a `Hint: No matching clause: circle` in the logs.

_No response_

_No response_

- Penpot Version: `2.13.3`
- Self-hosted via Proxmox 8.4.13 -> Portainer
- Using Claude Code and Codex on latest MacOS on Mac Mini M1

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

moorsecopers99:
### Related Ticket

Closes [#8606](https://github.com/penpot/penpot/issues/8606)

### Summary

Components created through the MCP integration can have a non-frame shape (rectangle, circle, text, …) as their root. Clicking such a component in **Assets → Components** crashed the panel with the JS error \`No matching clause: circle\` (or \`rect\` / \`text\`).

Both \`component-svg\` and \`component-symbol\` in \`frontend/src/app/main/render.cljs\` branched on the root shape's \`:type\` with a \`case\` that only had entries for \`:group\` and \`:frame\`. Any other root type fell through with no default and threw.

The fix reuses the existing \`shape-wrapper-factory\`, which is already the canonical wrapper used elsewhere in the rendering pipeline. Its internal \`case\` covers every shape type (\`:frame\`, \`:group\`, \`:rect\`, \`:circle\`, \`:text\`, \`:path\`, \`:image\`, \`:bool\`) and it adds its own \`shape-container\`, so:

- Components with a frame/group root render exactly as before (same wrapper, same container).
- Components with a non-frame root now render correctly instead of crashing.
- No new pattern-matching surface is introduced — bug reduction by removing duplicated, incomplete \`case\` expressions.

### Steps to reproduce

Before the fix:

1. Self-host or run Penpot locally with the MCP integration enabled.
2. Use Claude Code / Codex (or any MCP client) to create a component whose root is a rectangle, circle, or text shape — for example, ask it to "Create a circle component named foo".
3. Open the file in Penpot, switch to the **Assets** tab → **Components** section.
4. Click the newly created component.
5. Observe the Assets panel error overlay; the browser console shows \`Error: No matching clause: circle\` (or \`rect\` / \`text\`) thrown from \`render.cljs\`.

After the fix (verified locally on the devenv):

1. Repeat the same steps. The component preview/thumbnail renders correctly in the Assets panel for circle, rectangle, and text roots.
2. Manually-created components and frame-rooted components continue to render as before — no visual regressions in the Assets panel for the standard cases.
3. Dragging any of these components onto the canvas still creates a usable instance.

### Checklist

- [x] Choose the correct target branch; use \`develop\` by default.
- [x] Update the \`CHANGES.md\` file, referencing the related GitHub issue, if applicable.


niwinz:
Hi @moorsecopers99 
Thanks for the contribution.

I think the problem is other, the MCP should not be able to create components when the root is not a frame, so i think the bug is in an other part of code. I'm not going to close this but i will raise this on the  MCP team for properly address this. Be patient because this issure may require special attention and probably more work.

moorsecopers99:
Thanks for the careful read, @niwinz — that makes sense. You're right that the real root cause is upstream: the MCP shouldn't be letting a non-frame root through to create-component in the first place, and this PR only guards the symptom at the add-component call site.

Happy to hold off on further work here while the MCP team looks at it. If it helps, I can look at where in the MCP tool/schema side the frame-root invariant should be enforced (validation on the tool input or a server-side check in the RPC path before it ever reaches add-component) and open a separate PR scoped to that—just let me know if you'd prefer I wait for the MCP team's direction first so we don't duplicate effort.

Either way, happy to rework this one once you've heard back from them. Thanks!

niwinz:
I think is a safe default but still no response from the team, sorry for the delay :D

opcode81:
> the MCP should not be able to create components when the root is not a frame

@niwinz Penpot itself allows the creation of components from shapes like rectangles though (via the UI), so I think the issue is that the implementation of the API method `Library.createComponent` does not behave in the same way as the corresponding UI operation. The method should create a valid component (or raise an exception if that's not possible).

niwinz:
> > the MCP should not be able to create components when the root is not a frame
> 
> @niwinz Penpot itself allows the creation of components from shapes like rectangles though (via the UI), so I think the issue is that the implementation of the API method `Library.createComponent` does not behave in the same way as the corresponding UI operation. The method should create a valid component (or raise an exception if that's not possible).

Thanks @opcode81 for the hint, so we need to look the createComponent impl for fix the bug.

Alotor:
I close this MR, we'll solve the problem in the plugins API so it cannot create components without board.

PR #9011, opened 2026-04-15
:bug: Fix on nitrate leave org default org team must be deleted if empty

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot-nitrate/task/26270

### Summary

When an user leaves an org, the "Your penpot" team on the org must be deleted if it is empty

### Checklist

### DIFF
--- a/backend/src/app/rpc/commands/nitrate.clj
+++ b/backend/src/app/rpc/commands/nitrate.clj
@@ -107,13 +107,16 @@
 
         valid-teams-to-leave-ids    (into valid-teams-to-transfer-ids valid-teams-to-exit-ids)
 
-        all-teams                   (->> (:teams org-summary)
-                                         (map :id)
-                                         (into #{}))
-        selected-team-ids           (conj (into (into #{} teams-to-delete)
-                                                (map :id teams-to-leave))
-                                          default-team-id)
-        all-teams-selected?         (= all-teams selected-team-ids)
+        ;; Get all the teams ids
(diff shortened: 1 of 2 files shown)


backend/src/app/rpc/commands/nitrate.clj
```diff
@@ -100,6 +107,18 @@
 
         valid-teams-to-leave-ids    (into valid-teams-to-transfer-ids valid-teams-to-exit-ids)
 
+        all-teams                   (->> (:teams org-summary)
```
niwinz: `(into #{} d/xf:map-id (:teams org-summary))`

backend/src/app/rpc/commands/nitrate.clj
```diff
@@ -100,6 +107,18 @@
...
+                                         (map :id)
+                                         (into #{}))
+        selected-team-ids           (conj (into (into #{} teams-to-delete)
+                                                (map :id teams-to-leave))
+                                          default-team-id)
```
niwinz: All this looks very strange, if we expect teams to delete be a set, i think we can enforce it from input schema directly.

In any case this code should look like this I think
```clojure
(-> (into #{default-team-id} teams-to-delete)
    (into d/xf:map-id teams-to-leave))
```

pabloalba: No, we only need a set to check tah all the teams will be processed (are in one of the lists).
I've added comments to explain this

niwinz (approved):

PR #9025, opened 2026-04-16
:bug: Fix token validation failure caused by group nodes in token tree

### LINKED ISSUE #9010: Token validation fails when a malformed token exists in the Component category
### Steps To Reproduce

Token creation and editing fail in Penpot when a specific category (e.g. Component) is present. The issue does not appear to be related to the input format (hex, RGB, or color picker), but to an inconsistency or invalid structure somewhere in the token tree.

### Expected behavior

- Tokens should save correctly when using valid color inputs
- Validation should either:
- succeed if the new token is valid, or
- clearly identify which existing token is invalid

### Actual behavior

Penpot seems to validate the entire token set on save, so a single malformed token blocks all changes.

- Token creation/editing fails
- Token Manager behaves inconsistently

“Resolved value” shows unexpected structure like:

{"$meta$": null, "$cnt$": ...}

- Removing the Component category restores normal behavior

### Screenshots or video

Image

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

The issue appears to be caused by invalid or inconsistent token definitions within the token tree. Likely causes include:

- Mixed token formats (e.g. raw values like "color": "#fff" alongside $value / $type)
- Missing required fields (e.g. $type: "color")
- Broken or invalid references between tokens
- Presence of non-standard fields ($meta$, $cnt$, etc.), possibly introduced via MCP/Claude
- A category (likely Component) using a different or malformed schema

The “resolved value” structure suggests corrupted or non-standard token data rather than a UI issue.

moorsecopers99:
### Related Ticket

Closes https://github.com/penpot/penpot/issues/9010

### Summary

When a token set contains group categories (e.g. "Component"), `tokens-tree` produces a nested map where intermediate group nodes are plain maps, not `Token` records. StyleDictionary processes the entire tree and returns **all** nodes — both leaf tokens and group nodes. `process-sd-tokens` then tried to look up an origin token for every node, but group nodes have no matching entry in the flat token map, so `get-origin-token` returned `nil`.

This caused:
- `(:type nil)` → `nil`, falling through to the default parser which tried to parse a ClojureScript map as a value string, producing garbage like `{"$meta$": null, "$cnt$": ...}`
- The garbage entry (with `nil` name) polluted the resolved-tokens map
- Downstream validation saw the malformed entry and blocked all token creation/editing

**Fix:** In `process-sd-tokens`, skip any `sd-token` whose origin lookup returns `nil` — these are group nodes, not actual tokens.

### Steps to reproduce

1. Create a token set with a nested group (e.g. "Component.Primary.color")
2. Add another token in a different category
3. Try to create or edit any token — validation fails with no clear error
4. After the fix, tokens save correctly; the "Component" group no longer interferes

### Checklist


hirunatan:
Hi @moorsecopers99. Thanks for your contribution!

Looking at the PR, it seems correct at first. But I've tried to reproduce the error you describe, but I can't. Following your steps it seems to work ok (and we usually manage files with lots of tokens, with groups, and working well).

Could you please send a more detailed steps list to reproduce the error? Thanks again.

moorsecopers99:
Hi @hirunatan, thanks for taking a look!

You're right that groups alone don't trigger it — the issue only surfaces under a specific combination. Here are more precise steps:

Reproduction

1. Create a fresh file and open the Tokens panel.
2. Import a DTCG JSON token set that contains a group named Component (or any group) with nested tokens, e.g.:
  {
    "Component": {
      "Primary": {
        "color": { "$value": "#ff0000", "$type": "color" }
      }
    },
    "global": {
      "radius": { "$value": "4px", "$type": "borderRadius" }
    }
  }
3. After import, try to create or edit any token (not just the ones under Component).
4. Validation fails and the "Resolved value" for some entries shows {"$meta$": null, "$cnt$": ...}.

What I think is happening

The trigger seems to be when tokens-tree produces a nested map where StyleDictionary ends up emitting the intermediate group node as an sd-token itself. get-origin-token returns nil for that group node (no matching flat entry), then (:type nil)
falls through to parse-sd-token-general-value, which stringifies the CLJS map — that's where $meta$/$cnt$ come from.

If your test files import cleanly, it may be that your group structure gets flattened differently, or the StyleDictionary version in your env skips empty parents. Could you share a minimal JSON file you've been testing with? I'll diff it against the one from the original issue (#9010) and pinpoint the exact structural difference.

hirunatan:
Thanks for the revision. I'm merging it because I think it's a valid protection against unwanted behavior from SD. Anyway I haven't been able to reproduce the error. I imported exactly the json file you gave, and it works well. But it's ok, thanks indeed.

See https://github.com/penpot/penpot/pull/9825

PR #9034, opened 2026-04-16
:bug: Detect duplicated token names in the whole library

### LINKED ISSUE #9256: Crash on create/edit token named "white" and "black"
See Taiga https://tree.taiga.io/project/penpot/issue/13404

hirunatan:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/13404

### Summary

Forbids having tokens with conflicting paths (eg. a token "black" and a group "black.xxx") in different sets.

### Steps to reproduce 

- Create a set with a group "black.xxx" containing several tokens.
- Create another set, and inside it:
  - Try to create a token "black" using the token creation form.
  - Try to rename an existing set to "black".
  - Try to create a token "black" from a Penpot plugin.
  - Try to import a json file that has both the token and the groups.

In all cases Penpot allows to do it, but it should give an error and prevents it.

### Other bug

This PR uses the migration to also solve a nested bug under this issue
https://tree.taiga.io/project/penpot/issue/13849

(see comment from @NatachaMenjibar on 05 May 2026 10:49)

### Checklist

- [] Add or modify existing integration tests in case of bugs or new features, if applicable.
- [] Update the `CHANGES.md` file, referencing the related GitHub issue, if applicable.

### DIFF
--- a/common/test/common_tests/types/tokens_migrations_test.cljc
+++ b/common/test/common_tests/types/tokens_migrations_test.cljc
@@ -1,0 +1,214 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC
+
+(ns common-tests.types.tokens-migrations-test
+  #?(:clj
+     (:require
+      [app.common.data :as d]
+      [app.common.test-helpers.ids-map :as thi]
+      [app.common.time :as ct]
+      [app.common.types.tokens-lib :as ctob]
+      [clojure.datafy :refer [datafy]]
+      [clojure.test :as t])))
+
+#?(:clj
+   (t/deftest test-v1-5-fix-token-names
+
+     (t/testing "empty tokens-lib should not need any action"
+       (let [tokens-lib  (ctob/make-tokens-lib)
+             tokens-lib' (ctob/fix-conflicting-token-names tokens-lib)]
+         (t/is (empty? (d/map-diff (datafy tokens-lib) (datafy tokens-lib'))))))
+
+     (t/testing "tokens with valid names should not need any action"
+       (let [tokens-lib  (-> (ctob/make-tokens-lib)
+                             (ctob/add-set (ctob/make-token-set
+                                            :id (thi/new-id! :set1)
+                                            :name "set1"
+                                            :tokens {"name1" (ctob/make-token
+                                                              {:id (thi/new-id! :token1)
+                                                               :name "name1"
+                                                               :type :border-radius
+                                                               :value "1"})}))
+                             (ctob/add-set (ctob/make-token-set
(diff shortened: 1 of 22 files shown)


frontend/playwright/ui/specs/tokens/crud.spec.js
```diff
@@ -1687,29 +1687,58 @@ test.describe("Tokens - creation", () => {
     // Submit button should remain disabled when value is empty
     await expect(submitButton).toBeDisabled();
   });
+});
```
hirunatan, line 1691: Test for duplicate color token was duplicated 😆

common/test/common_tests/types/tokens_migrations_test.cljc
```diff
@@ -0,0 +1,214 @@
...
+  #?(:clj
+     (:require
+      [app.common.data :as d]
+      [app.common.test-helpers.ids-map :as thi]
+      [app.common.time :as ct]
+      [app.common.types.tokens-lib :as ctob]
+      [clojure.datafy :refer [datafy]]
+      [clojure.test :as t])))
+
+#?(:clj
```
niwinz: Why only :clj branch? do we have code that is not compatible in cljs?

frontend/translations/en.po
```diff
@@ -8366,13 +8366,21 @@ msgstr "Import Error: Invalid token data in JSON."
...
+
 #: src/app/main/data/workspace/tokens/errors.cljs:28
 msgid "workspace.tokens.invalid-json-token-name-detail"
 msgstr ""
 "\"%s\" is not a valid token name.\n"
 "Token names should only contain letters and digits separated by . "
 "characters and must not start with a $ sign."
 
+#: src/app/main/data/workspace/tokens/errors.cljs:33
+msgid "workspace.tokens.duplicated-json-token-name-detail"
```
niwinz, line 8476: This should be `errors.tokens.duplicated-json-token-name-detail`

niwinz: And this should be applied to other token errors. The key should be explcit, remember that this is the first entry point for a translator

common/test/common_tests/types/tokens_migrations_test.cljc
hirunatan: True! This file was copied from a test that had things with fressian. But this is not the case here. Removed.

frontend/translations/en.po
hirunatan: I can rename all keys in `main.data.workspace.tokens.errors`. But I think this should be in a different PR.

hirunatan: Here https://github.com/penpot/penpot/pull/9207

niwinz:
The next week i will want to look to do the same but without crating a new version for tokens lib serialization.

hirunatan:
@niwinz I have added another check and repair migration, for a different case found in the same file of the original bug.

Anyway, this still can't be merged because you said you want to do something to avoid generating a new data version of the library.

niwinz (approved):

PR #9050, opened 2026-04-17
:bug: Fix profile avatar not updating when account name changes

### LINKED ISSUE #9046: bug: Profile avatar isn't changed depend on name's first letter when name is changed
### Steps To Reproduce

1. Open Your Account
2. Change your profile name or name's first letter
3. Observe avatar
4. Move other tab (e.g. Settings) or reloading
5. Observe avatar again

### Expected behavior

The avatar should be changed depend on name's first letter.

### Actual behavior

The avatar is fixed.

### Screenshots or video

https://github.com/user-attachments/assets/f0f943fa-bd8d-4f55-8fc1-2542bb0a473f

_No response_

_No response_

https://design.penpot.app

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

moorsecopers99:
### Related Ticket

Closes https://github.com/penpot/penpot/issues/9046

### Summary

When a user with no uploaded photo renames their account from the Profile settings, the generated avatar (first letter initials) does not update. A full reload or tab switch is required before the new initial appears.

**Root cause.** `on-submit` in `frontend/src/app/main/ui/settings/profile.cljs` emits `update-profile` (optimistic) followed by `persist-profile` (RPC). `persist-profile`'s `WatchEvent` builds its request from `(:profile state)`, but the optimistic `set-profile` returned by `update-profile` has not been applied to state yet when that watch runs. The RPC therefore sends the *old* fullname, and when the server echoes that stale profile back, `set-profile` overwrites the optimistic update — the UI reverts to the old first letter.

**Fix.** `persist-profile` now accepts an explicit `:profile` kwarg and deep-merges it over state before selecting request params. Both submit handlers (`settings/profile.cljs` and `settings/options.cljs`, which has the same pattern) pass the already-cleaned form data in. The existing `(persist-profile)` call in `toggle-theme` still works: a nil `:profile` deep-merges to the state value.

**Bonus.** Also changes the generated avatar's background color to be derived from the first letter (12-color palette, char-code mod 12) so different initials are visually distinguishable, per discussion on the issue.

### Steps to reproduce

1. Log in with an account that has no uploaded photo.
2. Open **Your Account → Profile**.
3. Change the first letter of the name (e.g. `John → Kohn`) and click **Save**.
4. Before the fix: the avatar still shows the old letter until you navigate to another settings tab and back, or reload.
5. With the fix: the avatar updates to the new letter immediately, and the background color changes accordingly.

### Screenshots

https://github.com/user-attachments/assets/0f109aeb-b667-4a96-9eaf-25c4952dacce

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -51,6 +51,7 @@
 ### :bug: Bugs fixed
 
 - Fix Copy as SVG: emit a single valid SVG document when multiple shapes are selected, and publish `image/svg+xml` to the clipboard so the paste target works in Inkscape and other SVG-native tools [Github #838](https://github.com/penpot/penpot/issues/838)
+- Fix profile avatar not updating when the account name changes [Github #9046](https://github.com/penpot/penpot/issues/9046)
 - Reset profile submenu state when the account menu closes (by @eureka0928) [Github #8947](https://github.com/penpot/penpot/issues/8947)
 - Add export panel to inspect styles tab [Taiga #13582](https://tree.taiga.io/project/penpot/issue/13582)
 - Fix styles between grid layout inputs [Taiga #13526](https://tree.taiga.io/project/penpot/issue/13526)
--- a/frontend/src/app/main/data/profile.cljs
+++ b/frontend/src/app/main/data/profile.cljs
@@ -120,14 +120,15 @@
 ;; --- Update Profile
 
 (defn persist-profile
-  [& {:as opts}]
+  [& {:keys [profile on-success on-error]
+      :or {on-success identity
+           on-error rx/throw}}]
   (ptk/reify ::persist-profile
     ptk/WatchEvent
     (watch [_ state _]
-      (let [on-success (:on-success opts identity)
-            on-error   (:on-error opts rx/throw)
-            profile    (:profile state)
-            params     (select-keys profile [:fullname :lang :theme])]
+      (let [profile' (d/deep-merge (:profile state)
+                                   (dissoc profile :props))
+            params   (select-keys profile' [:fullname :lang :theme])]
         (->> (rp/cmd! :update-profile params)
              (rx/tap on-success)
              (rx/map set-profile)
--- a/frontend/src/app/main/ui/settings/options.cljs
+++ b/frontend/src/app/main/ui/settings/options.cljs
@@ -30,7 +30,8 @@
   [form _event]
   (let [data  (:clean-data @form)]
     (st/emit! (du/update-profile data)
-              (du/persist-profile {:on-success on-success}))))
+              (du/persist-profile :profile data
+                                  :on-success on-success))))
 
 (mf/defc options-form
   {::mf/wrap-props false}
--- a/frontend/src/app/main/ui/settings/profile.cljs
+++ b/frontend/src/app/main/ui/settings/profile.cljs
@@ -33,7 +33,8 @@
                      (swap! form assoc :touched {})
                      (st/emit! (ntf/success (tr "notifications.profile-saved"))))]
     (st/emit! (du/update-profile data)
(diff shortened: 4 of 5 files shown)


niwinz:
Hey @moorsecopers99 
Thanks for you contribution.

Can you explain me better the issue? I mean, i'm unable to reproduce the issue without your fix.

This video is from develop:
https://github.com/user-attachments/assets/4b661707-8930-4a61-b24b-cc4518b2ff81

As events work, the `(set-profile)` event is fired synchronously before persist-profile, if not, persist-profile will not work at all.

moorsecopers99:
Hi @niwinz 
Review again on `fix/profile-avatar-update` branch, please.

niwinz:
Maybe i have not explained well myself.
I can't reproduce the issue (without your PR). I mean i can't find the bug. And the changes proposed on this PR in my opinion do not proceed because they changes something that in essence is already working correctly from my point of view (as it stated on the attached video)

moorsecopers99:
Hi @niwinz — thanks for looking again.

I think we may be looking at different things. The bug isn't in the form field or in whether persist-profile runs; it's specifically about the avatar circle (the colored disc with the first letter of the name) shown in the top bar / sidebar when the user has no uploaded photo.
The form itself updates fine — that's why your video looks correct.

Exact repro on dev (no patch):

1. Log in with an account that has no uploaded photo, so the avatar is rendered as the first-letter circle.
2. Go to Your Account → Profile.
3. Change the first letter of the name (e.g., John → Kohn) and click Save.
4. Watch the avatar disc (not the input field). It keeps showing the old initial (J) until you switch settings tabs and come back or reload the tab.

On the ordering point: you're right that set-profile is dispatched before persist-profile, and the form reflects that optimistic update.
What I observed is that the RPC response then writes back a profile value that doesn't reflect the just-saved fullname for the avatar-deriving path, so the stale initial sticks until the next reload/refetch. That's the case the patch targets.

If it's useful, I can record a short video on clean develop (not my branch) showing the stale initial after save, so we're definitely looking at the same scenario. Happy to adjust or drop the PR if it turns out to be environment-specific on my side.

Want me to tighten it further, drop the root-cause paragraph, or reframe it as a question rather than a defense?

niwinz:
A video would be awesome, because i'm not seeing the issue but i'm on high context switching so is pretty normal for my situation. A small video showing the bad behavior and then other showing the good will make this more easy. Thank you very much.

moorsecopers99:
Absolutely, happy to put together a side-by-side video. I'll record two short clips on the same account (no uploaded photo):

Before (develop): open Your Account → Profile, change the first letter of the name (e.g., John → Kohn), and click Save—the initial in the avatar stays on the old letter.

After this PR: the same steps on the PR branch—the initial flips immediately on save, and the background color also shifts (since it's now derived from the new letter).

Thanks for taking a look!


Before(https://design.penpot.app): https://github.com/user-attachments/assets/c76e4944-5a7f-4855-8bd7-e70c4ae91a61

After(`fix/profile-avatar-update` branch): https://github.com/user-attachments/assets/bd53d21a-35c8-465b-86fc-eb4dee31a4b6

niwinz:
Hey @moorsecopers99 
Thanks a lot for your videos. But i'm still unable to reproduce this issue.
This is the video on `develop`:

https://github.com/user-attachments/assets/746ba3de-a958-4d2d-952e-36acad6f2507


I have tested it in serveral branches and environments and different browsers.
I mean this already works without the patch.

I will ask other teammates for check this but until we can reproduce the real issue, i put this PR on hold and in draft.

niwinz:
I asked and 3 other teammates confirmed that it already works as expected, so, unfortunatelly i will close the PR without merging.
In any case, thanks for your time.

PR #9098, opened 2026-04-22
Perf: band-aware tile scheduler rebuild optimizations (2–2.5× faster)

dhiaeddineTounekti:
### Related Ticket

_No ticket — follow-up perf work on the band-aware tile scheduler introduced in #3._

### Summary

Optimizes the rebuild path of the band-aware tile scheduler (`render-wasm/src/tile_grid.rs`) without changing scheduler behavior. All 73 unit tests pass, including the 6 `tile_grid::scheduling_tests` correctness tests.

**E2E rebuild benchmarks** (1k shapes, release, aarch64-apple-darwin):

| Scenario | Before | After | Speedup |
|---|---:|---:|---:|
| 0 gathers | 1.12 ms | 1.38 ms* | ≈ 1.0× |
| 5 gathers | 2.94 ms | 1.50 ms | **1.96×** |
| 20 gathers | **8.96 ms** | **3.84 ms** | **2.33×** |

<sub>* 0-gather case moves slightly because the merge of `develop` made the emit visibility check always use `extrect` (needed for scatter correctness). Pre-merge, this branch hit 0.93 ms on 0 gathers; post-merge it's the same logic as `develop` plus the cache.</sub>

**What changed (all in `render-wasm/src/tile_grid.rs`):**

- **`build_schedule`** now builds a per-tile cache of visible root children, but only for *productive* tiles (tiles with at least one shape-bearing band). A 6400×6400 viewbox has ~2 500 synthetic empty-tile bands whose emit body is a no-op — previously every one of them still iterated all 1 020 root children. Fixes a 5× regression that appeared when an earlier revision of this PR cached unconditionally.
- **`emit_shape_steps_checked`** takes a `skip_self_check` flag so the root-level caller skips the intersection test the cache already performed. Nested children still run the full check. Visibility check uses `extrect` unconditionally for scatter-shape correctness.
- **`build_dependency_graph`** replaces the O(G × B) "strictly-below-gather" backstop with a prefix walk over a once-sorted list of non-gather bands (`partition_point` by `max_paint_order`).
- **`compute_bands`** drops its defensive per-tile sort. The paint-order invariant on `self.grid[tile]` is maintained by `add_shape_at`'s binary-search insert and by a new final `sort_unstable_by_key` pass at the end of `renumber_paint_order` (which mutates entries in place).

Also in this PR: a clean merge of `develop` into this branch, picking up the noise/texture/scatter effects and the wasm loader fix from #4. Merge conflict resolutions:

- `render/surfaces.rs`, `render.rs` — accepted develop's scatter cache + texture module.
- `tile_grid.rs` — accepted develop's scatter-aware `RenderStep::Render` handler (which also fixes a drop-shadow TODO this branch had) and its extrect-based visibility, then re-applied the perf optimizations on top.
- `skia-rs-wasm/public/wasm/render-wasm.{js,wasm}` — took develop's build artifacts; they regenerate on the next wasm build.

### Steps to reproduce

From `render-wasm/`:

\`\`\`
cargo test --release --bin render_wasm --features tile-scheduler \\
  --target aarch64-apple-darwin tile_grid::bench::bench_rebuild_e2e \\
  -- --nocapture --test-threads=1
\`\`\`

Correctness:

\`\`\`
cargo test --release --bin render_wasm --features tile-scheduler \\
  --target aarch64-apple-darwin tile_grid::scheduling_tests
\`\`\`

### Checklist

- [x] Choose the correct target branch; use \`develop\` by default.
- [ ] Include screenshots or videos, if applicable. _(perf-only change, no visible output difference)_
- [x] Add or modify existing integration tests in case of bugs or new features, if applicable. _(pre-existing 6 scheduler correctness tests + 14 bench tests all pass)_
- [ ] Refactor any modified SCSS files following the refactor guide. _(n/a — Rust only)_
- [ ] Update the \`CHANGES.md\` file, referencing the related GitHub issue, if applicable.


dhiaeddineTounekti:
Oops — opened against the wrong upstream. Re-targeting the ux-plugin fork.

PR #9115, opened 2026-04-22
Add multi-page PAGES list to skia-rs-wasm LayersPanel

dhiaeddineTounekti:
## Summary
- Rework the left rail's page block into a **PAGES** section with a page count, a `+` add button, and per-page rows showing a document icon and name. The active page is highlighted with a rounded blue pill.
- Add a sibling **LAYERS** section below that shows the current page's layer tree (behavior unchanged — still uses `orderedNodesFromPage` and `setSelectedIds`).
- Wire the UI directly to the existing store APIs already exported from [skia-rs-wasm/src/lib/page-crud.ts](skia-rs-wasm/src/lib/page-crud.ts): click-to-switch uses `setActivePage`, the `+` button uses `addPage`, double-click a row to rename (committed via `commitPageMetadataUpdate`).
- Remove the inline Page name / Background editor from this panel — those fields live in the right-side panel, so keeping them here was duplicated UI.

No changes to the underlying document model, worker, or renderer — this is presentation-only wiring over already-exported store APIs.

## Test plan
- [x] Panel renders with `PAGES` header, count `1`, and `+` button on initial load
- [x] Clicking `+` creates a new page, increments the count, and makes the new page active (confirmed via `docProxy` inspection)
- [x] Clicking a different page row switches `docProxy.currentPageId` and swaps the LAYERS list to that page's shapes
- [x] Double-clicking a page row enters inline rename; Enter / blur commits, Escape cancels
- [x] Adding a shape to a page shows it in the LAYERS section; switching pages empties / repopulates the list as expected
- [x] Collapse / expand toggles on PAGES and LAYERS sections work independently
- [ ] Reviewer: sanity-check the active-pill colors under dark mode


dhiaeddineTounekti:
Created against wrong repo; reopening in ux-plugin/penpot-ai.

PR #9124, opened 2026-04-23
:sparkles: Add nitrate api endpoints to get and cancel org invitations

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot-nitrate/us/13487?milestone=512963

### Summary

Add nitrate api endpoints to get and cancel org invitations

### Checklist

### DIFF
--- a/backend/src/app/rpc/management/nitrate.clj
+++ b/backend/src/app/rpc/management/nitrate.clj
@@ -16,6 +16,7 @@
    [app.config :as cf]
    [app.db :as db]
    [app.media :as media]
+   [app.nitrate :as nitrate]
    [app.rpc :as-alias rpc]
    [app.rpc.commands.files :as files]
    [app.rpc.commands.nitrate :as cnit]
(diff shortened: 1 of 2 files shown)


backend/src/app/rpc/management/nitrate.clj
```diff
@@ -375,6 +376,87 @@ RETURNING id, name;")
...
+
+(def ^:private schema:get-org-invitations-params
+  [:map
+   [:organization-id ::sm/uuid]])
+
+(def ^:private schema:get-org-invitations-result
+  [:vector
+   [:map
+    [:id ::sm/uuid]
+    [:organization-id {:optional true} ::sm/uuid]
```
cocotime: In Malli, {:optional true} means the key may be absent from the map. But the SQL LEFT JOIN always returns organization_id and name columns as nil when there's no match.
The correct annotation for "can be nil" is [:maybe ...]:
`[:organization-id {:optional true} [:maybe ::sm/uuid]]`

Same will apply for name

backend/test/backend_tests/rpc_management_nitrate_test.clj
```diff
@@ -220,6 +220,209 @@
...
+          emails (->> result (map :email) set)
+          dedup  (->> result
+                      (filter #(= "dup@example.com" (:email %)))
+                      first)]
+      (t/is (th/success? out))
+      (t/is (= #{"dup@example.com" "valid@example.com"} emails))
+      (t/is (= 2 (count result)))
+      (t/is (some? (:id dedup)))
+      (t/is (some? (:sent-at dedup)))
+      (t/is (nil? (:org-id dedup)))
```
cocotime: Should be `:organization-id` (SQL alias)?

backend/src/app/rpc/management/nitrate.clj
pabloalba: Fixed

backend/test/backend_tests/rpc_management_nitrate_test.clj
pabloalba: Fixed

cocotime (approved):

PR #9127, opened 2026-04-23
skia-rs-wasm: drag-to-reparent (layers panel + canvas), per-type icons

dhiaeddineTounekti:
### Related Ticket

_None_

### Summary

Adds drag-to-reparent in skia-rs-wasm, mirroring Penpot's behaviour:

- **Layers panel** — drag a row onto another; top/bottom edges reorder as siblings, middle of a container reparents inside. Blue line / blue ring indicators match Penpot's layer_item.scss cues. A sticky "Drop here for root" zone appears at the bottom of the list while a drag is active so deep trees can escape back to root. Rows now show per-shape-type icons (frame/rect/circle/text/…) shared with `ShapeToolbar`.
- **Canvas** — on pointer release, selected shapes are reparented into the innermost container their new center sits inside (Penpot's drop-into-frame effect). Falls back to root when the center leaves every container, so shapes can escape past the right/bottom edges of the canvas.
- **Tile-scheduler fix (`render-wasm`)** — `ShapesPool::set_modifiers` now invalidates the cached extrect + bounds for all modified ancestors. Without this the tile index kept parents anchored to their pre-drag extrect, so children dragged beyond the parent's tile footprint were rendered only in the old tiles and cut off at the tile boundary. Adds 2 method calls per ancestor; benchmarks show <3% overhead at realistic depths (<3 µs/call at depth 10).
- **Worker bug fix** — `processMovObjects` was reading `shape.parentId` after already overwriting it with the new parent, so shapes were never removed from their original parent's `shapes[]` array. Drag-to-reparent would have corrupted state without this.

#### Layers panel drag
1. Load a page with a frame/group and some rects.
2. In the left rail, drag a rect onto the **middle** of a frame → becomes its child (indented, canvas clip now applies).
3. Drag a rect onto the **top** or **bottom** edge of a sibling → reorders without reparenting.
4. With all layers inside a frame, drag a row onto the sticky **"Drop here for root"** area that appears at the bottom → moves to root.
5. Try dropping a frame onto its own descendant → no-op (guarded by ancestor check).

#### Canvas drag-into-frame
1. Create a frame, and a rect outside it.
2. Drag the rect onto the frame and release → rect becomes a child of the frame (visible in the Layers panel).
3. Drag the rect back out past the right/bottom edge of the root canvas → rect returns to root.

#### Tile-scheduler fix (requires `TILE_SCHEDULER=1 ./build` in render-wasm)
1. Put a shape wider than its parent frame inside that frame (or turn clip off).
2. Drag it sideways far enough that it would cross tile boundaries.
3. Before fix: the shape is cut at the tile boundary. After fix: it stays fully visible.

### Checklist

- [x] Target branch: `develop`.
- [x] Summary of changes provided.
- [x] Detailed reproduction steps provided.
- [ ] Screenshots/videos.
- [x] Tests added/modified — `skia-rs-wasm/test/lib/components/LayersPanel/reparent.test.ts` (20 cases) + new `ShapesPool::set_modifiers` benches (5 cases) in `render-wasm/src/state/shapes_pool.rs`.
- [x] No SCSS changes.
- [ ] CI pending.
- [ ] `CHANGES.md` — not updated (no related issue).

### Notes for reviewer

- The `render-wasm` binaries under `skia-rs-wasm/public/wasm/` are the rebuilt `TILE_SCHEDULER=1` artifacts — rebuild locally with `cd render-wasm && TILE_SCHEDULER=1 ./build` if you want to regenerate.
- Undo/redo for canvas drag currently lands in **two separate frames** (move + reparent). Fine for correctness; folding into a single frame would require extending `applyModifiersAndCommit` to accept extra changes.
- During an active canvas drag the shape still renders under its old parent's clip (clip-on case). That's correct behaviour; the clip-escape only happens on release when the reparent commits. Discussed in the PR thread that led to this change.


dhiaeddineTounekti:
Mistakenly opened against upstream; re-creating in ux-plugin/penpot-ai.

PR #9130, opened 2026-04-23
:bug: Fix switching a team nitrate organization lose the background

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot-nitrate/task/26305

### Summary

* Fix switching a team nitrate organization lose the background
* For that, unify api schemas and how to set on a team its org info
* Fix also some minor css bugs

### Steps to reproduce 

* Requires an user that belongs several nitrate orgs
* Go to a team setings, and move it to an org
* Then move it to another org

The background image and/or org logo should be changed

### Checklist


backend/src/app/rpc/commands/nitrate.clj
```diff
@@ -249,22 +249,21 @@
...
 
 
 (def ^:private schema:add-team-to-org
   [:map
    [:team-id ::sm/uuid]
-   [:organization-id ::sm/uuid]
-   [:organization-name ::sm/text]])
+   [:organization-id ::sm/uuid]])
 
 (sv/defmethod ::add-team-to-org
```
niwinz: I think the method should be called `add-team-to-organization`

common/src/app/common/organization.cljc
```diff
@@ -0,0 +1,32 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC
+
+(ns app.common.organization)
```
niwinz: I think this NS should be on `app.common.types`
And the organization subscheme used on `app.common.types.team` should be here.

backend/src/app/rpc/commands/nitrate.clj
pabloalba: done

common/src/app/common/organization.cljc
pabloalba: done

niwinz (approved):

PR #9136, opened 2026-04-24
:bug: Fix crash pasting component with variants from shared library

### LINKED ISSUE #8144: bug: crash when copying a component from a library into a file using the library
### Steps To Reproduce

1. In a new file called "Lib" create a component with a variant (add a rectangle and Cmd+K twice).
2. Mark this file as Shared Library
3. In another file, call it "Using Lib" Connect to the "Lib" file
4. Back in the "Lib" file, copy the component you created (together with its variant).
5. Paste it in the "Using Lib" file.

### Expected behavior

I wanted the component + variants to be pasted, as if they had being created in the "Using Lib" file. This was the behaviour back in 2.12.1.

### Actual behavior

Crash:

```
Hint: error on validating file referential integrity
 [{:code :component-main-external,
   :hint "Main instance should refer to a component in the same file"
```

[Complete Report](https://github.com/user-attachments/files/24763559/report.29.txt)

### Screenshots or video

https://github.com/user-attachments/assets/806a968d-fec7-4688-9821-4535343e7366

self-hosted (develop) and design.penpot.app (2.12.13).
It doesn't happen in my self-hosted with 2.12.1

### Additional context

For the records, I ran into this while trying to isolate a component from a file in order to reproduce a bug. It is not something critical for me, but something I ran into. That said, crashes are always bad :)

For the records, if you only copy a component with no variants the crash doesn't happen.

FairyPigDev:
### Related Ticket

Closes [#8144 — crash when copying a component from a library into a file using the library](https://github.com/penpot/penpot/issues/8144).

### Summary

Copying a component with variants from a shared library file and pasting it into a file that uses that library crashed the destination with the referential-integrity validator error:

    {:code :component-main-external
     :hint "Main instance should refer to a component in the same file"}

This regressed sometime after 2.12.1 (reporter confirms the paste worked in that version and crashes in 2.12.13).

**Root cause.** The paste pipeline routes main-instance shapes through `duplicate-component` in [`common/src/app/common/logic/libraries.cljc`](common/src/app/common/logic/libraries.cljc#L119). Its `update-new-shape` helper already re-links the new outer main's `:component-id` to the freshly-created local component, but it never touches `:component-file`. The cloned shape therefore inherits `:component-file` from the source library while the new component is registered in the destination's local library (`:apply-changes-local-library? true`), leaving the main-instance dangling.

**Fix.** Extend `update-new-shape` with a clause sibling to the existing `:component-id` rewrite: when a destination file id is provided and differs from the new main's current `:component-file`, re-root the shape. The same `(= (:component-id new-shape) (:id component))` guard used for the id rewrite ensures only the outer main-instance is touched; nested shapes are unaffected.

The destination file id is threaded from the paste entry point through the two orchestration functions that already knew the source/destination distinction (`generate-duplicate-shape-change`, `generate-duplicate-component-change`) and on through `duplicate-variant` → `generate-duplicate-component` → `duplicate-component`. Same-file duplication is unaffected: callers that don't pass `:dest-file-id` skip the new clause entirely.

### Steps to reproduce

1. Create a new file called **Lib**. Inside it, create a component with a variant (add a rectangle and press `Cmd+K` twice).
2. Mark **Lib** as a Shared Library.
3. Create another file, **Using Lib**, and connect it to **Lib**.
4. Back in **Lib**, copy the component together with its variant container.
5. Paste into **Using Lib**.

**Before this PR:** destination file crashes with `:component-main-external`.

**After this PR:** the component + variants paste successfully, the destination file contains a new local component, and its main-instance references that local component (no external dangling reference).

Same-file variant duplication is unchanged and still covered by the existing `test-duplicate-variant-container` test.

### Screenshots / Video

https://github.com/user-attachments/assets/cc22eab3-b545-40e4-987a-d7b04227cecd

### Checklist

- [x] Target branch: `develop`.
- [x] Summary + reproduction steps above.
- [x] Video attached showing the fix in action.
- [x] Added unit tests in `common/test/common_tests/logic/comp_creation_test.cljc`: `test-duplicate-component-rewrites-component-file-to-destination` and `test-duplicate-component-keeps-component-file-without-dest` (regression + baseline).
- [x] No SCSS changes.
- [x] CI green.


FairyPigDev:
Hello, @niwinz 
Could you please check this PR?

common/src/app/common/logic/libraries.cljc
```diff
@@ -145,6 +154,16 @@
             (= (:component-id new-shape) (:id component))
             (assoc :component-id new-component-id)
 
+            ; Re-root the new main-instance to the destination file
+            ; when duplicating across files (Issue #8144). Mirrors
+            ; the :component-id rewrite above: only the outer main
+            ; matches `(:id component)`, so nested main-instances are
+            ; not touched here.
```
niwinz: 💄 This kind of comments should be always with `;;` (two `;`)

niwinz:
The code looks very nice and this is clearly a regression. Once merged i will backport it to the latest version for make it available to users sooner.

In any case i will ask the components team for make a second look on it.

common/src/app/common/logic/libraries.cljc
FairyPigDev: Thank you for your catch, @niwinz . I fixed this.

FairyPigDev:
Hello, @niwinz
Could you please check this PR and merge?

niwinz:
Still waiting internal response for this, sorry for the delay. Be patient please.

common/src/app/common/logic/libraries.cljc
```diff
@@ -118,8 +118,17 @@
...
+
+  When `dest-file-id` is provided and differs from the new main
+  instance's current `:component-file`, the new main-instance is
+  re-rooted to `dest-file-id`. This is needed when a component is
+  duplicated across files (e.g. pasting a component from an external
+  shared library into a file that uses that library): the new
+  component will be added to the destination file's local library, so
+  its main-instance must point there to satisfy the file-integrity
+  check `:component-main-external`. Issue #8144."
+  [component new-component-id library-data force-id delta variant-id dest-file-id]
```
hirunatan: We don't usually add comments for specific issues in the code, because this would quickly become crowded. The preferred way to register the historical decissions is to have well documented commits and pull requests (since you can later check in what commits were a particular line of code changed). And also in the regression tests you write for the bug.

You have already done both things (good work). So I think this docstring is no longer needed. I'll just write this:

"Clone the root shape of the component and all children. Generate new
  ids from all of them. Optionally set the component-file if the file where the
  new component will reside is different than the origin one."

And i would rename `dest-file-id` to `new-component-file` and put it just after `new-component-id`. The combination component-id + component-file is often managed together, and in this case both fields are handled in a similar way.

common/src/app/common/logic/libraries.cljc
```diff
@@ -145,6 +154,16 @@
             (= (:component-id new-shape) (:id component))
             (assoc :component-id new-component-id)
 
+            ;; Re-root the new main-instance to the destination file
+            ;; when duplicating across files (Issue #8144). Mirrors
+            ;; the :component-id rewrite above: only the outer main
+            ;; matches `(:id component)`, so nested main-instances are
+            ;; not touched here.
```
hirunatan: Same here. I think it's not necessary to reference the issue here. You can put the explanation about nested main instances in the line 153, because it refers to both id and file.

common/src/app/common/logic/libraries.cljc
```diff
@@ -187,8 +206,14 @@
...
 (defn generate-duplicate-component
-  "Create a new component copied from the one with the given id."
-  [changes library component-id new-component-id & {:keys [new-shape-id apply-changes-local-library? delta new-variant-id page-id]}]
+  "Create a new component copied from the one with the given id.
+
+  When `:dest-file-id` is provided and differs from the component's
+  source library, the new main-instance is re-rooted to that file so
+  the file-integrity validator (`:component-main-external`) accepts
+  the cross-file duplicate. Callers pasting from an external shared
+  library must pass the destination file id here. Issue #8144."
```
hirunatan: Same as above.

common/src/app/common/logic/libraries.cljc
```diff
@@ -2764,15 +2790,19 @@
...
-        component    (ctf/get-component libraries file-id component-id)
-        pos          (as-> (gsh/move main delta) $
-                       (gpt/point (:x $) (:y $)))
+  [changes objects page main parent-id frame-id delta libraries library-data ids-map & {:keys [dest-file-id]}]
+  (let [main-id        (:id main)
+        component-id   (:component-id main)
+        ;; Source library file id (where the component was originally
+        ;; defined). Renamed from `file-id` to make the contrast with
+        ;; `dest-file-id` explicit when duplicating across files.
+        ;; Issue #8144.
```
hirunatan: Same here. Just remove the issue # (the rest of the comment is ok).

common/test/common_tests/logic/comp_creation_test.cljc
```diff
@@ -304,6 +304,62 @@
     (t/is (= (thi/id :main1-child) (:id child1')))
     (t/is (not= (thi/id :main1-child) (:id child2')))))
 
+(t/deftest test-duplicate-component-rewrites-component-file-to-destination
+  ;; Regression test for Issue #8144. When a component is duplicated
```
hirunatan, line 308: This is indeed a good place to reference the issue, since this test is write to check its specific case. So it's perfect as is. Don't change it.

hirunatan:
I have reviewed this PR and it seems OK. Very good job, thanks!

I only have some little suggestions about the comments and docstrings, and the name of one parameter, to make them more efficient.

Please, rebase the branch to the latest `develop`, so I can test it working. Thanks very much.

hirunatan (changes requested):

FairyPigDev:
Hello, @hirunatan, @niwinz 
Thank  you for your feedback. I updated code with recent develop,  resolved and rebased with one commnet.
Please check this PR.

common/src/app/common/logic/libraries.cljc
FairyPigDev: Thank you for your feedback, @hirunatan 
I fixed this problem.

common/src/app/common/logic/libraries.cljc
FairyPigDev: Thank you for your feedback, @hirunatan 
I fixed this problem.

common/src/app/common/logic/libraries.cljc
FairyPigDev: Thank you for your feedback, @hirunatan 
I fixed this problem.

common/src/app/common/logic/libraries.cljc
FairyPigDev: Thank you for your feedback, @hirunatan 
I fixed this problem.

common/test/common_tests/logic/comp_creation_test.cljc
FairyPigDev: Thank you for your feedback, @hirunatan

niwinz (approved):

PR #9157, opened 2026-04-24
:sparkles: Allow alpha to be independent from hue in color styles

### LINKED ISSUE #3085: Allow alpha to be independent from hue in color styles/symbols

I've noticed this in both Penpot and Figma, but when I'm creating a color style in either application, the style stores and subsequently locks the opacity value to the style, meaning I'm forced to create color styles for every single instance of a color and its opacity options.

If you consider a brand having 3-4 main colors, then 6 other tints/shades for each, and you decide you only need opacity units in increments of 10%, you're looking at ~264 color styles.

It ends up looking something like this:

image

Because opacity is a separate control from hue in practice, I would like to see color styles have the option to lock/unlock opacity: 
- In a "locked" color style, the color style would act like it does currently - if you use the color style in your design, you can't change the opacity of the style. 
- In an "unlocked" color style, the style would only store the hue values, and not the alpha. This would allow the user to select a color style in the design and manually control its opacity value. Then, if the color style is updated to a different hue, the instances of that color style update the hue values but retain the designer's manual opacity setting.

The reason I'm suggesting locked/unlocked options and not simply removing alpha from color styles entirely is because I'm working under the assumption that some design teams WILL want to restrict alpha values to a select few, while other teams, like my own, don't need this functionality and want designers to have total control over the opacity of the library's color options.

_No response_

### Additional context

_No response_

davidv399:
## Related Ticket

Closes #3085

## Summary

Adds a `lock-alpha?` flag to color styles so the alpha channel can be independent from the hue. When a style has `:lock-alpha? false`, shapes that use it keep their own per-instance opacity, and later updates to the style's hue no longer overwrite that opacity. Locked styles (the default) behave exactly as they do today, so existing files are unaffected.

This unblocks the common workflow of maintaining a small set of brand color styles and adjusting opacity per shape, instead of duplicating a style for every opacity step.

## Steps to Reproduce

1. Open a file and create a color style (e.g. `#3B82F6` at 100% alpha).
2. Apply the style to a shape.
3. Try to change the shape's fill opacity to 50%.
4. Observe that the opacity control is locked to the style's value - the only workaround is to create a second color style at 50% alpha.
5. For a brand with 4 colors × 6 shades × 10 opacity steps, this forces designers to manage ~240 near-duplicate styles.

**Expected:** It should be possible to mark a color style as "alpha-unlocked" so shapes using it can have per-instance opacity without cloning the style.

## Checklist

- [ ] Color style editor exposes a "Lock alpha" toggle
- [ ] Locked styles preserve current behavior (default on)
- [ ] Unlocked styles do not overwrite shape alpha on hue updates
- [ ] Library panel shows a visual indicator for unlocked styles
- [ ] Copy-paste / duplicate / library import preserves the lock state
- [ ] Existing files migrated to `:lock-alpha? true` (no behavior change)
- [ ] Tested in Chrome and Firefox
- [ ] CHANGES.md entry added

## Video

https://github.com/user-attachments/assets/81e87cf8-069f-4d83-8aa0-5b408de25c9d


jack-stormentswe:
Hi @niwinz 👍
How are you? Please review this PR and let me know your recommendation.
Thanks.

jack-stormentswe:
Hi @niwinz , How are you?
Could you review this PR and share your recommendation?
Appreciate.

common/src/app/common/types/color.cljc
```diff
@@ -140,6 +140,10 @@
    [:name ::sm/text]
    [:path {:optional true} :string]
    [:opacity {:optional true} [::sm/number {:min 0 :max 1}]]
+   ;; When false, the style stores only hue; shapes using it keep their
+   ;; own per-instance alpha and hue updates do not overwrite it. When
+   ;; true (default, existing behavior), hue+alpha are bound together.
+   [:lock-alpha? {:optional true} :boolean]
```
niwinz, line 146: This should be `lock-alpha` (without `?`)
We should avoid usage `?` (special-chars) on data structure props, as this can be consumed by json/js where accesing this properties is "not convenient".

niwinz:
Hey @jack-stormentswe 
The code looks good. Thanks :heart: 

I leaved an inline comment.
Additionally i think we need to think on how this new propertly works with components and component synchronization. You can look it yourself or leave this to us (in this case we will priorize it on our global task list and I will not be able to give you an estimate for merging this).

common/src/app/common/types/color.cljc
davidv399: Got it. Will update asap.
Thanks for your feedback. ❤️

PR #9161, opened 2026-04-25
:bug: Fix plugin API fills/strokes arrays read-only

### LINKED ISSUE #8357: Fills and strokes array elements are read-only in plugin API instead of mutable
### Problem Description

The contents of the `fills` and `strokes` arrays are read-only.

The only way to modify fills and strokes is to fully replace the entire array, e.g.

    shape.fills = [{ fillColor: "#FF0000", fillOpacity: 1 }];

It is *not* possible to
  * change a property in an element, i.e.

        shape.fills[0].fillColor = "#FF0000";
  * replace an array element, i.e.

        shape.fills[0] = { fillColor: "#FF0000", fillOpacity: 1 };

This is undisclosed in the API documentation, where these properties are not declared as read-only, so I suppose this is likely a bug.
Notably, the current limitation can be problematic for LLM workflows, because if an object uses more complex fills, replacing a single property requires the entire array structure to be fully rebuilt, which is cumbersome and error-prone.

### Expected behavior

The aforementioned property/array element modifications can be applied.

### Actual behavior

The aforementioned property/array element modifications have no effect and fail silently.

### Screenshots or video

_No response_

Win11, Edge 144

_No response_

design.penpot.app

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

Current workaround in MCP: Instruct LLM to always reconstruct the arrays.

RenzoMXD:
## Related Ticket

Closes #8357

## Summary

Plugin reads of `shape.fills` / `shape.strokes` returned a detached JS snapshot, so per-element writes silently failed:

```js
shape.fills[0].fillColor = "#ff0000";        // no effect
shape.fills[0] = { fillColor: "#ff0000" };   // no effect
shape.fills.push({ fillColor: "#ff0000" });  // no effect
// same for shape.strokes
```

The whole-array setter `shape.fills = [...]` worked because it triggered the property setter; element-level mutations bypassed it. The TypeScript declaration in `plugins/libs/plugin-types/index.d.ts` exposes these as `Fill[]` / `Stroke[]`, so plugin authors expected mutation to stick.

## Root cause

`format-fills` / `format-strokes` in `frontend/src/app/plugins/format.cljs` produced a fresh JS array of fresh JS objects on every read, with no link back to shape state. Element mutations hit the snapshot, never the shape's setter path.

## Fix

Wrap the formatted array (and each element) in a `js/Proxy` whose mutation traps re-commit the parsed array through the existing setter path. The setter bodies in `shape.cljs` are extracted to top-level `commit-fills!` / `commit-strokes!` so the `:set` descriptor and the getter's `commit-fn` share one path. Whole-array assignment is unchanged.

## Tests

- Added six cases to `frontend/test/frontend_tests/plugins/context_shapes_test.cljs` covering per-element write, element replacement, and `push` / `pop` for both fills and strokes.
- Local run: **255 tests / 913 assertions / 0 failures**.
- Manually reproduced the bug and verified the fix in devenv against a rectangle.

## Checklist

- [x] Refactor any modified SCSS files following the refactor guide. _(N/A)_


niwinz:
Looks nice, thanks for contribution :heart: 
I will ask a second review of this to the plugins team.

frontend/src/app/plugins/format.cljs
```diff
@@ -26,6 +26,42 @@
...
+  (let [raw-arr (array)
+        commit! (fn [] (commit-fn raw-arr))]
+    (doseq [el coll]
+      (when-let [js-el (format-fn el)]
+        (.push raw-arr (wrap-mutable-element js-el commit!))))
+    (js/Proxy. raw-arr
+               #js {:set (fn [target prop value]
+                           (if (numeric-index? prop)
+                             (obj/set! target prop (wrap-mutable-element value commit!))
+                             (obj/set! target prop value))
```
Alotor: I think it's better to fail here, if the prop is not a number the user is trying to add a property over an array, seems fishy.

frontend/src/app/plugins/format.cljs
```diff
@@ -26,6 +26,42 @@
...
+                           (commit!)
+                           true)
+                    :deleteProperty (fn [target prop]
+                                      (js-delete target prop)
+                                      (commit!)
+                                      true)})))
+
+(defn mutable-proxy-array
+  [coll format-fn commit-fn]
+  (let [raw-arr (array)
```
Alotor: to create the `raw-arr` you can do it with one call like

```clojure
(let [raw-arr 
       (->> coll
               (keep (fn [el] (when-let [js-el (format-fn el)] 
                           (wrap-mutable-element js-el commit!))
               (format-array)
```

Alotor:
Hello, thanks for the contribution, looks really nice.

Just a couple of issues:

First, i've added some suggestions for you to check.

The other thing is that as it's it doesn't support changes to the gradients:

If the shape has already a fill in place and you try to add a gradient like so:

```typescript
penpot.selection[0].fills[0].fillColorGradient = {
    "type": "linear",
    "startX": 0.5,
    "startY": 0,
    "endX": 0.5,
    "endY": 1,
    "width": 1,
    "stops": [ { "color": "#b400ff", "opacity": 1, "offset": 0 }, { "color": "#0c3fd5", "opacity": 1, "offset": 1 } ]
};
```
This will fail because there cannot be a fillColor **and** a fillColorGradient at the same time.

Also, if there is a gradient already existing, you cannot change the properties because the mutable-proxy is not nested. It would be great if these were added as well.

frontend/src/app/plugins/format.cljs
```diff
@@ -198,16 +234,15 @@
...
-    "mixed"
-
-    (= fills "mixed")
-    "mixed"
-
-    (some? fills)
-    (format-array format-fill fills)))
+  ([fills] (format-fills fills nil))
+  ([fills commit-fn]
+   (cond
```
Alotor, line 294: I think this can be more legible like:

```clojure
(cond
     (= fills :multiple) "mixed"
     (= fills "mixed")   "mixed"

     (and (some? fills) (fn? commit-fn))
     (mutable-proxy-array fills format-fill commit-fn)

     :else
     (format-array format-fill fills))
```

frontend/src/app/plugins/format.cljs
```diff
@@ -240,9 +275,12 @@
           :strokeColorGradient (format-gradient stroke-color-gradient)})))
 
 (defn format-strokes
-  [strokes]
-  (when (some? strokes)
-    (format-array format-stroke strokes)))
+  ([strokes] (format-strokes strokes nil))
+  ([strokes commit-fn]
+   (when (some? strokes)
```
Alotor: You can remove one nesting:

```clojure
(if (and (some? strokes) (fn? commit-fn))
     (mutable-proxy-array strokes format-stroke commit-fn)
     (format-array format-stroke strokes))
```

Alotor (approved):

RenzoMXD:
Hi, @Alotor Thanks for your work and all CI tests passed.

RenzoMXD:
@Alotor @niwinz 
Could you please merge my PR if there's no update on it?

Alotor:
@RenzoMXD can you resolve the conflicts please, once this is done i'll merge the MR

Alotor (approved):

RenzoMXD:
@Alotor @niwinz Thanks.

PR #9177, opened 2026-04-27
:bug: Fix DTCG token import discriminator and group-level $type inheritance

### LINKED ISSUE #8342: Use  instead of  as DTCG token/group discriminator and make  optional to support group-level type inheritance
### Steps To Reproduce

I have been reading the DTCG and found what looks like discrepancies between the spec and Penpot validation that would cause incorrect behaviour when importing 3rd-party DTCG token files that use group-level type inheritance.

To reproduce: check https://github.com/penpot/penpot/blob/develop/common/src/app/common/types/tokens_lib.cljc (see for targeted links below)

### Expected behavior

Per the [DTCG Community Group Final Report (W3C, 2025-10-28) -- Group structure](https://www.w3.org/community/reports/design-tokens/CG-FINAL-format-20251028/#group-structure):

> The presence of a `$value` property definitively identifies an object as a token.

And from [4.2 Type](https://www.w3.org/community/reports/design-tokens/CG-FINAL-format-20251028/#type-0):

> A token's type can be specified by the optional `$type` property [...] Furthermore, the `$type` property on a group applies to all tokens nested within that group.

### Actual behavior

In [`common/src/app/common/types/tokens_lib.cljc`:
](https://github.com/penpot/penpot/blob/develop/common/src/app/common/types/tokens_lib.cljc)
1. **Parser discriminator** ([`flatten-nested-tokens-json`](https://github.com/penpot/penpot/blob/375608b44b6a9f84ddb3e7e5a0b78d77df31df32/common/src/app/common/types/tokens_lib.cljc#L1687)): uses `(not (contains? v "$type"))` to distinguish groups from tokens. 
2. **Validation schema** ([`schema:dtcg-node`](https://github.com/penpot/penpot/blob/375608b44b6a9f84ddb3e7e5a0b78d77df31df32/common/src/app/common/types/tokens_lib.cljc#L1540)): both `"$type"` and `"$value"` are required (non-optional) in the map schema.

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

## Impact
* as of this revision, there will be **no problem importing Penpot-exported files**, because Penpot always emits both `$type` and `$value` on every token and never sets `$type` on groups (AFAICT) 
* but we should **expect incorrect behavior when importing third-party DTCG token files** (those using group-level type inheritance.)

## Fix
To implement DTCG spec 

1. **Parser discriminator**  Use `(not (contains? v "$value"))`. A group carrying `$type` to set a default for child tokens would be misidentified as a token.
2.  **Validation schema** Makes`$type` optional on tokens when it can be inherited from a parent group, otherwise valid DTCG files using type inheritance will be deemed invalid

MilosM348:
## Summary

Closes #8342.

The DTCG Community Group Final Report (W3C, 2025-10-28) specifies:

> "The presence of a $value property definitively identifies an
>  object as a token."
> "A token's type can be specified by the optional $type property
>  [...] Furthermore, the $type property on a group applies to all
>  tokens nested within that group."

Two bugs in `common/src/app/common/types/tokens_lib.cljc` violate
the spec and silently break import of any third-party DTCG file
that uses group-level type inheritance:

1. `flatten-nested-tokens-json` used `(not (contains? v "$type"))`
   as the group-vs-token discriminator. A group node carrying only
   a `$type` (to set a default for child tokens) was misidentified
   as a token, then immediately discarded because it had no
   `$value`.

2. `schema:dtcg-node` declared both `$type` and `$value` as
   required, so even after the discriminator was fixed any leaf
   token that relied on group-level type inheritance failed
   `dtcg-node?` validation and never reached the parser.

The combined effect: importing a spec-compliant DTCG file that
expressed types at the group level produced a TokensLib with no
tokens at all, because every leaf was discarded as "unknown type".

Penpot-exported files are unaffected because Penpot always emits
both `$type` and `$value` on every token and never attaches `$type`
to a group, so the existing tests covered only the inline-type
shape.

## Fix

- `schema:dtcg-node`: mark `$type` optional.
- `flatten-nested-tokens-json`: use `$value` as the discriminator
  (anything without `$value` is a group), accept an optional
  `inherited-type` accumulator that carries the nearest enclosing
  group `$type` down through recursion, and resolve a token's type
  from its own `$type` first, falling back to the inherited type.
  A token's own `$type` always wins over the inherited one (per
  spec).

## Tests

Added `parse-dtcg-group-type-inheritance` to
`common/test/common_tests/types/tokens_lib_test.cljc`, covering
both cases:
- Group `$type` is inherited by tokens that don't declare their own
  (`colors.red`, `colors.blue`, `space.small`).
- Token-level `$type` overrides the inherited group `$type`
  (`colors.danger`, `space.large`).

Existing DTCG round-trip tests continue to pass because they all
declare `$type` at the token level, which the new code still
honours.

CHANGES.md entry added under the 2.17.0 Bugs-fixed section.

## Closes

Closes #8342.


niwinz:
Hey @MilosM348 
The code looks nice, and the PR looks like adressing a real issue. I asked it internally to tokens team for a feedback on this because i'm personally don't have enough context for proper review it

hirunatan:
Ok. I have tested it and it's all right.

I've fixed a small typo in the test, and removed the CHANGES.md update because our new policy. It has been merged in https://github.com/penpot/penpot/pull/9912 (your commit authorship is maintained). I'm closing this one now.

Thank very much for your contribution!!!

PR #9182, opened 2026-04-27
:sparkles: Preserve vector content when pasting from external tools

### LINKED ISSUE #546: copy/paste SVG from Inkscape
I can copy from Inkscape and paste into Penpot, but the graphic shows up as raster image instead of vector.

Is there a way to paste as vectors? (Perhaps this is on Inkscape's side, although it can copy/paste as vectors to itself.)

I'm on Fedora 33 (Linux) and using Firefox.

RenzoMXD:
Closes #546

## Summary

Tools like Inkscape place SVG on the clipboard alongside a raster preview. Penpot was picking the raster, so pasted graphics arrived as non-editable images. With certain tools (Inkscape included) the same paste also carried an empty `image/svg+xml` blob alongside the real SVG payload (which was sent as `text/plain`), which produced an extra "SVG is invalid or malformed" warning.

After this change pasting from Inkscape produces an editable vector shape with no warning.

## Root cause

`frontend/src/app/util/clipboard.js` had three gaps for the multi-MIME, multi-payload paste shape that vector tools tend to produce:

1. `parseText`'s SVG-detection regex only matched text starting with `<svg`, so SVG sent as `text/plain` with an XML declaration or   HTML comment in front (Inkscape's default) was tagged as plain text and dropped into a text shape.
2. The raster-vs-SVG preference in the reducer only consulted the    already-pushed accumulator, so the raster was preserved  hen the SVG came later in the priority sort.
3. Some tools advertise `image/svg+xml` and leave its payload empty, pointing the real SVG at `text/plain` instead. The empty blob was handed to `tubax/xml->clj`, which threw with an empty error message — surfacing as the "SVG is invalid or alformed" notification.

## Fix

* Move `image/svg+xml` to the front of `allowedTypes` so the priority sort ranks vector first.
* Add `svgTextPattern` constant matching SVG with optional XML declaration and HTML comments before the `<svg>` root, and use it inside `parseText` to reclassify text-as-SVG correctly.
* In the `fromDataTransfer` reducer, switch the raster-skip check from `filtered.find` to `all.find` so any SVG sibling — including
  ones that come later in the iteration — drops the raster.
* Filter out empty blobs (`item.size > 0`) before the reducer so the empty SVG blob never reaches `tubax/xml->clj`.

## Steps to Reproduce

1. In Inkscape, draw any shape and Edit → Copy.
2. In Penpot, click into a workspace file and Ctrl+V.

**Before:** the paste arrives as a static raster image (sometimes also as a text shape containing the SVG XML), with an "SVG is invalid or malformed" warning.

**After:** the paste arrives as an editable vector shape inside the standard SVG-import grouping, with no warning.

## Checklist

- [x] Refactor any modified SCSS files following the refactor guide. _(N/A — no SCSS)_
- [x] Update the `CHANGES.md` file, referencing the related GitHub issue.


niwinz:
Looks nice, thanks :heart:  

Ready to merge, lets wait the CI

RenzoMXD:
@niwinz Thanks again. 😊

Issue #4478, opened 2024-04-20
bug:  Cannot read properties of undefined (reading 'writeText')


greenbrainsecke:
### Steps To Reproduce

Select Board, Go to Specification, then Code and copy Code to clipboard

### Expected behavior

That the code is copied to the clipboard

### Actual behavior

Hello, I have the following problem: I'm creating something in Penpot 2 and would like to copy the CSS or HTML code to the clipboard and I get an error screen and the following error message:

### Screenshots or video

![Screenshot 2024-04-21 013809](https://github.com/penpot/penpot/assets/48941397/53f44688-5513-4917-b724-6752e1316efa)

### Environment (please complete the following information)

Hint:    Cannot read properties of undefined (reading 'writeText')
Prof ID: 681277dd-509d-810d-8004-2d6683e807b4
Team ID: 681277dd-509d-810d-8004-2d668445e021

Data:
{:hint "Cannot read properties of undefined (reading 'writeText')"}

Trace:
TypeError: Cannot read properties of undefined (reading 'writeText')
    at PENPOT.hO (http://192.168.1.52:9001/js/main.33bdf5e83.js:849:235)
    at http://192.168.1.52:9001/js/main.33bdf5e83.js:2453:426
    at Object.ob (http://192.168.1.52:9001/js/shared.33bdf5e83.js:1037:218)
    at Bb (http://192.168.1.52:9001/js/shared.33bdf5e83.js:1037:311)
    at Qb (http://192.168.1.52:9001/js/shared.33bdf5e83.js:1037:370)
    at ud (http://192.168.1.52:9001/js/shared.33bdf5e83.js:1064:469)
    at Xd (http://192.168.1.52:9001/js/shared.33bdf5e83.js:1065:420)
    at http://192.168.1.52:9001/js/shared.33bdf5e83.js:1076:57
    at Hi (http://192.168.1.52:9001/js/shared.33bdf5e83.js:1198:44)
    at Sa (http://192.168.1.52:9001/js/shared.33bdf5e83.js:1036:21)
    at Kc (http://192.168.1.52:9001/js/shared.33bdf5e83.js:1068:190)
    at Tc (http://192.168.1.52:9001/js/shared.33bdf5e83.js:1050:253)
    at Dc (http://192.168.1.52:9001/js/shared.33bdf5e83.js:1050:29)

Last events:
[":app.main.data.workspace.highlight/highlight-shape"
 ":app.main.data.workspace.collapse/expand-all-parents"
[...]

Issue #8423, opened 2026-02-20
"Show Guides" shortcut does not work on German keyboards (Ctrl+Shift+#)


lplath:
### Steps To Reproduce

After the recent updates I am no longer able to use the keyboard-shortcut  `CTRL + '` to hide and show guides. On german keyboards the `'` character is assigned to `SHIFT + #` so I am pressing `CTRL + SHIFT + #`.  Bellow you can see corresponding key-events.

<img width="1254" height="443" alt="Image" src="https://github.com/user-attachments/assets/7a15b338-82a7-4d43-9838-5fc62cf19db3" />

I have tested this shortcut with an english layout and it works fine.

### Desktop

Windows 11 & Microsoft Edge 145

madalenapmelo-kp:
Hi @lplath,

Thanks for reporting this! We've added this to our backlog on Taiga so that we can look further into it, you can find the details here: https://tree.taiga.io/project/penpot/issue/13469

daniel-herrero:
I've verified the error in PRE/PRO. For the German layout, the combination Ctrl+Shift+' is not hiding Penpot's guides.
The same keyboard combination works in HOURLY, hiding and showing the guides properly, as it's shown in the video.

https://github.com/user-attachments/assets/6fb52255-7a3c-448a-a5db-c54d83f4f3d1

PR #9188, opened 2026-04-27
:bug: Reject clipboard helpers gracefully on insecure origins

### LINKED ISSUE #6514: Use Clipboard API consistently across the application
Penpot mostly uses the clipboard with text/plain mime format via the `.writeText` interface. However, the new Clipboard API supports arbitrary MIME types via `.write`. This is already used in some places because it responds with a Promise.

However, one should probably be more consistent with MIME type support for better interop with other tools, perhaps copying to multiple mime formats at the same time (e.g. plain text and `image/svg+xml`)

Also, proper fallback / error handling when the API fails (e.g over insecure connections)

Related issues:

* #6510 Initial discussion
* #4478 Error with clipboard API in nonsecure context (HTTP) . See [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard)

MilosM348:
## Summary

Closes #6514. Resolves the user-visible crash originally reported in
#4478.

`app.util.clipboard/to-clipboard` and `to-clipboard-promise` called
`(unchecked-get js/navigator "clipboard")` and then immediately
invoked `.writeText` / `.write` on the result, with no guard for
the case where `navigator.clipboard` is `undefined`. The W3C
Clipboard API spec requires a "secure context" (HTTPS or
`localhost`), so a Penpot instance served over plain HTTP — which
the SSDP/LAN self-hosted setup in #4478 was — throws

  `TypeError: Cannot read properties of undefined (reading 'writeText')`

synchronously the moment the user clicks any copy button. The error
escapes the consuming function before any error-handling `rx/of`
arm runs, so the whole UI ends up on the error screen instead of
just the affected control showing a "could not copy" message.

A third helper (`to-clipboard-multi`) already guards `clipboard`
and `clipboard.write`, but if both are missing it silently returned
`nil` which is also surprising for callers expecting a Promise.

## Fix

Add a small `get-clipboard` accessor and an `unavailable-error`
factory that returns `Promise.reject(Error(...))` with a clear
message:

  *"Clipboard API is unavailable. This usually happens when the
  page is served over plain HTTP; serve Penpot over HTTPS to
  enable copy-to-clipboard."*

Wire all three helpers through the same defensive contract:

- `to-clipboard` — return the rejected Promise when
  `navigator.clipboard.writeText` is missing.
- `to-clipboard-promise` — return the rejected Promise when
  `navigator.clipboard.write` is missing.
- `to-clipboard-multi` — convert the existing `if` into a `cond`
  with three branches: prefer `clipboard.write` for true
  multi-MIME output, fall through to `writeText` with the
  `text/plain` payload when only the legacy text path is
  available, and finally reject with the unavailable error when
  neither path exists. Previously the no-API case fell off the
  `when-let` and silently returned `nil`.

The contract is now consistent: every helper either resolves or
rejects a Promise, never throws synchronously, and never returns
`nil`. Callers (which are already structured around `rx` streams
that call `rx/from` on the helper's return value) can chain
`.catch` / `rx/catch` to surface a status-bar message instead of
crashing.

The two stale `;; FIXME` comments on `to-clipboard` (rename to
`write-text`) and `to-clipboard-promise` (this API is very
confuse) are removed — the rename remains an open follow-up across
13+ call sites and is intentionally out of scope, but the API is
no longer "confuse" once the contract is documented and uniform.

## How to test

The cheapest way without redeploying:

1. Open Penpot in a browser tab (HTTPS / localhost is fine for setup).
2. In DevTools console, simulate an insecure origin for the rest of the tab:
   ```js
   Object.defineProperty(navigator, 'clipboard', { value: undefined });
3. Copy some text
4. Open workspace context menu and click paste

It should show a human readable message instead of generic error notification

## Closes

Closes #6514. Addresses #4478 (already closed; this prevents the
underlying crash from recurring on insecure origins).


niwinz:
Hey @MilosM348 
First of all, thanks for contribution.

Can you give me steps for reproduce this?
Right now when i paste code that uses clipboard api (Ctrl+V of some text copied externally) and have no permissions.  That is happens:
image
And has no console error. Is there any other path for reproducing the error that this PR tries to solve?
Thanks

MilosM348:
> Hey @MilosM348 First of all, thanks for contribution.
> 
> Can you give me steps for reproduce this? Right now when i paste code that uses clipboard api (Ctrl+V of some text copied externally) and have no permissions. That is happens: image And has no console error. Is there any other path for reproducing the error that this PR tries to solve? Thanks

Hi @niwinz — thanks for taking a look!

I think the screenshot you tried catches the *paste* path (`from-data-transfer`),
which already had its own permission-denied fallback (the orange toast).
This PR is about the *copy/write* path — `to-clipboard` /
`to-clipboard-promise` / `to-clipboard-multi` — which previously called
`.writeText` on `navigator.clipboard` without a nil-check.

The crash matches issue #4478 (linked from #6514): on insecure origins
`navigator.clipboard` is `undefined` per the W3C Secure Contexts spec, so
calling `.writeText` synchronously throws
`TypeError: Cannot read properties of undefined (reading 'writeText')`,
the copy silently fails, and the user gets no feedback.

### Repro

The cheapest way without redeploying:

1. Open Penpot in a browser tab (HTTPS / localhost is fine for setup).
2. In DevTools console, simulate an insecure origin for the rest of the tab:
   ```js
   Object.defineProperty(navigator, 'clipboard', { value: undefined });

niwinz:
Will look on it tomorrow, thanks!

niwinz:
I have tried for reproduce this with your suggestion and i get this:

image

image

MilosM348:
> I have tried for reproduce this with your suggestion and i get this:
> 
> image image

Good catch — that's the symmetric path I missed in the original
scope. Pushed 62b2ecf6f to add the same nil-check on
fromNavigator (clipboard.js): if navigator.clipboard?.read isn't a
function it now throws the same descriptive Error the copy direction
returns, so the rx subscription surfaces the unavailable-message
toast instead of TypeError → "Something wrong has happened".

Same defineProperty repro should now reject cleanly in both
directions. Also resolved the CHANGES.md conflict the develop merge

niwinz:
I think is almost ok for now. It prints on console the correct error but frontend still show generic error.
For properly fix we should add something like this:

```diff
diff --git a/frontend/src/app/main/data/workspace/clipboard.cljs b/frontend/src/app/main/data/workspace/clipboard.cljs
index e5d6dbd19b..427e8839a0 100644
--- a/frontend/src/app/main/data/workspace/clipboard.cljs
+++ b/frontend/src/app/main/data/workspace/clipboard.cljs
@@ -303,13 +303,30 @@
   (and (instance? js/DOMException cause)
        (= (.-name cause) "NotAllowedError")))
 
+(defn- clipboard-unavailable-error?
+  "Check if the given error is a clipboard API unavailable error
+  (thrown when navigator.clipboard is undefined, e.g. on insecure
+   origins per the W3C Secure Contexts spec)."
+  [cause]
+  (and (instance? js/Error cause)
+       (str/starts-with? (.-message cause) "Clipboard API is unavailable.")))
+
 (defn- on-clipboard-permission-error
   [cause]
-  (if (clipboard-permission-error? cause)
+  (cond
+    (clipboard-permission-error? cause)
     (rx/of (ntf/show {:content (tr "errors.clipboard-permission-denied")
                       :type :toast
                       :level :warning
                       :timeout 5000}))
+
+    (clipboard-unavailable-error? cause)
+    (rx/of (ntf/show {:content (tr "errors.clipboard-api-unavailable")
+                      :type :toast
+                      :level :warning
+                      :timeout 5000}))
+
+    :else
     (rx/throw cause)))
 
 (defn paste-from-clipboard
@@ -511,6 +528,12 @@
                                       :level :warning
                                       :timeout 5000}))
 
+                    (clipboard-unavailable-error? cause)
+                    (rx/of (ntf/show {:content (tr "errors.clipboard-api-unavailable")
+                                      :type :toast
+                                      :level :warning
+                                      :timeout 5000}))
+
                     (:not-implemented (ex-data cause))
                     (rx/of (ntf/warn (tr "errors.clipboard-not-implemented")))
 
diff --git a/frontend/translations/en.po b/frontend/translations/en.po
index 88b34b9c0b..460bfb193a 100644
--- a/frontend/translations/en.po
+++ b/frontend/translations/en.po
@@ -1413,6 +1413,10 @@ msgstr "Your browser cannot do this operation"
 msgid "errors.clipboard-permission-denied"
 msgstr "Clipboard access denied. Please allow clipboard permissions in your browser to paste content"
 
+#: src/app/main/data/workspace/clipboard.cljs
+msgid "errors.clipboard-api-unavailable"
+msgstr "Clipboard API is unavailable. Serve Penpot over HTTPS to enable clipboard access"
+
 #: src/app/main/errors.cljs:235
 msgid "errors.comment-error"
 msgstr "There was an error with the comment"
```

I can do it myself on merging this using CLI or may you want do it yourself.

MilosM348:
Thanks @niwinz, that's a really good catch — the rejected Promise was indeed only landing in the console because the workspace's `on-error` handlers all branched on `clipboard-permission-error?` (DOMException / `NotAllowedError`) and the unavailable case slipped through to the generic toast.

Applied your diff verbatim in c6b7f81 (alongside an `errors.clipboard-api-unavailable` entry in `frontend/translations/en.po` so the new `tr` lookup resolves):

- New `clipboard-unavailable-error?` predicate that matches the `Promise.reject(Error "Clipboard API is unavailable. ..."))` thrown by the `get-clipboard` / `unavailable-error` helpers added earlier in this PR.
- Used `str/starts-with?` against the message prefix (the file already aliases `cuerdas.core :as str`, so no extra `:require` was needed) to keep the predicate stable if the trailing "serve Penpot over HTTPS to enable clipboard access" advice text is ever reworded.
- Converted `on-clipboard-permission-error` from `if` to `cond` and added the third arm that fires `errors.clipboard-api-unavailable` as a warning toast.
- Mirrored the same arm in the second `cond` block inside `paste-from-clipboard`, before the existing `:not-implemented` and `:else` branches.

While I was rebasing I also noticed `develop` had cut a new `2.17.0 (Unreleased)` section and moved the Ctrl+' (#8423) and `team.features` race (#9197) bullets there, so the `CHANGES.md` merge conflict in `2.16.0` resolved by dropping those two duplicates and keeping only this PR's clipboard bullet in `2.16.0`. The merge is now clean and the `Files changed` view should be back to the minimal +33/-1 across `clipboard.cljs` + `en.po` + `CHANGES.md`. Ready for another look whenever you have a moment, thanks again for the careful review!

niwinz:
Just a suggestion before merge, Can you please make the changelog entry a bit shorter. In the same line as others. It should not contain all the description, only a one liner summary.

Thanks :heart:

MilosM348:
Good call, sorry — that was way too verbose for a `CHANGES.md` line. Trimmed it to a one-liner in `37e94bf33` to match the surrounding style:

```
- Fix copy and paste actions crashing the workspace on insecure origins (plain HTTP / non-`localhost`) where the Clipboard API is unavailable (by @MilosM348) [Github #6514](https://github.com/penpot/penpot/issues/6514)
```

While rebasing I also moved the bullet from the `2.16.0 (Unreleased)` section into `2.17.0 (Unreleased)` to match the milestone you set on the PR (and to be consistent with where you'd already relocated the Ctrl+' / `team.features` bullets from the prior `develop` cut). The `2.16.0` `### :bug: Bugs fixed` block is now back to its pre-PR state. Let me know if you'd rather have it back in `2.16.0` and I'll move it; happy either way.

niwinz:
For me it is ready to merge, good job. Do not worry about CI, because we experiencing issues with our runners

PR #9198, opened 2026-04-28
:bug: Fix lost-update race on team.features during concurrent file cr…

### LINKED ISSUE #9197: Team features have a lost-update race during concurrent file creation
### Steps To Reproduce

Preconditions:
- A Penpot backend with a no-migration / non-default feature flag enabled   globally (e.g. `:feature-fdata-pointer-map` or `:feature-fdata-objects-map`),   OR a freshly created team that has not yet had any file created.
- Two authenticated clients with edit permission on the same project of the same team.

Reproduction:
1. Pick a team T whose `team.features` row in Postgres does NOT yet contain the target feature F (e.g. `fdata/pointer-map`). For a fresh team this is trivially true; for an existing team you can verify with:
     SELECT features FROM team WHERE id = '<team-id>';
2. From two separate clients (or two parallel API calls) simultaneously invoke the `create-file` RPC against the same project in team T, both declaring support for feature F in the request `features` set.
3. Wait for both requests to commit.
4. Re-read `team.features`.

Expected to observe: both requests append F (and any other newly-introduced no-migration, team-inheritable feature) to `team.features`.

To make the race deterministic in a dev environment, add a small delay between `teams/get-team` (files_create.clj:90) and `db/update! :team` (files_create.clj:127) and fire the two requests within that window.

### Expected behavior

After both concurrent `create-file` calls commit, `team.features` should contain the union of every newly-introduced no-migration feature from both requests. No feature added by a successful create-file should silently
disappear from the team row.

### Actual behavior

Preconditions:
- A Penpot backend with a no-migration / non-default feature flag enabled   globally (e.g. `:feature-fdata-pointer-map` or `:feature-fdata-objects-map`),   OR a freshly created team that has not yet had any file created.
- Two authenticated clients with edit permission on the same project of the same team.

Reproduction:
1. Pick a team T whose `team.features` row in Postgres does NOT yet contain the target feature F (e.g. `fdata/pointer-map`). For a fresh team this is trivially true; for an existing team you can verify with:
     SELECT features FROM team WHERE id = '<team-id>';
2. From two separate clients (or two parallel API calls) simultaneously invoke the `create-file` RPC against the same project in team T, both declaring support for feature F in the request `features` set.
3. Wait for both requests to commit.
4. Re-read `team.features`.

Expected to observe: both requests append F (and any other newly-introduced no-migration, team-inheritable feature) to `team.features`.

To make the race deterministic in a dev environment, add a small delay between `teams/get-team` (files_create.clj:90) and `db/update! :team` (files_create.clj:127) and fire the two requests within that window.

### Screenshots or video

After both concurrent `create-file` calls commit, `team.features` should contain the union of every newly-introduced no-migration feature from both requests. No feature added by a successful create-file should silently disappear from the team row.

`team.features` is updated via a non-atomic read-modify-write that holds no row lock:

  backend/src/app/rpc/commands/files_create.clj:90-130

    let team     = teams/get-team(conn, ...)              ; plain SELECT
        features = ... compute candidate set ...
    when (candidate non-empty)
      db/update! conn :team {:features (team.features ∪ candidate − ...)}

The handler is wrapped in `::db/transaction true`, but the team row is read without `FOR UPDATE` and Postgres runs at READ COMMITTED. Two concurrent transactions on the same team observe the same `team.features = T`, each computes `T ∪ {f₁}` and `T ∪ {f₂}` independently, and both UPDATE. The second commit overwrites the first, so one of the newly-introduced features is lost from `team.features`.

Maintainers have already flagged this with an inline FIXME at line 115:
  ;; FIXME: IMPORTANT: this code can have race conditions, because
  ;; we have no locks for updating team so, creating two files
  ;; concurrently can lead to lost team features updating

Impact in practice:
- The created file row itself is correct (file.features is written independently in `create-file` → `bfc/insert-file!`), so no file data is lost or corrupted.
- `team.features` lags by one flag until the next non-racing create-file on the same team re-introduces it; downstream reads via `cfeat/get-team-enabled-features` re-union the global no-migration set, so most read paths mask the inconsistency.
- The exception is paths that compare `team.features` directly without going through `get-team-enabled-features`, notably `cfeat/check-teams-compatibility!` and `cfeat/check-paste-features!`, which can raise a spurious `:team-feature-mismatch` while the row is transiently lagging (e.g. when moving or pasting a file between teams in that window).

Realistic trigger windows:
- The first file ever created on a brand-new team (default features being  olded into team.features for the first time).
- An admin enabling a backend no-migration feature flag on a running deployment, followed by a burst of concurrent create-file calls on the same team before any of them lands the new flag.

N/A — backend race, not user-visible UI.

- Host: reproducible on any deployment, including https://design.penpot.app and self-hosted.
- Affected component: backend RPC `::create-file`
- Affected file: backend/src/app/rpc/commands/files_create.clj (lines 90-130)
- Affected branch: develop (verified at commit aa5bfe6dd)
- Database: PostgreSQL, default isolation level (READ COMMITTED)

### Frontend Stack Trace

```console
N/A
```

### Backend Stack Trace

```console
N/A — bug is a silent lost update, not an exception. No stack trace is produced.
```

### Additional context

Suggested fix (smallest local change):

The handler already runs inside `::db/transaction true`, so it is sufficient to read the team row with a row-level lock in the same transaction: 

    SELECT ... FROM team WHERE id = ? FOR UPDATE

i.e. extend `teams/get-team` (or use a dedicated helper) to accept a `:for-update` option and use it from `files_create.clj` before the read-modify-write block. This serializes concurrent create-file calls on the
same team only when a feature might actually need to be appended, so the performance cost is negligible.

Alternative: replace the read-modify-write with an atomic SQL merge, e.g.

    UPDATE team
       SET features = (SELECT array_agg(DISTINCT x)
                         FROM unnest(features || ?::text[]) x)
     WHERE id = ?

with the no-team-inheritable subtraction precomputed on the application
side. This avoids the explicit lock entirely.

Severity assessment: medium-low. The bug is real and worth fixing, but it is self-healing on the next non-racing write to the same team, does not corrupt file data, does not bypass any security check, and the trigger
window is narrow (first-ever create on a team, or shortly after an admin lips a backend feature flag). Not a release blocker.

Once the FIXME at files_create.clj:115 is addressed, please remove the comment.

web-dev0521:
### Related Ticket

Closes [Github #9197](https://github.com/penpot/penpot/issues/9197) 

### Summary

Fix a lost-update race in `team.features` during concurrent `create-file` calls on the same team.

`backend/src/app/rpc/commands/files_create.clj` previously read `team.features` (via `teams/get-team`) without a row lock, computed a new feature set in memory, and wrote it back. Two concurrent `create-file` calls on the same team — under PostgreSQL's default READ COMMITTED isolation — could both observe the same `team.features` value, each compute a different union, and the second `UPDATE` would silently overwrite the first, dropping a no-migration feature flag that one of the new files actually relies on. The maintainers had already flagged this with an inline `FIXME` at `backend/src/app/rpc/commands/files_create.clj:115`.

The handler is already wrapped in `::db/transaction true`, so the fix is to acquire a row-level lock on the team and re-read its `features` column inside the same transaction before the read-modify-write block:

```clojure
(let [team-features (-> (db/exec-one! conn
                                      ["SELECT features FROM team WHERE id = ? FOR UPDATE"
                                       team-id])
                        :features
                        (db/decode-pgarray #{}))]
  (when-let [new-features (-> features
                              (set/difference team-features)
                              (set/difference cfeat/no-team-inheritable-features)
                              (not-empty))]
    (let [features (-> new-features
                       (set/union team-features)
                       (set/difference cfeat/no-team-inheritable-features)
                       (into-array))]
      (db/update! conn :team
                  {:features features}
                  {:id team-id}
                  {::db/return-keys false}))))
```

Set algebra, write payload, `into-array` binding and `db/update!` call are byte-for-byte identical to the previous implementation — the only behavioural change is that `team-features` now comes from a locked re-read instead of the unlocked snapshot returned by `teams/get-team`. The inline `FIXME` is removed.

The pattern matches existing locking idioms in the codebase (`::db/for-update` opt and explicit `FOR UPDATE` in `backend/src/app/tasks/objects_gc.clj`, `backend/src/app/rpc/commands/profile.clj`, `backend/src/app/srepl/procs/fdata_storage.clj`, `backend/src/app/loggers/audit/archive_task.clj`). `db/exec-one!` and `db/decode-pgarray` are already public helpers in `app.db`, which is already aliased as `db` in this namespace, so no new imports are required. No schema changes, no API changes, no new dependencies, no behavioural change for non-racing callers.

### Steps to reproduce

**Preconditions**

- A Penpot backend with a non-default no-migration feature flag enabled globally (for example `:feature-fdata-pointer-map` or `:feature-fdata-objects-map`), or a team that has not yet had any file created.
- Two authenticated clients with edit permission on the same project of the same team.

**Reproduction (before the fix)**

1. Pick a team `T` whose `team.features` row in Postgres does **not** yet contain the target feature `F`:
   ```sql
   SELECT features FROM team WHERE id = '<team-id>';
   ```
2. From two separate clients (or two parallel API calls) simultaneously invoke the `create-file` RPC against the same project in team `T`, both declaring support for `F` in the request `features` set.
3. Wait for both requests to commit.
4. Re-read `team.features` — it contains only one of the two newly-introduced features; the other has been silently overwritten.

To make the race deterministic in a development environment, insert a small delay between `teams/get-team` and `db/update! :team` (around line 115 of `files_create.clj` before the patch) and fire the two requests within that window.

**Verification (after the fix)**

- Run the same reproduction. Both `create-file` calls now serialize on the team row; `team.features` ends up containing the union of every newly-introduced no-migration, team-inheritable feature from both requests.
- Existing test suite (`clojure -M:dev:test`) passes — semantics of the read-modify-write block are unchanged for non-racing cases.
- The `FIXME` comment at `backend/src/app/rpc/commands/files_create.clj:115` is gone.
- No deadlock regression: lock order in this handler is `team → project` (the implicit row lock taken by `db/update! :project` later in `create-file`), which matches the conceptual parent-child order used elsewhere in the codebase.

### Checklist


web-dev0521:
Hi, @niwinz, 
This is my first contribution to this repository.
Could you please review my first PR?
Thank you.

web-dev0521:
Hi @niwinz ,
Thank you for your attention.
This is my first contribution to penpot repository.
Could you please review my PR?
Thank you again.

niwinz:
Hey @web-dev0521 

Awesome, the code looks good to me!

Just to manage expectations since this is your first contribution: I dedicate specific blocks of time to reviewing PRs and I usually handle them in the order they arrive. Please keep in mind that reviews might take a few days, especially when I'm balancing other urgent tasks. Thanks for your patience and glad to have you contributing!

niwinz:
I think the only missing is the CHANGES.md entry.

web-dev0521:
Thank you @niwinz .

web-dev0521:
> I think the only missing is the CHANGES.md entry.
Hi @niwinz, thank you for the feedback. Just to clarify, would you prefer that I add the CHANGES.md entry in this PR, or include it in a follow-up PR?

web-dev0521:
Hello, @niwinz ,
Happy Friday.
Could you please review my PR?

wdeveloper16:
You need to add CHANGES.md entry for each PR, not follow-up.
cc: @niwinz

Lobster-0429:
Hi, @niwinz ,
Could you please review my PR when you are convenient?
Thanks.

Lobster-0429:
Hello, @niwinz ,
Could you please review my PR when you are convenient?
I really appreciate your feedback.
Thanks.

Lobster-0429:
Hi, @niwinz ,
Could you please review my PR?

niwinz:
Sorry for delay. Looks OK for me, lets wait CI
Thanks for the patience :heart:

Lobster-0429:
Thank you for your attention @niwinz.

Lobster-0429:
Hi, @niwinz ,
All tests are successfully passed.
Could you please review my PR again?
Thanks for your time.

niwinz:
Thanks :heart:

PR #9199, opened 2026-04-28
:bug: Lock team row to fix lost-update race in team.features on concurrent create-file

### LINKED ISSUE #9197: Team features have a lost-update race during concurrent file creation

MilosM348:
## Summary

Closes #9197.

`app.rpc.commands.files-create/::create-file` reads `team.features`
via `teams/get-team` at the start of the transaction (line ~90 of
files_create.clj), then unions the requested no-migration / team-
inheritable features into it and writes the result back via
`db/update! conn :team` (line ~127). The read and the write are not
serialised by any lock, and a `;; FIXME: IMPORTANT: this code can
have race conditions` comment in the body acknowledges the gap.

The race surfaces whenever two clients call `create-file` against
the same team within a single read-modify-write window:

- Request A and Request B both run `teams/get-team`. Both observe
  `team.features = X`.
- Request A computes `X ∪ {a}` and queues an UPDATE.
- Request B computes `X ∪ {b}` and queues an UPDATE.
- Whichever transaction commits second writes its own union and
  overwrites the other's. The losing feature is silently absent
  from `team.features` afterward, even though its `create-file`
  call returned success.

The reproduction in the issue body uses a fresh team plus two
concurrent `create-file` calls each declaring `:fdata/pointer-map`
in `params.features`; only one of the two writes survives.

## Fix

Acquire a row-level `SELECT ... FOR UPDATE` on the team row inside
the existing `::db/transaction true` envelope before computing the
final union, and re-derive the new-feature diff against the locked
snapshot rather than the cached one. The lock follows the standard
`::sql/for-update true` keyword used elsewhere in the codebase
(`projects.clj:252` rename-project, `comments.clj:538`+ for
comment threads, `binfile/v3.clj:899` for ref-file lookups), so
the pattern is familiar to reviewers.

Behaviour:

- The cheap unlocked check against the cached `team` is preserved
  on the outer `when-let`, so the common case of "team already has
  all my features" still avoids the FOR UPDATE round-trip.
- When the cheap check suggests an update is needed, we lock the
  team row, re-read its features, and re-derive the diff. If the
  intervening commit of a concurrent request already added our
  features, the inner `when-let` collapses to nil and we skip the
  redundant write.
- Union semantics unchanged (still filters
  `cfeat/no-team-inheritable-features` before `into-array`).

The `FIXME` comment block is removed and replaced with a concise
explanation of why the lock is held.

## Notes

Adds `[app.db.sql :as-alias sql]` to the require block (`:as-alias`
= no actual load, zero classpath impact).

No regression test added: a reliable concurrency test for this
race would need either a deliberate sleep injected into the
production code path or a custom test helper that holds two
transactions open while suspending one mid-flight; both approaches
add more complexity than the fix itself and there is no existing
concurrency-stress harness in the backend test suite to extend.
The issue body's reproduction recipe (insert a small delay between
`teams/get-team` and `db/update! :team` and fire two requests
within the window) reproduces deterministically on the unlocked
code and now serialises correctly with the lock.

## Closes

Closes #9197.


niwinz:
Hey, this issue is already taken https://github.com/penpot/penpot/pull/9198. 
Thanks for the contribution in any case. I try to review this in order and the other comes first.

MilosM348:
> Hey, this issue is already taken #9198. Thanks for the contribution in any case. I try to review this in order and the other comes first.

OK, Thanks, I will try another PR

PR #9200, opened 2026-04-28
:wrench: Change build and cargo settings

AzazelN28:
### Related Ticket

[Taiga Task #14082](https://tree.taiga.io/project/penpot/task/14082)

### Summary

Change build settings to improve performance globally.

- Changes compile-time optimization settings to more aggressive optimization (O3 and opt-level 3).
- Changes LTO (Link-Time Optimizations) to true when building in release mode. This increases build times but also allows the compiler to perform optimizations in link-time: better binary size, better function inlining, etc.
- Changes ALLOW_TABLE_GROWTH to false (as we're not using dynamic linking).
- Removes Emscripten ASSERTIONS. 

### Checklist

### DIFF
--- a/render-wasm/src/shapes/paths.rs
+++ b/render-wasm/src/shapes/paths.rs
@@ -1,7 +1,6 @@
-use skia_safe::{self as skia, Matrix};
-
 use crate::math;
 use crate::shapes::svg_attrs::{FillRule, SvgAttrs};
+use skia_safe::{self as skia, Matrix};
 
 mod subpaths;
 
--- a/frontend/src/debug.cljs
+++ b/frontend/src/debug.cljs
@@ -135,6 +135,14 @@
       (wasm.mem/free)
       text)))
 
+(defn ^:export wasmRenderStats
+  []
+  (let [module wasm/internal-module
+        f      (when module (unchecked-get module "_render_stats"))]
+    (if (fn? f)
+      (wasm.h/call module "_render_stats")
+      (js/console.warn "[debug] render-wasm module not ready or missing _render_stats"))))
+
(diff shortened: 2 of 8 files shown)


render-wasm/src/render/options.rs
```diff
@@ -7,7 +7,7 @@ const SHOW_WASM_INFO: u32 = 0x08;
 // Render performance options
 // This is the extra area used for tile rendering (tiles beyond viewport).
 // Higher values pre-render more tiles, reducing empty squares during pan but using more memory.
-const VIEWPORT_INTEREST_AREA_THRESHOLD: i32 = 3;
+const VIEWPORT_INTEREST_AREA_THRESHOLD: i32 = 1;
```
elenatorro: can we leave that to 3, so we keep this commit & PR only for build-related changes?

render-wasm/src/shapes/paths.rs
```diff
@@ -20,7 +19,7 @@ impl Segment {}
 #[derive(Debug, Clone, PartialEq)]
 pub struct Path {
     segments: Vec<Segment>,
-    skia_path: skia::Path,
+    pub skia_path: skia::Path,
```
elenatorro: we do not need the pub here, right?

elenatorro (changes requested):

render-wasm/src/render/options.rs
AzazelN28: sure

render-wasm/src/shapes/paths.rs
AzazelN28: yup

elenatorro (approved):

PR #9151, opened 2026-04-24
:sparkles: Add "Delete group" to assets panel context menu

FairyPigDev:
### Related Ticket

Closes [#9141 — Enable deleting entire asset groups in assets panel](https://github.com/penpot/penpot/issues/9141).

### Summary

When working with large asset groups, multi-select across hundreds of items is impractical and ungrouping first leaves the orphaned items scattered in the flat list. This PR adds a **Delete group** option to the assets-panel group context menu for the three asset types that already carry group structure: **components** (including variants), **colors**, and **typographies**.

**Behavior.** Right-click a group header → "Delete group" → confirmation modal showing how many assets will be removed → accept → every asset under that group path is deleted inside a single undo transaction. One Cmd+Z restores the whole group.

**Variants.** For component variants, sibling variants sharing a variant container are deduplicated so the container is deleted once (matching the per-item delete dispatch in [`file_library.cljs`](frontend/src/app/main/ui/workspace/sidebar/assets/file_library.cljs)).

**Reused vs new.**
- **Reused**: Penpot's generic `:confirm` modal ([`ui/confirm.cljs`](frontend/src/app/main/ui/confirm.cljs)) — no new modal component. Existing delete primitives (`dwl/delete-component`, `dwl/delete-color`, `dwl/delete-typography`, `dwsh/delete-shapes`) — no new delete logic.
- **New**: one menu entry in [`assets/groups.cljs`](frontend/src/app/main/ui/workspace/sidebar/assets/groups.cljs), one `on-delete-group` handler in each of the three asset-type files, three i18n strings. The menu entry only appears when the caller passes `on-delete-group`, so asset sections that don't implement it (e.g. graphics, which is a flat list) are unaffected.

### Steps to reproduce

1. Open any workspace file, open the **Assets** panel.
2. Under **Colors**, create 3 colors and drag them into a new group (e.g. `brand/primary`).
3. Right-click the group header → new **Delete group** option is visible.
4. Click it → confirm modal shows "Are you sure you want to delete these 3 assets?".
5. Accept → all 3 colors are removed from the assets panel.
6. Press **Cmd+Z** (Ctrl+Z) → all 3 colors return in a single undo step.
7. Repeat for **Typographies** and **Components** (including a component-with-variants case) to verify the same behavior across asset types.

[...]

PR #9211, opened 2026-04-28
:recycle: Extract delete-asset-group! helper for assets panel

### LINKED ISSUE #9141: feature: Enable deleting entire asset groups in assets panel

In the assets panel, users can delete individual assets or use multiselect to remove several at once. However, there is no way to delete an entire asset group in one action.

When working with large groups, such as many colors inside a group, this becomes time consuming and repetitive. Even with multiselect, users still need to manually select all items, which is not practical for large sets.

Allow users to delete an entire asset group directly from the assets panel. For example, when right clicking on a group or selecting it, there should be an option to delete the group. This action should remove all assets inside the group after a confirmation step.

The behavior should be consistent across different asset types such as colors, components, and typographies.

From the user point of view, this can be tested by creating a group with multiple assets, deleting the group, confirming the action, and verifying that all contained assets are removed without needing to manually select them first.

_No response_

### Additional context

From: https://community.penpot.app/t/delte-assets-groups-completly/10516

FairyPigDev:
## What and why

Reviewer follow-up on #9151. The "Delete group" handler in the assets panel was duplicated across three section files (colors, typographies, components), each carrying the same skeleton — filter assets by group path, build an `undo-id`, run the deletes inside one transaction, and show the same confirm modal — with only the path predicate and the per-asset delete event differing.

This PR extracts the shared scaffolding into `app.main.ui.workspace.sidebar.assets.common/delete-asset-group!` and
rewrites the three call sites to use it. No user-visible behavior change.

## How

Added a factory `delete-asset-group!` that accepts the differing parts as options and returns the `(fn [path] …)` body each `:on-delete-group` needs:

- `:assets`             — collection to filter.
- `:on-clear-selection` — invoked before the deletes.
- `:delete-events`      — `(fn [matching-assets] => seq-of-events)`.
- `:path-filter`        — predicate, defaults to `str/starts-with?`;
                          components override to `cpn/inside-path?` to match the existing ungroup/combine helpers.

Variant-container dedup (one `delete-shapes` per container, not per sibling variant) stays in the components section's `:delete-events` fn — it's genuinely component-specific.

The `:as i18n` alias is no longer needed in any of the three section files (its only use was `i18n/c` for the modal count, now handled by the helper); reduced to `:refer [tr]`.

## Diff size

4 files, **+69 / −112** (net −43 lines).

| File | Change |
| --- | --- |
| `assets/common.cljs` | new helper, +33 |
| `assets/colors.cljs` | call site simplified |
| `assets/typographies.cljs` | call site simplified |
| `assets/components.cljs` | call site simplified, variant dedup kept |

## Related

- Follow-up to #9151 (review comment suggesting extraction)
- Github #9141

## Testing notes

Manual smoke test in workspace assets panel:

- [ ] Right-click an asset group → "Delete group" → confirm modal shows with the correct count for **colors**.
- [ ] Same for **typographies**.
- [ ] Same for **components**, including a group containing a variant container with multiple sibling variants — confirm a single
      `Cmd+Z` restores everything in one step.
- [ ] Cancel on the confirm modal does not delete anything.
- [ ] Empty group path is a no-op (modal does not appear).

No new translation strings; existing `modals.delete-asset-group.*`
keys are reused.

## Breaking changes

None.


FairyPigDev:
Hello, @niwinz 
Could you please check this PR? I extracted common code and refactored for deleteing group functionality.

frontend/src/app/main/ui/workspace/sidebar/assets/common.cljs
```diff
@@ -198,6 +198,39 @@
          (run! st/emit!))
     (st/emit! (dwu/commit-undo-transaction undo-id))))
 
+(defn delete-asset-group!
```
niwinz: I suggest naming it acording to the return type: `make-delete-asset-group-fn`

frontend/src/app/main/ui/workspace/sidebar/assets/components.cljs
```diff
@@ -496,59 +496,33 @@
...
+                   variant-containers
+                   (->> variants
+                        (group-by :variant-id)
+                        (map (fn [[_ comps]] (first comps))))]
+               (concat
+                (map #(dwsh/delete-shapes (:main-instance-page %)
+                                          #{(:variant-id %)})
+                     variant-containers)
+                (map #(dwl/delete-component {:id (:id %)})
+                     non-variants))))}))
```
niwinz: I suggest transforming this mf/use-fn to mf/with-memo [...]
Rationale: right now independently of the deps, a new callback will created on each render and this all can be prevented just putting it all under mf/with-memo

Example: 

Image

niwinz:
Added several suggestions. Overall change is OK for me. Good job :heart:

FairyPigDev:
Hello, @niwinz 
Thank you for your feedback.
I fixed issue you mentioned. Please check this PR again.

niwinz:
Looks ok from my point of view, i just need to test it.
I don't have time for it today, try to do it tomorrow. It is almost ready to merge, just want a small check on it.

FairyPigDev:
Hello, @niwinz 
Thank  you. :)
Btw, is there any discussion room like Discord for Penpot?

carolportugal18:

Issue #9082, opened 2026-04-21
Devise concept for differentiation of agent (MCP) actions from user actions


opcode81:
An MCP beta user remarked that it would be nice to know which actions in the history were done by the agent vs. a human user. 
A similar idea should be applied to snapshots. Perhaps we should create snapshots after a series of agent actions that are identifiable as agent-induced snapshots.

This issue is about 
* deciding whether we want to do this and
* if so, creating a concept describing how we want to go about it.

edwin-rivera-dev:
Hi @opcode81 !
I'd like to share a quick concept proposal for this.

**Recommendation: Yes, we should do this.**

As MCP usage grows, users will mix their own edits with agent-driven changes in the same session. Without differentiation, it's hard to understand what happened especially when you need to roll back only the agent's actions.

**Proposed concept**

1. History tagging - Add an origin-type field (:user or :mcp) to each change entry. Purely additive, no changes to existing history semantics.

2. Agent snapshots - Auto-create a snapshot after MCP actions complete, with a recognizable prefix like "Agent: ..." so users have a clear restore point.

3. UI indicator - Show a small icon or label on agent-originated entries in the history panel so users can tell them apart at a glance.

**Questions**

- Should agent entries be collapsible so users can undo all agent actions in one step?
- Should agent snapshots be named after the MCP tool invoked, or just timestamped?

Happy to adjust based on feedback!

PR #9224, opened 2026-04-29
:sparkles: Create file snapshot when MCP connects to a workspace

### LINKED ISSUE #9219: Add new file version/snapshot upon MCP connection
As discussed in the context of #9082, we shall add a new (appropriately named) snapshot as soon as the MCP server is connected.
Snapshot creation is cheap (pointer only), so always doing this does not waste resources, but it could potentially create too many snapshots for MCP power users. We'll have to observe how we feel about this once it's deployed on develop.

jack-stormentswe:
## Related Ticket

Closes #9219
Related: #9082

## Summary

When an MCP client connects to a Penpot workspace and starts making changes, there's currently no clear restore point a user can return to if they want to undo everything the agent did. This PR adds an automatic, labeled snapshot at MCP connection time so users always have a one-click rollback target for "before the agent touched anything."

Snapshots in Penpot are pointer-based (cheap to create), so doing this on every connection has minimal storage cost.

## Why this matters

Penpot is the open-source design tool for everyone, including designers experimenting with AI agents for the first time. For that
audience, the moment an agent starts editing their file is the moment they most want a safety net. Forcing them to remember to snapshot manually before connecting MCP is a recipe for lost work and lost trust. Auto-snapshotting at connection time turns "I let the AI touch my file and now I can't get back" into a single click in version history.

## Steps to Reproduce (the gap)

1. Open a Penpot file in a workspace.
2. Connect an MCP client (e.g. via the Penpot MCP server).
3. Let the agent make several modifications via the plugin API.
4. Open **Version history**.
5. Observe: there is no marker showing "this is where MCP started" — the agent's changes are interleaved with prior user changes with no boundary.

**Expected:** A labeled snapshot at the connection point so the user can roll back to "just before the agent connected."

## Design Notes

- **Snapshot per session, not per change.** Avoids snapshot spam for power users who keep MCP connected for long sessions.
- **Pointer-based snapshot.** Penpot's snapshot machinery already uses copy-on-write pointers, so the storage cost is trivial even if many MCP sessions occur.
- **No frontend logic change required for triggering** — the existing `:origin-type :mcp` plumbing carries the signal end-to-end.
- **Auto-expiry honored** via the existing `created-by "agent"` deletion policy, so old MCP markers don't accumulate forever.

## Checklist

- [ ] First MCP-tagged update creates a labeled snapshot
- [ ] Subsequent updates in the same session do not create new snapshots
- [ ] Disconnecting and reconnecting MCP creates a new snapshot
- [ ] Non-MCP edits do not trigger this codepath
- [ ] Snapshot is visible and labeled clearly in version history
- [ ] Auto-expiry via `created-by "agent"` works as expected
- [ ] Tested in Chrome and Firefox
- [ ] CHANGES.md entry added

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -49,6 +49,7 @@
 - Preserve vector content when pasting from external tools such as Inkscape: recognise SVG sent as text/plain (with optional XML declaration and HTML comments), skip the raster preview when an SVG sibling is on the clipboard, and ignore empty SVG blobs that some tools advertise alongside the real payload, so pasted graphics arrive editable without spurious "SVG is invalid" warnings [Github #546](https://github.com/penpot/penpot/issues/546)
 - Add Shift+Numpad0/1/2 as aliases to Shift+0/1/2 for zoom shortcuts [Github #2457](https://github.com/penpot/penpot/issues/2457)
 - Adds a **Pixel grid color** picker in the viewport settings, next to the existing canvas color control [Github #7750](https://github.com/penpot/penpot/issues/7750)
+- Add an automatic, labeled file snapshot the first time an MCP agent connects to a workspace session, giving users a one-click rollback target for "before the agent touched anything" in the version history; the snapshot is taken once per `(file, session)` pair to avoid spam, and pointer-based so the storage cost is minimal [Github #9219](https://github.com/penpot/penpot/issues/9219)
 
 ### :bug: Bugs fixed
 
--- a/backend/src/app/rpc/commands/files_update.clj
+++ b/backend/src/app/rpc/commands/files_update.clj
@@ -47,6 +47,34 @@
 (declare ^:private take-snapshot?)
 (declare ^:private invalidate-caches!)
 
+;; In-memory marker of MCP sessions that have already received their
+;; one-shot "MCP connected" snapshot. Keyed by `[file-id session-id]`
+;; so a single MCP session that touches multiple files snapshots each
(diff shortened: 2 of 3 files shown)


niwinz:
hey @jack-stormentswe 

Looks very interesting, but i think the current approach does not takes in account scalability of the application. The current approach is pretty OK if penpot always runs on a single instance. But it is not the case, we on our saas per example usually run multiple instances and request can go to different instances. So maintaining the "per session" mark on the memory is not a good solution.

Additionally, it adds more complexity to the update-file which is already very complex.

In my opinion, we should take a very different approach for this. We should instruct MCP to perform the snapshot before proceed instead of making this implicit.

jack-stormentswe:
Thanks for your careful review. @niwinz ❤️ 
Understood. 👍 As for saas platform, maintaining the "per session" mark on the memory is not a good solution.

I think we should move the trigger to the MCP bridge. On `connect`, the bridge calls the existing `create-file-snapshot` RPC once per `(session, file)` and tracks that on its own side. `update-file` stays untouched, no backend-side state, scales naturally.

Please share your recommendation and assign this to me again.
Appreciate.

niwinz:
I don't know if on-connect is a good trigger. I mean, we can have connection problems and this will trigger multiple snapshots.
I think if this can be part of the plugins api, and MCP initial instructions so each MCP work can handle its own snapshots instead of making part depending on the connection event.

PR #9231, opened 2026-04-29
:bug: Preserve renamed layer name when re-entering edit mode

### LINKED ISSUE #9230: bug: Preserve renamed layer name when re-entering edit mode
### Summary

Renaming a layer twice in the layers panel silently discards the first
rename. When the user double-clicks (or otherwise enters rename mode)
on a layer that was previously renamed, the input opens with the
type-based default name (e.g. "Ellipse", "Rectangle") instead of the
user's saved name. If the user types a new name and confirms, the
saved name is overwritten with whatever they type, and the previous
rename is lost without warning.

### Steps To Reproduce

1. Create any shape in the workspace (e.g. an Ellipse).
2. In the **Layers** panel, double-click the layer name and rename it to **"sun"**. Press Enter.
3. Confirm the layer panel shows **"sun"**.
4. Double-click the same layer name again to rename it.
5. Observe: the rename input opens populated with **"Ellipse"** — the type-based default — not **"sun"**.

### Expected behavior

The rename input should open with the current saved name **"sun"**, so the user can edit from that starting point.

### Actual behavior

The rename input opens with the type-based default. If the user types a new value (e.g. "earth") and confirms, the original "sun" rename is silently lost.

### Screenshots or video

https://github.com/user-attachments/assets/25cd13a1-8a52-4f43-b98e-214ebf283e33

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

jack-stormentswe:
## Related Ticket

Closes #9230

## Summary

Fixes a silent data-loss bug in the layers panel: renaming a layer a second time would open the rename input with the type-based default name (e.g. "Ellipse") instead of the user's saved name (e.g. "sun"), allowing the previous rename to be silently overwritten on confirm. The rename input now correctly reads the shape's saved `:name` so the user's earlier work is preserved.

## Why this matters

Penpot is the open-source design tool for everyone, including users who do not carry Figma muscle memory with them. Renaming layers is one of the most fundamental interactions in any design tool. Silent loss of a rename is exactly the kind of "Penpot is unreliable" moment that pushes a hesitant first-time user back to commercial tools. Fixing it removes a small but trust-eroding bug from a core workflow.

## Steps to Reproduce (the bug)

1. Create an Ellipse on the canvas.
2. In the **Layers** panel, double-click the layer name. Rename it to **"sun"** and press Enter. The layer now shows "sun".
3. Double-click the same layer name again to rename it.
4. Observe: the rename input is populated with **"Ellipse"**, not **"sun"**.
5. Type **"earth"** and press Enter.
6. The layer now shows **"earth"**. The rename to "sun" was silently lost; there is no undo prompt or visual feedback.

## Expected

The rename input should always seed with the layer's current saved name. Editing it should be an edit *of that name*, not a replacement
of a default value.

## Attachments

https://github.com/user-attachments/assets/9c49450a-c476-452e-850a-fa78f4210067


jack-stormentswe:
Hi @niwinz 
How are you? ❤️ Please review this PR and share your recommendation.
Appreciate.

niwinz:
Hey @jack-stormentswe 
Thanks for the fix :heart: 

I have tested it and I think this code can be removed:

```clojure
    ;; Force-refresh the input's DOM value on every entry into edit
    ;; mode. The `:default-value` prop seeds an uncontrolled input only
    ;; on first mount, so without this effect a second rename would
    ;; reopen the input with the original (stale) name — silently
    ;; overwriting the user's previous rename on confirm.
    (mf/with-effect [edition? default-value]
      (when edition?
        (when-let [node (mf/ref-val ref)]
          (dom/set-value! node (d/nilv default-value ""))
          (dom/select-text! node))))
```

The issue is fixed with the other changes so i think we should not introduce more complexity here. At least that is my opinion; i mean maybe i'm missing something but the issue is fixed without that code.

jack-stormentswe:
Hi @niwinz 
You're right, thanks for testing. Removed the extra effect - the memo deps fix alone is enough. 👍 
Fixing is Done! Please review again and let me know your opinion.
Appreciate.

jack-stormentswe:
Hi @niwinz ,CI checks have passed.
Hope you merge this. ❤️

carolportugal18:

PR #9234, opened 2026-04-29
:bug: Resolve nested instance to immediate component in plugin Shape.component()

### LINKED ISSUE #9183: bug: API function .component() returns incorrect component for nested components
### Steps To Reproduce

1. Import the .penpot file from this issue: https://tree.taiga.io/project/penpot/issue/14014
2. Select the button saying "Read more"
3. Evaluate `return penpot.selection[0].component()` via the MCP REPL

### Expected behavior

The button component, which this button is an instance of, is returned.

### Actual behavior

The outer component `related-card`, which represents the entire card, is returned.

### Screenshots or video

_No response_

_No response_

_No response_

HOURLY

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

MilosM348:
`Shape.component()` currently walks all the way up to the OUTERMOST
instance root via `app.plugins.utils/locate-component`, so calling
`.component()` on a shape inside a nested copy returns the wrapping
outer component instead of the inner instance's own component.

## Repro (from the issue)

A `related-card` containing a `button` instance containing a "Read more"
text. Selecting "Read more" and evaluating
`penpot.selection[0].component()` returns the `related-card`
component instead of the expected `button` component.

## How

The OUTERMOST behavior is still correct for `:componentRoot` (which
documents itself as the "root") and for the legacy `:componentRefShape`
ref-resolution. So the fix adds a parallel `locate-immediate-component`
helper that walks up to the *nearest* `instance-head?` via
`ctn/get-head-shape` (with `:allow-main? true` to preserve the
main-instance code path) and uses it only at the `:component` getter
site. Other call sites of `locate-component` keep their existing
semantics.

## Defensive change

A `(when (and head component) ...)` guard is added at the call site so
a shape whose head/component cannot be resolved (e.g., partial-state
during construction) returns nil instead of constructing a
`lib-component-proxy` with nil arguments. Pre-existing
`locate-component` had the same fragility but was never reported.

## Files changed

- `frontend/src/app/plugins/utils.cljs` — new `locate-immediate-component`
  helper next to `locate-component`. Existing `locate-component` gets a
  docstring clarifying it returns the OUTERMOST.
- `frontend/src/app/plugins/shape.cljs` — `:component` getter now uses
  `locate-immediate-component`; nil-guard the proxy construction.
- `CHANGES.md` — Bugs-fixed entry under 2.17.0 (Unreleased).

Closes #9183


opcode81:
@MilosM348 please note that the ticket #9183 was assigned to me. It was actually already fixed on Tuesday (and PRs in my fork were already associated), but I hadn't yet issued the PR here (but I have now with #9298).
Please do not work on tickets that are already assigned to someone.

opcode81:
Issue was resolved via #9298; therefore, closing.

PR #9247, opened 2026-04-29
:bug: Expose Source Sans Pro 600 weight in builtin fonts list

### LINKED ISSUE #7378: Font selector is missing intermediate font weights for Source Sans Pro and similar fonts
### Steps To Reproduce

1. Create any text field
2. Go to font selection
3. Select Source Sans Pro
4. See font weight options
5. Font weights are stepping from 400 to 700, with no in-betweens

### Expected behavior

Source Sans Pro font option has the in-between font weights (500, 600, 800, etc.)

### Actual behavior

Source Sans Pro only has large steps of font weight (400, 700, 900) and does not have in-between options. Source Sans 3 has all the options though.

### Additional context

The font collection is open-source, so I'm not sure whether this is a licensing issue, or those options were not available at the time of it's inclusion into the application, but they should be available for import now.

dhgoal:
## Summary

The Source Sans Pro family in the builtin fonts list at `frontend/src/app/main/fonts.cljs` registers variants for weights 200, 300, 400, 700 and 900, but omits the semibold (600) entries — even though the font assets are already bundled and the CSS `@font-face` declarations that load them are already present. As a result, weight 600 cannot be picked from the font dropdown even though the bytes are downloadable; users see a 400 → 700 jump with no in-between.

This PR adds the two missing variant entries (600 and 600 italic) using the same `:suffix` style as the other numeric-id entries (200, 300), since the on-disk file-name component "semibold" doesn't match the weight-number `:id`.

## Reproduction

1. Create a text field on any board.
2. Open the font picker, select **Source Sans Pro**.
3. Open the weight dropdown.
4. Observe: the available weights are 200 → 300 → 400 → 700 → 900. **600 is missing**, even though `frontend/resources/fonts/sourcesanspro-semibold.{ttf,woff,woff2}` exist and `frontend/resources/styles/common/dependencies/fonts.scss:55-56` already declares both:
   ```scss
   @include font-face("sourcesanspro", "sourcesanspro-semibold", null, "600");
   @include font-face("sourcesanspro", "sourcesanspro-semibolditalic", null, "600", italic);
   ```

## Fix

Two new entries in the `:variants` vector:

```clojure
{:id "600" :name "600" :weight "600" :style "normal" :suffix "semibold" :ttf-url "sourcesanspro-semibold.ttf"}
{:id "600italic" :name "600 Italic" :weight "600" :style "italic" :suffix "semibolditalic" :ttf-url "sourcesanspro-semibolditalic.ttf"}
```

Inserted between the 400-italic and 700-bold entries to preserve weight-ordering. Same shape as the existing 200/300 numeric-id entries.

## Scope note

The issue body also mentions weights 500 and 800. No `sourcesanspro-medium` or `sourcesanspro-extrabold` assets exist in the repo, so this PR scopes to weight 600 — the recoverable subset. Adding 500/800 would require also bundling the asset files, which is out of scope here.

## Closes

Closes #7378.

## Precedent

Same file, same shape as merged PR #7087 (`:bug: Fix font variant names for source sans pro font`).


niwinz:
Do not worry if CI fails, we are experimenting some issues right  now and are not related to this PR

PR #9251, opened 2026-04-30
:bug: Skip add-recent-color when colorpicker has no completed color

jeffrey701:
## Summary

Closing the fill dialog while an image-fill upload is still in flight (or
while a gradient is mid-edit) leaves the colorpicker's `current-color`
with only `:opacity` — no `:image`, `:gradient`, or `:color`. The
`update-colorpicker-color` `WatchEvent` then constructed
`(add-recent-color partial)`, which runs the value through
`clr/check-color` and threw "expected valid color". The user saw an
**Internal Assertion Error** toast and lost the in-flight upload.

The existing `ignore-color?` guard reads `:type` from the *output* of
`get-color-from-colorpicker-state`, but that helper strips `:type` from
its result — so the guard never actually fires for any input. (Worth a
follow-up cleanup PR; this PR doesn't touch it because the new guard
makes it redundant rather than incorrect.)

## Fix

Wrap the `(rx/of (add-recent-color color))` call in a `clr/valid-color?`
gate — the same schema that `add-recent-color` itself uses via
`check-color`. A partial selection silently drops out of the recent-color
flow instead of crashing the workspace; behaviour for fully-valid colors
is unchanged.

```clj
(when (clr/valid-color? color)
  (rx/of (add-recent-color color)))
```

PR #9252, opened 2026-04-30
:sparkles: Preserve Inkscape labels when pasting SVGs

### LINKED ISSUE #7869: svg names are lost when file is pasted into penpot
### Steps To Reproduce

Download and unzip the test file: [generated.svg.zip](https://github.com/user-attachments/files/23858626/generated.svg.zip)

If you open it on inkscape for example, you will see groups and elements properly named:

Image

On penpot, however all the names are gone!

Image

### Expected behavior

Names and hierarchy to be preserved.

### Actual behavior

See steps to reproduce.

### Screenshots or video

_No response_

_No response_

_No response_

https://design.penpot.app as well as self-hosted

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

Last year (2024, around Septeper-October) I'm sure this wasn't the case. I actually relied in some of its names for [a penpot plugin](https://projects.blender.org/dfelinto/penpot-icon-generator-plugin) I wrote to import icons from Blender.

While trying to import new icons, I noticed that my plugin is no longer working.

It would be very helpful already to know whether this is considered a bug or the new expected behaviour. ./manage. This way I can decide whether I should find a different approach or wait/look into the issue.

Thanks

jeffrey701:
Steps to reproduce: paste an SVG authored in Inkscape (or any editor
that follows the inkscape:label convention) into a penpot file. The
group/element names visible in the source editor are dropped — penpot
shows generic auto-ids like `g1234` or `path5678` instead.

Root cause: `parse-svg-element` in
`common/src/app/common/files/shapes_builder.cljc` derived the shape
name from `(or (:id attrs) (tag->name tag))`. Inkscape stores
user-given element labels in the `inkscape:label` and `sodipodi:label`
namespaced attributes while `id` holds an auto-generated technical id,
so the operator's chosen name was always overridden by the technical
id when present.

`tubax/xml->clj` (the SVG parser the import pipeline already uses for
upload, paste, and library import) keeps namespaced attributes as
`:prefix:name` keywords — the same shape this file already reads
`:xlink:href` from on line 134, and that `app.common.svg` uses for
the `xlink:` namespace at lines 300-307.

Fix: extract the name-resolution logic into a public
`resolve-element-name` helper that prefers `:inkscape:label`, then
`:sodipodi:label`, then `:id`, then `(tag->name tag)`. Existing SVGs
that don't carry either label namespace fall through the same chain
as before, so the behaviour for non-Inkscape-authored SVGs is
unchanged.

This restores the behaviour dfelinto's
[penpot-icon-generator-plugin](https://projects.blender.org/dfelinto/penpot-icon-generator-plugin)
relies on — that plugin reads element names from the imported SVG to
map Blender icons to penpot components.

## Tests

`common/test/common_tests/files/shapes_builder_test.cljc` (new):

- `resolve-element-name-prefers-inkscape-label`
- `resolve-element-name-prefers-sodipodi-label-when-no-inkscape-label`
- `resolve-element-name-falls-back-to-id-when-no-label-namespace`
- `resolve-element-name-falls-back-to-tag-name-when-no-id-and-no-label`
- `resolve-element-name-inkscape-label-wins-over-sodipodi-and-id`
- `resolve-element-name-empty-attrs-uses-tag-fallback`

Registered in `common-tests.runner`.

## Test plan

- [x] Code change is purely additive on the priority chain (two new
      preferred keys before the existing `:id` fallback); cannot
      regress SVGs that don't carry a namespaced label.
- [x] Tests added covering the priority order and each fallback step.
- [ ] Local Clojure test execution skipped — penpot's `common/test`
      runs via Clojure CLI which isn't available in my workstation;
      I'd appreciate a CI run on the PR. Same caveat as my prior
      penpot PR #9251 (which CI passed cleanly).
- [x] Manual repro path: paste any SVG with `inkscape:label="My Name"`
      attributes into a penpot file. Before this fix → element shows
      as `path1234`; after → element shows as `My Name`.

Closes #7869

### DIFF
--- a/common/src/app/common/files/shapes_builder.cljc
+++ b/common/src/app/common/files/shapes_builder.cljc
@@ -677,6 +677,22 @@
          (remove is-style-fragment?)  ;; Filter style fragments and hex colors
          (filter #(contains? defs %)))))  ;; Only existing defs
 
+(defn resolve-element-name
+  "Pick the most user-meaningful name for an SVG element.
+
+  Inkscape (and editors following the same convention) write the
+  operator-given label to ``inkscape:label``/``sodipodi:label`` while
+  ``id`` holds an auto-generated technical id like ``path1234``.
+  Preferring the namespaced label keeps the layer/group/element names
+  the operator sees in their source editor across a paste/import
+  (#7869); the existing ``id`` and ``(tag->name tag)`` fallbacks keep
+  legacy SVGs that don't carry a label working unchanged."
+  [tag attrs]
+  (or (:inkscape:label attrs)
+      (:sodipodi:label attrs)
+      (:id attrs)
+      (tag->name tag)))
+
 (defn parse-svg-element
   [frame-id svg-data {:keys [tag attrs hidden] :as element} unames]
 
@@ -684,7 +700,7 @@
   ;; think we should handle this case early and avoid some code
   ;; execution
 
-  (let [name         (or (:id attrs) (tag->name tag))
+  (let [name         (resolve-element-name tag attrs)
         att-refs     (csvg/find-attr-references attrs)
         defs         (get svg-data :defs)
         valid-refs   (filter-valid-def-references att-refs defs)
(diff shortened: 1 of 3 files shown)


niwinz:
Looks very nice.
Thanks :heart: 
Ready to merge; waiting CI

PR #9254, opened 2026-04-30
:bug: Only fall back to anonymous on :not-found in get-profile

### LINKED ISSUE #9253: The get-profile rpc method silently masks DB errors as "Anonymous User" instead of re-raising
### Summary 

`::get-profile` in `backend/src/app/rpc/commands/profile.clj` wraps its DB lookup in `(try ... (catch Throwable _ ...))`. The in-code
comment immediately above the `try` reads:

> We need to return the anonymous profile object in two cases, when
> no profile-id is in session, and when db call raises not found. In
> all other cases we need to reraise the exception.

But the code catches `Throwable` — every exception, including `org.postgresql.util.PSQLException` for transient DB conditions
(connection-pool checkout timeout, statement timeout, replica lag, network blip). Those all get silently mapped to `{:id uuid/zero :fullname "Anonymous User"}`.

Consequences during a transient DB outage:

- Logged-in users with valid session cookies are silently treated as anonymous → redirected to login → confusing UX and audit logs.
- Real DB outages are hidden from on-call: the endpoint returns 200 instead of 5xx, so error-rate alerts don't fire.

### Steps To Reproduce

1. Make any DB call to `get-profile` fail with something other than `:not-found` — e.g. simulate a pool exhaustion by setting `::db/pool` size to 0, or pause the DB during a request.
2. The endpoint returns the anonymous user payload

### Expected behavior

Match the comment: only `:type :not-found` should fall through to the anonymous response. Any other exception should propagate.

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

jack-stormentswe:
## Related Ticket

Closes #9253 

## Summary

`::get-profile` had a `(catch Throwable _ ...)` that silently mapped every error to the anonymous user payload, despite the comment one line above saying:

> In all other cases we need to reraise the exception.

This made the endpoint a black hole for transient DB errors - connection-pool checkout timeouts, replica lag during failover,
statement timeouts, network blips. All of them returned HTTP 200 with `{:id uuid/zero :fullname "Anonymous User"}` instead of a 5xx, and logged-in users got redirected to the login screen.

The fix narrows the catch: only `ex-info` with `:type :not-found` falls through to the anonymous response; anything else propagates to the normal error pipeline.

## Changes

**`backend/src/app/rpc/commands/profile.clj:113`** — narrowed thecatch from `Throwable _` to `Throwable cause` and gate the fallback on `(= :not-found (-> cause ex-data :type))`. Anything else rethrows.

## Design Rationale

- **Match the in-code comment**, which already documents the intended behavior. No design discussion needed.
- **No new error code, no new schema, no new endpoint.**
- **Preserves the original intent**: anonymous fallback for unauthenticated / not-found, real errors for everything else.

## Checklist

- [ ] Logged-in user with valid session unaffected
- [ ] Unauthenticated request still returns anonymous payload
- [ ] Genuinely missing profile (no row) still returns anonymous payload
- [ ] DB outage now surfaces as 5xx instead of anonymous-200
- [ ] CHANGES.md updated
- [ ] `./scripts/lint` passes


jack-stormentswe:
@niwinz Wishing you a joyful May Day! ❤️

davidv399:
Hi, @niwinz
I have opened this PR last week and hope you review it
Thanks.❤️

davidv399:
Hi @niwinz please review my PR. Appreciate your feedback. Thanks

PR #9278, opened 2026-04-30
:bug: Stop swallowing unrelated exceptions in get-profile RPC

### LINKED ISSUE #9276: bug: `get-profile` RPC swallows all exceptions despite the documented contract
### Steps To Reproduce

1. Read the comment block above the `try`/`catch` in `get-profile` at `backend/src/app/rpc/commands/profile.clj:102` to `114`. The contract states that the anonymous profile object should only be returned when no `profile-id` is in session and when the database call raises not found, and that all other cases must reraise.
2. Trigger any unrelated failure inside `get-profile`. The simplest reproduction is to drop the database connection mid call (so the JDBC layer raises a `SQLException`) or to temporarily monkey patch `get-profile` to throw `IllegalStateException`.
3. Issue a `::get-profile` request from a logged in client.

### Expected behavior

Database outages and other unexpected throwables propagate to the caller as a 5xx error. Only the two documented cases fall back to the anonymous profile object.

### Actual behavior

Any throwable raised inside the body, including `SQLException`, `OutOfMemoryError`, and arbitrary runtime errors, is caught by `(catch Throwable _ ...)` and replaced with `{:id uuid/zero :fullname "Anonymous User"}`:

```clojure
(try
  (let [profile (-> (get-profile pool profile-id)
                    (strip-private-attrs)
                    (update :props filter-props))]
    (if (contains? cf/flags :nitrate)
      (nitrate/add-nitrate-licence-to-profile cfg profile)
      profile))
  (catch Throwable _
    {:id uuid/zero :fullname "Anonymous User"})))
```

The session caller silently downgrades to anonymous, masking real outages, and downstream code that compares the response against the session `profile-id` sees the two disagree.

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

Dexterity104:
### Related Ticket

Closes #9276 

### Summary

The `::get-profile` RPC was wrapping its body in `(catch Throwable _ ...)` and returning `{:id uuid/zero :fullname "Anonymous User"}` for every failure. Database outages, `OutOfMemoryError`, and any other runtime exception were silently downgrading logged in users to anonymous, which masked real incidents and caused downstream code that compares the response against the session `profile-id` to disagree with itself.

The comment above the block already described the intended behavior: return the anonymous profile only when no `profile-id` is present in session, or when the database call raises `:type :not-found`. The implementation now matches that contract.

* The nil `profile-id` precondition is checked before entering the `try`, so the documented "no profile in session" case no longer
  relies on an exception path.
* The `catch` is narrowed from `Throwable` to `ExceptionInfo`, and inside the handler the fallback only runs when `(:type (ex-data
  cause))` is `:not-found`. Anything else is rethrown.
* `clojure.lang.ExceptionInfo` is added to the namespace `:import`, matching the pattern already used in `app.rpc.climit`.

### Steps to reproduce

1. Read the comment block above the `try` in `get-profile` at `backend/src/app/rpc/commands/profile.clj`. Confirm the contract:
   anonymous profile only when there is no `profile-id` in session, or when the database call raises not found.
2. Trigger an unrelated failure inside `get-profile`, for example by dropping the database connection mid call so the JDBC layer raises a `SQLException`, or by temporarily monkey patching `get-profile` to throw `IllegalStateException`.
3. Issue a `::get-profile` request from a logged in client.

Before the fix, the response was the anonymous profile object regardless of the underlying failure. After the fix, only the two documented cases return the anonymous profile, and every other throwable propagates to the caller as a 5xx error.

### Checklist

- [x] Add a detailed explanation of how to reproduce the issue and verify the fix.

### DIFF
--- a/backend/src/app/rpc/commands/profile.clj
+++ b/backend/src/app/rpc/commands/profile.clj
@@ -31,7 +31,9 @@
    [app.tokens :as tokens]
    [app.util.services :as sv]
    [app.worker :as wrk]
-   [cuerdas.core :as str]))
+   [cuerdas.core :as str])
+  (:import
+   clojure.lang.ExceptionInfo))
 
 (declare check-profile-existence!)
 (declare decode-row)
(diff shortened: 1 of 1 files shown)


madalenapmelo-kp:

niwinz:
I have merged this one https://github.com/penpot/penpot/pull/9254 which is a bit simplier. In any case, thanks for the contribution.

PR #9281, opened 2026-04-30
:bug: Fix conditional use-ctx hook violation in shape-wrapper

### LINKED ISSUE #9280: The use-ctx hook is called under a conditional in shape-wrapper, violating rules of hooks
### Steps To Reproduce

1. Open `frontend/src/app/main/ui/workspace/shapes.cljs` and read the `shape-wrapper` definition starting at line 95.
2. Inspect lines 103 to 106:

```clojure
;; FIXME: WARN: this breaks react rule of hooks (hooks can't be under conditional)
active-frames
(when (cfh/root-frame? shape)
  (mf/use-ctx ctx/active-frames))
```

3. Construct a scenario where the same `shape-wrapper` instance re renders with a shape whose root frame status differs across renders (rare in production thanks to `mf/memo`), but the underlying violation is still latent).

### Expected behavior

Hooks are invoked unconditionally and in the same order on every render of a single component instance, as required by the rules of hooks. Conditional selection happens on the value, not on the hook call.

### Actual behavior

mf/use-ctx is only invoked on renders where cfh/root-frame? shape is truthy. If the same instance ever re renders with a non root frame shape, the hook index shifts and React's internal hook bookkeeping for that instance corrupts. The maintainer FIXME on line 103 already acknowledges the violation; the mf/memo' wrapper masks it most of the time because shape identity changes generally remount the component.

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

Dexterity104:
### Related Ticket

Closes #9280 

### Summary

The `shape-wrapper` component in `frontend/src/app/main/ui/workspace/shapes.cljs` called `mf/use-ctx ctx/active-frames` inside a `when` form gated by `(cfh/root-frame? shape)`. That places a hook under a conditional, which violates React's rules of hooks: if the same component instance ever re renders with a shape whose root frame status differs from the previous render, the hook order shifts and React's internal bookkeeping for that instance corrupts.

The `mf/memo'` wrapper masks the problem most of the time because shape identity changes generally remount the component, but the violation is latent and was already flagged with a FIXME by a previous maintainer.

The fix hoists `mf/use-ctx` so it runs unconditionally on every render, and moves the `(cfh/root-frame? shape)` check onto the resulting value. Observable behavior is preserved: `active-frames` is still the context value when the shape is a root frame, and `nil` otherwise.

### Steps to reproduce

1. Open `frontend/src/app/main/ui/workspace/shapes.cljs`.
2. Inspect the `shape-wrapper` definition and confirm that `mf/use-ctx ctx/active-frames` is invoked at the top of the `let` binding, with the conditional applied to the resulting value.
3. Open a workspace containing both root frames and loose shapes, and confirm that frame thumbnails and shape rendering continue to behave as before.

### Checklist


madalenapmelo-kp:

niwinz:
This if look simple but it affect a lot related to performance, because it makes watching that context on all shapes, not only root. This hook rule violation is preferred. I think the solution should be a bit more sophisticated, maybe have a different component for root shapes.

Dexterity104:
Hi @niwinz 
Thanks for the review, the performance point is fair.

I pushed a follow-up that splits the component:

- `root-frame-shape-wrapper` subscribes to `ctx/active-frames` and computes `thumbnail?`.
- `shape-wrapper` dispatches to it for root frames; non-root shapes render directly through a small `render-shape-content` helper.

Hooks are called unconditionally in each component, and only the root-frame path watches the context, so non-root shapes are not invalidated when `active-frames` changes.

niwinz:
Looks nice, let me test it.

niwinz:
image
Looks like it has some issues. 

I've looked at the code but I can't figure out where the problem might be at a glance.

Dexterity104:
Hi @niwinz 
I've fixed the issue. Could you please review this again? Thank you!

niwinz:
Ok, looks ok for me. Thanks

PR #9283, opened 2026-04-30
:recycle: Make ShapeImageIds byte conversion fallible

### LINKED ISSUE #9282: The ShapeImageIds From impl carries unwrap paths that can panic
### Steps To Reproduce

1. Open `render-wasm/src/wasm/fills/image.rs` at line 73.
2. Read the `From` impl and the FIXME comment on line 75:

```rust
impl From<[u8; IMAGE_IDS_SIZE]> for ShapeImageIds {
    fn from(bytes: [u8; IMAGE_IDS_SIZE]) -> Self {
        // FIXME: this should probably be a try_from instead
        let shape_id = Uuid::try_from(&bytes[0..16]).unwrap();
        let image_id = Uuid::try_from(&bytes[16..32]).unwrap();
        ShapeImageIds { shape_id, image_id }
    }
}
```
3. Note that the `TryFrom<Vec<u8>>` impl right below (line 82) routes its input back through this `From`, so it inherits the same brittle conversion path.

### Expected behavior

Fallible conversions live in `TryFrom` and surface a `Result`. `From `impls contain no `unwrap` paths, so the type system honestly reflects which conversions can fail.

### Actual behavior

The `From` impl unwraps two `Uuid::try_from` calls. The slice lengths happen to be statically guaranteed today, so no real input panics in practice, but the impl shape misleads readers and any caller that constructs a `[u8; IMAGE_IDS_SIZE]` from a less safe source inherits the panic risk. The maintainer FIXME on line 75 already flags this.

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

Dexterity104:
### Related Ticket

Closes #9282 

### Summary

The `From<[u8; IMAGE_IDS_SIZE]>` impl on `ShapeImageIds` decoded its input via `Uuid::try_from(...).unwrap()` twice. The accompanying `TryFrom<Vec<u8>>` impl just routed through that same `From`, and additionally panicked in `copy_from_slice` when the input length did not match. Nothing in the current call graph actually triggers those panics, but the impl shape misleads readers and any future caller building the buffer from a less safe source inherits the risk.

Both impls collapse into a single `TryFrom<&[u8]>` that returns `Result<Self>` and surfaces an `Error::CriticalError` when the input is shorter than 32 bytes. The two call sites in `store_image` and `store_image_from_texture` now pass slices directly, which also removes a redundant `.to_vec()` allocation on each call.

Behavior on valid 32-byte input is unchanged. Inputs shorter than 32 bytes now return an error instead of panicking; longer inputs are accepted and only the first 32 bytes are read.

### Steps to reproduce

1. Build the `render-wasm` crate.
2. Load a file containing image fills in the workspace and confirm images render as before.
3. Inspect `render-wasm/src/wasm/fills/image.rs` and confirm the `From` and `TryFrom<Vec<u8>>` impls are gone in favor of a single `TryFrom<&[u8]>`.

### Checklist


madalenapmelo-kp:

belen-albeza:
We would like to merge some PRs related to a refactor before rebasing & merging this one. Sorry for the delay!

belen-albeza:
Thanks for your contribution!

PR #9289, opened 2026-04-30
:bug: Populate is-indirect flag on file libraries from relation graph

### LINKED ISSUE #9506: Compute is-indirect flag from library relation graph instead of hardcoding it
## Summary

The `get-file-libraries` function in the backend was hardcoding the `is-indirect` flag to `false` for every library, making the field effectively useless. This is a technical/code fix to properly compute the flag from the library relation graph — the UI behavior is preserved and still shows both direct and indirect libraries.

## Problem

In `backend/src/app/binfile/common.clj`, the `get-file-libraries` function was running every row through `(map #(assoc % :is-indirect false))`. This made it impossible for any consumer to distinguish direct library dependencies from transitive ones (libraries reached through other libraries). There was already a FIXME comment in the code acknowledging the field was effectively useless.

No user-facing behavioral change results from this fix — the frontend UI continues to show both direct and indirect libraries as before. The change is purely a correctness improvement to the backend data model.

## Solution

The `is-indirect` flag is now computed in SQL via `NOT EXISTS` against the `file_library_rel` table:
- A library is **direct** when a row exists in `file_library_rel` for the given `(file_id, library_file_id)` pair.
- A library is **indirect** otherwise.

This keeps the existing recursive CTE intact, adds a single indexed lookup per row, and removes the hardcoded assignment along with the FIXME.

## Affected Files

- `backend/src/app/binfile/common.clj` — `get-file-libraries` now computes `is-indirect` via SQL `NOT EXISTS`

## Related

- PR: #9289

Dexterity104:
### Related Ticket

https://github.com/penpot/penpot/issues/9506

### Summary

`get-file-libraries` in `backend/src/app/binfile/common.clj` was running every row through `(map #(assoc % :is-indirect false))`, so consumers could never distinguish direct libraries from libraries reached transitively through the recursive CTE. The maintainer FIXME on the same block already flagged the field as effectively useless.

The flag is now computed in SQL via `NOT EXISTS` against `file_library_rel`: a library is direct when a row exists for the given `(file_id, library_file_id)` pair, indirect otherwise. This keeps the existing recursive query intact, adds a single indexed lookup per row, and removes the hardcoded assignment along with the FIXME.

The fix unblocks the workspace assets sidebar in `frontend/src/app/main/ui/workspace/sidebar/assets.cljs`, which calls `(remove :is-indirect)` to hide transitive libraries from the panel.

### Steps to reproduce 

1. Create file F with a library L1.
2. Make L1 use another library L2 so that L2 is reachable from F only through L1.
3. Open F in the workspace and check the assets sidebar.

Before the fix: L2 appears next to L1 even though it is not a direct library of F. After the fix: only L1 appears, and L2 is correctly flagged as indirect.

### Checklist


madalenapmelo-kp:

niwinz:
The code looks ok. I have created the issue. I will test it ASAP

niwinz:
The same here, we have introduced changes on bugfixes that conflicts on this.
The change is interesting for us, we can approach it on two ways: resolve conflct ourselves or you fix the conflicts.

In any case, thanks :heart:

Dexterity104:
Hi @niwinz 
I appreciate your time and effort for penpot project.
I'm sorry for my late update. I've just fixed the conflicts. Could you please review this again?

PR #9291, opened 2026-04-30
:sparkles: Validate shape on add-object to catch malformed inputs early

### LINKED ISSUE #9507: Validate shape on add-object to catch malformed inputs early
## Summary

Resolves the long-standing `FIXME: add shape validation` in `app.common.files.changes-builder/add-object`. Until now, any map handed to `add-object` was wrapped into a commit record without further checking, so malformed shapes (missing `:id`, mismatched `:parent-id`, geometry that breaks the type schema) flowed straight into the commit log and only surfaced later inside the file validator, where the source of the bad data was hard to pinpoint.

## Problem

`add-object` in `app.common.files.changes-builder` had no input validation. Malformed shapes could enter the commit log silently and only trigger errors later in the file validator, making debugging difficult. A `FIXME: add shape validation` comment had been in the code to track this gap.

## Solution

The fix runs `cts/check-shape` (already defined in `app.common.types.shape` and used elsewhere — `app.common.files.builder`, `app.main.data.workspace.shapes/add-shape`) on the incoming object before any change record is built. The call is inlined into the existing `obj` binding so validation happens on the user-supplied shape, before the internal `::index` key is grafted on.

`add-objects` benefits for free since it delegates to `add-object`.

## Steps to Reproduce

1. Hand a malformed map to `pcb/add-object` (e.g. a shape without `:type`, or a plain map that does not satisfy the `shape?` predicate).
2. Confirm a validation error is raised at the call site instead of leaking into the commit log and resurfacing later in the file validator.
3. Verify existing flows that build shapes through `cts/setup-shape` (workspace, plugins API, SVG upload, bool/group creation, library duplication) keep working unchanged.

## Affected Files
- `common/src/app/common/files/changes_builder.cljc` — Added `cts/check-shape` validation in `add-object`

## Related
- PR: #9291

Dexterity104:
### Related Ticket

https://github.com/penpot/penpot/issues/9507

### Summary

Resolves the long-standing `FIXME: add shape validation` in `app.common.files.changes-builder/add-object`. Until now, any map handed to `add-object` was wrapped into a commit record without further checking, so malformed shapes (missing `:id`, mismatched `:parent-id`, geometry that breaks the type schema) flowed straight into the commit log and only surfaced later inside the file validator, where the source of the bad data was hard to pinpoint.

The fix runs `cts/check-shape` (already defined in `app.common.types.shape` and used by `app.common.files.builder` and `app.main.data.workspace.shapes/add-shape`) on the incoming object before any change record is built. The call is inlined into the existing `obj` binding so validation happens on the user-supplied shape, before the internal `::index` key is grafted on. `add-objects` benefits for free since it delegates to `add-object`.

### Steps to reproduce

1. Hand a malformed map to `pcb/add-object` (e.g. a shape without `:type`, or a plain map that does not satisfy the `shape?`
   predicate).
2. Confirm a validation error is raised at the call site instead of leaking into the commit log and resurfacing later in the file
   validator.
3. Verify existing flows that build shapes through `cts/setup-shape` (workspace, plugins API, SVG upload, bool/group creation, library duplication) keep working unchanged.

### Checklist


madalenapmelo-kp:

Issue #9260, opened 2026-04-30
:recycle: Refactor penpot legacy components


niwinz:
## Summary

The main idea is refactor step by step legacy components from penpot codebase to use modern syntax. The modern syntax additionally gives an efficiency improvements because removes the overhead of converting props from js to clojure data structures on each render.

This task is very amenable to be done in small tasks and probably with help of AI agents. (you can note that penpot includes several AGENTS.md files across the modules and preconfigured OpenCode Agents under .opencode (the `Penpot Engineer` is a good candidate to use).

## Advanced prompt for perform a one step migration run

https://gist.github.com/niwinz/d0fc272d0c9ff321d57911113f80ace4

I have run this with MiMo V2.5 Pro and Kimi K2.6 models for test it and it works pretty well, so i think cheap models can be used easily for perform this kind of tasks.

FairyPigDev:
Hello, @niwinz 
I am workiing on this issue. I created one PR for this issue. Please check it.

niwinz:
> Hello, [@niwinz](https://github.com/niwinz) I am workiing on this issue. I created one PR for this issue. Please check it.

Nice! This is a big one.

Offtopic: Just FYI, i will not be able to review so much this days, because tomorrow is holiday  and i will get rest days Monday and Tuesday, so be patient, i will come back Wednesday

wdeveloper16:
Interested!

FairyPigDev:
@tmimmanuel 
I suggest combining your open PRs into a single PR. Since the code changes are minimal, having one PR will make it easier to review.

wdeveloper16:
Hi @niwinz,
I really appreciate your review and your patience with the many contributions.
I’d be grateful if you could take a look at the PRs above and share any feedback you may have.
Thanks!

Dexterity104:
Hi @niwinz Could you please review these PRs? Thanks.
#9428 
#9429 
#9433 
#9436

PR #9293, opened 2026-04-30
:recycle: Migrate fontfaces and viewer thumbnails components to modern syntax

FairyPigDev:
## What and why

Step toward #9260 (incremental migration of legacy UI components to the modern `*`-suffixed syntax, which removes the per-render JS-to-Clojure props conversion overhead).

Two unrelated namespaces are bundled here for review efficiency — both are clean Case-A migrations with no behavior change.

## Pre-flight checks

| Check | fontfaces | thumbnails |
| --- | --- | --- |
| `::mf/register` present | no | no |
| `::mf/wrap-props false` | yes (Case A) | no |
| `unchecked-get` / `obj/merge!` | no | no |
| `obj/get props "..."` (legacy props read) | yes — collapses to destructure | no |
| `?`-suffixed props | no | yes — renamed |
| External call sites | 4 (`render.cljs` × 3 + `workspace/shapes.cljs`) | 1 (`viewer.cljs`) + require |

### `shapes/text/fontfaces.cljs` (3 components)

All three previously used `::mf/wrap-props false` with a custom memoizer (`#(mf/memo' % (mf/check-props ["fonts"]))`) and read `fonts` via `(obj/get props "fonts")`. The custom memoizer existed because legacy components received raw JS-object props, so the default `mf/memo` Clojure-equality compare would always miss. With modern destructuring the props are Clojure data, so plain `mf/memo` is structurally correct — and a stronger guarantee than the previous shallow JS-prop check.

```diff
-(mf/defc fontfaces-style
-  {::mf/wrap-props false
-   ::mf/wrap [#(mf/memo' % (mf/check-props ["fonts"]))]}
-  [props]
-  (let [;; Retrieve the fonts ids used by the text shapes
-        fonts (obj/get props "fonts")]
-    (when (d/not-empty? fonts)
-      [:> fontfaces-style-render {:fonts fonts}])))
+(mf/defc fontfaces-style*
+  {::mf/wrap [mf/memo]}
+  [{:keys [fonts]}]
+  (when (d/not-empty? fonts)
+    [:> fontfaces-style-render* {:fonts fonts}]))


madalenapmelo-kp:
Thanks for your contribution! We've received your pull request and assigned it for review. We have a lot on our plate right now, so it may take us a little while to get to it but we appreciate your patience and will be in touch as soon as we can!

FairyPigDev:
Hello, @madalenapmelo-kp 
Thank you.
Btw, any community sever on discord for Penpot?

PR #9295, opened 2026-04-30
:zap: Cache OIDC provider records to skip per-login discovery

### LINKED ISSUE #9294: Cache OIDC provider record to avoid rebuilding discovery and JWKS on every login
`get-provider` in `backend/src/app/auth/oidc.clj:699` rebuilds the OIDC provider record on every call, which makes each login pay the cost of an OIDC discovery fetch and a JWKS fetch, and inherits the provider's availability:

```clojure
;; TODO: add cache layer for avoid build an discover each time

(defn get-provider
  [cfg id]
  (try
    (when-let [params (some->> (db/get* cfg :sso-provider {:id id :is-enabled true})
                               (decode-row))]
      (case (:type params)
        "oidc" (prepare-oidc-provider cfg params)))
    (catch Throwable cause
      (l/err :hint "unable to configure custom SSO provider"
             :provider (str id)
             :cause cause))))
```
The TODO already calls this out. A short hiccup at the provider's discovery endpoint currently turns into a Penpot login outage.

Dexterity104:
### Related Ticket

Closes #9294 

### Summary

`get-provider` in `app.auth.oidc` rebuilt the provider record on every call, so every login paid the cost of a fresh OIDC discovery fetch and a JWKS fetch. A short outage at the provider's `.well-known/openid-configuration` endpoint translated directly into a Penpot login outage. The TODO on `oidc.clj:697` already flagged this.

The fix wraps the lookup in a 10 minute TTL cache keyed on provider id. `app.util.cache` is now built on Caffeine's synchronous API (`.build()` in place of `.buildAsync().synchronous()`). With the synchronous variant, `Cache.get` runs the load on the calling thread, and entries are never stored when the load throws or returns nil. A transient discovery failure therefore cannot poison the cache, while concurrent logins for the same provider still deduplicate to a single load.

This also corrects a latent issue in `app.util.cache`: single key `invalidate!` was calling `.invalidateAll(Object)`, which throws `ClassCastException` at runtime. It now calls `.invalidate(K)`, the correct Caffeine method for a single key.

`climit`, the only other caller of `app.util.cache`, is functionally unchanged: its load fn is fast and never throws or returns nil, so the move from async to sync makes no observable difference.

### Steps to reproduce

1. Configure a custom OIDC provider in the `sso_provider` table and enable it.
2. Trigger one successful login through that provider so the cache warms.
3. Block egress to the provider's `.well-known/openid-configuration` endpoint.
4. Confirm that subsequent logins keep working for up to 10 minutes, instead of failing on every attempt.
5. Restore connectivity and confirm that after the TTL the cache reloads with current discovery and JWKS data.

### Checklist


madalenapmelo-kp:

niwinz:
The code looks Ok, i put this on my queue for testing

niwinz:
Looks awesome! thanks!
Ready to merge!

PR #9297, opened 2026-05-01
:bug: Fall back to execCommand copy when navigator.clipboard is unavailable

### LINKED ISSUE #8496: Internal Error when clicking Copy button on Access Token page
## Description

Clicking the **Copy** button after creating an Access Token in Settings → Access Tokens causes an Internal Error page (unhandled exception crash).

## Root Cause

The Clipboard API (`navigator.clipboard.writeText`) is `undefined` when accessing a self-hosted Penpot instance over plain HTTP on a LAN IP (non-localhost). Browsers restrict `navigator.clipboard` to secure contexts (HTTPS or `localhost`).

The error is unhandled, causing a full-page crash instead of a graceful fallback.

## Steps to Reproduce

1. Self-host Penpot via Docker with `enable-access-tokens` flag
2. Access the instance from a LAN IP (e.g., `http://192.168.1.83:3100`) — NOT `localhost`
3. Go to **Settings → Access Tokens**
4. Create a new token
5. Click the **Copy** button on the token modal

## Expected Behavior

- Copy should work, OR
- A graceful fallback (e.g., select-to-copy, or a user-friendly message saying clipboard requires HTTPS)

## Actual Behavior

Full "Internal Error" crash page with:
```
Hint: can't access property "writeText", navigator.clipboard is undefined
```

Stack trace:
```
$APP.$app$util$clipboard$to_clipboard$$@.../js/shared.js
$app$main$ui$settings$access_tokens$access_token_modal$$/$copy_token$$<@.../js/main-settings.js
```

## Suggested Fix

In `app.util.clipboard/to-clipboard`, add a guard for `navigator.clipboard` availability and fall back to `document.execCommand('copy')` or show a notification that HTTPS is required for clipboard access.

## Environment

- **Penpot version**: 2.13.3 (Docker, latest as of Feb 27 2026)
- **Host OS**: Linux ARM64 (NVIDIA DGX Spark)
- **Browser**: Chromium-based, accessed via LAN IP over HTTP
- **Docker Compose**: Self-hosted with `PENPOT_PUBLIC_URI: http://192.168.1.83:3100`

## Workaround

The token value is visible on screen before clicking Copy — manually select and copy the text, or access Penpot via `https://` with a reverse proxy.

jeffrey701:
## Summary

When self-hosting Penpot over plain HTTP from a LAN IP (e.g.
`http://192.168.1.83:3100`), clicking the **Copy** button on the Access
Token modal crashed the UI with "Internal Error" and the underlying
exception:

> `can't access property "writeText", navigator.clipboard is undefined`

Browsers gate `navigator.clipboard` to **secure contexts** (HTTPS or
`localhost`), so on LAN HTTP the property is `undefined`. The
synchronous `(.writeText ^js clipboard data)` call in
`app.util.clipboard/to-clipboard` then threw a `TypeError`, which the
React error boundary surfaced as a full-page crash. Reported and
root-caused by @NerdBase-by-Stark in #8496.

## Fix

`frontend/src/app/util/clipboard.cljs` now:

- Exposes two pure availability predicates (`clipboard-write-text-supported?`,
  `clipboard-write-supported?`) so callers and tests can reason about the
  Clipboard API surface without poking `js/navigator` directly.
- Falls back to a private `legacy-write-text!` that uses
  `document.execCommand('copy')` via a temporary off-screen textarea when
  `navigator.clipboard.writeText` is missing. This works in non-secure
  contexts and is the standard pre-async-Clipboard polyfill.
- `to-clipboard` now always returns a Promise (resolved on success,
  rejected with a descriptive `Error` if both the async and legacy paths
  fail), so the synchronous-throw path that bubbled up to the React error
  boundary is gone.
- `to-clipboard-promise` now rejects with a clear error message when
  `navigator.clipboard.write` is unavailable, instead of synchronously
  throwing on `.write` of `undefined`. There is no good
  `document.execCommand` fallback for arbitrary `ClipboardItem` blobs.
- `to-clipboard-multi`'s existing fallback was buggy: when
  `navigator.clipboard` was missing the guard fell through and called
  `.writeText` on the same nil/undefined `clipboard`. It now routes the
  text fallback through `to-clipboard`, which has the legacy execCommand
  path.

## Tests

`frontend/test/frontend_tests/util_clipboard_test.cljs` — unit tests for
the two predicates covering: undefined/nil clipboard, empty object,
clipboard with only `writeText`, clipboard with only `write`, and
clipboard with both. Predicates are intentionally independent so a
partial polyfill exposing one method but not the other does not crash.

Tests are registered in `frontend/test/frontend_tests/runner.cljs`.

## Reproduction

1. Self-host Penpot via Docker with `enable-access-tokens`.
2. Access from a LAN IP (e.g. `http://192.168.1.83:3100`) — **not**
   `localhost`.
3. **Settings → Access Tokens →** create a token → click **Copy**.
4. Before this PR: full "Internal Error" crash page.
5. After this PR: copy succeeds via the `execCommand` fallback; the
   existing success notification fires.

## Closes

#8496

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -95,6 +95,7 @@
 - Fix crash in share-link viewer when a team member's email is missing `@` or has no domain TLD (by @boskodev790) [Github #9120](https://github.com/penpot/penpot/pull/9120)
 - Fix crash when pasting a component with variants from an external shared library into a file that uses that library (by @FairyPigDev) [Github #8144](https://github.com/penpot/penpot/issues/8144)
 - Remove `corepack` from the MCP local launcher so it runs on Node.js 25+, where corepack is no longer bundled (by @TheAifam5) [Github #8877](https://github.com/penpot/penpot/issues/8877)
+- Fix Internal Error crash when clicking the Copy button on a newly-created Access Token while accessing a self-hosted Penpot over plain HTTP from a LAN IP — `navigator.clipboard` is `undefined` outside secure contexts, so the synchronous `.writeText` call threw and bubbled up to the React error boundary; the clipboard helpers now fall back to `document.execCommand('copy')` and always return a Promise so callers can handle write failures uniformly (by @jeffrey701) [Github #8496](https://github.com/penpot/penpot/issues/8496)
 - Fix Copy as SVG to produce a valid document for multi-shape selections and use `image/svg+xml` MIME type (by @RenzoMXD) [Github #9066](https://github.com/penpot/penpot/pull/9066)
 - Reset profile submenu state when the account menu closes (by @eureka0928) [Github #8947](https://github.com/penpot/penpot/issues/8947)
 - Preserve OpenType variant name table for custom fonts in the dashboard (by @rutherfordcraze) [Github #8924](https://github.com/penpot/penpot/issues/8924)
--- a/frontend/src/app/util/clipboard.cljs
+++ b/frontend/src/app/util/clipboard.cljs
@@ -70,12 +70,72 @@
   ([event options]
    (from-data-transfer (.-dataTransfer ^js event) options)))
 
+(defn clipboard-write-text-supported?
+  "True when `clipboard` is a non-nil object exposing the async writeText API.
+  In non-secure browsing contexts (e.g. self-hosted Penpot served over plain
+  HTTP from a LAN IP) `navigator.clipboard` is `undefined`, in which case this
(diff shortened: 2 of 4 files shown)


madalenapmelo-kp:

niwinz:
I think this issue is already fixed. Thanks in any case.
Please reopen if you consider i'm wrong.

PR #9301, opened 2026-05-02
:sparkles: Add getVariants() method to LibraryComponent plugin API

### LINKED ISSUE #9185: Expose Variants retrieval for LibraryComponent in plugin API

In the plugin API, there appears to be no easy way to retrieve the `Variants` object corresponding to a `LibraryComponent` instance, for which `isVariant()` returns true.
Retrieving the `Variants` instance is essential, because it allows to determine which variant properties and values are admissible.

Add a way to directly retrieve the `Variants` instance.

Right now, it appeared as if the `Variants` instance could only be retrieved by looking at ancestors of the `mainInstance()` and checking them with `isVariantContainer()`, causing LLMs to use cumbersome implementations.

### Additional context

It looks like `.variants` is actually already there (in plain JS) but not exposed in the TypeScript API.

We will use this as a demo issue for the MCP self-improvement project, therefore assigned to myself.

opcode81:
## Summary

Add a `getVariants()` method to the `LibraryComponent` plugin API interface, allowing plugin developers to easily retrieve the `Variants` object for a component that is a variant.

Closes penpot/penpot#9185

## Problem

Previously, there was no direct way to get the `Variants` object from a `LibraryComponent` instance where `isVariant()` returns true. Developers had to either:
- Use type narrowing via `isVariantComponent()` to access the `variants` property on `LibraryVariantComponent`
- Manually traverse ancestors of `mainInstance()` and check with `isVariantContainer()`

## Solution

Added `getVariants(): Variants | null` to `LibraryComponent`:
- Returns the `Variants` object if the component is a variant
- Returns `null` otherwise

## Changes

- **`plugins/libs/plugin-types/index.d.ts`**: Added `getVariants()` method declaration to the `LibraryComponent` interface
- **`frontend/src/app/plugins/library.cljs`**: Added `:getVariants` implementation to `lib-component-proxy`, reusing the same logic as the existing (non-enumerable) `:variants` property
- **`plugins/CHANGELOG.md`**: Added changelog entries
- **`CHANGES.md`**: Added changelog entry

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -60,6 +60,7 @@
 - Add a search bar to filter board size presets (by @eureka0928) [Github #4658](https://github.com/penpot/penpot/issues/4658)
 - Allow customising the OIDC login button label (by @wdeveloper16) [Github #7027](https://github.com/penpot/penpot/issues/7027)
 - Add page separators in Workspace [Taiga #13611](https://tree.taiga.io/project/penpot/us/13611?milestone=262806)
+- Add `getVariants()` method to plugin API `LibraryComponent` interface [Github #9185](https://github.com/penpot/penpot/issues/9185)
 - Preserve vector content when pasting SVG from external tools such as Inkscape (by @RenzoMXD) [Github #9182](https://github.com/penpot/penpot/pull/9182)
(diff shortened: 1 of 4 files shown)


madalenapmelo-kp:

PR #9302, opened 2026-05-02
:sparkles: Expose variants retrieval via `isVariant()` type guard on `LibraryComponent`

### LINKED ISSUE #9185: Expose Variants retrieval for LibraryComponent in plugin API

opcode81:

## Problem

In the plugin API, there was no easy way to go from a `LibraryComponent` to its `Variants` object. Although `isVariant()` existed, it returned a plain `boolean`, so TypeScript users had no way to access `variants`, `variantProps`, or `variantError` without resorting to workarounds like walking ancestor shapes and checking `isVariantContainer()`, or using the separate `penpot.utils.isVariantComponent()` type guard.

The underlying JS proxy already exposed these properties — they just weren't surfaced through the TypeScript types after an `isVariant()` check.

## Solution

Change the return type of `LibraryComponent.isVariant()` from `boolean` to `this is LibraryVariantComponent`, making it a type guard. This lets TypeScript narrow the type inline:

```typescript
if (component.isVariant()) {
  // component is now LibraryVariantComponent
  const variants = component.variants;
  const props = component.variantProps;
}
```

No runtime changes are needed — the ClojureScript proxy already implements the correct behavior.

## Changes

- **`plugins/libs/plugin-types/index.d.ts`**: Change `isVariant()` return type to `this is LibraryVariantComponent` and update its docstring.
- **`mcp/packages/server/data/initial_instructions.md`**: Improve variant navigation guidance in MCP instructions.
- **`CHANGES.md`** / **`plugins/CHANGELOG.md`**: Add changelog entries.

## Relevant Issues

- #9185

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -64,6 +64,7 @@
 - Add Shift+Numpad0/1/2 as aliases to Shift+0/1/2 for zoom shortcuts (by @RenzoMXD) [Github #9063](https://github.com/penpot/penpot/pull/9063)
 - Add pixel grid color picker in viewport settings (by @Yakehira) [Github #7750](https://github.com/penpot/penpot/issues/7750)
 - Add HEX, HSB and HSL support to the color picker with a model switcher that persists across sessions (by @edwin-rivera-dev) [Github #9133](https://github.com/penpot/penpot/issues/9133)
+- Expose `variants` retrieval on `LibraryComponent` via `isVariant()` type guard in plugin API [Github #9185](https://github.com/penpot/penpot/issues/9185)
 - Show specific invitation-link error messages for expired, email-mismatch and invalid token cases [Github #9220](https://github.com/penpot/penpot/issues/9220)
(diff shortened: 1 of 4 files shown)


madalenapmelo-kp:

Alotor (approved):

PR #9312, opened 2026-05-04
:tada: Add prototype swap interaction for shapes

### LINKED ISSUE #6884: feature: Add prototype interaction to "swap" an object for another

It's not a "problem" exactly but, the current way that we do *prototypes* in Penpot is a very unintuitive, slow and overall annoying way to work.

I will describe something simple as an example (and the simpler the thing you want to do is, the worse the current system is). Let's say that I create a navbar that has an SVG for changing themes. So, I start with my "base" navbar, been used for the light theme. If I want to *toggle* the dark theme, I will copy my navbar and use it as an "overlay". When the object that changes themes (which is an SVG) in the light theme is clicked, the overlay opens. When the object (which, again, an SVG) in the dark theme is clicked, the overlay closes.

The above mechanism is relative simple but, it is a lot of work and, it has limitations! What happens if we want to "toggle" both a theme and a language? No way to do this currently. You'll need to have a different page for every language, and all of them will need to have multiple overlays for everything. Add even a 3rd parameter to the mix and things have already gone out of hand...

Even if we don't care about our development experience (which we of course should because if not, why use a design tool to begin with and not just write code, right!?), presenting the above setup to a client, where giving them 2-3 design options to choose from, will annoying, to say the least...

What if we, instead of having to do all that and facing all those limitations, we had a very simple system that would add an interaction to *change* an object for a new one? Basically, the way that would work is by creating an interaction (let's call it "swap") that replaces an object for another. That interaction could take two parameters:

1. `source`: which tells us which object will be changed. In our example, that would be set to **self** as, we want to change the current SVG, but it could be useful if we want to change something else (like a language switcher that in addition to changing its own SVG, it would also have multiple interactions to change different text objects)
2. `destination` which tells us what this object will be changed to (where, the option **self** should not be available if **self** has been used as a `source`).

With the new system, in the above example, when we clicked the **light-switcher**, it would just be set to be replaced with the **dark-switcher**. Then, the **dark-switcher** would be set to be replaced with the **light-switcher** when pressed. So simple and so powerful! No overlays and other "hacks" to do our work.

By implementing the above change, we win the following:

* A more intuitive workflow that is easier to learn and use.
* No limitations as everything can change everything and nothing is *tight* to anything.
* We work faster, doing less repetitive work and having a more clean design.
* Better performance and memory consumption (especially in bigger projects and weaker machines that will probably notice the difference).

After the above is implemented, and we have a proper way to handle "events", the overlay mechanism can be complete dropped as it will no longer be necessary. With the much-needed addition of my other request to [allow design tokens to be used in prototypes](https://github.com/penpot/penpot/issues/6883), Penpot will be truly next level!

claytonlin1110:
### Related Ticket

Closes #6884

## What and why
Implements GitHub enhancement request #6884: a prototype action that swaps the visual of one shape for another without overlay boards, including an optional alternate “layer to replace” when the trigger is not that layer.

### Steps to reproduce 

1. Open a file in the workspace. On a board, place two shapes (e.g. “light” / “dark” icons).
2. Select the trigger shape → Prototype / Interactions → add an interaction.
3. Action: Swap appearance. Leave Layer to replace as This shape (trigger) (or pick another layer).
4. Replace with: choose the second shape. Save.
5. Open Play / view mode for that board. Click the trigger (or the wired flow).
6. Confirm the target slot shows the replacement shape; the standalone replacement should not draw in its original position while the swap is active.
7. (Optional) Add a second reciprocal swap on the other shape to toggle back.

### Checklist


madalenapmelo-kp:

NatachaMenjibar:
Hey @claytonlin1110 , thanks for your contribution! Could you attach a video for a better understanding of the feature?

niwinz:
I'm closing this for inactivity, if you still interested on contirbuting this, please reopen and answer the @NatachaMenjibar question.

In any case, thanks for your time.

PR #9353, opened 2026-05-05
:sparkles: Improve nitrate checkout failure feedback on subscription …

juanfran:
…page

  When the Stripe checkout fails to start, the subscription page now
  shows an inline error in the Business Nitrate card under the CTA
  instead of a toast. When the post-payment activation fails, the toast
  message is updated to point users to support@penpot.app.

  The nitrate-form modal also passed a URI object to
  build-nitrate-callback-urls while the underlying append-query-param
  relied on lambdaisland's u/parse, which only accepts strings. Switched
  to the local u/uri helper so both strings and URI records work, so
  failures opened from the modal land on the subscription page.

### Checklist

### DIFF
--- a/frontend/src/app/main/data/nitrate.cljs
+++ b/frontend/src/app/main/data/nitrate.cljs
@@ -13,6 +13,7 @@
    [app.util.i18n :refer [tr]]
    [app.util.storage :as storage]
    [beicon.v2.core :as rx]
+   [cuerdas.core :as str]
    [potok.v2.core :as ptk]))
 
 (def ^:private nitrate-entry-active-key ::nitrate-entry-active)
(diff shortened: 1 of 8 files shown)


frontend/src/app/main/data/nitrate.cljs
```diff
@@ -74,12 +75,46 @@
...
+  a `subscription` query param identifying the outcome."
+  [base-url]
+  (let [build (fn [token]
+                (append-query-param base-url :subscription token))]
+    {:success-callback      (build "subscribed-to-penpot-nitrate")
+     :error-callback        (build nitrate-checkout-error-token)
+     :finish-error-callback (build nitrate-checkout-finish-error-token)
+     :cancel-callback       (build nitrate-checkout-cancelled-token)}))
+
 (defn go-to-buy-nitrate-license
```
pabloalba, line 109: You are passing all the callbacks to this function. But everytime, you call to build-nitrate-callback-urls fo generating those callbacks.

So it would be cleaner to pass only the base-url, and call to build-nitrate-callback-urls here directly. 

Also, I think the arity with only one parameter is not used anymore

frontend/translations/en.po
```diff
@@ -5162,6 +5162,15 @@ msgstr "Inviting people while on the Unlimited plan"
 msgid "subscription.dashboard.upgrade-plan.power-up"
 msgstr "Power up"
 
+msgid "subscription.error.nitrate.checkout-cancelled"
```
pabloalba, line 5169: Missing spanish translations

frontend/src/app/main/data/nitrate.cljs
juanfran: done

frontend/translations/en.po
juanfran: done

pabloalba (approved):

PR #9366, opened 2026-05-06
:bug: Remove stray println from onboarding team_choice success handler

Dexterity104:
### Related Ticket

N/A

### Summary

The success callback in [frontend/src/app/main/ui/onboarding/team_choice.cljs](frontend/src/app/main/ui/onboarding/team_choice.cljs) had a leftover `(println go-to-team)` sitting inside the variadic argument list of `st/emit!`. Two side effects came out of that:

* Every successful submit logged the boolean value of `go-to-team` to the browser console.
* `println` returns `nil`, so a `nil` event was being passed as a positional argument to `st/emit!` and forwarded into the store on each onboarding submit.

The fix drops the `println` form and lifts the conditional navigation into its own `st/emit!` call, so the redirect is only dispatched when `go-to-team` is truthy and the variadic list never receives a `nil` event. 

### Steps to reproduce

1. Open Penpot in the browser with the developer console visible.
2. Sign up as a new profile and progress through onboarding until the team choice step.
3. Submit the form.

Before the fix, the console prints the value of `go-to-team` and a `nil` event reaches the store. After the fix, the console stays clean and only the intended events (`update-profile-props`, plus a conditional `go-to-dashboard-recent`) are emitted.

### Checklist


madalenapmelo-kp:

PR #9368, opened 2026-05-06
:bug: Fix plugin schema-error formatter losing field name

### LINKED ISSUE #9290: applyToken fails with "Field message is invalid" on color/dimension/opacity tokens (SDK 2.15)
## Setup

| | |
|---|---|
| Penpot version | 2.15.0 (production, design.penpot.app) |
| MCP server | `@penpot/mcp@2.15.0-rc.2` (tagged `latest` + `stable` on npm) |
| MCP mode | Remote (`design.penpot.app/mcp/stream?userToken=…`) |
| Client | LLM with tool-use via `mcp__penpot__execute_code` |

## Bug — `applyToken` rejects all token types with generic error

```javascript
const core = tokens.sets.find(s => s.name === 'core');
const colorTok = core.tokens.find(t => t.name === 'palette.gold.50');
// colorTok = { id, name: 'palette.gold.50', type: 'color', value: '#563f00', description }

r.x = 100; r.y = 100;
r.resize(80, 80);

r.applyToken(colorTok, ['fill']);
// throws: [PENPOT PLUGIN] Value not valid: Field message is invalid: . Code: :error
```

**Tested signatures, all fail with the same generic error** (or a similar one for `applyToShapes`):

| Signature | Result |
|---|---|
| `r.applyToken(tok, ['fill'])` | `Field message is invalid` |
| `r.applyToken(tok)` (default property) | `Field message is invalid` |
| `r.applyToken(tok, 'fill')` (string instead of array) | `Field message is invalid: . Field message is invalid: .` |
| `tok.applyToShapes([r], ['fill'])` | `Doesn't support name: 0` |

**Not specific to color tokens** — same error on `dimension` token (`applyToken(dimTok, ['width'])`) and `opacity` token (`applyToken(opTok, ['opacity'])`). Bug appears to affect the entire `applyToken` API.

The token referenced is validly resolved:
- has an `id` and `type: 'color' | 'dimension' | 'opacity'`
- the parent set is `active === true` (verified after activating the corresponding theme)
- the rectangle is validly created (appears on canvas with default fill)

Tested on a file with 11 sets / 13 themes / 478 tokens imported via drag-drop UI (the SDK bulk import path has its own bugs which I worked around by using the drag-drop path).

## Expected vs Observed

| | |
|---|---|
| Expected | shape.fills bound to the token, `shape.tokens.fill === 'palette.gold.50'`, `getStyle` reflects the resolved value |
| Observed | Generic error message with no actionable detail; shape stays unmodified |

## Why blocking

`applyToken` is the only documented mechanism to programmatically bind a shape to a design token via the plugin SDK. Without it:
- Cannot generate visual documentation of design tokens (swatches, samples) that stays linked to the token's source value
- Cannot author/modify component instances tied to tokens via plugin
- Cannot integrate token systems with external automation (Style Dictionary sync, MCP-driven design system tooling, etc.)

## Workaround used

For color documentation we fell back to `Library.createColor()` + `LibraryColor.asFill()`, which works correctly. Trade-off: duplication between Library Colors and Tokens Studio tokens, both must be kept in sync via external script. For non-color tokens (dimension, opacity, etc.) no equivalent workaround — values inlined as plain strings, breaking the lineage to source.

## Suggestions

1. **Improve error messages** — `Value not valid: Field message is invalid: . Code: :error` provides no diagnostic signal. Minimum: include the field name, received value, expected schema.
2. **Document the matrix of supported `(token type × shape type × property)` combinations** in the plugin API docs.
3. **Possibly the same async state race as #8698** — the token has a valid `id` and the active set is `active === true`, but `applyToken` may still fail because the internal state map hasn't propagated yet. Worth checking if the same fix that addressed `theme.addSet()` async race applies here.

## Related

- #8698 (`theme.addSet` async state race — closed, fixed)
- #7916 (Plugin API for tokens — initial feature)
- Forum thread ["API and MCP issues"](https://community.penpot.app/t/api-and-mcp-issues/9962) — similar `200 OK` / persistence-fails class of bugs

Happy to file separate issues for related bugs encountered while investigating this one (`set.remove()` followed by `addSet({same name})` rejects for collision; `theme.activeSets` ghost entries with `id: ""` after drag-drop import) if useful.

tmimmanuel:
### Related Ticket

Fixes #9290 — `Plugin API: applyToken fails with "Field message is invalid" on color/dimension/opacity tokens (SDK 2.15)`.

### Summary

Plugins that exercised any schema-validated method (`shape.applyToken`, the `Library.createColor` parser, `font*` setters, etc.) saw a useless console error whenever a single-level schema check failed:

```
[PENPOT PLUGIN] Value not valid: Field message is invalid: . Code: :error
```

The literal token `message` came from the *internal* `:message` key of `interpret-schema-problem` — not from any user-facing field name — and the trailing colon-with-no-message came from the destructure binding to `nil`.

Root cause is in `app.plugins.utils/error-messages` (`frontend/src/app/plugins/utils.cljs`):

```clojure
;; before
(->> (:errors explain)
     (reduce csm/interpret-schema-problem {})
     (mapcat (comp seq val))
     (map (fn [[field {:keys [message]}]]
            (tr "plugins.validation.message" (name field) message)))
     (str/join ". "))
```

`interpret-schema-problem` builds a nested map and `assoc-in`s `{:message <string>}` under the offending field path. For the typical single-level failure (`{:applyToken {:message "…"}}`) the previous walker iterated the *inner* map, so:

- `field` was bound to `:message` (the leaf key) instead of the actual field name.
- `{:keys [message]}` was destructured against the *string* leaf, so `message` was `nil`.

The result was a fixed `"Field message is invalid: "` for every plugin schema failure — exactly the symptom in the bug report.

### Fix

Replaced the inline `mapcat` chain with a small recursive walker, `collect-schema-error-leaves`, that emits `[field-path message]` pairs from every leaf in the post-reduce tree. The renderer joins the path with `.` and falls back to `"value"` when the schema rejects the whole input. Numeric path segments (tuple positions) are stringified safely so callers like `applyToken` (whose schema is a `[:tuple …]`) now read `"applyToken.1"` instead of crashing through `name`.

After the fix, a `r.applyToken(tok, ['fill'])` call that hits the validator surfaces something like:

```
[PENPOT PLUGIN] Value not valid: Field applyToken.1 is invalid: Expected a set of token attributes. Code: :error
```

…which actually points at the offending argument.

### Steps to reproduce

The reporter's repro applies verbatim:

```js
const core   = tokens.sets.find(s => s.name === 'core');
const tok    = core.tokens.find(t => t.name === 'palette.gold.50');
const r      = penpot.createRectangle();
r.x = 100; r.y = 100; r.resize(80, 80);
r.applyToken(tok, ['fill']); // before: "Field message is invalid: ."
```

After the patch, if the call still fails (e.g. permission, attribute kind), the console message names the actual offending field and includes the schema's own message body. The original token-application bug surfaces *underneath* this fix and is out of scope here.

### Tests

Added `frontend/test/frontend_tests/plugins/utils_test.cljs` covering the walker:

- single-level keyword path
- single-level numeric (tuple) position — would previously crash through `name`
- single-level string path
- nested keyword and mixed paths
- multiple sibling errors
- empty input
- defensive nil-leaf handling

Plus a smoke test that `error-messages` returns `""` for an empty explain.

The new test namespace is registered in `frontend/test/frontend_tests/runner.cljs`.

`clj-kondo` lints clean on every touched file.

### Checklist

- [ ] Include screenshots or videos, if applicable. *(Console-only error string change.)*
- [ ] Refactor any modified SCSS files following the refactor guide. *(No SCSS touched.)*
- [ ] Update the `CHANGES.md` file, referencing the related GitHub issue, if applicable. *(Plugin-API error message detail — happy to add an entry if maintainers prefer.)*

### DIFF
--- a/frontend/src/app/plugins/utils.cljs
+++ b/frontend/src/app/plugins/utils.cljs
@@ -276,13 +276,68 @@
   (let [s (set values)]
     (if (= (count s) 1) (first s) "mixed")))
 
+(defn- error-path-segment->str
+  "Stringify one segment of a schema-explain path.
+
+  Keywords lose the leading colon, ints are kept verbatim (so tuple positions
+  read as ``\"applyToken.0\"`` rather than crashing through ``name``), and
+  anything else falls through ``str``."
+  [seg]
+  (cond
+    (keyword? seg) (name seg)
+    (string? seg)  seg
+    :else          (str seg)))
+
+(defn- error-path->label
+  "Render the dotted field path that the user sees in the error message.
+
+  Empty paths fall back to ``\"value\"`` so the formatted message never reads
+  ``\"Field  is invalid: …\"`` when the schema rejects the whole input."
+  [path]
+  (if (seq path)
+    (str/join "." (map error-path-segment->str path))
+    "value"))
+
+(defn collect-schema-error-leaves
+  "Walk the nested error map produced by ``csm/interpret-schema-problem`` and
+  emit a sequence of ``[field-path message]`` pairs, one per leaf.
+
+  Each leaf is the ``{:message <string>}`` map that the reducer assoc-ins under
+  the offending field path. Returning the *full* path (rather than just the
+  innermost key, as the prior implementation did) keeps the eventual user
+  message anchored to the property the plugin actually called — for example
+  ``applyToken.1`` instead of the unhelpful ``message``."
+  ([m] (collect-schema-error-leaves [] m))
+  ([path m]
+   (cond
+     (and (map? m)
+          (contains? m :message)
+          (== 1 (count m)))
+     [[path (:message m)]]
+
+     (map? m)
+     (mapcat (fn [[k v]] (collect-schema-error-leaves (conj path k) v)) m)
+
+     :else
+     [])))
+
 (defn error-messages
+  "Render an ``::sm/explain`` map into a list of human-readable plugin errors.
+
+  The previous implementation iterated entries of the inner ``{:message …}``
+  map and destructured each value, which silently produced a literal
+  ``\"Field message is invalid: \"`` string for any single-level schema failure
+  (issue #9290). The walker now collects every leaf with its full field path so
+  the formatted line names the offending property and the message body is the
+  schema's own explanation."
   [explain]
   (->> (:errors explain)
        (reduce csm/interpret-schema-problem {})
-       (mapcat (comp seq val))
-       (map (fn [[field {:keys [message]}]]
-              (tr "plugins.validation.message" (name field) message)))
+       (collect-schema-error-leaves)
+       (map (fn [[path message]]
+              (tr "plugins.validation.message"
+                  (error-path->label path)
+                  (or message ""))))
        (str/join ". ")))
 
 (defn handle-error
--- a/frontend/test/frontend_tests/plugins/utils_test.cljs
+++ b/frontend/test/frontend_tests/plugins/utils_test.cljs
@@ -1,0 +1,109 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC
+
+(ns frontend-tests.plugins.utils-test
+  "Regression coverage for the schema-explain walker that backs the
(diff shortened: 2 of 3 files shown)


madalenapmelo-kp:

niwinz:
Hey @tmimmanuel 

thanks for the contribution.
I think this issue is already addressed by this PR https://github.com/penpot/penpot/pull/9486

Closing for now, if you consider is still not fixed, consider reopening.
:heart:

PR #9370, opened 2026-05-06
:recycle: Refactor spiral rendering

AzazelN28:
### Related Ticket

### Summary

Changes how spiral is rendered.

### Checklist

### DIFF
--- a/render-wasm/_build_env
+++ b/render-wasm/_build_env
@@ -80,6 +80,9 @@ function copy_artifacts {
 
     cp target/wasm32-unknown-emscripten/$BUILD_MODE/render_wasm.js $DEST/$BUILD_NAME.js;
     cp target/wasm32-unknown-emscripten/$BUILD_MODE/render_wasm.wasm $DEST/$BUILD_NAME.wasm;
+    if [ -f target/wasm32-unknown-emscripten/$BUILD_MODE/render_wasm.wasm.map ]; then
+        cp target/wasm32-unknown-emscripten/$BUILD_MODE/render_wasm.wasm.map $DEST/$BUILD_NAME.wasm.map;
+    fi
 
     sed -i "s/render_wasm.wasm/$BUILD_NAME.wasm?version=$VERSION_TAG/g" $DEST/$BUILD_NAME.js;
 
--- a/render-wasm/src/tiles.rs
+++ b/render-wasm/src/tiles.rs
@@ -22,43 +22,95 @@ impl Tile {
 pub struct TileRect(pub i32, pub i32, pub i32, pub i32);
 
 impl TileRect {
+    pub fn empty() -> Self {
+        Self(0, 0, 0, 0)
+    }
+
+    #[inline]
     pub fn x1(&self) -> i32 {
         self.0
     }
 
+    #[inline]
     pub fn y1(&self) -> i32 {
         self.1
     }
 
+    #[inline]
     pub fn x2(&self) -> i32 {
         self.2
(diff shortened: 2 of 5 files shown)


render-wasm/src/tiles.rs
```diff
@@ -195,43 +247,76 @@ impl TileHashMap {
...
+            if direction_current == direction_total {
+                if direction % 2 == 0 {
+                    direction_total_x += 1;
+                } else {
+                    direction_total_y += 1;
+                }
+                direction = (direction + 1) % 4;
+                direction_current = 0;
+            }
+            current += 1;
```
elenatorro: now generate_spiral produces total + 1 tiles, right? but we have:

```
if self.spiral.len() != (spiral_rect.width() * spiral_rect.height()) as usize {  
```

so this is never going to happen 🤔

render-wasm/src/tiles.rs
```diff
@@ -247,7 +332,22 @@ impl PendingTiles {
...
+        //     return;
+        // }
+
+        self.spiral_rect = *spiral_rect;
+
+        // We do not regenerate spiral if the spiral_rect
+        // doesn't change. The spiral_rect is based on the
+        // viewbox so, if the viewbox doesn't change
+        // the spiral should not change.
+        if self.spiral.len() != (spiral_rect.width() * spiral_rect.height()) as usize {
```
elenatorro: Related to the +1 change, comparing against width * height might be wrong. For example:
- `TileRect`  is inclusive, so `TileRect(0, 0, 2, 1)` -> 3 * 2 = 6 tiles.
- However, for spiral_rect, width() * height() = 2 * 1 = 2 tiles

render-wasm/_build_env
```diff
@@ -50,7 +50,7 @@ export CARGO_PARAMS="${@:2}";
 
 if [ "$BUILD_MODE" = "release" ]; then
     export CARGO_PARAMS="--release $CARGO_PARAMS"
-    export EMCC_CFLAGS="-O3 -sASSERTIONS=0 $EMCC_CFLAGS"
+    export EMCC_CFLAGS="-O3 -sASSERTIONS=0 --profiling $EMCC_CFLAGS"
```
elenatorro: we should remove that for this PR, right?

render-wasm/src/tiles.rs
```diff
@@ -259,7 +359,9 @@ impl PendingTiles {
         let mut interest_cached = Vec::new();
         let mut interest_uncached = Vec::new();
 
-        for tile in spiral {
+        let center_tile = Tile(spiral_rect.center_x(), spiral_rect.center_y());
+        for spiral_tile in self.spiral.iter() {
```
elenatorro, line 353: Here with this center (center_x, center_y) only works if the pattern is symmetric. I added some logs to test it, for instance:

```
  For rect=(-3,-2 -> 3,2) (35 tiles), translated by center=(0,0):                                                                                                                                                                                             
  [SPIRAL] coverage: rect=(-3,-2->3,2) center=(0,0) ->
    tiles_outside_rect=0,
    tiles_in_rect_never_rendered=10
    (first few missing: [Tile(-3,-2), Tile(-3,-1), Tile(-3,0), Tile(-3,1), Tile(-3,2), Tile(-1,-2), Tile(0,-2), Tile(1,-2)])
```

elenatorro (changes requested):

elenatorro (approved):

PR #9385, opened 2026-05-06
:recycle: Migrate fo-text and html-text renderers to modern component syntax

FairyPigDev:
## What and why

Step toward #9260 (incremental migration of legacy UI components to the modern `*`-suffixed syntax, removing the per-render JS-to-Clojure props conversion overhead).

Twin namespaces with parallel structure migrated together. Both files were uniformly legacy: every component carried `::mf/wrap-props false` and read its props with `(obj/get props "key")`.

### `shapes/text/fo_text.cljs` (6 components)

- `render-text`, `render-root`, `render-paragraph-set`,   `render-paragraph` — renamed with `*` suffix; switched to keyword   destructuring; `::mf/wrap-props false` removed.
- `render-node` - renamed `render-node*`; uses `::mf/props :obj` +   `:as props` to keep the props symbol available; `obj/clone` +   `obj/set!` chain replaced with `mf/spread-props`.
- `text-shape` - renamed `text-shape*`; `::mf/props :obj` removed,  `::mf/forward-ref true` kept.
- `app.util.object` require dropped.

### `shapes/text/html_text.cljs` (6 components)

- Same six-component shape and same migrations as `fo_text.cljs`.
- The `code?` boolean prop was renamed to `is-code` per the migration prompt (`?`-suffixed booleans → `is-` prefix). The rename is applied at every read site and at the `text-shape*` → `render-node*` call.
- `app.util.object` require dropped.

### External call sites

- `shapes/text.cljs` — `[:> fo/text-shape props]` → `[:> fo/text-shape* props]`. The wrapper `text-shape` itself is intentionally kept legacy (see "Not touched" below).
- `code_gen/markup_html.cljs` — `text/text-shape` → `text/text-shape*` and `:code? true` → `:is-code true` in the `#js {…}` literal.
- `workspace/shapes/text/viewport_texts_html.cljs` — `[:& html/text-shape …]` → `[:> html/text-shape* …]`.

## Diff size

**5 files, +85 / −131** (12 components migrated, 2 requires removed, 4 external call sites updated).

## Testing notes

Manual smoke test in workspace:

- [ ] Open a file with text shapes — both renderers (foreignObject and HTML) render correctly.
- [ ] Edit text (type, change font, weight, line-height) — recursive `render-node*` keeps cursor/selection intact, covering the new
      `mf/spread-props` path.
- [ ] Generate code via the inspect "Code" tab on a text shape —  covers `markup_html.cljs` calling `text-shape*` with `:is-code true`.
- [ ] `pnpm run lint:clj` and `pnpm run fmt:clj` from `frontend/`pass with no manual fixes.

## Breaking changes

None. Pure internal refactor; same render output, same props contract aside from the contained `code?` → `is-code` rename inside `html_text` and its single caller.


madalenapmelo-kp:

Alotor (approved):

PR #9386, opened 2026-05-06
:recycle: Refactor GpuState and RenderState

AzazelN28:
### Related Ticket

### Summary

1. I've moved the GPU State outside the render state because in the future maybe we could  use a non-gpu renderer, so the RenderState can be decoupled more easily from GPUState.
2. I've moved the RenderState outside the State.
3. I've added some options to the `_build_env` as @superalex suggested.

The idea of `get_gpu_state` and `get_render_state` is to have a global static pointer that it is initialized when the `init` function is called and that GPU_STATE and RENDER_STATE are just pointers so access to them doesn't have any overhead. 

### Checklist

### DIFF
--- a/render-wasm/src/wasm/text_editor.rs
+++ b/render-wasm/src/wasm/text_editor.rs
@@ -2,7 +2,6 @@ use macros::{wasm_error, ToJs};
 
 use crate::get_text_editor_state;
 use crate::math::{Matrix, Point, Rect};
-use crate::mem;
 use crate::render::text_editor as text_editor_render;
 use crate::render::SurfaceId;
 use crate::shapes::{Shape, TextAlign, TextContent, TextPositionWithAffinity, Type, VerticalAlign};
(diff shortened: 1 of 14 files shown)


render-wasm/src/wasm/text_editor.rs
```diff
@@ -920,16 +920,16 @@ pub extern "C" fn text_editor_render_overlay() {
             return;
         };
 
-        let canvas = state.render_state.surfaces.canvas(SurfaceId::Target);
-        let viewbox = state.render_state.viewbox;
+        let canvas = get_render_state().surfaces.canvas(SurfaceId::Target);
```
elenatorro, line 883: maybe we can do this as in other places:

let render_state = get_render_state()
let canvas = render_state.surfaces.canvas(SurfaceId::Target);

and same on other places, so `get_render_state()` is only called once

elenatorro (changes requested):

render-wasm/src/wasm/text_editor.rs
AzazelN28: Do you mind if I do this in a different PR? `get_render_state` have a `#[inline(always)]` so it really doesn't have any performance impact (but I agree that it is more readable).

elenatorro (approved):

PR #9389, opened 2026-05-06
:recycle: Refactor State into DesignState

AzazelN28:
### Related Ticket

⚠️ PLEASE! First merge [this other PR](https://github.com/penpot/penpot/pull/9386)

### Summary

1. Renames State to DesignState
2. Changes the State Option<Box> for a mutable pointer. 
3. Move globals into a file called globals.rs
4. Adds `get_design_state` and modifies macros to support it. 

### Checklist

### DIFF
--- a/render-wasm/src/globals.rs
+++ b/render-wasm/src/globals.rs
@@ -1,0 +1,138 @@
+use macros::wasm_error;
+
+use crate::mem;
+use crate::render::{gpu_state::GpuState, RenderState};
+use crate::state::{State, TextEditorState};
+
+static mut DESIGN_STATE: *mut State = std::ptr::null_mut();
+
+/// Design State.
+pub(crate) fn get_design_state() -> &'static mut State {
+    unsafe {
+        debug_assert!(!DESIGN_STATE.is_null(), "Design State is null");
+        &mut *DESIGN_STATE
+    }
+}
+
+/// GPU State.
+static mut GPU_STATE: *mut GpuState = std::ptr::null_mut();
+
+#[inline(always)]
+pub(crate) fn get_gpu_state() -> &'static mut GpuState {
+    unsafe {
+        debug_assert!(!GPU_STATE.is_null(), "GPU State is null");
+        &mut *GPU_STATE
+    }
+}
+
+/// Render State.
+static mut RENDER_STATE: *mut RenderState = std::ptr::null_mut();
+
+#[inline(always)]
+pub(crate) fn get_render_state() -> &'static mut RenderState {
+    unsafe {
+        debug_assert!(!RENDER_STATE.is_null(), "Render State is null");
+        &mut *RENDER_STATE
+    }
+}
+
+/// Text Editor State
+static mut TEXT_EDITOR_STATE: *mut TextEditorState = std::ptr::null_mut();
+
+#[inline(always)]
+pub(crate) fn get_text_editor_state() -> &'static mut TextEditorState {
+    unsafe {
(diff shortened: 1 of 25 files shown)


render-wasm/src/globals.rs
```diff
@@ -0,0 +1,156 @@
...
+        use $crate::globals::get_design_state;
+        let $state = get_design_state();
+        if let Some($shape) = $state.current_shape() {
+            $block
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! with_state_mut_current_shape {
```
elenatorro: This is not needed anymore, right?

render-wasm/src/globals.rs
```diff
@@ -0,0 +1,156 @@
...
+}
+
+#[no_mangle]
+#[wasm_error]
+pub extern "C" fn clean_up() -> Result<()> {
+    // Cancel the current animation frame if it exists so
+    // it won't try to render without context
+    let render_state = get_render_state();
+    render_state.cancel_animation_frame();
+    unsafe { DESIGN_STATE = std::ptr::null_mut() }
```
elenatorro, line 135: should we drop here first drop(Box::from_raw(DESIGN_STATE))? 

maybe since we have a design_init we can also have a design_end method for this, and even better now that we have different stats, have an end method for them all

render-wasm/src/state.rs
```diff
@@ -38,26 +38,26 @@ impl State {
...
-    pub fn start_temp_objects(mut self) -> Result<Self> {
+    pub fn start_temp_objects(&mut self) -> Result<()> {
         if self.saved_shapes.is_some() {
             return Err(Error::CriticalError(
                 "Tried to start a temp objects while the previous have not been restored"
                     .to_string(),
             ));
         }
-        self.saved_shapes = Some(self.shapes);
+        self.saved_shapes = Some(self.shapes.clone());
```
elenatorro, line 48: To avoid cloning all the shapes:

```rs
    pub fn start_temp_objects(&mut self) -> Result<()> {
        if self.saved_shapes.is_some() {                                                                                                                                                                                                                        
            return Err(Error::CriticalError(
                "Tried to start a temp objects while the previous have not been restored".into(),                                                                                                                                                               
            ));                                                                                                                                                                                                                                                 
        }                                                                                                                                                                                                                                                       
        self.saved_shapes = Some(std::mem::replace(&mut self.shapes, ShapesPool::new()));                                                                                                                                                                       
        Ok(())                                                                                                                                                                                                                                                  
    }

    pub fn end_temp_objects(&mut self) -> Result<()> {                                                                                                                                                                                                          
        self.shapes = self.saved_shapes.take().ok_or(Error::CriticalError(
            "Tried to end temp objects but not content to be restored is present".into(),                                                                                                                                                                       
        ))?;
        Ok(())
    }
```

render-wasm/src/globals.rs
```diff
@@ -0,0 +1,156 @@
...
+macro_rules! with_state_mut {
+    ($state:ident, $block:block) => {{
+        use $crate::globals::get_design_state;
+        let $state = get_design_state();
+        $block
+    }};
+}
+
+#[macro_export]
+macro_rules! with_state {
```
elenatorro, line 54: this is the same as `with_state_mut`, right? maybe we can delete it and replace it when needed?

elenatorro (changes requested):

render-wasm/src/state.rs
AzazelN28: I think we should postpone this change until we can have time to get a deeper knowledge on what's std::mem::replace exactly doing in here.

elenatorro (approved):

PR #9390, opened 2026-05-06
:bug: Harden outbound HTTP requests against SSRF and restrict assets handlers

niwinz:
### Summary

- Add app.util.ssrf URL/host validator that resolves hostnames and blocks loopback, link-local, site-local, cloud metadata, and operator-supplied CIDRs
- Add app.media.sanitize image EOF truncator that strips trailing data after PNG IEND, JPEG EOI, GIF trailer, and WebP RIFF markers
- Disable HTTP client auto-redirect; add req-with-redirects! helper that revalidates every redirect hop against the SSRF blocklist
- Wire SSRF validation and EOF sanitization into media/download-image
- Validate webhook URLs and OIDC profile picture URLs against SSRF
- Restrict /assets/by-id to require authentication for non-public buckets (profile) while keeping public access for file-media-object, file-object-thumbnail, team-font-variant, and file-data-fragment
- Add config knobs: ssrf-protection-enabled, ssrf-allowed-hosts, ssrf-extra-blocked-cidrs


backend/src/app/util/ssrf.clj
```diff
@@ -0,0 +1,229 @@
...
+  (:import
+   com.google.common.net.InetAddresses
+   java.net.InetAddress
+   java.net.UnknownHostException
+   java.net.URI))
+
+(def ^:private allowed-schemes
+  #{"http" "https"})
+
+(def ^:private cloud-metadata-ips
```
Alotor, line 23: Is it safe to have this ips hardcoded? a change in these will requiere a new version

Alotor (approved):

backend/src/app/util/ssrf.clj
```diff
@@ -0,0 +1,229 @@
...
+    - scheme must be http or https,
+    - host must resolve to at least one address, and
+    - **every** resolved address must NOT be in the blocklist
+      (loopback, link-local, site-local, multicast, any-local,
+      cloud-metadata 169.254.169.254, IPv6 ULA fc00::/7, IPv4-mapped
+       IPv6 of any blocked IPv4, plus operator-supplied CIDRs).
+   When the host is an IP literal (decimal/octal/hex/IPv6) it is
+   normalized via `com.google.common.net.InetAddresses` before the
+   check.
+   Hosts in `:ssrf-allowed-hosts` (case-insensitive exact match) bypass
```
superalex, line 171: It says insensitive but then there is a lower

CHANGES.md
```diff
@@ -3068,4 +3059,9 @@ is a number of cores)
 
 ## 1.0.0-alpha
 
+
+
+### :boom: Security
+
+- Fix SSRF in media URL import and restrict unauthenticated asset access to public buckets only.
```
superalex: This should be associated to the last release, not the 1-0.0.0-alpha

backend/src/app/util/ssrf.clj
niwinz: Yep, are the standard ones

backend/src/app/util/ssrf.clj
niwinz: This is because the checking set is also all lowercase, so it is ok.

CHANGES.md
niwinz: True

PR #9265, opened 2026-04-30
:recycle: Migrate navigation-bullets to modern component syntax

FairyPigDev:
## What and why

Step toward #9260 (incremental migration of legacy UI components to the modern `*`-suffixed syntax, which removes the per-render JS-to-Clojure props conversion overhead).

This PR migrates one small, self-contained component:
[`app.main.ui.releases.common/navigation-bullets`](frontend/src/app/main/ui/releases/common.cljs)
— the slide-progress dots rendered by every release-notes modal.

## Pre-flight checks

The component satisfies all the criteria in the migration prompt:

- [x] No `::mf/register` metadata
- [x] No `::mf/wrap-props false`
- [x] No `unchecked-get` — already uses standard keyword destructuring (`[{:keys [slide navigate total]}]`)
- [x] No `?`-suffixed props
- [x] No `obj/merge!` / `mf/spread-props` / exotic prop chains at any callsite
- [x] Callsites pass plain literal maps, not `#js {…}`

So the migration reduces to a mechanical name + invocation rewrite.

### `releases/common.cljs` (1 file)

Component definition renamed:

```diff
-(mf/defc navigation-bullets
+(mf/defc navigation-bullets*
   [{:keys [slide navigate total]}]
   …)

PR #9392, opened 2026-05-06
:recycle: Migrate inspect fill/stroke deprecated blocks to modern syntax

tmimmanuel:
### Related Ticket

Refs #9260 — Refactor penpot legacy components.

### Summary

Migrates the two deprecated-but-still-used `fill-block` and `stroke-block` components in the inspect attribute panels from the legacy `mf/defc` form (with `{::mf/wrap-props false}`) to the modern `mf/defc*` form, following the canonical pattern from #9264 and merged #9265.

Both components are explicitly marked `;; DEPRECATED, use *-block-styles* instead` in the source. They are still exercised when the `:inspect-styles` flag is **off** (the default branch), so this is a syntactic migration with zero behaviour change — the deprecation pathway stays alive for the flag-disabled fallback.

For each component:
- Drop `{::mf/wrap-props false}` from the metadata map.
- Append `*` to the component name.
- Update the two internal `[:&` call-sites (the in-file fallback branches in `fill-panel*` and `stroke-panel*`) to `[:>`.

Both components are *internally* used only — the deprecated names are not referenced from any other namespace, so this PR has no external call-site cascade.

### Steps to reproduce

The deprecated branches render whenever the experimental `:inspect-styles` feature flag is **not enabled**. To verify:

1. Run the frontend in dev mode with the default flag set (no `:inspect-styles`).
2. Open the inspect panel on a shape that has a fill and a stroke.
3. Confirm the Background and Border-color rows render with the expected per-property values and that the copy buttons still work.
4. Toggle `:inspect-styles` on and confirm the modern `*-block-styles*` branch still renders identically (it was untouched by this PR).

### Checklist

- [ ] Include screenshots or videos, if applicable. *(No visual changes — purely a syntax / contract migration.)*
- [ ] Add or modify existing integration tests in case of bugs or new features, if applicable. *(No behaviour change.)*
- [ ] Refactor any modified SCSS files following the refactor guide. *(No SCSS touched.)*
- [ ] Update the `CHANGES.md` file, referencing the related GitHub issue, if applicable. *(Internal refactor only — not user-visible.)*

### DIFF
--- a/frontend/src/app/main/ui/inspect/attributes/fill.cljs
+++ b/frontend/src/app/main/ui/inspect/attributes/fill.cljs
@@ -27,8 +27,7 @@
 
 ;; DEPRECATED, use fill-block-styles* instead.
 ;; This component is kept for backward compatibility
-(mf/defc fill-block
-  {::mf/wrap-props false}
+(mf/defc fill-block*
   [{:keys [objects shape]}]
   (let [format*   (mf/use-state :hex)
         format    (deref format*)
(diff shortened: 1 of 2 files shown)


madalenapmelo-kp:

PR #9421, opened 2026-05-07
:bug: Route render fallback errors through the project logger

Dexterity104:
### Related Ticket

N/A

### Summary

The render entry point caught every exception during component rendering and wrote schema validation output to `js/console.log`, bypassing the configured loggers (Sentry, server forwarders) and surfacing as plain "Info" entries in DevTools. Non-schema errors were swallowed silently, with no log emitted at all.

Both catch sites in `frontend/src/app/render.cljs` now go through `log/error`, which already handles humanizing and sanitizing `::sm/explain` data via `app.common.exceptions/explain`. The duplicated catch logic is consolidated into a small `handle-render-error` helper that also returns the user facing fallback markup, so each catch stays a single line.

### Steps to reproduce

1. Open an export URL such as `/render.html?route=objects&file-id=...&page-id=...&object-id=...` with malformed parameters (for example, an invalid UUID).
2. Watch the browser DevTools console.

Before: the failure prints as two unstructured `console.log` lines and is not delivered to the configured loggers; non schema failures produce no output at all.

After: the failure is reported via `log/error` with the standard hint, sanitized humanized explain, and stack trace, matching the format used elsewhere in the codebase.

### Checklist


madalenapmelo-kp:

niwinz:
Ready to merge!
Thanks

PR #9424, opened 2026-05-07
:bug: Translate layout panel header and add-layout options

Dexterity104:
### Related Ticket

N/A

### Summary

The Layout panel header in the design sidebar bypassed i18n in three spots: the panel title "Layout", and the "Flex layout" / "Grid layout" buttons inside the add-layout dropdown (which appeared twice, once in the variant shown when a layout already exists and once in the variant shown before any layout is added).

The title now reads from the existing `labels.layout` key. Two new keys, `labels.flex-layout` and `labels.grid-layout`, back the dropdown options so the English wording stays the same while every other locale picks up a translatable string. The duplicated dropdown markup was extracted into a single local binding to keep the two branches in sync.

### Steps to reproduce

1. Switch the UI language to anything other than English (e.g. Spanish or French) from the user settings.
2. Open the design panel for a frame.
3. Click the add layout icon next to the "Layout" title.
4. Confirm the panel title and both dropdown options now render in the selected language.

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/options/menus/layout_container.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/menus/layout_container.cljs
@@ -1236,17 +1236,18 @@
         (mf/use-fn #(reset! show-dropdown* false))
 
         add-layout-dropdown
-        [:& dropdown {:show show-dropdown?
-                      :on-close on-hide-dropdown}
-         [:div {:class (stl/css :layout-options)}
-          [:button {:class (stl/css :layout-option)
-                    :data-type "flex"
-                    :on-click on-add-layout}
-           (tr "labels.flex-layout")]
-          [:button {:class (stl/css :layout-option)
-                    :data-type "grid"
-                    :on-click on-add-layout}
-           (tr "labels.grid-layout")]]]]
+        (mf/html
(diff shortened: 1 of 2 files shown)


madalenapmelo-kp:

frontend/src/app/main/ui/workspace/sidebar/options/menus/layout_container.cljs
```diff
@@ -1233,15 +1233,28 @@
         (mf/use-fn #(swap! show-dropdown* not))
 
         on-hide-dropdown
-        (mf/use-fn #(reset! show-dropdown* false))]
+        (mf/use-fn #(reset! show-dropdown* false))
+
+        add-layout-dropdown
```
niwinz, line 1238: This, defined as let will need the use of (mf/html ...)

Because component macro only compiles the body  of the function and not let binding.
I suggest moving this to a separate component or use mf/html

Dexterity104: Thanks, @niwinz  
I switched the let binding to wrap the hiccup in mf/html so it compiles to real React elements instead of staying as a raw vector. Both call sites stay untouched.

PR #9429, opened 2026-05-07
:recycle: Migrate perf/profiler to modern component syntax

Dexterity104:
### Related Ticket

Part of #9260 

### Summary

Migrates the `profiler` debug component in `app.util.perf` to modern Rumext syntax. The component is renamed to `profiler*`, the `unchecked-get` prop reads are replaced with `:keys` destructuring, and the `::mf/wrap-props false` metadata is dropped. The default for the `enabled` prop is now expressed declaratively via `:or {enabled true}`, which removes the redundant local rebind that the old syntax required.

The doc-comment example at the top of the file is also updated to the modern `[:> perf/profiler*]` form so the documentation stays in sync with the component.

### Steps to reproduce

Not applicable. The change is an internal refactor of an unused debug helper. The component has no active callsites in the codebase, so runtime behavior is preserved for all realistic invocations.

### Checklist


madalenapmelo-kp:

PR #9440, opened 2026-05-07
:recycle: Migrate color-item asset component to modern syntax

Dexterity104:
### Related Ticket

Part of #9260 

### Summary

Update the `color-item` component in the colors assets sidebar to the modern Rumext component conventions:

* Rename `color-item` to `color-item*` and remove `::mf/wrap-props false` (the body already destructures props).
* Rename boolean props to the `is-` prefix form: `local?` becomes `is-local`, `multi-colors?` becomes `is-multi-colors`, and `multi-assets?` becomes `is-multi-assets`. Every reference inside the component body is updated accordingly.
* Update the only callsite (inside `colors-group`) from `[:& color-item {…}]` to `[:> color-item* {…}]`, with the matching key renames.

`colors-group` and `colors-section*` are intentionally left as before, since they are tracked in their own follow-up migration. They continue to forward the original values through the new prop names.

### Steps to reproduce

Open any file with a local color library and verify that the colors panel still works as before:

* Single click on a color applies it to the current selection.
* Double click on a local color enters rename mode and saves on blur or `Enter`.
* Right click opens the context menu and the rename, edit, duplicate, delete, and group actions all work.
* Drag and drop a color onto another color, group, or empty area.
* Select multiple colors and confirm the bulk delete path is taken.

### Checklist


madalenapmelo-kp:

PR #9442, opened 2026-05-07
:recycle: Migrate viewer/interactions viewport components to modern syntax

Dexterity104:
### Related Ticket

Continues the Rumext modern syntax migration tracked under #9260.

### Summary

Migrates the two remaining legacy components in `frontend/src/app/main/ui/viewer/interactions.cljs` to modern Rumext syntax: `viewport-svg` becomes `viewport-svg*`, and `viewport` becomes `viewport*`. Both drop `::mf/wrap-props false` and switch from `(unchecked-get props "...")` accesses to map destructuring on the props argument. The `mf/memo` wrap stays in place on both.

The boolean prop `fixed?` is renamed to `is-fixed` per the migration convention. The `(or delta (gpt/point 0 0))` fallback continues to live in the body of `viewport-svg*`. Inside `viewport*`, `h/use-equal-memo` still wraps `interactions-mode`, `frame-offset`, and `size` so reference identity is preserved across renders.

The four callsites in `frontend/src/app/main/ui/viewer.cljs` (overlay viewports plus the main viewer wrapper) switch from `[:&]` to `[:>]` accordingly. None of those callers passed `:fixed?`, so no key rename was needed at the callsite level.

The neighboring `flows-menu*` and `interactions-menu*` were already migrated in earlier PRs and remain untouched.

### Steps to reproduce

1. Open a Penpot project that includes prototype interactions.
2. Open the viewer in interactions mode.
3. Navigate between frames, trigger overlay open and close, and confirm fixed-position elements still render the same way they did before.
4. Cycle through the interaction modes (Show, Hide, Show on click) using the header dropdown.
5. Run `./scripts/check-fmt` and `./scripts/lint` over the modified files and confirm both pass.

### Checklist


madalenapmelo-kp:

PR #9458, opened 2026-05-07
:recycle: Migrate flex-controls padding/margin/gap to modern syntax

Dexterity104:
### Related Ticket

Part of #9260.

### Summary

Continues the incremental migration of legacy UI components to the modern `*`-suffixed Rumext syntax. This change covers the three flex-layout overlay controls rendered on top of the workspace viewport (padding, margin, gap), which together expose nine related components.

Each component gains the `*` suffix, and boolean props ending in `?` were renamed to the `is-` prefix across the destructured signatures, every body reference (including `mf/deps` arrays), and every callsite that passes them.

- `padding-display*` and `margin-display*`: `hover-all?`, `hover-v?`, `hover-h?`, `hover?`, `selected?` become `is-hover-all`, `is-hover-v`, `is-hover-h`, `is-hover`, `is-selected`
- `gap-display*`: `hover?`, `selected?` become `is-hover`, `is-selected`
- `padding-rects*`, `margin-rects*`, `padding-control*`, `margin-control*`: `alt?`, `shift?` become `is-alt`, `is-shift`
- `gap-rects*` and `gap-control*` carry no boolean props, so only the `*` rename applies

Internal callsites between the trio switched from `[:&` to `[:>`. The two viewport consumers (`viewport.cljs`, `viewport_wasm.cljs`) and the `flex_controls.cljs` aggregator that re-exports the controls were updated to match.

None of the migrated components carried `::mf/wrap-props false`, `::mf/register`, `::mf/wrap`, or `::mf/forward-ref`, and none spread props to children. That makes this a clean Case-A migration with no metadata changes and no need for `::mf/props :obj`. Local `let` bindings whose names happen to end in `?` (the computed `hover-all?` / `hover-v?` / `hover-h?` inside the `*-rects*` components, the local `hover?` predicate, and helpers such as `negate?` and `resizing?`) were left untouched, since the rename rule targets props only.

`pnpm run fmt:clj` was run after the edits.

### Steps to reproduce

1. Open a workspace and add a flex-layout frame with at least two children.
2. Hover the frame to surface the gap, padding, and margin overlays.
3. Drag the padding, margin, and gap handles in turn, with and without `Alt` (apply to all sides) and `Shift` (apply to opposite pair) held.
4. Confirm that the resize cursor, hover highlight, value pill, and modifier behavior all match the pre-migration build.

### Checklist


madalenapmelo-kp:

EvaMarco:
Thanks for your patience. We've been juggling several priorities and workstreams recently, which has limited the time we've been able to dedicate to PR reviews.

We're getting back to this now and will continue the review shortly.

Thanks again for your contribution and for sticking with us while we work through the backlog 🙌

PR #9483, opened 2026-05-10
:tada: Add telemetry anonymous event collection

### LINKED ISSUE #9467: Add anonymous telemetry event collection for audit_log
## Description

When the `:telemetry` flag is ON and `:audit-log` is OFF, collect frontend and backend events anonymously in the `audit_log` table. Events are shipped in compressed batches by the existing telemetry task.

## Requirements

- **Anonymous storage**: Stored rows must be fully anonymous — `ip-addr=0.0.0.0`, `props={}`. Timestamps truncated to day precision to avoid leaking exact event timing.
- **Safe context subset**:
  - Backend events: `initiator`, `version`, `client-version`, `client-user-agent`
  - Frontend events: `browser`, `os`, `locale`, screen metrics, and `event-origin`
- **Backend** (`app.loggers.audit`): Store backend telemetry events with `source='telemetry'` and the safe context subset.
- **Frontend RPC** (`app.rpc.commands.audit`): Filter and anonymise frontend events before inserting them. `push-audit-events` should accept events when telemetry is active.
- **Telemetry task** (`app.tasks.telemetry`):
  - GC: enforce a 100,000-row safety cap by dropping the oldest rows first.
  - Batch collection: fetch up to 10,000 rows per iteration, encode, send, delete on success; stop on failure leaving remaining rows for retry.
  - Send via `fressian+zstd` batch (base64-encoded) to the telemetry endpoint.
- **Database**: Add index on `audit_log (source, created_at ASC)` for efficient queries.
- **Tests**: Comprehensive coverage of all code paths (disabled/enabled, no-op, happy-path, failure retention, PII stripping, timestamp precision, GC cap, etc.)

niwinz:
## Summary

Rewrite the audit logging subsystem to support two operating modes and
add anonymous telemetry event collection:

Modes:
- A (audit-log only): events persisted with full context
- B (audit-log + telemetry): same as A, plus events are collected for
  telemetry shipping
- C (telemetry-only): events stored anonymously with PII stripped,
  telemetry flag active, audit-log flag inactive

This is no new, telemetry already existed, we just make it more robust and send more useful information for understand how application features are used without sending private data.

Audit system refactoring (app.loggers.audit):
- Replace qualified map keys (::audit/name etc.) with plain keywords
- Rename submit! -> submit, insert! -> insert, prepare-event -> prepare-rpc-event
- Add submit* as a lower-level public API
- Add process-event dispatch function that handles all three modes and webhooks in a single tx-run!
- Add :id to event schema (auto-generated if omitted)
- Add filter-telemetry-props: anonymises event props per event type.
  Keeps UUID/boolean/number values; for login/identify events preserves
  lang, auth-backend, email-domain; for navigate events preserves route,
  file-id, team-id, page-id; instance-start trigger passes through.
- Add filter-telemetry-context: retains only safe context keys.
  Backend: version, initiator, client-version, client-user-agent.
  Frontend: browser, os, locale, screen metrics, event-origin.
- Timestamps truncated to day precision via ct/truncate for telemetry storage

Config (app.config):
- Derive :enable-telemetry flag from telemetry-enabled config option

Email utilities (app.email):
- Add email/clean and email/get-domain helper functions for domain
  extraction from email addresses

Setup (app.setup):
- Emit instance-start trigger event at system startup
- Simplify handle-instance-id (remove read-only check)

RPC layer (app.rpc):
- wrap-audit now activates when :telemetry flag is set
- Add :request-id to RPC params context for event correlation

RPC commands (management, teams_invitations, verify_token, OIDC auth,
webhooks): migrate all audit call sites to use the new plain-key API

SREPL (app.srepl.main):
- Migrate all audit/insert! calls to audit/insert with plain keys

Telemetry task (app.tasks.telemetry):
- Restructure legacy report into make-legacy-request; distinguish payload type as :telemetry-legacy-report
- Add collect-and-send-audit-events: loop fetching up to 10,000 rows per iteration, encodes and sends each page, deletes on success, stops immediately on failure for retry
- Add send-event-batch: POSTs fressian+zstd batch (base64 via blob/encode-str) to the telemetry endpoint with instance-id per event
- Add gc-telemetry-events: enforces 100,000-row safety cap by dropping oldest rows first
- Add delete-sent-events: deletes successfully shipped rows by id

Blob utilities (app.util.blob):
- Add encode-str/decode-str: combine fressian+zstd encoding with URL-safe base64 for JSON-safe string transport

Database:
- Add migration 0145: index on audit_log (source, created_at ASC) for efficient telemetry batch collection queries

Frontend:
- Always initialize event system regardless of :audit-log flag
- Defer auth events (signin identify) to after profile is set
- Refactor event subsystem for telemetry support

Tests (21 test vars, 94 assertions in tasks-telemetry-test):
- Cover all code paths: disabled/enabled telemetry, no-events no-op, happy-path batch send and delete, failure retention, payload anonymity, context stripping, timestamp day precision, batch encoding round-trip,
  multi-page iteration, GC cap enforcement, partial failure handling
- blob encode-str/decode-str round-trip tests (14 test vars)
- RPC audit integration tests (5 test vars)

Signed-off-by: Andrey Antukh <niwi@niwi.nz>

### DIFF
--- a/backend/src/app/loggers/audit.clj
+++ b/backend/src/app/loggers/audit.clj
@@ -16,12 +16,12 @@
    [app.common.uuid :as uuid]
    [app.config :as cf]
    [app.db :as db]
+   [app.email :as email]
    [app.http :as-alias http]
    [app.http.access-token :as-alias actoken]
    [app.loggers.audit.tasks :as-alias tasks]
    [app.loggers.webhooks :as-alias webhooks]
    [app.rpc :as-alias rpc]
-   [app.rpc.retry :as rtry]
    [app.setup :as-alias setup]
    [app.util.inet :as inet]
    [app.util.services :as-alias sv]
@@ -32,6 +32,60 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; HELPERS
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(def ^:private safe-backend-context-keys
+  #{:version
+    :initiator
+    :client-version
+    :client-user-agent})
+
+(def ^:private safe-frontend-context-keys
+  #{:version
+    :locale
+    :browser
+    :browser-version
+    :engine
+    :engine-version
+    :os
+    :os-version
+    :device-type
+    :device-arch
+    :screen-width
+    :screen-height
+    :screen-color-depth
+    :screen-orientation
+    :event-origin
+    :event-namespace
+    :event-symbol})
+
+(def profile-props
+  [:id
+   :is-active
+   :is-muted
+   :auth-backend
+   :email
+   :default-team-id
+   :default-project-id
+   :fullname
+   :lang])
+
+(def ^:private event-keys
+  #{:id
+    :name
+    :type
+    :profile-id
+    :ip-addr
+    :props
+    :context
+    :source
+    :tracked-at
+    :created-at})
+
+(def reserved-props
+  #{:session-id
+    :password
+    :old-password
+    :token})
 
 (defn extract-utm-params
   "Extracts additional data from params and namespace them under
@@ -47,29 +101,12 @@
                 (assoc (->> sk str/kebab (keyword "penpot")) v))))]
     (reduce-kv process-param {} params)))
 
-(def profile-props
-  [:id
-   :is-active
-   :is-muted
-   :auth-backend
-   :email
-   :default-team-id
-   :default-project-id
-   :fullname
-   :lang])
-
 (defn profile->props
   [profile]
   (-> profile
       (select-keys profile-props)
       (merge (:props profile))
       (d/without-nils)))
-
-(def reserved-props
-  #{:session-id
-    :password
-    :old-password
-    :token})
 
 (defn clean-props
   [props]
@@ -121,15 +158,16 @@
 
 (def ^:private schema:event
   [:map {:title "AuditEvent"}
-   [::type ::sm/text]
-   [::name ::sm/text]
-   [::profile-id ::sm/uuid]
-   [::ip-addr {:optional true} ::sm/text]
-   [::props {:optional true} [:map-of :keyword :any]]
-   [::context {:optional true} [:map-of :keyword :any]]
-   [::tracked-at {:optional true} ::ct/inst]
-   [::created-at {:optional true} ::ct/inst]
-   [::source {:optional true} ::sm/text]
+   [:id {:optional true} ::sm/uuid]
+   [:type ::sm/text]
+   [:name ::sm/text]
+   [:profile-id ::sm/uuid]
+   [:props [:map-of :keyword :any]]
+   [:context [:map-of :keyword :any]]
+   [:tracked-at ::ct/inst]
+   [:created-at ::ct/inst]
+   [:source ::sm/text]
+   [:ip-addr {:optional true} ::sm/text]
    [::webhooks/event? {:optional true} ::sm/boolean]
    [::webhooks/batch-timeout {:optional true} ::ct/duration]
    [::webhooks/batch-key {:optional true}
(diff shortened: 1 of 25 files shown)


wdeveloper16:
That's perfect feature.

backend/src/app/loggers/audit.clj
```diff
@@ -141,7 +179,157 @@
...
+  (let [params (-> params
+                   (assoc :id (uuid/next))
+                   (update :props db/tjson)
+                   (update :context db/tjson)
+                   (update :ip-addr db/inet))
+        params (select-keys params event-keys)]
+    (db/insert! cfg :audit-log params)))
+
+(def ^:private xf:filter-telemetry-props
+  "Transducer that keeps only map entries whose values are UUIDs."
```
Alotor: This message is not true, the filter keeps uuids, booleans and numbers

backend/src/app/loggers/audit.clj
```diff
@@ -141,7 +179,157 @@
...
+                       (assoc ::wrk/label label)
+                       (assoc ::wrk/params (-> event
+                                               (dissoc :source)
+                                               (dissoc :context)
+                                               (dissoc :ip-addr)
+                                               (dissoc :type)))))))
+  event)
+
+(defn submit*
+  "A public API, lower-leve lhan submit, assumes all required fields are filled"
```
Alotor: Typo: `lower-leve lhan`

backend/src/app/loggers/audit.clj
```diff
@@ -141,7 +179,157 @@
...
+;; PUBLIC API
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn filter-telemetry-props
+  [{:keys [source name props type] :as params}]
+  (cond
+    (or (and (= source "frontend")
+             (= type "identify"))
+        (and (= source "backend")
+             (or (= name "login-with-oidc")
```
Alotor: maybe this list of events can be stored in a set?

```
(def filter-auth-events #{"login-with-oidc" "login-with-password" "register-profile" "update-profile"})
```

backend/src/app/loggers/audit.clj
```diff
@@ -141,7 +179,157 @@
...
+                 (= name "login-with-password")
+                 (= name "register-profile")
+                 (= name "update-profile"))))
+
+    (let [props' (into {} xf:filter-telemetry-props props)
```
Alotor, line 303: I think what this is really doing is removing private information from the telemetry events, isn't it? maybe a better name for this is `strip-auth-data` or something around those lines. `filter-telemetry-props` is too generic

backend/src/app/rpc/commands/audit.clj
```diff
@@ -138,17 +161,31 @@
...
-        (catch Throwable cause
-          (l/error :hint "unexpected error on persisting audit events from frontend"
-                   :cause cause)))
-
-      (rph/wrap nil))))
+  (let [telemetry? (contains? cf/flags :telemetry)
+        audit-log? (contains? cf/flags :audit-log)
+        enabled?   (and (not (db/read-only? pool))
+                        (or audit-log? telemetry?))]
+    (if-not enabled?
```
Alotor: both branches of the `if` end with the same `(rph/wrap nil)` you can simplify this and remove both `do` statements

backend/src/app/tasks/telemetry.clj
```diff
@@ -207,6 +182,124 @@
...
+  "Delete rows by their ids after a successful send."
+  [conn ids]
+  (let [arr (db/create-array conn "uuid" ids)]
+    (db/exec-one! conn ["DELETE FROM audit_log WHERE id = ANY(?)" arr])))
+
+(defn- collect-and-send-audit-events
+  "Collect anonymous telemetry-mode audit events and ship them to the
+  telemetry endpoint in a loop. Each iteration fetches one page of
+  `batch-size` rows, encodes and sends them, then deletes the rows on
+  success. The loop stops as soon as a send fails, leaving remaining
```
Alotor: I think it's better to say "when the send returns false". That way is clear that you don't relay on exceptions to end the loop (as i thought at the beginning)

backend/src/app/tasks/telemetry.clj
```diff
@@ -218,46 +311,45 @@
...
+          subs     (get-subscriptions cfg)]
+
+
+      ;; If we have telemetry enabled, then proceed the normal
+      ;; operation sending legacy report
+
+      (if enabled?
+        (when send?
+          (db/run! cfg gc-events)
+          (px/sleep (rand-int 10000))
```
Alotor, line 331: Maybe add a comment on why this sleep is necesary?

backend/src/app/loggers/audit.clj
niwinz: I'm not sure if I agree on this 100% but in part you are right. The problem is that strip-auth-data also can lead to confusion, because it may mean username and password.

With `filter-telemetry-props` it intend to say that telemetry props have their own rules, which one consist on removing identifiable data. If is not a problem, for now i will keep the original names.

backend/src/app/loggers/audit.clj
niwinz: True

backend/src/app/loggers/audit.clj
niwinz: True

backend/src/app/loggers/audit.clj
niwinz: True

backend/src/app/rpc/commands/audit.clj
niwinz: True

backend/src/app/tasks/telemetry.clj
niwinz: OK

backend/src/app/tasks/telemetry.clj
niwinz: OK

Alotor (approved):

PR #9486, opened 2026-05-11
:bug: Fix plugin API error message for nested malli validation paths

### LINKED ISSUE #9417: Plugin API shows incorrect error messages for invalid operations
### Steps To Reproduce

Some times error messages for invalid operations in plugins show incorrectly.

Steps to reproduce:

1. From a plugin, add a TokensTheme with an invalid group:
```javascript
  const tokensCatalog = penpot.library.local.tokens;
  const theme = tokensCatalog?.addTheme({ group: 234, name: "Bad Theme" });
```
2. Now add a TokensTheme with some non existing sets:
```javascript
  const tokensCatalog = penpot.library.local.tokens;
  const theme = tokensCatalog?.addTheme({ group: "Bad Group", name: "Bad Theme", sets: ["Not Existing Set"] });
```

### Expected behavior

In both cases we should get a correct error message in devtools console:
`[PENPOT PLUGIN] Value not valid: Field group is invalid: 234 Name should be at least 1 character. Code: :error`
`[PENPOT PLUGIN] Value not valid: Field sets is invalid: The theme refers to some not existing sets: Not Existing Set. Code: :error`

### Actual behavior

In the second case the field and validation message is lost:
`[PENPOT PLUGIN] Value not valid: Field message is invalid: Code: :error`

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

The issue can be pinned down to [frontend.src.app.plugins.utils/error-messages](https://github.com/penpot/penpot/blob/develop/frontend/src/app/plugins/utils.cljs#L283). The called function `interpret-schema-problem`, in the first case it returns a nested structure that is flattened by the `(mapcat (comp seq val))`. In the other, the nesting is inexistent so the mapcat destroys the structure.

The `interpret-schema-problem` should be normalized so it always return a flattened structure, in any case of schema validation.

bitcompass:
When a plugin call fails malli validation, the frontend renders one "plugins.validation.message" line per error via
`app.plugins.utils/error-messages`, which reduces the explain via `csm/interpret-schema-problem` and then destructures each entry as `[field {:keys [message]}]` for translation.

That works only when the underlying malli error path has a single element. `interpret-schema-problem` calls `(assoc-in acc field ...)` where `field` can be a multi-element vector (e.g. `[:sets 0 :name]`). For single-element paths the resulting map is flat (`{:group {:message "..."}}`); for multi-element paths it is nested (`{:sets {0 {:name {:message "..."}}}}`). The destructure assumes the flat shape, so for a nested error the consumer reads:

    field   -> :sets
    message -> nil (the nested entry has no :message at the top level)

and the produced i18n line resolves to `Field sets is invalid: ` -- or, when several errors are merged together at the same outer key, to the user-facing `Field message is invalid` that the bug report calls out, because `:message` then becomes the field name of the deepest nested entry.

The original consumer carried a `#_(mapcat (comp seq val))` FIXME that hinted at the missing flattening but did not implement one, because the data shape produced by `interpret-schema-problem` is not uniform.

Fix
---

Add a private `flatten-error-map` helper inside `app.plugins.utils` that walks the error map produced by `interpret-schema-problem` and yields `[path message]` pairs where `path` is the dot-joined field path. Keywords use `(name k)`, strings pass through, anything else (such as numeric indices from vector positions in the malli path) is coerced via `str`. The recursion descends until it hits a leaf that carries `:message`, which matches what
`interpret-schema-problem` produces in every branch.

The producer side (`csm/interpret-schema-problem` in `common/src/app/common/schema/messages.cljc`) is left alone: it already has another consumer (`collect-schema-errors` + the form-validators pipeline) that depends on the keyed-by-field-path shape, so normalising it at the source would require auditing every validator. Flattening at the plugin consumer is the narrowest fix.

The FIXME comment is removed because the new helper supersedes it.

Tests
-----

`frontend-tests.plugins.utils-test` (new file, registered in `runner.cljs`) covers:

- flat single-segment paths (`{:group {:message "..."}}`)
- nested multi-segment paths (`{:sets {0 {:name {:message "..."}}}}`) -- the case from #9417
- mixed single- and multi-segment paths at the same explain
- mixed key types (keyword / string / numeric index)
- empty explain (no validation errors)

Closes #9417

### Steps to reproduce 

In a plugin, run:

```js
tokens?.addTheme({ group: 234, name: "Bad Theme" });
tokens?.addTheme({ group: "Bad Group", name: "Bad Theme", sets: ["Not Existing Set"] });
Before:


[PENPOT PLUGIN] Value not valid: Field group is invalid: 234 … . Code: :error
[PENPOT PLUGIN] Value not valid: Field message is invalid: . Code: :error
After:


[PENPOT PLUGIN] Value not valid: Field group is invalid: 234 Name should be at least 1 character. Code: :error
[PENPOT PLUGIN] Value not valid: Field sets is invalid: The theme refers to some not existing sets: Not Existing Set. Code: :error
````
### Checklist


madalenapmelo-kp:

niwinz:
Nice fix, thanks! :heart: 

I have marked it ready to merge, but we having some problems with our CI, so we waiting to fix the CI before.

PR #9488, opened 2026-05-11
:bug: Plugin API theme.addSet/removeSet accept proxy or set ID

### LINKED ISSUE #9369: Theme.addSet() and theme.removeSet() always throw "Field message is invalid" in plugin API
### Steps To Reproduce

1. Open a Penpot file and load a plugin with access to the Token API (`penpot.library.local.tokens`).

2. Create token sets and a theme:

```js

const core = tokens.addSet({ name: 'Core' });
const light = tokens.addSet({ name: 'Light' });

const theme = tokens.addTheme({ group: 'mode', name: 'Light' });
// Note: addTheme throws the same error, but the theme IS created.
```

3. Attempt to link a set to the theme:

```js
theme.addSet(core);
// → Error: Value not valid: Field message is invalid. Code: :error

theme.addSet(light);
// → Error: Value not valid: Field message is invalid. Code: :error
```

4. Check that no sets were linked:

```js
theme.activeSets;
// → [] (always empty)
```

5. Open the Penpot UI → Tokens panel → Themes tab. Manually add sets to the theme. **This works correctly.**

The same error occurs for `theme.removeSet(set)`.

### Additional failing variations

All produce the identical error:

```js
theme.addSet(core.id);                                    // pass string ID
theme.addSet({ id: core.id, name: core.name });           // pass plain object
tokens.themes[0].addSet(tokens.sets[0]);                  // direct chain
core.toggleActive(); theme.addSet(core);                   // activate set first
theme.toggleActive(); theme.addSet(core);                  // activate theme first
```

### Expected behavior

`theme.addSet(set)` should link the given `TokenSet` to the `TokenTheme`, making it appear in `theme.activeSets` and enabling proper theme switching between token sets.

The operation should work identically to the manual UI flow (Tokens → Themes → add sets).

### Actual behavior

`theme.addSet(set)` **always** throws:

```
Error: Value not valid: Field message is invalid. Code: :error
```

- `theme.activeSets` remains `[]` permanently.
- No sets are linked to the theme.
- The theme object itself is valid and can be read, activated, and removed — only `addSet()` and `removeSet()` fail.

- **Penpot version**: Latest (cloud)
- **Browser**: Chrome
- **OS**: macOS
- **Context**: Plugin Token API via MCP (Model Context Protocol), but the bug is reproducible in any standard Penpot plugin using `penpot.library.local.tokens`

ai-mountain:
###### Related Ticket

Closes #9369

###### Summary

`theme.addSet()` and `theme.removeSet()` on `TokenThemeProxy` always threw `"Value not valid: Field message is invalid. Code: :error"`, leaving `theme.activeSets` permanently empty even though the same operation worked from the UI.

The schemas were `[:tuple [:fn token-set-proxy?]]`, where `token-set-proxy?` checks `(get proxy (Symbol.for "penpot.reify:type"))` — a JS Symbol slot stamped by `obj/reify`. That slot doesn't survive the plugin sandbox boundary (structured-clone strips Symbol-keyed properties), so by the time validation runs in the host realm the predicate always returns `false`.

This change drops the broken predicate and looks up the set by ID instead, accepting any of the input shapes the issue lists:

* a `TokenSetProxy` → reads its `.id` getter
* a UUID string → used directly
* a plain `{ id }` object → reads its `id` property

A small `resolve-token-set` helper performs the ID extraction (`uuid/parse*` for safe parsing) and the `u/locate-token-set` lookup, so both `addSet` and `removeSet` share the same normalization path. `enable-set` / `disable-set` then run with the canonical set name from the lookup, preserving the existing nil-guard behavior.

No `:schema` is declared on these two methods anymore — invalid input (`42`, `null`, etc.) silently no-ops rather than throwing, matching the rest of the proxy's nil-guard pattern.

###### Steps to reproduce

Run in a plugin (or via MCP) with access to `penpot.library.local.tokens`:

```js
const core   = tokens.addSet({ name: 'Core' });
const theme  = tokens.addTheme({ group: 'mode', name: 'Light' });

theme.addSet(core);             // proxy
theme.addSet(core.id);          // UUID string
theme.addSet({ id: core.id });  // plain object
```

**Before:** every call throws `Value not valid: Field message is invalid. Code: :error`. `theme.activeSets` stays `[]`.

**After:** each call links the set to the theme; `theme.activeSets` contains the linked sets and reads back as `TokenSetProxy` instances.

`theme.removeSet(set)` and its other input shapes succeed the same way. `theme.addSet(42)` or `theme.addSet(null)` no-op silently. Manual add/remove from the Tokens → Themes UI is unchanged.

###### Checklist

- [x] Refactor any modified SCSS files following the refactor guide. _(N/A — no SCSS changes)_


madalenapmelo-kp:

frontend/src/app/plugins/tokens.cljs
```diff
@@ -353,6 +353,16 @@
...
 
+(defn- resolve-token-set-name
+  [file-id set-arg]
+  (or (when (and (some? set-arg) (not (string? set-arg)))
+        (let [n (obj/get set-arg "name")]
+          (when (string? n) n)))
+      (when-let [id (cond
+                      (string? set-arg) (uuid/parse* set-arg)
+                      (some? set-arg)   (some-> (obj/get set-arg "id") uuid/parse*))]
+        (some-> (u/locate-token-set file-id id) ctob/get-name))))
```
niwinz: I think we should preserve the error if no valid data is provided

RenzoMXD: I've fixed. Thanks.

filipsajdak:
Thanks for this @RenzoMXD — I hit exactly this bug driving Penpot's Token API from an MCP (`execute_code`) plugin, where `theme.addSet(core)` is called with a real `TokenSet` proxy, and your `resolve-token-set-name` (proxy `id` getter + name fallback) fixes our case cleanly.

While verifying locally against current `develop` (PR branch checked out via `gh pr checkout 9488`, `pnpm run build:test` + `node target/tests/test.js --focus frontend-tests.plugins.tokens-test`), I ran into one issue worth flagging:

**`token-theme-add-set-accepts-token-set-id` currently fails on develop** with:

```
TypeError: app.main.store.emit_BANG_.cljs$core$IFn$_invoke$arity$1 is not a function
```

`app.main.store/emit!` is multi-arity (`([] ) ([event]) ([event & events])`), so the call site in `tokens.cljs` compiles to `…emit_BANG_.cljs$core$IFn$_invoke$arity$1(event)`. The test's `with-redefs [st/emit! (fn [event] …)]` installs a *single-arity* anonymous fn, which doesn't carry `.cljs$core$IFn$_invoke$arity$1`, so the redef'd call throws — and the proxy's `on-error` handler turns it into a `[:error nil]` invalid entry, failing the assertions. (`token-theme-add-set-preserves-validation-errors` passes because it never reaches `emit!`.)

One-line fix — make the redef multi-arity:

```clojure
;; before
st/emit! (fn [event] (swap! emitted conj event))
;; after
st/emit! (fn ([event]    (swap! emitted conj event))
             ([event & _] (swap! emitted conj event)))
```

With that, the namespace is fully green here (9 tests / 32 assertions, 0 failures).

I also added a small regression test covering the **proxy argument** path specifically (the issue's primary repro / the MCP case), since the current tests cover a string id and the invalid-input path but not an actual `TokenSet` proxy:

```clojure
(t/deftest token-theme-add-remove-set-accepts-token-set-proxy
  (let [plugin-id "plugin-id"
        file-id   (uuid/next)
        theme-id  (uuid/next)
        set-id    (uuid/next)
        token-set (ctob/make-token-set :id set-id :name "Core")
        theme     (ctob/make-token-theme :id theme-id :group "mode" :name "Light")
        emitted   (atom [])
        invalid   (atom [])]
    (with-redefs [u/locate-token-set   (fn [_ id] (when (= id set-id) token-set))
                  u/locate-token-theme (fn [_ id] (when (= id theme-id) theme))
                  u/not-valid          (fn [_ code value] (swap! invalid conj [code value]))
                  dwtl/update-token-theme (fn [id theme] {:id id :theme theme})
                  st/emit!             (fn ([event]    (swap! emitted conj event))
                                          ([event & _] (swap! emitted conj event)))]
      (let [theme-proxy (ptok/token-theme-proxy plugin-id file-id theme-id)
            set-proxy   (ptok/token-set-proxy plugin-id file-id set-id)]
        (.addSet theme-proxy set-proxy)
        (t/is (= #{"Core"} (-> @emitted first :theme :sets)))
        (.removeSet theme-proxy set-proxy)
        (t/is (= #{} (-> @emitted second :theme :sets)))
        (t/is (empty? @invalid))))))
```

Happy to push these as a small commit to your branch if that's easier, or you can fold them in. Thanks again for fixing this!

Alotor:
Thanks for this! The resolution logic looks right and nice job adding the regression tests. 

We've been talking internally and there a few tweaks before we merge:

1. **TypeScript types are missing**. The whole point here is accepting a set id too, but plugins/libs/plugin-types/index.d.ts still says addSet(tokenSet: TokenSet): void / removeSet(tokenSet: TokenSet): void, so TS plugin folks can't actually use it without a type error. Can you bump both signatures (+ JSDoc) to TokenSet | string and drop a line in plugins/CHANGELOG.md?

2. **Let's keep the schema, just widen it.** Instead of removing :schema [:tuple [:fn token-set-proxy?]] and validating by hand, I'd rather extend it to accept a proxy or an id, e.g. [:tuple [:or [:fn token-set-proxy?] ::sm/uuid]]. That keeps validation where the rest of the API does it and saves us the manual cond + u/not-valid dance.

3. **Rename resolve-token-set-name → resolve-token-set and have it return a set.** If it's a proxy, just return it as-is; if it's a string, locate and return the set. Callers can grab the name off that. Keeps it nice and symmetric.

4. **Let's not accept a plain object.** Only a real TokenSet proxy or an id string — can you drop that fallback and its test case? Keeps the contract simple.

5. **Remove the /CHANGES.md entry.** It's no longer necesary (careful! not to mistake with the one inside plugins mentioned on 1., that is still used)

Once the schema handles validation and we only allow proxy/id, the invalid-input handling should get a lot simpler. 

And, please, while you're at it squash the commits with the same message.

Alotor:
Thanks for your contribution!

We’re going to merge your changes as part of this PR: https://github.com/penpot/penpot/pull/10503

I've made the changes that I've mentioned by amending your PR.

I’m going to close this PR, but your contribution is included in the other one. Thanks again!

PR #9501, opened 2026-05-11
:bug: Fix two plugin error i18n keys broken by leading whitespace in en.po

MilosM348:
## Summary

Fix two plugin error i18n keys in `frontend/translations/en.po` that were
silently broken by a leading space before `msgid`. Per the gettext format
spec, `msgid` lines must start at column 0; leading-whitespace `msgid`
lines are treated as continuations of the previous entry, so the keys
were never registered as translation entries.

Affected keys (both shown in the plugin error UI to end users):

- `workspace.plugins.error.manifest` — surfaces in the plugins panel when
  a plugin manifest cannot be parsed
  (`frontend/src/app/main/ui/workspace/plugins.cljs`).
- `workspace.plugins.error.need-editor` — surfaces in the main menu
  tooltip and as a `ntf/warn` notification when a non-editor user tries
  to launch an edit-permission plugin
  (`frontend/src/app/main/data/plugins.cljs`,
  `frontend/src/app/main/ui/workspace/main_menu.cljs`,
  `frontend/src/app/main/ui/workspace/plugins.cljs`).

Without this fix, end users see the raw key string (e.g.
`workspace.plugins.error.manifest`) instead of the intended English text.
This is the same class of bug as #9320.

## Fix

Remove the leading whitespace from the two `msgid` lines (and the
adjacent `msgstr` for the manifest entry) so they parse as proper
gettext entries. Same template as the `labels.open` fix from #9320.

## Verification

- `grep -P '^\s+msgid' frontend/translations/en.po` returns 0 matches
  after the fix (was 2 before).
- Both keys now appear as standalone entries during a normal `gettext`
  parse pass.
- Other 47 language `.po` files in `frontend/translations/` already
  reference these keys with proper translations and are unaffected.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@
 - Fix plugin API `LibraryTypography.remove()` failing with a UUID assertion error [Github #8223](https://github.com/penpot/penpot/issues/8223)
 - Fix MCP SSE sessions leaking memory on zombie connections by adding inactivity timeout parity with Streamable HTTP sessions (by @bitloi) [Github #9432](https://github.com/penpot/penpot/issues/9432)
 - Fix missing `labels.open` translation (by @MilosM348) [Github #9320](https://github.com/penpot/penpot/pull/9320)
+- Fix two plugin error i18n keys broken by leading whitespace before `msgid` in `en.po` (by @MilosM348)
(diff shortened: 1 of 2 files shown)


madalenapmelo-kp:

PR #9441, opened 2026-05-07
:recycle: Migrate template-item to modern component syntax

Dexterity104:
### Related Ticket

Part of #9260 

### Summary

Migrates the `template-item` component in `frontend/src/app/main/ui/onboarding/templates.cljs` to the modern Rumext form. This component renders a single template card (image, title, and "+ Add to drafts" action) inside the onboarding templates modal.

Changes in this PR:

* Renamed `template-item` to `template-item*`.
* Updated both internal callsites inside `onboarding-templates-modal` from `[:& template-item {...}]` to `[:> template-item* {...}]`.

The sibling `onboarding-templates-modal` component is intentionally left untouched in this PR. It carries `::mf/register modal/components` and is excluded from this migration pass per the migration guide; converting modal-registered components is handled separately.

The leaf component already used keyword destructuring and carried no `::mf/wrap-props false` metadata, no `?`-suffixed boolean props, and no external `:refer` imports, so this is a clean Case A migration with no other adjustments required. `cljfmt` was run and produced no further changes.

### Steps to reproduce

1. Sign up for a new account so the onboarding flow triggers the templates modal, or open the modal manually via `:onboarding-templates`.
2. Confirm both cards ("Penpot Design System" and "Material Design Kit") render with their image, title, and "+ Add to drafts" action.
3. Click "+ Add to drafts" on either card and verify that the import modal opens with the correct template name and project.
4. While the download is in flight, verify that the action label switches to "Fetching...".

### Checklist

- [x] Choose the correct target branch; use `develop` by default.
- [x] Provide a brief summary of the changes introduced.
- [x] Add a detailed explanation of how to reproduce the issue and/or verify the fix, if applicable.
- [ ] Include screenshots or videos, if applicable.
- [ ] Add or modify existing integration tests in case of bugs or new features, if applicable.
- [ ] Refactor any modified SCSS files following the refactor guide.
- [ ] Check CI passes successfully.
- [ ] Update the `CHANGES.md` file, referencing the related GitHub issue, if applicable.

PR #9363, opened 2026-05-05
:bug: Log template download failures via console.error

Dexterity104:
### Related Ticket

N/A

### Summary

The error subscriber on the onboarding *Templates* download stream was emitting failures through `js/console.log`. Because that logs at info level, the failure was hidden from the *Errors* filter in browser devtools and the template silently failed to import: `downloading?` was reset, the modal never opened, and nothing appeared as an error in the console.

The handler now uses `js/console.error`, matching the pattern already used elsewhere (for example the dashboard font download handler at `frontend/src/app/main/ui/dashboard/fonts.cljs:368`).

### Steps to reproduce

1. Open Penpot and navigate to the onboarding *Templates* step.
2. Force the template download request at `frontend/src/app/main/ui/onboarding/templates.cljs:41` to fail (for example, by blocking the configured template host in the browser devtools network panel).
3. Click a template to start the download.
4. Open the browser console and check the *Errors* filter.

Before the fix, the failure appeared as an info log under `console.log` and was not visible under the *Errors* filter. After the fix, the failure appears as a proper console error.

### Checklist

- [x] Choose the correct target branch; use `develop` by default.
- [x] Provide a brief summary of the changes introduced.
- [x] Add a detailed explanation of how to reproduce the issue and/or verify the fix, if applicable.
- [ ] Include screenshots or videos, if applicable.
- [ ] Add or modify existing integration tests in case of bugs or new features, if applicable.
- [ ] Refactor any modified SCSS files following the refactor guide.
- [x] Check CI passes successfully.
- [ ] Update the `CHANGES.md` file, referencing the related GitHub issue, if applicable.

PR #9504, opened 2026-05-11
:bug: Fix onboarding template spinner stuck after failed template download

### LINKED ISSUE #9931: Fix onboarding template spinner stuck after failed template download
### Description

When onboarding as a new user, clicking "+ Add to drafts" on a template tile triggers a download of the template archive. If that download fails (e.g., network outage, CORS rejection, or server error), the tile gets stuck displaying "Fetching..." indefinitely. The user cannot retry the download without reloading the entire page.

### Steps to reproduce

1. Open the onboarding *Templates* modal as a fresh user.
2. In DevTools, throttle the network or block requests to `*.penpot.app/github/penpot-files/*` to force the template fetch to fail.
3. Click `+ Add to drafts` on any template tile.
4. Observe that the tile switches to `Fetching...` and never returns to its clickable state.

### Expected behavior

After a failed template download, the tile should return to its `+ Add to drafts` state so the user can retry without reloading the page.

### Affected versions

Unreleased (2.17.0)

jeffrey701:
### Related Ticket

Closes #9931

### Summary

`template-item` in `frontend/src/app/main/ui/onboarding/templates.cljs` fetches a template archive over HTTP and tracks an in-flight `downloading?` atom so the click target swaps to a `Fetching...` label. The current `rx/subs!` only resets the atom in the **on-complete** handler:

```clojure
(->> (http/send! {:method :get :uri link :response-type :blob :mode :no-cors})
     (rx/subs! (fn [{:keys [body] :as response}] ...)
               (fn [error] (js/console.log "error" error))
               (fn [] (reset! downloading? false))))
```

When the observable errors (network failure, CORS rejection, 4xx/5xx surfaced as a stream error, etc.), the on-complete callback never fires. As a result the item stays in `Fetching...` forever and the user has no way to retry without reloading the page.

This PR resets `downloading?` to `false` from the on-error handler too, so a failed download returns the item to its clickable `+ Add to drafts` state. The behavior on the success and complete paths is unchanged.

### Notes for reviewers

- Single one-line addition in the error handler; no behavior change on the happy path.
- This is intentionally **complementary** to PR #9363 (which upgrades `console.log` -> `console.error` for the same error handler) and they should rebase trivially against each other; my change adds a sibling line and does not touch the log line.
- Not touching `templates.cljs` for the modern-syntax migration in PR #9441.

### How to validate manually

1. Open the onboarding *Templates* modal as a fresh user.
2. In DevTools, throttle network or block `*.penpot.app/github/penpot-files/*` to force the template fetch to error.
3. Click `+ Add to drafts` on a template tile.
   - Before this patch: tile flips to `Fetching...` and stays there until the page is reloaded.
   - After this patch: tile flips to `Fetching...`, then returns to `+ Add to drafts` once the error reaches the subscriber, and a second click can be retried.

### Checklist

- [x] Bug fix
- [x] Updated `CHANGES.md` under `2.17.0 (Unreleased)` with a one-line bullet matching the format of nearby community fixes.
- [x] Targets `develop`.


madalenapmelo-kp:

PR #9512, opened 2026-05-11
:bug: Highlight first matching font when searching the font picker

### LINKED ISSUE #3204: Highlight first found font at the fonts list when searching

At the design workspace, the first found font at the fonts list when searching should be highlighted/focused so it can be easily applied.

At the workspace fonts list, when searching for a font:

1. The first result should be highlighted using the same style as when focused (the one with the tick).
2. Pressing enter selects the font highlighted by the search and closes the fonts list.

image

ai-mountain:
## Related Ticket

Closes #3204

## Summary

In the workspace **font picker**, typing into the search box currently narrows the list but leaves `selected` pinned to whatever font the text element had when the picker opened. That breaks two of the picker's existing behaviours:

* The filtered list shows **no tick mark** on any row (the previously-selected font is filtered out and no replacement is chosen).
* Pressing **Enter** calls `on-close` without applying anything — and since the live-preview effect (`mf/with-effect [@selected] (on-select @selected)`) only fires on `selected` changes, the canvas keeps the original font.

Whenever the filtered `fonts` list changes, if the current `@selected` is no longer in it the effect resets to the first match. The existing live-preview effect immediately chains: `on-select` fires → parent's `change-font` runs → the text element previews the first match in real-time. Pressing **Enter** then closes the picker; the font is already committed because the live preview already applied it.

Both behaviours requested in #3204 are satisfied with one effect (no change to the keydown handler is needed — the live preview was already doing the "Enter commits" work, but it was unreachable for the search case because `selected` never moved).

## Edge cases handled

* Search matches the current selection (e.g. `wor` with Work Sans current) → `selected` stays, tick stays, no jarring preview change.
* Search yields no matches (e.g. `zzzz`) → `(when (seq fonts)` skips the reset, last preview stays applied, no flicker.
* User clears the search → larger filtered list still contains `selected` → no reset, preserves prior exploration.
* Arrow-key navigation still works unchanged — `select-next` / `select-prev` already keep `selected` inside the filtered list.

## Steps to reproduce

1. Open a workspace file, add a text element (Text tool `T`, drag a frame, type a few characters, `Esc`).
2. Select the text element. In the right sidebar **Design** tab → **Typography** section, click the font name to open the font picker.
3. In the search box, type a few characters that **do not** match the currently selected font (e.g. `mono` when the default font is "Work Sans").

**Before:**
* The filtered list has no tick mark anywhere.
* Press Enter → picker closes; the text on canvas still has the original font.

**After:**
* The first matching font in the filtered list shows the tick mark and is previewed live on the canvas.
* Press Enter → picker closes; the previewed font stays applied.

## Screenshots

Before:

https://github.com/user-attachments/assets/412dbea4-56fa-413f-841a-659c9d4cadc3

After:

https://github.com/user-attachments/assets/3243c524-5940-4cd7-a8a1-df23bfef8cce

## Checklist

- [x] Refactor any modified SCSS files following the refactor guide. _(N/A — no SCSS changes)_


madalenapmelo-kp:

RenzoMXD:
Hi, @niwinz
This is a simple UI enhancement. Could you review my PR when you have a sec.
Thanks.

myfunnyandy:
Hi @RenzoMXD,

I was able to test it and found out that the font selector is not positioning itself properly. The search field's top position varies depending on the viewport height and can even be overlapped by the navigation tabs. The width adjustments aren't correct either. This should keep working as in design.penpot.app. Thanks!

image

NatachaMenjibar:
@niwinz @myfunnyandy I think we can approve this pr since the issue @myfunnyandy pointed out is not related to this enhacement and it's logged here https://github.com/penpot/penpot/issues/10083

myfunnyandy:
@NatachaMenjibar fine by me :+1:

PR #9551, opened 2026-05-12
:tada: Add chunked upload support for font variants

### LINKED ISSUE #9560: Support uploading larger font files via chunked upload
### Description

Uploading large font files (e.g. variable fonts or font families with many weights) could fail due to size limits in the existing upload mechanism. The font upload process now supports chunked uploads, enabling larger font files to be uploaded reliably.

### Use case

Designers working with large font files (variable fonts, comprehensive font families) can now upload them without hitting size limits, making font management more robust.

### Affected versions

2.15.1

niwinz:
## Summary

- **Font variant upload**: Extend the existing `create-font-variant` RPC endpoint to support the standardized chunked upload protocol (previously only available for media and binfile uploads), removing the previous client-side chunking approach and enabling larger font files.
- **File media logging**: Add structured logging to the file media assembly and chunk processing paths for better observability.

### Backend
- Add `:font-max-file-size` config default (30 MiB) and schema entry
- Add `validate-font-size!` in `media.clj` (mirrors `validate-media-size!`)
- Extend `schema:create-font-variant` to accept either `:data` (legacy) or `:uploads` (chunked sessions)
- Add `prepare-font-data-from-uploads`: assembles chunked sessions, validates type+size
- Add `prepare-font-data-from-legacy`: normalises legacy byte/chunk entries, writes to tempfile, validates
- Add structured logging in `create-font-variant` with `:size`, `:mtypes`, `:elapsed`
- Add logging in media chunk assembly for better debuggability

### Frontend
- `upload-blob-chunked` accepts a per-caller `:chunk-size` option
- Add `font-upload-chunk-size` (10 MiB) and `upload-font-variant` fn
- Dashboard fonts now uses chunked upload path instead of raw RPC call
- `process-upload` stores raw `ArrayBuffer` instead of chunking client-side

### Tests
- 25 tests / 224 assertions covering all three upload paths, size validation, and media type validation

## Related
- Follows the pattern established in PR #8909

Fixes #9560

### DIFF
--- a/backend/src/app/rpc/commands/media.clj
+++ b/backend/src/app/rpc/commands/media.clj
@@ -8,6 +8,7 @@
   (:require
    [app.common.data :as d]
    [app.common.exceptions :as ex]
+   [app.common.logging :as l]
    [app.common.schema :as sm]
    [app.common.time :as ct]
    [app.common.uuid :as uuid]
@@ -58,8 +59,8 @@
   (db/run! cfg (fn [{:keys [::db/conn] :as cfg}]
                  ;; We get the minimal file for proper checking if
                  ;; file is not already deleted
-                 (let [_     (files/get-minimal-file conn file-id)
-                       mobj  (create-file-media-object cfg params)]
+                 (let [_      (files/get-minimal-file conn file-id)
+                       mobj   (create-file-media-object cfg params)]
 
                    (db/update! conn :file
                                {:modified-at (ct/now)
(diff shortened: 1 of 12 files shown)


backend/src/app/rpc/commands/media.clj
```diff
@@ -149,20 +150,49 @@
...
-
-    (db/exec-one! conn [sql:create-file-media-object
-                        (or id (uuid/next))
-                        file-id is-local name
-                        (:id image)
-                        (:id thumb)
-                        (:width result)
-                        (:height result)
-                        (:mtype result)])))
+   {:keys [id file-id is-local name content from-url from-chunks]}]
```
hirunatan: Being a local function, I suppose we can name these `from-url?` and `from-chunks?` and don't need the `(true? xxx)`. What do you think?

common/src/app/common/media.cljc
```diff
@@ -12,8 +12,8 @@
 (def font-types
   #{"font/ttf"
     "font/woff"
-    "font/otf"
-    "font/opentype"})
+    "font/woff2"
+    "font/otf"})
```
hirunatan, line 16: Is this correct? In the description you didn't mention you are adding a font type and removing other one.

backend/src/app/rpc/commands/media.clj
niwinz: agree

common/src/app/common/media.cljc
niwinz: The font/opentype is font/otf
This set should match the accepted ones on the backend.
So yes, this looks ok

PR #9564, opened 2026-05-12
:bug: Fix plugin API token application defects

### LINKED ISSUE #9561: ## Token application API: `applyToShapes` throws "Doesn't support name: 0", `shape.tokens` setter is silent no-op
**Describe the bug**

Two documented methods for applying design tokens to shapes are non-functional:
1. **`token.applyToShapes([shape])`** throws `Error: "Doesn't support name: 0"` regardless of shape type (Board, Text, Rectangle) or property specification (`['fill']`, `['all']`, or omitted).
2. **`shape.tokens.fill = 'tokenName'`** silently discards the assignment. The `tokens` property is a readonly getter — writes don't throw but are never persisted.
The only working method is `penpot.selection = [shape]; token.applyToSelected(['fill'])` (selection must be an array).

**Steps to Reproduce**

1. Create a color token:

```js
const set = penpot.library.local.tokens.sets.find(s => s.name === 'Light');
await set.addToken({ type: 'color', name: 'colors.primary', value: '#3525CD' });
const token = set.tokens.find(t => t.name === 'colors.primary');
```
2. Attempt applyToShapes:
```js
token.applyToShapes([shape]);           // Error: "Doesn't support name: 0"
token.applyToShapes([shape], ['fill']); // Same error
token.applyToShapes([shape], ['all']);  // Same error
```
3. Attempt direct setter:
```js
shape.tokens.fill = 'colors.primary';
console.log(shape.tokens.fill); // undefined — silently discarded
```
**Expected behavior**

- token.applyToShapes([shape], ['fill']) should apply the token to the shape's fill color property.
- shape.tokens.fill = 'tokenName' should bind the token (or throw if unsupported), with the value resolving based on currently active token sets and themes.

**Working workaround**

The only functional method found:
```js
penpot.selection = [shape];  // Must be an ARRAY, not a single shape
token.applyToSelected(['fill']);
```

Rene0422:
### Related Ticket

Fixes [#9561](https://github.com/penpot/penpot/issues/9561) — Token Application API Defects (plugin API).

### Summary

Two coupled defects made the plugin design-token API effectively unusable from JavaScript:

1. `token.applyToShapes([shape], ["fill"])` threw `"Doesn't support name: 0"`.
The `[:vector [:fn shape-proxy?]]` schema failed for hardened SES shape proxies, and then `error-messages` crashed calling `(name 0)` on the resulting tuple-index path — masking the real validation error.

- `applyToShapes` now validates manually with `array?` + `shape-proxy?` (the same pattern `penpot.selection` and `replaceColor` already use).
- `error-messages` has a new `format-field` helper that renders integer / vector paths as strings instead of throwing.

2. `shape.tokens.fill = "colors.primary"` was silently discarded.
`:tokens` exposed only a getter that returned a fresh plain object on every access, so per-property assignments vanished.

- The getter now returns a `js/Proxy` whose `set` trap looks up the named token in the file's active token sets and dispatches the existing `toggle-token` action. The trap always returns `true` so strict-mode SES plugins don't get TypeErrors on invalid input — validation failures surface through `u/not-valid` like the rest of the plugin API.

Also:

Extended `token-attr-plugin->token-attr` to kebab-case string input so camelCase property names from the read side of `shape.tokens` (e.g. `"borderRadiusTopLeft"`) round-trip into a working setter call.
Dropped readonly from ShapeBase.tokens in the TypeScript plugin types and documented the new write semantics.
Added regression coverage in `frontend/test/frontend_tests/plugins/tokens_test.cljs`: unit tests for `format-field` and the camelCase normalisation, plus async end-to-end tests via `tohs/run-store-async` that exercise `applyToShapes`, the `shape.tokens` setter, and the unknown-token rejection path.

### Steps to reproduce 

Inside any plugin running against a file that has a color token `colors.primary` (value `#3525CD`) in an active set:

`const shape = penpot.selection[0];`
`const token = penpot.library.local.tokens`
`  .getSetById(setId)`
`  .getTokenById(tokenId);`

`// Before this PR: throws 'Doesn't support name: 0'.`
`// After this PR: token is applied to the shape's fill.`
`token.applyToShapes([shape], ["fill"]);`

`// Before this PR: silently discarded, shape.tokens.fill stays undefined.`
`// After this PR: the named token is resolved against the active sets`
`// and applied; reading shape.tokens.fill returns "colors.primary".`
`shape.tokens.fill = "colors.primary";`

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@
 - Harden Nginx responses with standard security headers and hide upstream `X-Powered-By` headers
 - Expose Source Sans Pro semibold (weight 600) variants in the builtin fonts list, matching the bundled font assets and CSS @font-face declarations [Github #7378](https://github.com/penpot/penpot/issues/7378)
 - Fix plugin API `shape.fills` and `shape.strokes` arrays being read-only [Github #8357](https://github.com/penpot/penpot/issues/8357)
+- Fix plugin API `token.applyToShapes()` failing with "Doesn't support name: 0" and `shape.tokens` per-property assignment being silently discarded [Github #9561](https://github.com/penpot/penpot/issues/9561)
 - Fix `get-profile` masking transient DB errors as anonymous user (by @jack-stormentswe) [Github #9253](https://github.com/penpot/penpot/issues/9253)
 - Fix `Ctrl+'` "Show guides" shortcut on non-US keyboard layouts by matching the physical key location (by @RenzoMXD) [Github #8423](https://github.com/penpot/penpot/issues/8423)
 - Fix lost-update race on `team.features` during concurrent file creation (by @web-dev0521) [Github #9197](https://github.com/penpot/penpot/issues/9197)
--- a/frontend/src/app/plugins/shape.cljs
+++ b/frontend/src/app/plugins/shape.cljs
@@ -31,6 +31,7 @@
    [app.common.types.shape.radius :as ctsr]
    [app.common.types.shape.shadow :as ctss]
    [app.common.types.text :as txt]
+   [app.common.types.tokens-lib :as ctob]
    [app.common.uuid :as uuid]
    [app.config :as cf]
    [app.main.data.exports.wasm :as wasm.exports]
@@ -1357,12 +1358,52 @@
               (let [applied-tokens
                     (-> (u/locate-shape file-id page-id id)
                         (get :applied-tokens)
-                        (applied-tokens-plugin->applied-tokens))]
-                (reduce
-                 (fn [acc [prop name]]
-                   (obj/set! acc (json/write-camel-key prop) name))
-                 #js {}
-                 applied-tokens)))}
+                        (applied-tokens-plugin->applied-tokens))
+
+                    target
+                    (reduce
+                     (fn [acc [prop name]]
+                       (obj/set! acc (json/write-camel-key prop) name))
+                     #js {}
+                     applied-tokens)]
+                ;; Wrap the plain object in a Proxy so that
+                ;; `shape.tokens.fill = "tokenName"` actually applies
+                ;; the token instead of silently failing (#9561).
+                ;; Reads continue to fall through to the underlying
+                ;; object built from the current shape state.
+                ;; The handler always returns `true` so strict-mode
+                ;; assignment from SES-sandboxed plugins doesn't throw
+                ;; a TypeError on invalid input. Validation failures are
+                ;; surfaced through the plugin's standard `u/not-valid`
+                ;; channel (console / throw-validation-errors flag).
+                (js/Proxy.
+                 target
+                 #js {:set
+                      (fn [target-obj prop value]
+                        (let [attr (token-attr-plugin->token-attr prop)
+                              tokens-lib (u/locate-tokens-lib file-id)
+                              token (when (and tokens-lib (string? value))
+                                      (-> (ctob/get-tokens-in-active-sets tokens-lib)
+                                          (get value)))]
+                          (cond
+                            (not (token-attr? prop))
+                            (u/not-valid plugin-id :tokens
+                                         (str "unknown token property: " prop))
+
+                            (nil? token)
+                            (u/not-valid plugin-id :tokens
+                                         (str "token not found in active sets: " value))
+
+                            :else
+                            (do
+                              (st/emit!
+                               (-> (dwta/toggle-token {:token token
+                                                       :attrs #{attr}
+                                                       :shape-ids [id]
+                                                       :expand-with-children false})
+                                   (se/add-event plugin-id)))
+                              (obj/set! target-obj prop value)))
+                          true))})))}
 
            :applyToken
            {:enumerable false
--- a/frontend/src/app/plugins/tokens.cljs
+++ b/frontend/src/app/plugins/tokens.cljs
@@ -55,13 +55,16 @@
 
   Accepts either a Clojure keyword (the canonical form, e.g. `:r1`,
   `:fill`) or a string (the natural shape that arrives from a JS plugin
-  call such as `shape.applyToken(token, [\"fill\"])`). Converts strings
-  to keywords first, then maps verbose plugin-side aliases (e.g.
-  `:border-radius-top-left`) to their internal short form (e.g. `:r1`).
-  Inputs that are already in canonical form (`:r1`, `:fill`, `\"fill\"`,
-  …) pass through unchanged."
+  call such as `shape.applyToken(token, [\"fill\"])`). String input is
+  kebab-cased so that camelCase property names produced by the read
+  side of `shape.tokens` (e.g. `\"borderRadiusTopLeft\"` via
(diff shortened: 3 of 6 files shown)


madalenapmelo-kp:

niwinz:
Hey @Rene0422 
Thanks for contribution, and sorry for so late reply, we are looking PR's in order and mixing that with other priorities.

This PR add several distinct improvements. And  i think one of them is conflicting with already implemented (merged in other pr) enhancement on error reporting.

If you think that the enhancement (at least partially) still valid, please, can you rebase the changes and resolve conflicts? Thanks.

niwinz:
marking as draft meanwhile

CHANGES.md
```diff
@@ -19,6 +19,7 @@
 - Harden Nginx responses with standard security headers and hide upstream `X-Powered-By` headers
 - Expose Source Sans Pro semibold (weight 600) variants in the builtin fonts list, matching the bundled font assets and CSS @font-face declarations [Github #7378](https://github.com/penpot/penpot/issues/7378)
 - Fix plugin API `shape.fills` and `shape.strokes` arrays being read-only [Github #8357](https://github.com/penpot/penpot/issues/8357)
+- Fix plugin API `token.applyToShapes()` failing with "Doesn't support name: 0" and `shape.tokens` per-property assignment being silently discarded [Github #9561](https://github.com/penpot/penpot/issues/9561)
```
filipsajdak, line 22: If this PR is split per the request above, this changelog line should split into two bullets (one for the `applyToShapes` validation fix, one for `shape.tokens` write support). Also worth adding `(by @Rene0422)` to match the changelog attribution convention.

frontend/src/app/plugins/utils.cljs
```diff
@@ -276,13 +276,29 @@
   (let [s (set values)]
     (if (= (count s) 1) (first s) "mixed")))
 
+(defn- format-field
```
filipsajdak, line 279: Heads-up: on current `develop`, `error-messages` was rewritten to use `flatten-error-map` and no longer calls `(name field)`, so the `"Doesn't support name: 0"` crash this addresses is already fixed upstream. This `format-field` hunk (and the `format-field` / `error-messages-does-not-crash-on-integer-field` tests) will conflict on rebase -- I'd drop them. This is the piece that conflicts with the merged error-reporting work.

frontend/src/app/plugins/tokens.cljs
```diff
@@ -55,13 +55,16 @@
...
+  kebab-cased so that camelCase property names produced by the read
+  side of `shape.tokens` (e.g. `\"borderRadiusTopLeft\"` via
+  `json/write-camel-key`) round-trip correctly. After normalisation,
+  verbose plugin-side aliases (e.g. `:border-radius-top-left`) are
+  mapped to their internal short form (e.g. `:r1`). Inputs already in
+  canonical form (`:r1`, `:fill`, `\"fill\"`, …) pass through
+  unchanged."
   [k]
-  (let [k (cond-> k (string? k) keyword)]
+  (let [k (cond-> k (string? k) json/read-kebab-key)]
```
filipsajdak, line 67: Good fix. Since `token-attr?` delegates to this fn, switching to `read-kebab-key` also makes `token-attr?` / `applyToken` accept camelCase property names -- not just the new setter. Worth calling out in the commit body so the broader effect is visible.

frontend/src/app/plugins/tokens.cljs
```diff
@@ -203,11 +206,33 @@
...
+     ;; spurious validation failure. The error formatter then crashed
+     ;; with "Doesn't support name: 0" on the tuple-index path,
+     ;; obscuring the real cause and making the method completely
+     ;; unusable (#9561). Manual validation mirrors the pattern used by
+     ;; `penpot.selection` / `replaceColor` and works reliably for
+     ;; JS-array inputs.
      :fn (fn [shapes attrs]
-           (apply-token-to-shapes plugin-id file-id set-id id (map #(obj/get % "$id") shapes) attrs))}
+           (cond
+             (or (not (array? shapes)) (not (every? shape-proxy? shapes)))
```
filipsajdak, line 221: This `array?` + `every? shape-proxy?` guard now mirrors the pattern in `penpot.selection` and `replaceColor`. Consider extracting a shared `valid-shape-proxy-array?` helper so the three stay in sync (optional). Minor: `(every? shape-proxy? [])` is `true`, so an empty array is accepted and no-ops -- fine, just intended behaviour worth a one-line comment.

frontend/src/app/plugins/shape.cljs
```diff
@@ -1357,12 +1358,52 @@
...
+                ;; a TypeError on invalid input. Validation failures are
+                ;; surfaced through the plugin's standard `u/not-valid`
+                ;; channel (console / throw-validation-errors flag).
+                (js/Proxy.
+                 target
+                 #js {:set
+                      (fn [target-obj prop value]
+                        (let [attr (token-attr-plugin->token-attr prop)
+                              tokens-lib (u/locate-tokens-lib file-id)
+                              token (when (and tokens-lib (string? value))
```
filipsajdak, line 1385: Minor: `token` (and the `get-tokens-in-active-sets` scan) is evaluated eagerly even when `prop` is invalid; moving it into the `nil? token` / `:else` branch avoids the wasted lookup on the reject path. Also, a non-string `value` currently falls through to "token not found in active sets" -- a dedicated "token name must be a string" message would read friendlier. Both optional.

filipsajdak (commented):
Took a look while going through plugin/token PRs (disclosure: this review was prepared with AI assistance as part of the Penpot MCP self-improvement initiative). The two core fixes -- manual `applyToShapes` validation and the `shape.tokens` write proxy -- look solid and the added test coverage is good. Echoing @niwinz's earlier note, the blocker is structural, and I verified the specifics against current `develop`:

- The `utils.cljs` error-formatter change is now **superseded on `develop`**: `error-messages` was rewritten to use `flatten-error-map` and no longer calls `(name field)`, so the `"Doesn't support name: 0"` crash is already fixed. This hunk (plus the `format-field` / `error-messages-does-not-crash-on-integer-field` tests) will conflict on rebase -- I'd drop it. This looks like the part that "conflicts with merged error-reporting" mentioned above.
- The remaining work is two independent concerns in a single empty-body commit. Splitting them keeps review/bisect tractable and lets the lower-risk validation fix land even if the setter needs more discussion:
  - **`:bug:`** -- `applyToShapes` manual validation (`tokens.cljs`) + its e2e test.
  - **`:sparkles:`** -- `shape.tokens` per-property assignment: `shape.cljs` proxy + `token-attr-plugin->token-attr` kebab-case + `index.d.ts` + the setter/camelCase tests + CHANGES.md. This adds a new *write* capability, so per CONTRIBUTING it's worth its own PR with a quick discussion-first nod.

A couple of small inline notes below. Nice work overall -- the always-return-`true` SES handling and the "bypass the schema pipeline" rationale are both the right calls.

PR #9575, opened 2026-05-13
:bug: Re-render after font load so first paint uses the chosen font

### LINKED ISSUE #9574: Text shapes render with Times New Roman fallback on initial load until hover repaints the tile cache
### Steps To Reproduce

1. Open or refresh a Penpot file that contains text shapes using a non-default font (any Google Font such as `Inter` or `Roboto Slab`, or a custom uploaded font).
2. Watch the workspace canvas the moment the file finishes loading **without moving the mouse**.
3. Then move the mouse over any shape on the canvas.

### Expected behavior

Every text shape is rendered with the font it was authored in, immediately on the first paint after the file finishes loading. No mouse interaction should be required to display the correct font.

### Actual behavior

On the first paint, text shapes are drawn with a Times-New-Roman-style serif (Skia's fallback when the requested font has not yet been stored in the WASM font registry). Once the mouse moves over the canvas — which triggers a hover halo / `set-modifiers` invalidation and forces the affected tiles to be re-painted — the correct font appears. Without that mouse interaction the cached fallback rendering can persist indefinitely.

### Screenshots or video

Easy to reproduce on any document with a Google or custom font; user-reported on Firefox and Chrome. No clip attached in this initial report.

- OS: Windows 11; also reproduced on Linux
- Browser & version: Firefox 149, Chrome 130 — same behavior

_n/a_

- Host: self-hosted, develop branch built mid-May 2026 (penpot 2.15.x line).

### Frontend Stack Trace

_No error thrown._

### Backend Stack Trace

_n/a_

#### Diagnosis

Penpot's WASM renderer caches each tile after it paints. On initial workspace load `set-objects-async` chunks through shape construction and issues a render via `request-render "set-objects-complete"` **before** any of the pending font TTFs have been fetched and stored into WASM via `_store_font`. The tiles therefore contain text rendered with Skia's fallback font.

When the pending font-fetch rx streams finally resolve, the code in `set-objects-async`'s background block does try to call `update-text-layouts` + `request-render "images-loaded"`, but this path is fragile — anything along the chain that skips or coalesces the redraw (page transition snapshot still up, `pending-render` flag previously set, etc.) leaves the cached fallback paint in place. The next invalidation that does invalidate text tiles is typically a hover, which goes through `set-modifiers` and forces ancestor tile rebuild — which is exactly why the bug presents as "fonts only become correct when I move my mouse over the canvas."

#### Proposed fix

A global subscriber on `:font-loaded` data-events that calls `update-text-layout` for every text shape on the current page (touching them so their tiles get marked dirty in the WASM renderer) and then `request-render`s. The render is rAF-coalesced, so multiple font loads in the same frame produce a single paint. PR follows.

fluffyspace:
### Related Ticket

Fixes #9574

### Summary

When the WASM tile renderer paints text shapes before their custom / Google fonts have been stored into the WASM font store, tiles get cached with Skia's fallback (Times-New-Roman-style serif). `set-objects-async` does try to redraw after the pending fetches resolve, but that path is fragile — any flake along the chain leaves the cached fallback paint in place until an unrelated tile invalidation (most commonly hovering) clears it.

This adds a global subscriber on `:font-loaded` data-events in `app.render-wasm.api`. Whenever one fires, it re-runs `update-text-layout` for every text shape on the current page (which touches them so the WASM renderer marks the tiles dirty), then `request-render`s. The request is rAF-coalesced, so a burst of font loads at startup produces a single paint.

Also reorders the `:font-loaded` emit in `store-font-buffer` to fire **after** the `_store_font` WASM call, so any subscriber that re-renders sees the font already present in the WASM font store.

### Steps to reproduce

1. Run a self-hosted Penpot built from this branch.
2. Open a file that contains text shapes using a non-default font (Google or custom).
3. Refresh the page and watch the text on the first paint. Before the patch the text appears in Times-New-Roman-style serif and only "fixes" itself when the mouse moves over the canvas. With this patch the correct font is on the first frame.

### Checklist

- [x] Target branch is `develop`.
- [x] Brief summary of the changes included.
- [x] Reproduction / verification steps included.
- [ ] Screenshots / videos — easy to reproduce; happy to attach a clip on request.
- [ ] Integration tests — the existing harness does not cover WASM tile-cache vs font-load timing. Happy to add one if maintainers point at the right entry-point.
- [x] No SCSS changes.
- [ ] CI status — will check after CI runs on this PR.
- [ ] `CHANGES.md` — left untouched for now; will add an entry referencing #9574 if maintainers prefer.


madalenapmelo-kp:

superalex:
Hi @fluffyspace 

Thank you very much for the pull request. I'm trying to reproduce the original issue on the current `develop` branch, but I can't reproduce it. Am I missing any steps?

I'm also including a video showing the steps I'm following.

https://github.com/user-attachments/assets/4b5d5510-c3fa-4284-b755-8423fa937d5a

belen-albeza:
Hi @fluffyspace , we cannot reproduce this bug anymore in `develop`, so I'm closing the PR. Thanks for your help!

PR #9601, opened 2026-05-13
:bug: Sanitize font-family name on custom uploaded fonts

### LINKED ISSUE #9643: Custom uploaded font family names are not sanitized
### Description

Font family names on custom uploaded fonts are not sanitized, potentially allowing CSS injection or other issues through crafted font names.

### Steps to reproduce

1. Upload a custom font with a crafted font family name containing unsafe characters
2. The font name is stored and rendered without sanitization

### Expected behavior

Font family names should be validated to only allow safe characters (letters, digits, spaces, hyphens, underscores, dots) on both client and server side.

### Affected versions

2.15.3

niwinz:
### Summary

Add a shared `schema:font-family` whitelist validator in app.common.types.font that only allows letters, digits, spaces, hyphens, underscores, and dots in font family names. Apply the schema to create-font-variant and update-font RPC endpoints on the backend, and add client-side validation in the dashboard fonts UI. Include unit tests for the schema and integration tests for the RPC handlers.

Fixes #9643

### DIFF
--- a/frontend/translations/en.po
+++ b/frontend/translations/en.po
@@ -1439,6 +1439,10 @@ msgstr "Invalid text"
 msgid "errors.team-name-invalid-chars"
 msgstr "The team name can't contain any of the following characters:'.', ':' or '/'"
 
+#: common/src/app/common/types/font.cljc
+msgid "errors.font-family-invalid-chars"
+msgstr "The font family name can only contain letters, numbers, spaces, hyphens, underscores, and dots."
+
 #: src/app/main/ui/static.cljs:74
 msgid "errors.invite-invalid"
 msgstr "Invite invalid"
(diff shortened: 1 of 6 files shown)


frontend/translations/en.po
```diff
@@ -1439,6 +1439,10 @@ msgstr "Invalid text"
 msgid "errors.team-name-invalid-chars"
 msgstr "The team name can't contain any of the following characters:'.', ':' or '/'"
 
+#: common/src/app/common/types/font.cljc
+msgid "errors.font-family-invalid-chars"
+msgstr "The font family name can only contain letters, numbers, spaces, hyphens, underscores, and dots."
```
elenatorro, line 1444: We need to support it also in Spanish

niwinz: This is a security fix. The spanish translation can be added later.

PR #9611, opened 2026-05-13
:tada: Feat tile atlas

AzazelN28:
### Related Ticket

[Taiga Task #14198](https://tree.taiga.io/project/penpot/task/14198)

### Summary

1. Changed the way frames are queued so that there is now a single ‘source of truth’ for frame and request identifiers (ClojureScript).
2. Now render returns the type of frame that has been generated: 0 (NONE), 1 (PARTIAL), 2 (FULL). If PARTIAL is returned, a new frame is queued from ClojureScript (this gives us a bit more control over the state of the current render).
3. Now Viewbox is passed by reference to all render functions.
4. I’ve added a few things like `#[inline(always)]` (I know the compiler is super smart, but this way we guarantee that things like getters and the like are definitely compiled inline).
5. I’ve added a new surface called `tile_atlas`, which is an 8192px texture (though it would be great if this were entirely hardware-dependent and had no limitations) and now this `tile_atlas` is rendered using `draw_atlas` so tiles are rendered in parallel.

### Checklist

### DIFF
--- a/frontend/playwright/ui/specs/workspace.spec.js
+++ b/frontend/playwright/ui/specs/workspace.spec.js
@@ -414,7 +414,8 @@ test("[Taiga #9929] Paste text in workspace", async ({ page, context }) => {
     .getByText("Lorem ipsum dolor");
 });
 
-test("[Taiga #9930] Zoom fit all doesn't fit all shapes", async ({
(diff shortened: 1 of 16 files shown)


AzazelN28:
https://github.com/penpot/penpot/pull/9891 replaces this

PR #9617, opened 2026-05-14
:bug: Fix Plugin API applyToken/applyToShapes destroying proxy objects

### LINKED ISSUE #9290: applyToken fails with "Field message is invalid" on color/dimension/opacity tokens (SDK 2.15)

SkoomaBreath04:
## Summary

- **Root cause**: `json/->clj` converts all objects with `constructor === Object` into Clojure maps, but Penpot's proxy objects (TokenProxy, ShapeProxy) created via `obj/reify` also match this check. When `add-properties!` decoded arguments through `->clj`, proxy objects were destructured into plain maps, stripping the `Symbol.for("penpot.reify:type")` property. This caused `token-proxy?` and `shape-proxy?` to fail during Malli schema validation.
- **Fix**: Guard `->clj` to skip objects carrying the `penpot.reify:type` Symbol, preserving proxy identity through the validation pipeline.
- **Secondary fix**: `error-messages` in `utils.cljs` called `(name field)` on numeric tuple indices from Malli's `:in` path, causing a crash. Changed to `(str field)`.

## Affected methods

- `shape.applyToken(token, properties)` — threw *"Value not valid: Field message is invalid: . Code: :error"*
- `token.applyToShapes(shapes, properties)` — threw *"Doesn't support name: 0"*

Both methods now work correctly for all token types (color, borderRadius, opacity) and shape types.

## Test plan

- [x] `shape.applyToken(opacityToken, ["opacity"])` succeeds
- [x] `token.applyToShapes(shapes, ["opacity"])` succeeds  
- [x] `shape.applyToken(colorToken, ["fill"])` succeeds
- [x] No regressions in `json/->clj` for plain JS objects (the guard only skips objects with the reify Symbol)

## Related issues

Fixes #9290

Signed-off-by: Miguel Dominguez <miguel.dominguez@zohocorp.com>


madalenapmelo-kp:

tknatwork:
## Production verification — same fix works for self-hosted 2.15.2

Confirming this PR's approach end-to-end on Penpot 2.15.2 self-hosted (Tailscale + MCP plugin). Hand-applied the same `Symbol.for("penpot.reify:type")` guard to `app.common.json/->clj`'s Object branch via a `patches/shared.js` volume mount; backported into `penpotapp/frontend:2.15.2`'s compiled bundle. After 24h of exercise via the MCP plugin runtime, all three plugin paths pass:

- `shape.applyToken(token, ["fill"])` — color, dimension, opacity ✅
- `token.applyToShapes([shapes], ["fill"])` ✅
- `lib.tokens.addTheme({name, group})` ✅ (separate bug, fixed via PR #9034 + an arity correction in compiled output)

The proxy-Symbol guard is the load-bearing fix — once `TokenProxy`/`ShapeProxy` survive `->clj` intact, `token-proxy?` / `shape-proxy?` predicate validation passes and the rest of the plugin schema chain works as designed.

### Secondary observation — `error_messages` empty-result path

Your `(name field) → (str field)` fix in `utils.cljs` is correct for the numeric-tuple-index case. We also hit an adjacent edge in production: when `interpret-schema-problem` returns no entries (or all are filtered out), `(str/join ". " [])` produces `""` → `handle-error` calls `(not-valid plugin-id :error nil)` → user sees bare `"[PENPOT PLUGIN] Value not valid. Code: :error"` with no schema info. The bare-error surface was what hid an unrelated plain-JS-Error in our deployment for several rounds of debugging.

A defense-in-depth fallback chain in `handle-error`:

```clojure
(let [message (if-let [explain (-> cause ex-data ::sm/explain)]
                (or (error-messages explain)
                    (sm/humanize-explain explain)
                    (pr-str explain))
                (or (ex-data cause)
                    (.-message cause)
                    (str cause)))]
  ...)
```

…would surface the schema-validation details (or the raw JS error message) instead of swallowing them. Happy to open a follow-up PR if you'd like it as a separate change — orthogonal to the proxy fix here.

SkoomaBreath04:
Closing — the core team's #9632 (merged May 14) fixes the same root cause via `frontend/src/app/util/object.cljc`. Different file path but same fix: preventing proxy objects from being destructured during schema validation.

The secondary `(name field) → (str field)` fix in `utils.cljs` (for numeric tuple indices in Malli's `:in` path) may still be relevant if that code path is hit independently. Happy to open a focused follow-up PR for that if useful.

Thanks @Alotor for the quick turnaround on the core fix.

Issue #9620, opened 2026-05-14
Incorrect error message when trying to apply a token to a shape - "Tokens can't be applied while editing text. Select the text layer instead."


madalenapmelo-kp:
https://tree.taiga.io/project/penpot/issue/14191

**Describe the bug**          
          
An incorrect error message is displayed when trying to apply a token to a shape that is another set and is not a text: Tokens can't be applied while editing text. Select the text layer instead. 

**To reproduce (What did you do?)** 

1.  Create a new file
2.  Navigate to TOKENS tab
    1.  Create set 1 with a Border Radius token (active set)
    2.  Create set 2 with a Border Radius token (active set)
3.  Add a Rectangle
4.  Apply token from set 2
5.  Click to set 1
6.  Click on token applied
7.  **Observe**: an error message is displayed Tokens can't be applied while editing text. Select the text layer instead.

**Actual behavior (What happened?)**

An incorrect message is displayed and token can not be applied.

**Expected behavior (What should have happened?)**          
          
No errors should be displayed and token should be applied

**Screenshots/Videos**          
          
Attached.

**Browser and OS (if different than Linux+Chrome)**

**Environment (please complete the following information):**          
          
PRE

Design:  
External source: 
External source 2:

joanaclr:
Hello,
I have exactly the same issue, how to correct it ?

Thank you

EdgarVerona:
Ah, fantastic!  I just started using the product and ran into this - that's great to see it already has a fix coming in!  I really love the app so far, thank you folks for all the hard work you've put into this.

PR #9627, opened 2026-05-14
:bug: Fix misleading text-edition warning on token apply

Rene0422:
### Related Ticket

#9620 

### Summary

`[apply-token](vscode-webview://1nl4ntrg7gcer6k0cm6mqulhep1iptit3jmp76c6peucvqbg0dki/frontend/src/app/main/data/workspace/tokens/application.cljs#L660)` paired the "Tokens can't be applied while editing text" toast with a combined guard:

`(if (and (some? token) (not text-editing?))`
`  ...apply...`
`  ...show editing-text toast...)`

so the same wording fired for two unrelated failure modes — actually editing text and the `token` argument being nil. After applying a token from one set and switching to another, stale UI references can resolve to a nil token at click time, leaving the user staring at a "select the text layer" message on a rectangle.

- [frontend/src/app/main/data/workspace/tokens/application.cljs](vscode-webview://1nl4ntrg7gcer6k0cm6mqulhep1iptit3jmp76c6peucvqbg0dki/frontend/src/app/main/data/workspace/tokens/application.cljs#L670) — gate the toast on `text-editing?` only. When `token` is nil and we are not editing text, do nothing; the existing `assert` at the top of `apply-token` still catches programmer errors in development.
- [frontend/test/frontend_tests/tokens/logic/token_actions_test.cljs](vscode-webview://1nl4ntrg7gcer6k0cm6mqulhep1iptit3jmp76c6peucvqbg0dki/frontend/test/frontend_tests/tokens/logic/token_actions_test.cljs#L106) — regression test that pre-populates `[:workspace-local :edition]` with a rectangle's id (the case the old code conflated with "editing text") and verifies the border-radius token still applies.
- [CHANGES.md](vscode-webview://1nl4ntrg7gcer6k0cm6mqulhep1iptit3jmp76c6peucvqbg0dki/CHANGES.md) — changelog entry referencing the issue.

### Steps to reproduce 

Before this PR, the toast appears on every step 8 below; after, the token applies normally.

1. Create a new file.
2. Open the Tokens tab.
3. Create Set 1 with a Border Radius token (active).
4. Create Set 2 with a Border Radius token (active).
5. Add a rectangle.
6. Apply the token from Set 2 to the rectangle.
7. Switch to Set 1.
8. Click on the applied token.

Expected: the token applies without errors.
Before this PR: a warning toast appears reading "Tokens can't be applied while editing text. Select the text layer instead." even though the selected shape is a rectangle and no text editor is open.
After this PR: no toast; the token is applied. The same wording is now reserved for the case it actually describes (an open text editor on a text shape).

### Checklist


madalenapmelo-kp:

EvaMarco:
Thanks a lot for your contribution and for taking the time to investigate this issue 🙌

After reviewing the proposed solution, we noticed that it does not address the underlying problem itself, but rather prevents the error message from appearing. The root cause of the issue is still present.

I've already worked on a [separate PR ](https://github.com/penpot/penpot/pull/9708) that resolves the actual issue, so we're going to move forward with that approach and close this one.

We really appreciate the effort and the time you put into helping improve the project. Thanks again for contributing!

Issue #9498, opened 2026-05-11
Component variant switching behaves differently on two identical copies


madalenapmelo-kp:
https://tree.taiga.io/project/penpot/issue/14070

NatachaMenjibar:
Tested and working fine now

PR #9434, opened 2026-05-07
:bug: Fix float comparison #14070

MischaPanch:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/14070
https://github.com/penpot/penpot/issues/9498

### Summary

Overly strict equality comparison of floats caused the difficult to identify intermittent bug described in the ticket.
The new test fails on develop and is fixed by the commit following it.

### Steps to reproduce 

See ticket and the new test

PR #9691, opened 2026-05-15
:bug: Preserve target-positioned composite geometry when variant switch keeps size overrides

### LINKED ISSUE #9749: Fix variant switching preserving stale absolute positions in nested instance overrides
### Description

When switching between variants, geometry overrides (such as resized nested component instances) can carry over stale absolute position data from the previous variant. This causes the nested instance to appear at the wrong location after the switch — its size override is preserved correctly, but its position (`:selrect` / `:points`) still reflects the source variant's absolute coordinates rather than the target variant's.

### Steps to reproduce

1. Create two variants that contain the same nested component instance.
2. The nested instance has the same base size in both variants, but different relative positions.
3. Instantiate one variant.
4. Apply a real geometry override to the nested instance, e.g. resize its width from 100 to 150.
5. Switch to the other variant while keeping the override.

### Expected behavior

The width/height override is preserved, but the nested instance should be correctly positioned at the target variant's position. The scalar geometry (`:x`/`:y`) and composite geometry (`:selrect`/`:points`) should remain internally consistent after the switch.

### Affected versions

2.17.0

MischaPanch:
### Related Ticket

https://tree.taiga.io/project/penpot/issue/14070
https://github.com/penpot/penpot/issues/9498

Related to the fix in #9434 and presents an alternative fix. The code was created by reverting the original fix and letting Codex find a solution. The result of this was the discovery of this additional failure path. The summary below is also mostly written by Codex.

The existing fix handles the drift-triggered reproduction, but variant switching still has an unsafe composite-geometry preservation path for real geometry overrides.

That is a bug because it can create invalid shape geometry under plausible user behavior, even if we do not currently have a confirmed UI reproduction beyond the synthetic common test.

### Summary
When switching variants, `update-attrs-on-switch` may preserve touched geometry from the previous instance. The existing float-tolerance fix prevents stale composite geometry from being copied when the previous shape only differs from its source master by sub-pixel drift.

There is a related but distinct case: the previous shape can have a real geometry override, such as a resized nested component instance. In that case the override should be preserved. However, :selrect and :points contain absolute coordinates, so copying them verbatim from the previous variant can also copy the previous variant’s absolute position.

Expected behavior:

Preserve the user’s real size override.
Keep the switched shape anchored at the target variant’s position.
Keep scalar geometry and composite geometry internally consistent, especially :x/:y vs :selrect.x/:selrect.y.

### Steps to reproduce 

1. Create two variants that contain the same nested component instance.
2. The nested instance has the same base size in both variants, but different relative positions.
3. Instantiate one variant.
4. Apply a real geometry override to the nested instance, e.g. resize width from 100 to 150.
5. Switch to the other variant with keep-touched? behavior.

Without anchoring preserved composite geometry to the target shape, the switch can preserve the width override while also carrying stale absolute :selrect / :points coordinates from the source variant. This risks a malformed shape where scalar :y follows the target variant, but :selrect.y still reflects the source variant.

### Additional Context

The previous fix solved that by making equal-geometry? tolerant, so the stale-copy path is skipped for drift-only changes.

I focused on the downstream invariant instead: if update-attrs-on-switch ever does preserve :selrect / :points, it should not copy old absolute position. That also fixes the reverted reproduction, because it makes the bad path safe rather than avoiding it.

So my fix addressed the same observed symptom because it was applied lower in the pipeline:

Previous fix: prevent entering the stale-copy branch for float drift.
My fix: make the stale-copy branch rebuild target-positioned composite geometry.
With the previous fix restored, my change is no longer necessary for the exact original Alt-drag reproduction, but it still protects the same invariant for real geometry overrides.

@pabloalba FYI

Closes #9749

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
 
 ### :bug: Bugs fixed
 
+- Fix malformed component variants after Alt-drag duplication and variant switching
 - Fix plugin API `LibraryTypography.remove()` failing with a UUID assertion error [Github #8223](https://github.com/penpot/penpot/issues/8223)
 - Fix MCP SSE sessions leaking memory on zombie connections by adding inactivity timeout parity with Streamable HTTP sessions (by @bitloi) [Github #9432](https://github.com/penpot/penpot/issues/9432)
 - Fix missing `labels.open` translation (by @MilosM348) [Github #9320](https://github.com/penpot/penpot/pull/9320)
--- a/common/src/app/common/logic/libraries.cljc
+++ b/common/src/app/common/logic/libraries.cljc
@@ -2101,6 +2101,39 @@
            (grc/rect->center selrect)
            (or (:transform current-shape) (gmt/matrix)))))))
 
+
+(defn- switch-geom-change-value
+  [prev-shape current-shape attr]
+  ;; Composite geometry stores absolute coordinates. When preserving a size
+  ;; override across variants, keep the target variant's position and only carry
+  ;; the previous dimensions; otherwise :x/:y can disagree with :selrect/:points.
+  (let [prev-selrect (:selrect prev-shape)
+        current-selrect (:selrect current-shape)
+        final-width (:width prev-selrect)
+        final-height (:height prev-selrect)
+        x (:x current-selrect)
+        y (:y current-selrect)
+        selrect (assoc current-selrect
+                       :width final-width
+                       :height final-height
+                       :x x
+                       :y y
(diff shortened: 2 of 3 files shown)


madalenapmelo-kp:

niwinz:
hey @MischaPanch Can you rebase this and remove the changelog entry (it will be generated automatically)

MischaPanch:
@niwinz done, thanks for the review!

PR #9693, opened 2026-05-15
:sparkles: Dismiss workspace search panel with the Escape key

### LINKED ISSUE #9540: Close workspace search with the Escape key

**Panels in scope**

- History
- Shortcuts

dhgoal:
This PR closes the workspace **Search layers / Search on canvas** panel when the user presses <kbd>Escape</kbd> while the search field is focused, returning keyboard control to the canvas.

## References

Resolves https://github.com/penpot/penpot/issues/9540

## Code changes

- `search-bar*` gains an optional `on-escape` callback. The component already swallowed <kbd>Escape</kbd> to blur its input; it now also invokes `on-escape` when provided. All existing `search-bar*` callers are unaffected (the prop is optional and `fn?`-guarded).
- The layers search panel passes a `close-search` handler that resets search state and hides the panel.

## Behaviour / design notes

- **Focus scoping** — because the handler fires from the search field’s own `keydown`, Escape only dismisses the panel when focus is inside it. If focus has moved elsewhere (e.g. back to the layers list), the search input never receives the event, so the panel is left open — matching the issue’s requirement.
- **Progressive Escape** — when the filter dropdown is open, the existing document-level handler closes just the menu; `close-search` is suppressed in that case, so the panel is dismissed on the next Escape press rather than both at once.
- **Return focus to canvas** — the search field is blurred (existing behaviour); with no element retaining focus, the workspace’s document-level shortcut handlers regain control, which is how the canvas “has focus” in this codebase.

## How to verify

Open the workspace, open search with its shortcut, place the cursor in the search input, and press <kbd>Escape</kbd> — the panel dismisses. Confirm that pressing <kbd>Escape</kbd> while focus is outside the panel (e.g. in the layers list) does not affect it, and that with the filter menu open the first <kbd>Escape</kbd> closes only the menu.

## Backwards-incompatible changes

None.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
 
 ### :sparkles: New features & Enhancements
 
+- Dismiss the workspace layers search panel with the Escape key while the search field is focused (by @dhgoal) [Github #9540](https://github.com/penpot/penpot/issues/9540)
 - Show a read-only W × H size badge below the bounding box of the current selection (by @bittoby) [Github #9205](https://github.com/penpot/penpot/issues/9205)
 - Expose `variants` retrieval on `LibraryComponent` via `isVariant()` type guard in plugin API [Github #9185](https://github.com/penpot/penpot/issues/9185)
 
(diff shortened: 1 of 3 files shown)


madalenapmelo-kp:

niwinz:
Hey @dhgoal 

Several other changes entered  and we have code conflicts. Can you please fix the conficts, and rebase or merge with the develop branch. 

Thanks

PR #9695, opened 2026-05-15
:bug: Preserve explicit hide-in-viewer when adding prototype interactions

### LINKED ISSUE #9049: Prototype interaction targets appear in View Mode automatically when linked
### Steps To Reproduce

1. Create a board, change its fill color to #f00, call it Red.
2. Disable its "Show in view mode" option (why is this on by default anyways?) 👀
3. Create another board, call it White.
4. Go to the Prototype tab and drag an arrow from the White board into the Red board.
5. Now go back into the Design tab and check the Red board.

### Expected behavior

No changes on it not being shown in the View mode. In fact I would much prefer if NOTHING would show in View Mode unless explicitely marked as such.

### Actual behavior

Things "randomly" show up in the View Mode, polluting the View mode picker.

### Screenshots

The show in View Mode I'm referring to is the play-like icon at the bottom at this image:

Image

https://design.penpot.app and self-host.

### Additional context

I believe this was designed like this so any time you set an element as a target to a `Navigate to`, this element becomes a valid new start point for interactions.

In practise (for me anyways) what happens is that even if I want to set a Show Overlay interaction I first drag the arrow to connect the elements, and only then switch the interaction type. So by the time I made the interaction no-longer a Navigate To it is too late. The element was already set to visible in View Mode.

Now why is this really a problem?

1. It gets really polluted real quick. And when using components inside components, ... I have a lot of repeated elements. It is annoying/hard to clean them. Having a filter in the outliner to show only elements in "View Mode" would help with this. This feels like fixing the wrong issue though.

Real example (only 2 board were actually intended as start point):

Image

2. When I share the prototype link with someone else I have to be extra careful to make sure the initial/default visible page/board is the intended one. It is often not the case since it is pretty random what is the "default board" that is shown. If I didn't have to remove the extra boards from View Mode my original link would always be safe.

---

Thanks :)

### LINKED ISSUE #10289: Boards hidden from View Mode reappear when adding prototype interactions
### Description

When a user explicitly hides a board from View Mode (via the `hide-in-viewer` toggle in the measures/design panel), then later drags a prototype arrow (adds an interaction) to that same board, the board becomes visible again in View Mode. The user's explicit intent to keep the board hidden during presentation is silently overridden by the interaction wiring.

### Steps to reproduce

1. Create two boards (e.g., "White" and "Red").
2. Select the Red board, and in the design panel, toggle **Hide in viewer** on.
3. Verify Red is hidden in View Mode.
4. Select the White board, and drag a prototype arrow to the Red board (adding an interaction).
5. Enter View Mode — the Red board is now visible despite the user having explicitly hidden it.

### Expected behavior

Boards that a user has explicitly marked as hidden from View Mode should remain hidden when adding or updating prototype interactions. The user's explicit toggle should take precedence over the default behavior of auto-showing new interaction targets.

### Affected versions

2.17.0

jeffrey701:
## Summary

- `cls/show-in-viewer` in `common/src/app/common/logic/shapes.cljc` unconditionally `dissoc`'ed `:hide-in-viewer` from the destination. Every `add-interaction` / `add-new-interaction` / `update-interaction` call in `frontend/src/app/main/data/workspace/interactions.cljs` (lines 145, 172, 211) routes the destination shape through this helper, so dragging a prototype arrow at a board the user had deliberately hidden silently flipped the board back to visible-in-viewer. Reporter (#9049) hit this exact case: hid a Red board, dragged an arrow from White → Red, Red reappeared in View Mode.
- Make `show-in-viewer` a no-op when the destination already has `:hide-in-viewer true`. New prototype targets that don't have the flag set still get auto-unhidden (preserves the original UX — most users expect a brand-new destination to be visible in View Mode), but explicit user intent is now preserved across later interaction add / update operations targeting the same shape.
- Behaviour change is consistent with the surrounding `measures.cljs` UI that already lets the user toggle `:hide-in-viewer` directly — once they make that toggle, the prototype-interaction wiring no longer overrides it.

## Closes

Closes #10289

## Verification

- Function-level walk against the reporter's repro: with `:hide-in-viewer true` set on Red board, calling `(show-in-viewer red-board)` now returns `red-board` unchanged, so `dwsh/update-shapes [red-board-id] cls/show-in-viewer` is a no-op transaction — Red stays hidden after the arrow is drawn. With the flag absent (default frame), `(show-in-viewer frame)` still returns `(dissoc frame :hide-in-viewer)` — the original auto-unhide path for fresh destinations is intact.
- No other call sites: `cls/show-in-viewer` (the function, not the property name) is only called from those three lines in `interactions.cljs`. Everywhere else in the UI uses `cls/change-show-in-viewer` (the explicit-toggle path) or reads `:hide-in-viewer` directly. The patch does not affect any of those.
- No CHANGES.md edit (per `AGENTS.md` — contributors leave the changelog to maintainers).

## Out of scope

- Reporter also suggested *"NOTHING would show up in View Mode unless explicitly marked as such"* — i.e. flipping the default for fresh destinations too. That's a broader product decision (every existing prototype destination would start hidden) and is outside this PR's narrow regression-fix scope. Happy to revisit in a follow-up if maintainers want that default.


madalenapmelo-kp:

niwinz:
Ready to merge, thanks.
Lets wait the CI

Issue #8897, opened 2026-04-07
bug: IPv4 only not working anymore


tomhouweling1987:
### Steps To Reproduce

Run Penpot 2.14.1 on a IPv4 only environment and the Frontend image will break.

### Expected behavior

The image is also working on an environment with IPv4 only.

### Actual behavior

The frontend image is not starting because it doesnt have IPv6. IPv6 is enabled in #8285 

### Frontend Stack Trace

```console
[emerg] 1#1: socket() [::]:8080 failed (97: Address family not supported by protocol)
nginx: [emerg] socket() [::]:8080 failed (97: Address family not supported by protocol)
```

### Backend Stack Trace

```console

```

### Additional context

Create a flag to turn off ipv6 :)

tomhouweling1987:
Is there some more information if this is going to be fixed?

sancfc:
Hi!
We tested penpotapp/frontend:2.14.1, but we could not reproduce the issue in our environment because the host has IPv6 enabled and the frontend **container starts correctly there.**
Since your report is specifically about an IPv4-only host, this still looks consistent with nginx trying to bind an IPv6 listener on a system where
  **IPv6 is not available:**

 ` socket() [::]:8080 failed (97: Address family not supported by protocol)
`

Could you share a bit more about the failing environment?

  - Host OS / distro
  - Docker / Docker Compose version
  - Whether IPv6 is disabled at the host/kernel level
  - Exact steps or a minimal setup to reproduce it

  That would help us reproduce it under the same conditions.

tomhouweling1987:
Goodmorning,

### Info:

- Ubuntu 24.04
- Docker v29.4.3 
- Docker compose 5.1.3
- IPv6 is disabled on kernel and host

### Steps:

1. Create a host with IPv4 disabled on the host and kernel
2. Spin up a new penpot compose with image version penpotapp/frontend:2.14.1
3. Frontend service wil fail because it cant bind to [::]8080

### Possible solution:

I think a solution could be:

- Create a new variable `PENPOT_DISABLE_IPV6`
- Add  an if statement, based on the disable variable to the nginx-entrypoint.sh to delete the listen `[::]:8080 default_server;` in the /etc/nginx/nginx.conf

```
# NEW: strip IPv6 listen directives on IPv4-only hosts
if [[ "${PENPOT_DISABLE_IPV6:-false}" == "true" ]]; then
  sed -i '/listen \[::\]/d' /etc/nginx/nginx.conf
[...]

PR #9738, opened 2026-05-19
:whale: Fix frontend startup on hosts without IPv6 support

sancfc:
S
This change allows self-hosted frontend containers to start on hosts where IPv6 is disabled at kernel boot level.

  The frontend nginx config currently includes:

  ```nginx
  listen [::]:8080 default_server;
```
  On systems booted with IPv6 disabled, nginx fails to start with:
```
  socket() [::]:8080 failed (97: Address family not supported by protocol)
```
  This PR makes the IPv6 listen directive configurable through `PENPOT_DISABLE_IPV6_LISTEN=true`, while preserving the current behavior by default.


niwinz:
It should target `staging` or `main` if you consider to use this on the next release. Merging it to `develop` will make it for 2.17, just FYI

yamila-moreno (approved):

bameda:
This PR will fix this https://github.com/penpot/penpot/issues/8897

PR #9748, opened 2026-05-19
:sparkles: Bound MCP memory consumption by limiting parallel exports & response size

### LINKED ISSUE #9493: Add concurrency limiter for MCP Server Plugin Communications
## Summary

Add a configurable concurrency limiter (semaphore) to the MCP server's plugin communication layer to prevent overwhelming the plugin WebSocket connection with too many concurrent heavy operations (e.g., image exports, code execution with large state).

## Problem

The MCP server (`mcp/packages/server`) maintains persistent WebSocket connections to Penpot plugin instances. Currently, every tool call that touches the plugin is immediately dispatched over the WebSocket with **no backpressure mechanism**. If an AI client issues many rapid tool calls (e.g., export multiple shapes, execute several code snippets in parallel), all tasks are sent concurrently to the plugin. This can:

1. **Overwhelm the plugin** — it has to process many heavy operations simultaneously, each potentially involving image data, large state queries, or expensive DOM manipulations.
2. **Saturate WebSocket bandwidth** — image exports and imports can transfer megabytes of data per operation; concurrent transfers compound the problem.
3. **Exhaust memory** — in-flight task responses (especially PNG/SVG image blobs) accumulate in the server process until consumed.
4. **Degrade latency for all users** — in multi-user mode, one aggressive client can degrade responsiveness for others due to plugin-side resource contention.

## Proposed Solution

Introduce a **per-connection semaphore** inside `PluginBridge.executePluginTask()` — the single chokepoint through which all plugin-bound operations flow:

1. Each plugin WebSocket connection gets its own semaphore with a configurable concurrency limit (`PENPOT_MCP_MAX_CONCURRENT_TASKS`, default: `5`).
2. When a task arrives and the connection is under the limit, it dispatches immediately (current behavior unchanged).
3. When at the limit, the task is **transparently queued** — the caller receives a Promise that resolves when the task eventually completes. No code changes needed in tools or upstream callers.
4. When a dispatched task completes (success, error, or timeout), the semaphore releases a slot and the next queued task for that connection is automatically dispatched.
5. The queue is **per-connection** (not global) — in multi-user mode, each user's plugin connection has its own independent queue, preventing cross-user starvation.

### Key design decisions

- **Per-connection (not global) queue**: Each user's plugin connection is independent. A global queue would let one user's heavy exports starve another user's lightweight queries.
- **Timeout starts at dispatch time**: A queued task's timeout only begins counting when it is sent to the plugin, not when it entered the queue. This prevents premature timeout for legitimately backlogged tasks.
- **Disconnect cleanup**: If a WebSocket disconnects, all queued tasks for that connection are immediately rejected (with an appropriate error), and in-flight tasks are left to their existing timeout mechanism.
- **No API changes**: The semaphore is entirely internal to `PluginBridge`. Tools, the MCP SDK wrapper, and clients are completely unaware of the queueing.

### Affected areas

- `mcp/packages/server/src/PluginBridge.ts` — primary changes (semaphore logic, queue, dispatch extraction)
- `mcp/packages/server/src/PenpotMcpServer.ts` — read and pass the new env var
- New tests in `mcp/packages/server/src/__tests__/PluginBridge.test.ts`

### Configuration

Proposal:

```bash
# Maximum concurrent intercommunications per plugin WebSocket connection (default: 5)
export PENPOT_MCP_MAX_CONCURRENT_TASKS=5
```

## Open Questions

- Should we also add a maximum queue depth (`PENPOT_MCP_MAX_QUEUED_TASKS`), rejecting tasks when the queue is full? Or is unbounded queuing acceptable (bounded by memory)?
- Should the limit be configurable per-connection (e.g., different limits for different user tiers)?
- Should we expose current queue depth and active count as metrics (e.g., a Prometheus endpoint or the REPL server)?
- **Should we add also a global limit of task as an additional measure for prevent unbound memory consumption (this is an addtional semaphore to the already proposed)?**

opcode81:
### Related Ticket

* #9493 

### Summary

The changes in this PR seek to bound memory consumption in the centrally provided remote MCP server.
A main driver of memory consumption is large plugin task responses, which originate from the `export_shape` tool (which returns images to the caller).
This PR takes two measures to control memory usage:

1. It introduces a **limit L on the task response size in the plugin** (which is enabled for the integrated remote MCP server only). The current limit is set to 15 MB. Any larger response is not returned to the server; an error response is returned instead.  
   The bound pertains to the JSON-serialised format; the actual image size that can be exported is lower (~11 MB).
   This limit is well above limits used by current mainstream LLMs and we can consider reducing it if needed.
3. The total **number N of parallel requests to the `export_shape` tool** (across all users in multi-user mode) is optionally limited (configured via env. var. `PENPOT_MCP_EXPORT_SHAPE_MAX_PARALLEL_REQUESTS`; default 0 = no limit).  
   Bounding the requests directly in the tool allows the semaphore to cover the entire unit of work, including decoding.
   Any additional requests that are received are queued; saturation is discernible via new log messages.

In combination, this results in a total bound of N * L for these requests.
Other task responses are typically fairly small (text only). This PR addresses the main driver of memory usage.

The implementation was manually tested:
* To test the bounded parallelism of the tool, a script simulating parallel client requests was added (integration test).
* The response limit was tested by configuring a smaller threshold.

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -264,14 +264,15 @@ The Penpot MCP server can be configured using environment variables.
 
 ### Server Configuration
 
-| Environment Variable               | Description                                                                | Default      |
-|------------------------------------|----------------------------------------------------------------------------|--------------|
-| `PENPOT_MCP_SERVER_HOST`           | Address on which the MCP server listens (binds to)                         | `localhost`  |
-| `PENPOT_MCP_SERVER_PORT`           | Port for the HTTP/SSE server                                               | `4401`       |
-| `PENPOT_MCP_WEBSOCKET_PORT`        | Port for the WebSocket server (plugin connection)                          | `4402`       |
-| `PENPOT_MCP_REPL_PORT`             | Port for the REPL server (development/debugging)                           | `4403`       |
-| `PENPOT_MCP_REMOTE_MODE`           | Enable remote mode (disables file system access). Set to `true` to enable. | `false`      |
-| `PENPOT_MCP_DEVENV`                | Enable Penpot development environment tools. Set to `true` to enable.      | `false`      |
+| Environment Variable                             | Description                                                                | Default        |
+|--------------------------------------------------|----------------------------------------------------------------------------|----------------|
+| `PENPOT_MCP_SERVER_HOST`                         | Address on which the MCP server listens (binds to)                         | `localhost`    |
+| `PENPOT_MCP_SERVER_PORT`                         | Port for the HTTP/SSE server                                               | `4401`         |
+| `PENPOT_MCP_WEBSOCKET_PORT`                      | Port for the WebSocket server (plugin connection)                          | `4402`         |
+| `PENPOT_MCP_REPL_PORT`                           | Port for the REPL server (development/debugging)                           | `4403`         |
+| `PENPOT_MCP_REMOTE_MODE`                         | Enable remote mode (disables file system access). Set to `true` to enable. | `false`        |
+| `PENPOT_MCP_DEVENV`                              | Enable Penpot development environment tools. Set to `true` to enable.      | `false`        |
+| `PENPOT_MCP_EXPORT_SHAPE_MAX_PARALLEL_REQUESTS`  | Maximum number of parallel export shape requests (multi-user mode only).   | `0` (no limit) |
 
 ### Logging Configuration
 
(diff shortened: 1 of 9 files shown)


niwinz:
Is this ok for SQUASH merge this PR?
It looks OK for me, i mark this as candidate for backport.

opcode81:
Yes, squashing is fine for this one.

niwinz:
Merging, integration tests looks unrelated to this PR.

PR #9789, opened 2026-05-20
:zap: Improve performance and fix orphan detection in validate-file

### LINKED ISSUE #9323: Internal Error opening IOs library
See https://tree.taiga.io/project/penpot/issue/14049

### LINKED ISSUE #9324: Internal error opening the wikimedia Design System library
See https://tree.taiga.io/project/penpot/issue/14048

### LINKED ISSUE #9790: Improve file validation performance and fix orphan shape detection
### Description

File validation (the background integrity checker that detects broken references,
orphan shapes, and structural inconsistencies) was running a full sweep of every
page and component on every change batch, regardless of what was actually
modified. On large, component-heavy files this caused measurable overhead on
every save.

Additionally, orphan shape detection was silently broken: the detection pipeline
produced empty results, meaning shapes whose parent no longer existed in the file
were never flagged.

This improves both correctness and performance of the file validation subsystem:

- Orphan shapes are now correctly detected and reported.
- Validation is scoped to only the pages and components touched by the incoming
  change batch, avoiding full-file sweeps on partial updates.
- Internal lookups (parent-child containment, ref-shape resolution, duplicate
  detection, swap-slot checks) have been optimised to reduce redundant work on
  large object trees.

### Use case

Users with large files containing many components and deeply nested copies will
experience lower latency on saves. The fix to orphan detection ensures that
structural integrity issues are surfaced correctly rather than silently ignored.

### Affected versions

main / staging

niwinz:
## Summary
- Add `*ref-shape-cache*` dynamic var to memoize `find-ref-shape` lookups per page, avoiding repeated O(depth) ancestor walks.
- Add `*children-sets*` pre-computed maps for O(1) parent-child containment checks, replacing linear `some` scans.
- Short-circuit `inside-component-main?` when the shape context already implies a main component.
- Use single-pass reduce with early exit for duplicate detection (children, swap slots) instead of count/distinct or frequencies.
- Guard `check-missing-slot` to skip expensive `find-near-match` when the shape already has a swap slot.
- Refactor variant-set validation to use `run!` with direct `get`.
- Refactor `check-ref-cycles` to use a single `reduce-kv` pass.
- Fix `get-orphan-shapes`: the original `map` pipeline produced nils so orphan shapes were never validated; rewrite with `reduce-kv` for correct results.
- Add `validate-file-affected!` for change-scoped validation, replacing full file validation in `process-changes-and-validate` to only validate pages and components touched by the changes.

Closes #9790 #9324 #9323

### DIFF
--- a/backend/src/app/rpc/commands/files_update.clj
+++ b/backend/src/app/rpc/commands/files_update.clj
@@ -379,19 +379,16 @@
       (l/error :hint "file schema validation error" :cause cause))))
 
 (defn- soft-validate-file!
-  [file libs]
+  [file libs changes]
   (try
-    (val/validate-file! file libs)
+    (val/validate-file-affected! file libs changes)
     (catch Throwable cause
       (l/error :hint "file validation error"
                :cause cause))))
 
-
 (defn- process-changes-and-validate
   [cfg file changes skip-validate]
-  (let [;; WARNING: this ruins performance; maybe we need to find
-        ;; some other way to do general validation
-        libs
+  (let [libs
         (when (and (or (contains? cf/flags :file-validation)
                        (contains? cf/flags :soft-file-validation))
                    (not skip-validate))
@@ -419,14 +416,14 @@
 
     (binding [pmap/*tracked* nil]
       (when (contains? cf/flags :soft-file-validation)
-        (soft-validate-file! file libs))
+        (soft-validate-file! file libs changes))
 
       (when (contains? cf/flags :soft-file-schema-validation)
         (soft-validate-file-schema! file))
 
       (when (and (contains? cf/flags :file-validation)
                  (not skip-validate))
-        (val/validate-file! file libs))
+        (val/validate-file-affected! file libs changes))
 
       (when (and (contains? cf/flags :file-schema-validation)
                  (not skip-validate))
--- a/common/src/app/common/files/validate.cljc
+++ b/common/src/app/common/files/validate.cljc
@@ -17,7 +17,6 @@
    [app.common.types.components-list :as ctkl]
    [app.common.types.container :as ctn]
    [app.common.types.file :as ctf]
-   [app.common.types.pages-list :as ctpl]
    [app.common.types.shape-tree :as ctst]
    [app.common.types.variant :as ctv]
    [app.common.uuid :as uuid]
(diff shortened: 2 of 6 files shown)


common/src/app/common/files/validate.cljc
```diff
@@ -638,11 +721,12 @@
 
 (defn check-ref-cycles
   [component file]
-  (let [cycles-ids (->> component
-                        :objects
-                        vals
-                        (filter #(= (:id %) (:shape-ref %)))
-                        (map :id))]
+  (let [cycles-ids (reduce-kv (fn [acc _ shape]
+                                (if (= (:id shape) (:shape-ref shape))
```
niwinz: We can avoid this lookup on shape

common/src/app/common/files/validate.cljc
```diff
@@ -703,10 +787,24 @@
...
+  shape itself is excluded since it is always validated separately.
+
+  Implemented with `reduce-kv` rather than a `map`/`filter` pipeline.
+  The previous implementation mapped over `objects` and returned a
+  lazy seq that could contain `nil` entries (for shapes that were
+  *not* orphans), meaning `check-shape` was called with `nil` IDs and
+  orphaned shapes were silently skipped.  The `reduce-kv` approach
+  builds a plain vector of IDs and never yields `nil` entries."
+  [{:keys [objects] :as _page}]
+  (reduce-kv (fn [result id shape]
```
niwinz: We can use transients maybe?

common/src/app/common/files/validate.cljc
```diff
@@ -744,6 +861,97 @@
...
+(defn- extract-affected-ids
+  "Single reduce pass over a changes batch.
+  Returns {:page-ids #{uuid …} :component-ids #{uuid …}}.
+
+  Only entities that need re-validation after applying `changes` are
+  included.  Deleted entities and pure file-level changes (colors,
+  tokens, typography…) produce no entries because there is nothing left
+  to check."
+  [changes]
+  (reduce
```
niwinz: I think implementing this using loop will be more efficient

common/src/app/common/files/validate.cljc
```diff
@@ -149,15 +194,32 @@
                     shape file page)
       (do
         (when-not (cfh/root? shape)
-          (when-not (some #(= shape-id %) (:shapes parent))
+          ;; Fast path: O(1) set lookup when `*children-sets*` is pre-computed
+          ;; for the current page.  Falls back to an O(k) linear scan via
+          ;; `some` when the var is unbound (e.g. in isolated single-shape
+          ;; validation calls outside a full page pass).
+          (when-not (if-let [cs *children-sets*]
+                      (contains? (get cs (:id parent)) shape-id)
```
elenatorro, line 202: Since we're removing parents without children, this could fail if `(get cs (:id parent))` returns `nil`, right? 🤔 should it return `false` in this situation to prevent raising an error?

niwinz: The `*children-set*` is generated thanks to `build-children-sets`

And it already expectes that the get can return nil
image

common/src/app/common/files/validate.cljc
Alotor: i've made the change. Also i've used transients for this loop.

common/src/app/common/files/validate.cljc
Alotor: done

common/src/app/common/files/validate.cljc
Alotor: i've changed it for a loop

backend/src/app/rpc/commands/files_update.clj
```diff
@@ -419,14 +416,14 @@
 
     (binding [pmap/*tracked* nil]
       (when (contains? cf/flags :soft-file-validation)
-        (soft-validate-file! file libs))
+        (soft-validate-file! file libs changes))
```
Alotor, line 419: we have validate and soft-validate but they call the same code?

PR #9828, opened 2026-05-22
:bug: Fix problem with position-data not present

Alotor:
### Summary
Fix performance problem with the calculation of missing `position-data` in text shapes. 

This is done through two methods: first the workspace will calculate the position data for the current page only. 

For possible missing position data in the viewer there is a second process that makes the calculation on the viewer.

### Checklist

### DIFF
--- a/frontend/src/app/main/data/viewer.cljs
+++ b/frontend/src/app/main/data/viewer.cljs
@@ -193,30 +193,26 @@
                                           :all-frames (ctt/get-viewer-frames (:objects data) {:all-frames? true}))])))
                  (into {}))]
 
-        (prn "update viewer state")
-
         (-> state
             (assoc-in [:viewer :file] file)
             (assoc-in [:viewer :pages] pages))))))
 
 (defn- generate-update-position-data-changes
-  [state file-id page-id]
-  (reduce-kv
-   (fn [result _ shape]
-     (if (and (cfh/text-shape? shape) (nil? (:position-data shape)))
-       (conj result
-             {:type :mod-obj
-              :id (:id shape)
-              :page-id page-id
-              :operations
-              [{:type :set
-                :attr :position-data
-                :val (wasm.api/calculate-position-data shape)
-                :ignore-touched true
-                :ignore-geometry true}]})
-       result))
+  [shapes page-id]
+  (reduce
+   (fn [result shape]
+     (conj result
+           {:type :mod-obj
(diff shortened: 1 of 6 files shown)


frontend/src/app/main/data/viewer.cljs
```diff
@@ -171,6 +174,79 @@
...
+            pages
+            (->> (dm/get-in file [:data :pages])
+                 (map (fn [page-id]
+                        (let [data (get-in file [:data :pages-index page-id])]
+                          [page-id (assoc data
+                                          :frames (ctt/get-viewer-frames (:objects data))
+                                          :all-frames (ctt/get-viewer-frames (:objects data) {:all-frames? true}))])))
+                 (into {}))]
+
+        (prn "update viewer state")
```
elenatorro: print

frontend/src/app/render_wasm/api.cljs
```diff
@@ -1263,6 +1276,15 @@
         {:thumbnails pending_thumbnails
          :full pending_full}))))
 
+;; Check if the shape is already in the wasm memory
+;; if it's not it will copy it into the memory
+(defn assert-shape
+  [shape]
+  (when wasm/context-initialized?
+    (when-not (has-shape (:id shape))
+      (set-object shape))
```
elenatorro: set-object is async, if the font is not loaded at this point we might prefer to wait until the font is loaded to correctly calculate the position data

elenatorro (approved):

frontend/src/app/main/data/viewer.cljs
Alotor: ups :)

frontend/src/app/main/data/viewer.cljs
```diff
@@ -171,6 +174,88 @@
...
+   []
+   shapes))
+
+(defn update-page-position-data
+  [file-id page-id]
+  (ptk/reify ::update-page-position-data
+    ptk/WatchEvent
+    (watch [_ state _]
+      (if (features/active-feature? state "render-wasm/v1")
+        (let [objects (dsh/lookup-page-objects state file-id page-id)
```
elenatorro, line 223: we have both objects and shapes here, right? we're doing this call twice: `(dsh/lookup-page-objects state file-id page-id)`

elenatorro (approved):

PR #9860, opened 2026-05-25
:sparkles: Add single-file export to PDF using the WebGL render

### LINKED ISSUE #10022: Support single-file PDF export using WebGL render

📜 Description
--------------

Use the WebGL render to export to PDF format

**Context:**

Now that we've released the WebGL render, we're moving to having a consistent rendering experience across Penpot. We're working on migrating the Export features. We currently support exporting to images. This feature is to support the PDF format.

It should work for now under the `wasm-export` config flag only, and only when the WebGL render is active.

Initially, we'll start with single-file exports, and then we'll move to multiple-file exports as this would also require backend changes and a bigger refactor.

### How to Test

**Prerequisites**:

- WebGL render must be active
-  wasm-export config flag must be enabled

**Steps:**

- Open a file in Penpot with the WebGL renderer active
- Select an object or frame
- Go to Export and choose PDF as the format
- Click export and verify a valid PDF is downloaded

**What to verify:**

-  Exported PDF renders correctly and matches the canvas

elenatorro:
### Related Ticket

* https://github.com/penpot/penpot/issues/10022
* Taiga original ticket (do not close): https://tree.taiga.io/project/penpot/task/13939

### Summary

https://github.com/user-attachments/assets/d5cd3517-2cd0-4b71-9327-b30bc2a95566

**Vector PDF export via render-wasm**

Adds vector PDF export for **single-file** exports, rendered in the browser through render-wasm's Skia PDF backend instead of rasterising a screenshot. Paths, text, and fills come out as real PDF vector operations; only inherently pixel-based effects (blur, blurred shadows) are rasterised.

**What's included**

  - New vector render path driving Skia's PDF canvas (render/vector.rs → render/pdf.rs), the render_shape_pdf WASM export, and the CLJS wiring.
  - :pdf export uses the WASM path only when render-wasm is active and the :wasm-export flag is set, otherwise it falls back to the normal server export (only for single-file exports)
  - Matches the on-screen render for shapes, strokes, text, shadows, images, masks, and groups, but there could be some non-identified mismatches as this is the first step of the iteration

 **Next steps**

  - Add parity tests (vector vs GPU render)
  - PDF support for multiple/batch export

### Steps to reproduce

- Add `enable-wasm-export` on config.js public file
- Test exporting shapes to PDF (single shapes / boards!) 

### Checklist

### DIFF
--- a/frontend/src/app/main/data/exports/assets.cljs
+++ b/frontend/src/app/main/data/exports/assets.cljs
@@ -163,31 +163,46 @@
       (when (= status "ended")
         (dom/trigger-download-uri filename mtype resource-uri)))))
 
+;; TODO: Remove once we support WASM SVG export
+(def ^:private wasm-export-types #{:jpeg :webp :png :pdf})
+
+(defn- wasm-export-enabled?
+  "WASM export is available: the flag is set AND render-wasm is active for the
+  current file. When render-wasm is inactive its shape tree isn't loaded, so a
+  client-side WASM render would crash."
+  [state]
+  (and (contains? cf/flags :wasm-export)
+       (features/active-feature? state "render-wasm/v1")))
+
+(defn- use-wasm-export?
+  "Whether to take the client-side WASM export path for `export`."
+  [state export]
+  (and (wasm-export-enabled? state)
+       (contains? wasm-export-types (:type export))))
+
 (defn request-simple-export
   [{:keys [export]}]
-  (if (and (contains? cf/flags :wasm-export)
-           (contains? #{:jpeg :webp :png} (:type export)))
-    (ptk/reify ::request-simple-export-wasm
-      ptk/EffectEvent
-      (effect [_ _ _]
-        (wasm.exports/export-image export)))
+  (ptk/reify ::request-simple-export
+    ptk/UpdateEvent
+    (update [_ state]
+      (cond-> state
+        (not (use-wasm-export? state export))
+        (update :export assoc :in-progress true :id uuid/zero)))
 
-    (ptk/reify ::request-simple-export
-      ptk/UpdateEvent
-      (update [_ state]
-        (update state :export assoc :in-progress true :id uuid/zero))
-
-      ptk/WatchEvent
-      (watch [_ state _]
+    ptk/WatchEvent
+    (watch [_ state _]
+      (if (use-wasm-export? state export)
+        (do
+          (case (:type export)
+            :pdf (wasm.exports/export-pdf export)
+            (wasm.exports/export-image export))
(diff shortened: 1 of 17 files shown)


niwinz:
Small suggestion, change the commit message to something like:

```
:sparkles: Add single-file export to PDF using the WebGL render
```

frontend/src/app/main/data/exports/assets.cljs
```diff
@@ -163,31 +163,50 @@
...
+       (contains? wasm-export-types (:type export))))
+
+(defn- request-simple-export-wasm
+  [export]
+  (ptk/reify ::request-simple-export-wasm
+    ptk/EffectEvent
+    (effect [_ _ _]
+      (case (:type export)
+        :pdf (wasm.exports/export-pdf export)
+        (wasm.exports/export-image export)))))
```
niwinz: Do we really need this event? this only has 3 lines, i think we just inline it

frontend/src/app/main/data/exports/assets.cljs
```diff
@@ -163,31 +163,50 @@
       (when (= status "ended")
         (dom/trigger-download-uri filename mtype resource-uri)))))
 
+(def ^:private wasm-export-types #{:jpeg :webp :png :pdf})
```
niwinz, line 167: This types are different from the default ones? I mean if they are the same set, maybe we just can reuse it.

frontend/src/app/main/data/exports/wasm.cljs
```diff
@@ -26,3 +26,18 @@
...
+  (let [bytes (wasm.api/render-shape-pdf object-id (or scale 1))
+        blob (wapi/create-blob bytes "application/pdf")]
+    (wapi/create-uri blob)))
+
+(defn export-pdf
+  [{:keys [suffix name] :as params}]
+  (let [url (export-pdf-uri params)
+        filename (str name (or suffix "") ".pdf")]
+    (dom/trigger-download-uri filename "application/pdf" url)
+    (wapi/revoke-uri url)
```
niwinz: I think this revoke should be executed in an micro-task, for avoid any sync issues. I mean revoking on the middle of download.

elenatorro: done ✅

frontend/src/app/main/data/exports/assets.cljs
elenatorro: Yes, SVG is not there yet. I'll unify it at some point, but right now it will keep using the SVG render. I've added a comment to clarify it 👍🏽

PR #9869, opened 2026-05-26
:bug: Fix several boolean new render issues

Alotor:
### Summary

Fixed some problems with booleans and text to path in the render. 

- Text-to-path had some sharp angles instead of smooth curves
- Boolean operations produce wrong results on letters with interior holes
- Wrong calculation of quadratic bezier curves
- False intersection detection
- Replaced BTreeMap for Vector. The current BTree was skipping result because the segments cannot have a strict total order. This caused some glitches in edge cases with boolean shapes.

### Checklist

### DIFF
--- a/render-wasm/src/math/bools.rs
+++ b/render-wasm/src/math/bools.rs
@@ -381,6 +381,7 @@ pub fn bool_from_shapes(bool_type: BoolType, children_ids: &[Uuid], shapes: Shap
 
         let (segs_a, segs_b) = split_segments(&current_path, &other_path);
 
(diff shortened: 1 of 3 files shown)


render-wasm/src/shapes/shape_to_path.rs
elenatorro, line 242: shouldn't we also check if is even_odd here? We do set it at `join_paths` but I guess we should include it here too

render-wasm/src/math/bools.rs
elenatorro: same here: should we check even_odd here?

elenatorro: and here?

PR #9876, opened 2026-05-26
:sparkles: Dismiss History and Shortcuts panels with Escape

Atefoub:
## Summary

- Close the workspace **History** panel with Escape when focus is inside the panel.
- Close the workspace **Shortcuts** panel (`?`) with Escape when focus is inside the panel.
- Version rename inputs in History keep their existing Escape behaviour.

Partial fix for [#9540](https://github.com/penpot/penpot/issues/9540) (layers search is handled separately in [#9693](https://github.com/penpot/penpot/pull/9693)).

## Test plan

- [ ] Open workspace, press `?`, focus the shortcuts search field, press Escape — panel closes
- [ ] Shortcuts open, focus on canvas/layers, press Escape — panel stays open
- [ ] Open History, focus inside the panel (tab or list), press Escape — panel closes
- [ ] In History, start renaming a version, press Escape — rename cancels only, panel stays open


madalenapmelo-kp:

niwinz:
Looks like it has some syntax errors

PR #9879, opened 2026-05-26
:whale: Remove the configuration of the mcp from Nginx if it is not enabled

bameda:
<div align="center">

</div>

These changes are intended to prevent the penpot-frontend Nginx instance from loading the MCP configuration if the `enable-mcp` flag is not set. To do this, move the locations to a file that will be deleted at the entry point if the MCP is not enabled. 

## How to test: 

You need to build docker image for penpot-frontend first

1. With `enable-mcp`: start penpot with the default behavior and verify that:
    - the MCP is working 
    - the Nginx configuration includes the file containing the MCP locations
      ```bash
      docker compose exec penpot-frontend cat /etc/nginx/overrides/server.d/mcp-locations.conf
      ```
2. Without `enable-mcp`: Remove `enable-mcp` in the docker compose file, also remove the penpot-mcp service (and the depends-on for penpot-frontend) and verify that:
    - the Nginx configuration does not contain the file with the MCP locations.

### DIFF
--- a/docker/images/files/nginx-entrypoint.sh
+++ b/docker/images/files/nginx-entrypoint.sh
@@ -34,15 +34,23 @@ update_flags /var/www/app/js/config.js
 export PENPOT_BACKEND_URI=${PENPOT_BACKEND_URI:-http://penpot-backend:6060}
 export PENPOT_EXPORTER_URI=${PENPOT_EXPORTER_URI:-http://penpot-exporter:6061}
 export PENPOT_NITRATE_URI=${PENPOT_NITRATE_URI:-http://penpot-nitrate:3000}
-export PENPOT_MCP_URI=${PENPOT_MCP_URI:-http://penpot-mcp:4401}
-export PENPOT_MCP_URI_WS=${PENPOT_MCP_URI_WS:-http://penpot-mcp:4402}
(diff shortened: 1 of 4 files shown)


docker/images/files/nginx-entrypoint.sh
```diff
@@ -34,15 +34,23 @@ update_flags /var/www/app/js/config.js
...
+        < /tmp/nginx.conf.template > /etc/nginx/nginx.conf
+
+if [[ $PENPOT_FLAGS == *"enable-mcp"* ]]; then
+    export PENPOT_MCP_URI=${PENPOT_MCP_URI:-http://penpot-mcp:4401}
+    export PENPOT_MCP_URI_WS=${PENPOT_MCP_URI_WS:-http://penpot-mcp:4402}
+
+    envsubst "\$PENPOT_MCP_URI,\$PENPOT_MCP_URI_WS" \
+             < /tmp/nginx-mcp-locations.conf.template > /etc/nginx/overrides/server.d/mcp-locations.conf
+else
+    rm -f /etc/nginx/overrides/server.d/mcp-locations.conf
```
yamila-moreno, line 52: this rm should point to this file or the `/tmp/nginx-mcp-locations.conf.template` file?

bameda: Imagine we start up with `enable_mcp`; the entrypoint will create the file. Then, If we restart with `disable_mcp` while reusing the container (without destroying it), we must ensure that the file does not exist.

yamila-moreno (approved):

PR #9898, opened 2026-05-27
:sparkles: Add waitForLayoutUpdate plugin method

### LINKED ISSUE #27: Asynchronous Update of Properties
I encountered an issue while investigating the design linter use case #26.

When a property of a shape is updated, (computed) properties affected by the change are not updated synchronously. Instead, they will update asynchronously at some point in the future!

This is a problem, because when the LLM wants to determine how a property change affects the design, it will often read incorrect information. This makes it difficult for the LLM to reason about the effects of its actions.

The key question is: **Is there anything we could do?**
 * Could we add a method `waitForSync` that waits for everything to be synchronised?
 * Could we perhaps even add a synchronous API mode?

The current workaround is to inform the LLM about asynchronous updates and make it wait for e.g. 100ms before reading the effects - but this is error-prone for two reasons:
 1. The LLM can forget to wait/be unaware that it needs to wait in a particular case
 2. The wait time could be too short (not solving the problem) or too long (wasting time).

### Example: Updating the `growType` property of a text shape

In this design, the text element saying "Bounce parade" uses the "fixed" grow type, evident from the mismatch between the bounding box and the actual text bounds.

Image

If we now update the growType and inspect the width/height of the bounding box, the result depends on the time that has passed between the property change and the read operation.
Reading the bounding box immediately will typically show no change of the bounding box.
Waiting 100ms, however, will reliably return the updated bounding box.

Given that the text element is selected in Penpot, this code snippet demonstrates this:

```typescript
// You can execute this in the MCP server's API REPL

const shape = penpot.selection[0];

const getState = () => {
  return {growType: shape.growType, width: shape.width, height: shape.height};
};

const results = [];

// observe initial state
results.push(getState());

// apply change
shape.growType = "auto-height";

// read state immediately
results.push(getState());

// read state again after 100 ms
await new Promise(resolve => setTimeout(resolve, 100));
results.push(getState());

return results;
``` 

The result:
```json
[
  {
    "growType": "fixed",
    "width": 135.00000000000006,
    "height": 20
  },
  {
    "growType": "auto-height",
    "width": 135.00000000000006,
    "height": 20
  },
  {
    "growType": "auto-height",
    "width": 135.00000000000006,
    "height": 96
  }
]
``` 

Notice that while the `growType` property itself changes immediately, the `height` does not change if we have not waited long enough.

### LINKED ISSUE #10136: Add waitForLayoutUpdate method to plugin API
### Description

Plugins can now call `waitForLayoutUpdate()` to obtain a promise that resolves once all pending layout operations — including flex/grid layout recalculations, WASM text resize, and font-change reflow — have completed. This is useful for plugins that make multiple sequential changes (e.g., set text content, modify sizing) and need to wait for the UI layout to settle before taking the next action, such as capturing a screenshot or reading final geometry.

### Use case

Plugin authors who need to:

- Apply several shape modifications and wait for the layout to fully stabilize before proceeding.
- Capture accurate screenshots or export operations after layout-affecting changes.
- Sequence operations that depend on final computed layout values.

### Example

```ts
await penpot.waitForLayoutUpdate(5000);
// Layout is now settled; safe to read bounds or export.
```

### Affected versions

2.17.0

### LINKED ISSUE #10207: Text width and height stay stale after setting growType in the Plugin API
## Steps To Reproduce

1. From the plugin:
   ```js
   const t = penpot.createText("A label that should size to its content");
   t.fontSize = "16";
   t.growType = "auto-width";
   return { width: t.width, height: t.height }; // not yet the measured size
   ```
2. Read `t.width` again in a later execution — now it is correct.

## Expected behavior

After setting `growType` to `"auto-width"` / `"auto-height"`, the shape's `width`/`height`/`textBounds` should reflect the measured content size, or the API should provide an awaitable way to get the settled size.

## Actual behavior

The recalculated size is not available synchronously. Immediately after setting `growType`, `width`/`height` report stale values; the correct size only appears on a later execution (the docs mention ~100ms). Because there is no awaitable signal, any layout math performed in the same pass (e.g. centering a label by `x = cx - t.width/2`) uses the wrong width and mis-positions the element.

Additionally, `resize()` forces `growType` back to `"fixed"`, so the auto setting must be re-applied after every resize — easy to miss and interacts badly with the delay.

## Screenshots or video

N/A — stale numeric values; visible as mis-centered text if used immediately.

## Desktop

- OS: Linux (Ubuntu 22.04)
- Browser & version: [Brave 1.91.172 (Official Build) (64-bit)](https://brave.com/latest/)
Chromium: 149.0.7827.115

## Smartphone

N/A

## Environment

- Host: self-hosted Penpot instance ("PRE"); also expected on https://design.penpot.app
- Surface: Penpot Plugin API (`Text.growType`, `Text.width/height/textBounds`)
- Penpot version: 2.16

## Frontend Stack Trace

```console
No exception — stale measurements.
```

## Backend Stack Trace

```console
N/A
```

## Additional context

Suggested fix: recompute text bounds synchronously on `growType`/`characters`/`fontSize` change, or expose `await text.measure()` / a promise that resolves with settled bounds. Related: "stale reads after a mutation".

Alotor:
### Related Ticket

Closes https://github.com/penpot/penpot-mcp/issues/27

### Checklist

Closes #10136

### DIFF
--- a/frontend/src/app/main/data/workspace/reflow.cljs
+++ b/frontend/src/app/main/data/workspace/reflow.cljs
@@ -8,13 +8,14 @@
   "Tracks the shape ids that have layout/reflow work in flight, broken down by
   the kind of work so we can tell which type of reflow is blocking each shape.
 
-  Pending work is stored as a nested refcount map `{shape-id -> {kind -> count}}`:
-  a shape is pending while it has at least one kind with a count greater than
-  zero. The count (instead of a plain set) is required because several reflow
-  operations of the same kind can target the same shape while an earlier one is
-  still in flight; each `mark-pending!` must be balanced by exactly one
-  `mark-done!` of the same kind and ids so an operation finishing doesn't resolve
-  waiters that are still blocked on another operation for the same shape.
+  Pending work is stored as `{shape-id -> {kind -> #{task-id}}}`. Every producer
+  opens an exact task with `start!` and closes that same task with `finish!`.
+  Tasks belong to a workspace generation, so a delayed completion from a
+  finalized workspace cannot drain work opened after the workspace reloads.
+
+  Observable producers should use `with-pending`; imperative renderer work
+  should use `run-pending!`. Direct task lifecycle calls are reserved for Potok
+  pipelines whose operation has to cross event or batching boundaries.
 
   Kinds correspond to the pipelines that schedule the work:
     :layout        flex/grid layout reflow      (shape-layout)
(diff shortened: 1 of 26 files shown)


frontend/src/app/main/data/workspace/shape_layout.cljs
```diff
@@ -119,6 +119,8 @@
                                            :undo-group undo-group}))))
           (rx/empty))))))
 
+(defonce layout-pending (atom false))
```
niwinz: Would be nice to have a concise comment, because looking at this code it is not evident that this is used from plugins.

frontend/src/app/main/data/workspace/texts.cljs
```diff
@@ -391,11 +391,13 @@
                         (update-content-range start end attrs))]
     (assoc shape :content new-content)))
 
+(defonce font-pending (atom nil))
```
niwinz: The same here

niwinz (changes requested):

frontend/src/app/main/ui/workspace/shapes/text/viewport_texts_html.cljs
```diff
@@ -27,9 +27,18 @@
...
    [promesa.core :as p]
    [rumext.v2 :as mf]))
 
+;; Re-emit browser-side font loads as :font-loaded store events so consumers
+;; filtering with `font-loaded-event?` (app.main.data.workspace.texts) also
+;; wake up for fonts loaded through `fonts/ensure-loaded!`, not only for the
+;; ones loaded by the wasm font store.
+(set! fonts/on-loaded-font
+      (fn [font-id]
+        (st/emit! (ptk/data-event :font-loaded {:font-id font-id}))))
```
niwinz: This is set at namespace level, so it will trigger always. So in this case why just not put it on app.main.fonts namespace on the on-load and use namespaced key for the event name.

frontend/src/app/main/data/workspace/reflow.cljs
```diff
@@ -0,0 +1,117 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC
```
niwinz, line 5: Incorrect copyright

niwinz (approved):

PR #9906, opened 2026-05-27
:sparkles: Add the ability to launch parallel devenv instances

### LINKED ISSUE #9777: Enable Parallel Devenvs

MischaPanch:
### Related Ticket

https://github.com/penpot/penpot/issues/9777

### Brief Summary by Human (Mischa)

The PR brings in convenient startup of multiple instances and related agent ergonomics. The documentation and manage.sh CLI has been expanded and reworked around that. With just two commands you get a complete setup (for supported agents):

```bash
./manage.sh run-devenv-agentic --ws 1
./manage.sh start-coding-agent claude
# in a new tab
./manage.sh start-coding-agent claude --ws 1
```

The `start-coding-agent` command automatically wires the devenv's MCP servers into claude code (in addition to any custom config that you have for this agent). It was also set up for vscode, opencode and codex, but I haven't fully tested it with these agents. This feature should be considered in alpha stage, but it's very useful, as otherwise one has to manage agent config in a cumbersome way.

 What I have tested:
 -  things work as before for a single devenv
 - two devenvs start up correctly and work well
 - all new CLI commands and flags
 - MCP config wiring for claude code
 
I cannot properly test parallelized work on two devenvs because they together with two penpot intances open in chrome tabs exhaust my RAM. I don't see what could prevent it from working (given what I already tested), but someone should test this yourself on a more capable machine - maybe one of you can do this @opcode81 @Alotor @niwinz ?

The quickest way to see how to work with this is to read the updated developer docs or to just read the manage.sh usage info.

### Design Doc

The text below was created and maintained in collaboration with various AI agents, it's a technical summary of the implementation.

 # Parallel Penpot Devenvs — Design and Implementation Summary

A summary of the new "parallel devenv" feature for circulating with the team.

## Motivation

A single developer plus multiple agents need to work against Penpot in parallel without stepping on each other's source edits, builds, ports, tmux sessions, Valkey Pub/Sub, or MCP/plugin connections. The previous single-project devenv only supported one running instance and assumed exclusive use of a fixed set of host ports and container names.

## What you get

- N parallel runtime instances of Penpot on one developer machine.
- One shared infrastructure stack (Postgres, MinIO, mailer, LDAP) underneath them.
- One shared Penpot database and one shared MinIO bucket, so users, teams, files, MCP tokens, and onboarding state created in any instance are visible from every instance.
- Per-instance source checkout, build outputs, Valkey, tmux session, and host-port range.
- Same-origin MCP routing (`<public-uri>/mcp/ws`) so the in-browser plugin always reaches the right MCP server without per-tab configuration.
- A single-instance CLI (`run-devenv-agentic [--ws N]`) that brings up one agentic devenv at a time, auto-starting ws0 first when a ws1+ target is requested.
- Auto-generated per-workspace MCP-client configs for four AI coding agents (Claude Code, opencode, VS Code Copilot, Codex CLI) plus a `start-coding-agent` launcher. The generated config *enhances* the developer's existing global client config rather than replacing it.
- A git identity wired into each container on bring-up, so commits made by an agent inside the devenv carry a real author/committer instead of the bare `penpot@<container>` fallback.

## Architecture

Two layers of `docker compose` projects:

- **`penpotdev-infra`** — shared services: `postgres`, `minio`, `minio-setup` (one-shot user/policy provisioner), `mailer`, `ldap`. Brought up once and reused by every running instance.
- **`penpotdev-wsN`** (N = 0, 1, …) — one project per runtime instance, each containing its own `main` container and Valkey. `ws0` binds the live repository (`$PWD`); each `wsN` (N ≥ 1) binds a workspace clone under `~/.penpot/penpot_workspaces/wsN/`.

All projects join a single external Docker network, `penpot_shared`, created idempotently by `manage.sh`. The network is never removed by lifecycle commands.

Configuration lives in `docker/devenv/defaults.env` (the ws0 baseline) plus auto-generated per-instance overlays under `docker/devenv/instances/wsN.env` (gitignored). Backend-internal defaults (secret keys, `PENPOT_FLAGS`, `JAVA_OPTS`, MinIO bucket setup) remain in `backend/scripts/_env`.

### Workspace clones

Workspaces for `wsN` (N ≥ 1) are scratch clones of the live repo: seeded from the developer's current working tree on first bring-up, and re-seeded on demand with `run-devenv-agentic --ws N --sync`. The sync uses `git ls-files --cached --others --exclude-standard` to enumerate exactly the files Git considers part of the working tree (this matters: rsync's gitignore filter would drop committed files under gitignored parent directories, like `.clj-kondo/config.edn`). Workspace `.git` is mirrored separately. Gitignored caches in the workspace (`node_modules`, `target`, `.shadow-cljs`, etc.) survive between syncs.

Before any sync, manage.sh refuses if the live repo is in a fragile Git state (rebase / merge / cherry-pick in progress, or `.git/index.lock` present). Hard refusal, not warning.

Each workspace switches to a stable per-workspace branch (`wsN/<current-branch>`) immediately after seeding, so commits made inside a workspace can be pushed under a unique name without colliding with the live repo's branch.

### Port layout

Container-internal ports stay fixed across all instances. Host ports are offset by `10000 × N`:

| Service           | ws0   | ws1    | ws2    |
| ----------------- | ----- | ------ | ------ |
| Penpot UI (HTTPS) | 3449  | 13449  | 23449  |
| MCP HTTP stream   | 4401  | 14401  | 24401  |
| MCP REPL          | 4403  | 14403  | 24403  |
| Serena MCP        | 14181 | 24181  | 34181  |
| Serena dashboard  | 14182 | 24182  | 34182  |

Everything else (frontend dev server, backend API, exporter, Storybook, REPLs, plugin dev, MCP inspector, MCP WebSocket) is in-process or reachable same-origin through the per-instance Caddy/nginx on the public HTTPS port. Infra services publish singleton ports: mailer 1080, LDAP 10389/10636.

The port bases (3449, 4401, 4403, 14181, 14182) and the 10000-wide stride live as named shell constants (`PENPOT_PORT_BASE_*`, `PENPOT_INSTANCE_PORT_STRIDE`) at the top of `manage.sh`. An `instance-port <instance> <base>` helper does the `base + N × stride` math; `write-instance-env` (which bakes the values into the per-instance Compose env-file) and `print-instance-info` (which prints the URLs at startup) both go through it, so there is exactly one place to change a port number.

## CLI

```
./manage.sh run-devenv-agentic [--ws N] [--sync] [--serena-context CTX] \
                               [--git-user-name NAME] [--git-user-email EMAIL]
```

Brings up a single agentic instance (MCP + Serena are always on). `--ws` takes a non-negative integer (0..5; the upper bound keeps the offset ports inside the 16-bit range), default `0`. The command:

1. Fails fast if `frontend/resources/public/js/config.js` is missing in the live repo (gitignored; carries the `enable-mcp` frontend flag — read directly on ws0, seeded into wsN on initial sync). This runs before any side effects.
2. Ensures the shared infra stack is up and `minio-setup` has completed.
3. If the target is ws1+ and ws0 is not already running, brings ws0 up first (workers must run on ws0; see *Worker policy* below).
4. For ws1+, performs an initial workspace sync from the live repo when the workspace directory does not yet exist; `--sync` re-seeds it unconditionally. `--sync` is rejected on ws0 (it binds the live repo directly).
5. Generates the per-instance Compose env-file overlay (ws1+ only) and the per-workspace AI-client MCP configs (committed templates + envsubst pass; see *AI coding-agent integration* below), brings the project up with `up -d --no-deps main redis`, and starts the tmux session detached.
6. Wires a git author/committer identity into the container's global config (defaults to the host's effective `git config user.{name,email}`; `--git-user-name` / `--git-user-email` override).
7. Errors out if the requested target is already running.
8. Prints the per-instance URLs (Penpot UI, MCP stream, Serena MCP, Serena dashboard), the attach command, and the `start-coding-agent` invocation.

To run more than one instance, invoke `run-devenv-agentic` once per target (e.g. once for ws0, then once with `--ws 1`). There is no single "reconcile to N instances" command.

Every `--ws` flag accepts a bare non-negative integer only (`--ws 0`, `--ws 1`, …); `main`, `ws0`, `ws1` spellings are rejected so the flag shape is uniform across commands.

Other commands:

- `./manage.sh start-coding-agent <claude|opencode|vscode|codex> [--ws N] [args…]` — launch the chosen AI coding agent against a workspace with the right MCP-config wiring (Claude Code: `--mcp-config`; opencode: `OPENCODE_CONFIG=`; VS Code: auto-discovers `.vscode/mcp.json`; Codex: auto-discovers `.codex/config.toml` for trusted projects). `cd`s into the target workspace itself and refuses to launch if the instance is not running. Extra args are forwarded to the client. See *AI coding-agent integration*.
- `./manage.sh attach-devenv [--ws N]` — pure attach to an instance's tmux session. Fails fast if the instance or session is missing.
- `./manage.sh run-devenv-shell [--ws N] [cmd…]` — bash shell in the target instance's running `main` container (`docker exec`). Defaults to `bash` and lands in the source-tree root.
- `./manage.sh run-devenv [-e VAR=value …]` — ws0-only, non-agentic tmux, attached. Brings ws0 up first if needed. Takes no workspace flag.
- `./manage.sh isolated-shell [--ws N] [cmd…]` — fresh one-shot `docker run` (not `docker exec`) against the target workspace's user-data volume and source tree; the workspace need not be running.
- `./manage.sh stop-devenv [--ws N | --all]` — stop instances. With no flag (or `--ws 0`), stops ws0 plus shared infra, refusing if any ws1+ is still running. `--ws N` (N≥1) stops just that workspace. `--all` stops every wsN highest-first, then ws0, then infra.
- `./manage.sh drop-devenv [--ws N | --all]` — same CLI and invariants as `stop-devenv`, plus removal of the shared devenv image on full teardowns (no flag, `--ws 0`, or `--all`). A single `--ws N` (N≥1) keeps the image. Never removes data volumes.
- `./manage.sh start-devenv` — start ws0 + infra in the background (no tmux, no MCP/Serena).
- `./manage.sh log-devenv [--ws N]` — tail a workspace's compose logs (default ws0).

## State-sharing model

| State                                                              | Sharing                  |
| ------------------------------------------------------------------ | ------------------------ |
| Postgres database                                                  | Shared across all wsN    |
| MinIO bucket (`penpot`)                                            | Shared across all wsN    |
| Users, teams, files, access tokens, MCP tokens, onboarding state   | Shared (in the DB)       |
| Valkey (Redis-compatible) Pub/Sub                                  | Per-instance container   |
| Backend workers (cron + async task runners)                        | ws0 only (see below)     |
| Source checkout                                                    | ws0 = live repo; ws1+ = scratch clones |
| Build outputs (`node_modules`, `target`, `.shadow-cljs`, …)        | Per-workspace            |

This is intentional: code/build/runtime is isolated per instance, but Penpot product data is shared so a developer doesn't need to recreate users, tokens, and sample files for each instance.

The main caveat is concurrent writes to the same Penpot file from different instances. Penpot uses incrementing file versions, so stale writes should fail rather than silently overwriting, but the workflow recommendation is: agents should not intentionally edit the same Penpot file through different backend instances unless conflict behavior is the test.

## Worker policy and the ws0 invariant

Backend workers (the `enable-backend-worker` flag) handle scheduled cron tasks (DB/storage GC, telemetry, audit-log archive) and ad-hoc async work submitted by RPC handlers (`:delete-object` cascades for files/teams/projects, webhook delivery, `:sendmail`, file-data offloading). Submission is fire-and-forget: `wrk/submit!` inserts a row into the shared Postgres `task` table and returns; RPC handlers never poll, and workers don't publish completion notifications anywhere.

Workers run only on ws0: ws0 keeps `enable-backend-worker`, ws1+ overlays set `PENPOT_BACKEND_WORKER=false`. ws1+ RPC handlers still submit into the shared `task` table — ws0's dispatcher claims via `FOR UPDATE SKIP LOCKED` and runs the work against the shared DB and MinIO. The CLI enforces a matching lifecycle invariant: **ws0 is always running whenever any ws1+ is**. `run-devenv-agentic --ws N` (N≥1) auto-starts ws0 first if it isn't up; `stop-devenv` refuses to stop ws0 while any ws1+ is still running.

The reason for the single worker-bearer is to avoid cross-instance worker races. The Postgres-level row-locking on `task` is safe for arbitrarily many workers, but the cron dedup primitive (`SELECT FOR UPDATE SKIP LOCKED` on `scheduled_task` for the duration of the task body) is best-effort — two cron timers firing the same scheduled instant with a gap larger than the body's runtime can both execute it. For most cron entries this is harmless (the GC tasks are idempotent), but `:telemetry` and `:audit-log-archive` would double-fire. Keeping workers on one instance side-steps the question entirely.

## MCP routing

Each Penpot frontend computes its plugin's MCP WebSocket URL as `<public-uri>/mcp/ws`. The per-instance nginx proxies that path to the MCP server running in the same `main` container on the fixed internal port 4401. Because each browser tab opens against a different `<public-uri>` (one per instance), the hidden plugin automatically connects to the right MCP server with no per-tab configuration.

External AI clients (Claude Desktop, Claude Code) connect directly to the per-instance MCP HTTP stream endpoint, `http://localhost:<4401 + 10000·N>/mcp`. The startup output prints these URLs per instance.

## AI coding-agent integration

Bringing a workspace up also produces everything an AI coding agent needs to drive that specific workspace with a single command.

**Per-workspace MCP-client configs.** A committed top-level `.devenv/` directory holds templates (`templates/{claude-code,opencode,vscode}.json`, `templates/codex.toml`, each with `${PENPOT_MCP_PORT}` / `${SERENA_MCP_PORT}` placeholders) plus shared, workspace-independent entries (`shared/*` — currently just the Playwright server). On every `run-devenv-agentic` pass, `write-instance-mcp-configs` merges shared + the port-substituted template per tool (via `.devenv/scripts/merge-mcp-config.py` — JSON deep-merge for the three JSON schemas, textual concat for Codex's TOML) and writes the result where each client expects it:

| Client          | Generated file                        | Loaded via                  |
| --------------- | ------------------------------------- | --------------------------- |
| Claude Code     | `<ws>/.devenv/mcp/claude-code.json`   | `--mcp-config` flag         |
| opencode        | `<ws>/.devenv/mcp/opencode.json`      | `OPENCODE_CONFIG=` env var  |
| VS Code Copilot | `<ws>/.vscode/mcp.json`               | auto-discovered             |
| Codex CLI       | `<ws>/.codex/config.toml`             | auto-discovered (trusted)   |

All generated outputs are gitignored; only the templates and shared entries are tracked.

**Launcher.** `./manage.sh start-coding-agent <client> [--ws N]` `cd`s into the target workspace, refuses to launch if that instance is not running, and exec's the client with the right flag/env. The generated config is loaded *on top of* the developer's global client config — their existing Playwright/GitHub/etc. entries stay active alongside the Penpot + Serena entries we ship (each client merges by its own precedence rules; the agentic-devenv guide documents the per-client override paths).

**Git identity.** `run-devenv-agentic` writes a git author/committer identity into the container's global config on every bring-up so commits made inside the devenv aren't attributed to the bare `penpot@<container>` fallback. The default is the host's effective `git config user.{name,email}` (local repo override wins over `~/.gitconfig`); `--git-user-name` / `--git-user-email` override it.

## Migration steps for colleagues

The previous devenv ran as a single compose project (`penpotdev`) with containers `penpot-devenv-main` and `penpot-devenv-valkey`. The new layout uses `penpotdev-infra` plus one `penpotdev-wsN` per instance. When you pull the new code, you must remove the legacy containers and the auto-generated network once. The named data volumes (`penpotdev_postgres_data_pg16`, `penpotdev_minio_data`, `penpotdev_user_data`, `penpotdev_valkey_data`) are pinned by explicit name in the new compose files and survive the rename — your Postgres DB, MinIO objects, and home cache are preserved.

```bash
# Stop and remove the old single-project containers.
docker stop penpot-devenv-main penpot-devenv-valkey 2>/dev/null
docker rm   penpotdev-postgres-1 penpotdev-minio-1 penpotdev-minio-setup-1 \
            penpotdev-mailer-1   penpotdev-ldap-1 \
            penpot-devenv-main   penpot-devenv-valkey 2>/dev/null

# Remove the orphaned auto-generated network.
docker network rm penpotdev_default 2>/dev/null

# Bring up infra + ws0 under the new project layout.
./manage.sh run-devenv-agentic
```

If you previously ran `./manage.sh build-devenv --local`, no rebuild is required for the parallel feature: `start-tmux.sh` and the devenv image already contain the updates from the prior compose-split work.

## What was actually built

In rough order of dependency:

1. **Compose split.** `docker-compose.yaml` was decomposed into `docker-compose.infra.yml` (shared) and `docker-compose.main.yml` (per-instance), joined by the external `penpot_shared` network.
2. **defaults.env.** A single env file for the ws0 baseline replaced the inline default values that were previously scattered across `_env`, compose, and `manage.sh`. Compose substitutes from `--env-file`; backend reads from the container env.
3. **Volume name pinning.** Each volume is given an explicit `name:` in compose, decoupled from `COMPOSE_PROJECT_NAME`. ws0 keeps the pre-existing physical names so the project rename didn't require a data migration; ws1+ overlay env files set distinct names.
4. **Network alias removal.** Removed `aliases: - main` / `- redis` from `docker-compose.main.yml` to avoid same-name DNS collisions across instances on the shared network. Cross-container DNS uses container_name; backend's `PENPOT_REDIS_URI` is set explicitly per instance to point at its unique Valkey container.
5. **Same-origin MCP routing.** Confirmed the frontend's `<public-uri>/mcp/ws` path lands in per-instance nginx → MCP, and removed a stale Host-matched server block from the Caddyfile that referenced the old container name.
6. **Worker gating.** Added `PENPOT_BACKEND_WORKER` env var; `_env` conditionally includes `enable-backend-worker` in `PENPOT_FLAGS` based on it. Per-instance overlays for ws1+ set it to `false`.
7. **Workspace orchestration.** New `sync-workspace` helper in `manage.sh` performs the pre-flight Git-state check, mirrors `.git` via rsync, and copies the working tree using `git ls-files`. Each workspace is then switched to its unique `wsN/<branch>` name.
8. **Per-instance env-file generation.** `write-instance-env` writes `docker/devenv/instances/wsN.env` deterministically from the instance index: project name, container names, volume names, port offsets, public URI, Redis URI, worker flag, source path.
9. **Single-instance agentic startup.** `run-devenv-agentic [--ws N]` brings one instance up (default ws0), auto-starting ws0 first whenever a ws1+ target is requested (worker invariant). The compose project comes up with `up -d --no-deps main redis` and the tmux session starts detached with MCP and Serena always enabled; the command errors out if the target is already running. `--sync` re-seeds a ws1+ workspace from the live repo; ws1+ also sync implicitly the first time when their workspace directory does not yet exist. Stopping an instance is handled by `stop-devenv --ws N` (or `--all`), which protects the ws0 invariant.
10. **Compose helper hardening.** `infra-compose` and `instance-compose` wrap `docker compose` with `env -i` to strip the shell env before invocation. Without this, the shell env from sourcing `defaults.env` at startup would shadow per-instance overlay `--env-file` (Compose gives shell precedence over `--env-file`).
11. **`PENPOT_DATABASE_MAX_POOL_SIZE` lowered** from 60 to 20 per instance, so the connection budget fits N instances within Postgres's default `max_connections=100` with headroom for admin sessions.
12. **Tmux session idempotence.** `start-tmux.sh` was made idempotent earlier in the project: if the target session exists it attaches (or exits when `PENPOT_TMUX_ATTACH=false`); otherwise it creates the full window set including optional MCP and Serena windows. MCP/Serena windows are only added on session create.
13. **CLI ergonomics.** Every workspace-targeting command takes `--ws N` (integer-only, validated by `parse-ws-integer`, 0..5): `run-devenv-agentic`, `attach-devenv`, `stop-devenv`, `drop-devenv`, `log-devenv`, `run-devenv-shell`, `start-coding-agent`, `isolated-shell`. `run-devenv` is ws0-only (no workspace flag). Strict argument parsing rejects unknown flags. Per-instance startup prints URLs (Penpot UI, MCP stream, Serena MCP, Serena dashboard), the attach command, and the `start-coding-agent` invocation.
14. **Caddy reverse proxies same-container traffic via `localhost`/`127.0.0.1`** (nginx already did). No baked-in container-name references remain, so the image is reusable across instances without templating.
15. **Port-allocation single source of truth.** Per-service port bases and the 10000-wide stride live as named shell variables at the top of `manage.sh` (`PENPOT_PORT_BASE_PUBLIC`, `_MCP`, `_MCP_REPL`, `_SERENA`, `_SERENA_DASHBOARD`, `PENPOT_INSTANCE_PORT_STRIDE`). An `instance-port <instance> <base>` helper does the math, and both `write-instance-env` (per-instance Compose env-file) and `print-instance-info` (startup URLs) consume it — one place to change a port number.
16. **AI-client MCP config plumbing.** A new `.devenv/` directory holds committed templates (`templates/{claude-code,opencode,vscode}.json` and `templates/codex.toml` with `${PENPOT_MCP_PORT}` / `${SERENA_MCP_PORT}` placeholders) and committed shared entries (matching `shared/*` files for Playwright). A new `write-instance-mcp-configs` helper merges shared + envsubst'd templates per tool on every `run-devenv-agentic` pass — via `.devenv/scripts/merge-mcp-config.py` (JSON deep-merge for the three JSON formats, textual concat for Codex's TOML) — writing to `<workspace>/.devenv/mcp/{claude-code,opencode}.json` (loaded via `--mcp-config` / `OPENCODE_CONFIG=`), `<workspace>/.vscode/mcp.json` (auto-discovered by Copilot), and `<workspace>/.codex/config.toml` (auto-discovered by Codex for trusted projects). All generated paths are gitignored. A new `./manage.sh start-coding-agent <claude|opencode|vscode|codex> [--ws N]` wrapper `cd`s into the target workspace, refuses to launch if the instance is not running, and launches the corresponding client with the right flag/env wiring; the generated config enhances the developer's global client config rather than replacing it. Manual configuration is still documented in `docs/technical-guide/developer/agentic-devenv.md` as the path for clients we don't cover. See the *AI coding-agent integration* section above.
17. **Container git identity.** `run-devenv-agentic` seeds the container's global git config from the host on every bring-up (`start-instance` runs `docker exec … git config --global user.{name,email}`), so commits made inside the devenv carry a real author. Default is the host's effective `git config user.{name,email}`; `--git-user-name` / `--git-user-email` override.
18. **config.js pre-flight.** `run-devenv-agentic` fails fast (before any side effects) if `frontend/resources/public/js/config.js` is missing in the live repo — the gitignored file carries the `enable-mcp` frontend flag, is read directly on ws0, and is seeded into a wsN clone only on its initial sync.

## Verification status

Verified live end-to-end:

- Migration from the legacy single-project layout to the split projects, with all four data volumes preserved.
- N=1 (ws0 alone): bring-up, all 6 tmux windows, backend at `https://localhost:3449`, MCP at 4401, Serena at 14181 (dashboard 14182), service discovery for shared infra.
- N=2 (ws0 + ws1): ws1 workspace at `~/.penpot/penpot_workspaces/ws1` on branch `ws1/feature/...`, container `penpot-devenv-ws1-main`, ports offset to 13449/14401/14403/24181/24182, ws0 left untouched.
- Valkey isolation: a key written to one instance's Valkey is invisible to the other's.
- Shared DB and MinIO: profile count and bucket contents identical from both instances.
- Scale-down N=2 → N=1: ws1 containers stopped, workspace directory and named volumes preserved, ws0 unchanged.
- `attach-devenv --ws 1` fail-fast after scale-down.
- Strict arg parsing (unknown flags rejected with clear errors; `run-devenv-agentic` refuses to run against an already-running target and refuses `--sync` on ws0).
- Published port surface exactly the five expected mappings per instance.

Code-reviewed but not exercised live (would have required disturbing the running state):

- `stop-devenv` graceful all-down + restart loop.
- `create-devenv` (rarely-used path).
- `drop-devenv` non-destructive variant (data-preservation rule precludes live testing).
- `run-devenv-isolated-shell` (one-shot docker run, no compose involvement; see audit doc — this command may need rework anyway).

## Known trade-offs and limitations

- **Async-task notifications miss ws1+ tabs.** A direct consequence of the workers-on-ws0-only policy. See *Worker policy* above.
- **Concurrent edits of the same Penpot file from different backend instances** are not safe in general. Penpot's version counter should make stale writes fail rather than silently overwriting, but the recommendation is to avoid this unless testing conflict behavior.
- **No locking on `manage.sh` invocations.** Two simultaneous `./manage.sh run-devenv-agentic` runs against the parallel set are out of scope; the CLI assumes a single developer at the wheel.
- **Schema drift across workspaces.** If the developer commits a migration on the live repo while ws1 is running, ws1's older snapshot of backend code may race against the shared DB schema. Out of scope; workspaces sync from the live repo so they're schema-aligned at the moment of their last sync.
- **Workspace sync is one-way.** Edits made inside a ws1+ workspace are lost on the next sync (the workspace is a scratch clone, by design). The unique branch (`wsN/<base-branch>`) lets a workspace commit and push, but the next `--sync` will overwrite uncommitted edits.
- **Storybook is not host-published.** It still runs in-tmux for parity, but external access to it from the host requires either `docker exec` in or manually publishing port 6006.

## Where to read more

- **Developer guides** — `docs/technical-guide/developer/devenv.md` and `docs/technical-guide/developer/agentic-devenv.md` include the user-facing parts (parallel workspaces section, port table, agentic-mode flags, upgrade steps).
- **Agent memory** — `mem:devenv/core` carries the dense reference (project layout, invariants, port table, worker policy, workspace orchestration, CLI surface). Trigger conditions for reading it live in `mem:critical-info`.
- **MCP wiring** — `mem:mcp/core` has the plugin↔server connection model (how the frontend computes the URL, the in-memory connection registry, why DB-mediated routing isn't needed).


docker/devenv/defaults.env
```diff
@@ -0,0 +1,69 @@
...
+# storybook, exporter, REPLs, plugins, MCP inspector/websocket, aux) is
+# accessed in-process or through the same-origin Caddy/nginx proxy at
+# PENPOT_PUBLIC_HTTP_PORT. Container-internal ports remain fixed; per-instance
+# overlays may offset these host-side values.
+PENPOT_PUBLIC_HTTP_PORT=3449
+PENPOT_MCP_SERVER_PORT=4401
+PENPOT_MCP_REPL_PORT=4403
+
+# Serena (agentic devenv). Internal ports are fixed by Serena itself.
+SERENA_EXTERNAL_PORT=14281
```
MischaPanch: Note for self: these are probably unused, check it

docker/devenv/files/start-tmux.sh
```diff
@@ -6,6 +6,23 @@ cd ~;
 
 source ~/.bashrc
 
+PENPOT_TMUX_SESSION="${PENPOT_TMUX_SESSION:-penpot}"
```
MischaPanch: Note for self: maybe changes can/should be rolled back

docs/technical-guide/developer/devenv.md
```diff
@@ -45,13 +45,149 @@ This is an incomplete list of devenv related subcommands found on
 manage.sh script:
 
 ```bash
-./manage.sh build-devenv-local # builds the local devenv docker image (called by run-devenv automatically when needed)
-./manage.sh start-devenv       # starts background running containers
-./manage.sh run-devenv         # enters to new tmux session inside of one of the running containers
-./manage.sh stop-devenv        # stops background running containers
-./manage.sh drop-devenv        # removes all the containers, volumes and networks used by the devenv
+./manage.sh build-devenv-local # builds the local devenv docker image
```
niwinz: it should be `./manage.sh build-devenv --local`

docker/devenv/defaults.env
MischaPanch: Removed duplication/overrides of values here in manage.sh

docs/technical-guide/developer/devenv.md
MischaPanch: Done

docker/devenv/files/start-tmux.sh
MischaPanch: Removed everything that sets PENPOT_TMUX_SESSION outside of here

docs/technical-guide/developer/agentic-devenv.md
```diff
@@ -5,18 +5,74 @@ desc: Dive into agentic Penpot development.
...
+
+4. **Launch your AI client** against the workspace you want it to drive:
+
+   ```bash
+   ./manage.sh start-coding-agent claude              # ws0
+   ./manage.sh start-coding-agent claude --ws 1       # ws1
+   ```
+
+   Supported clients: `claude` | `opencode` | `vscode` | `codex`. The
+   launcher loads a per-workspace MCP config *on top of* your global config.
```
opcode81: This makes it sound as if all other clients are unsupported. We should mention the manual configuration option.

MischaPanch: There is a section on the manual option below

opcode81: I know, but this doesn't change the fact that this part, which is read first, is misleading.

opcode81: Also, doesn't ws0 (the default) have to be excluded from the auto-configuration? After all, ws0 has the user's own configuration, which is not safe to overwrite.

MischaPanch: Adjusted the wording here and linked to the manual config section.

Regarding your question: this was adjusted to inject additional config instead of changing files for all clients except vscode. There, on ws0 we merge the new config into the user's project-level .vscode/mcp.json.

The mechanism is now explained in the docs, mentioning the behavior for vscode on ws0

PR #9921, opened 2026-05-28
:recycle: Refactor font upload to process variants sequentially

### LINKED ISSUE #9922: Prevent concurrent font uploads from causing excessive simultaneous requests
### Description

Batch uploading fonts no longer sends all font variant uploads simultaneously, preventing excessive simultaneous upload requests. Font variants are now processed one at a time.

### Use case

Users who upload multiple fonts at once will experience more reliable uploads without overwhelming the server with concurrent requests.

### Affected versions

2.15.3

niwinz:
## Summary

Change the batch upload handler so fonts are uploaded one at a time instead of all concurrently, preventing excessive simultaneous upload requests.

Previously `on-upload-all` used `run!` which fired all font variant uploads simultaneously. Now it uses `rx/from` combined with `rx/mapcat` to process each font sequentially.

As part of this change, extract the upload logic into a standalone `handle-font-upload` helper for reuse between single and batch upload paths, and remove the separately memoized `on-upload*` hook.

Also fix error logging to use `js/console.error` instead of `js/console.log` for consistency with project conventions.

Closes #9922

## How to Test

- Upload multiple fonts at once, you should verify that fonts are uploaded one by one
- Upload single font, should work as before, with no regression

### DIFF
--- a/frontend/src/app/main/ui/dashboard/fonts.cljs
+++ b/frontend/src/app/main/ui/dashboard/fonts.cljs
@@ -51,6 +51,13 @@
   (and (contains? font :font-family-tmp)
        (str/blank? (:font-family-tmp font))))
 
+(defn- handle-font-upload
+  [{:keys [id] :as item}]
+  (->> (df/upload-font-variant item)
+       (rx/delay-at-least 2000)
+       (rx/map (fn [font]
+                 [id font]))))
+
 (mf/defc header*
   {::mf/props :obj
    ::mf/memo true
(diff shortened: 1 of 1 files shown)


frontend/src/app/main/ui/dashboard/fonts.cljs
```diff
@@ -51,6 +51,13 @@
   (and (contains? font :font-family-tmp)
        (str/blank? (:font-family-tmp font))))
 
+(defn- handle-font-upload
+  [{:keys [id] :as item}]
+  (->> (df/upload-font-variant item)
+       (rx/delay-at-least 2000)
```
Alotor, line 60: Is there a reason for this delay? if so, i think a comment would be necessary to explain it

niwinz: done

Alotor (approved):

PR #9943, opened 2026-05-29
:sparkles: Batch multiple thumbnail deletions into a single RPC call

### LINKED ISSUE #9944: Batch multiple thumbnail deletions into a single RPC call
### Description

When clearing thumbnails for multiple shapes (e.g. when shapes are deleted or modified in quick succession), the system now batches those deletion requests into a single RPC call instead of sending one request per shape. This eliminates bursts of requests and reduces the total number of RPC calls needed.

### Use case

Operations that clear thumbnails for several objects at once — such as batch-deleting shapes, file operations, or rapid shape modifications — previously triggered a separate RPC request for each object, causing a burst of network traffic. With this change, the frontend accumulates deletion requests over a short debounce window and flushes them together in a single batched call, reducing server load and improving responsiveness.

### How to Test

- Create a component
- Create two copies
- Update the color of main component
- Look at devtools network, you should see a single request to `delete-file-object-thumbnails` instead of multiple `delete-file-object-thumbnail`
- The thumbnails works as before

### Affected versions

All versions (no user-facing version-specific impact)

niwinz:
## Summary

Replace the old per-object immediate thumbnail deletion with a debounced batched approach. Previously, clearing thumbnails for multiple shapes sent one RPC request per object-id, causing bursts of requests in a short amount of time. Now the frontend queues deletions and flushes them in a single batched RPC call, reducing the number of requests and eliminating request bursts.

**DO NOT MERGE**: just approve it, we will wait a bit after release for merging this

### Frontend (`frontend/src/app/main/data/workspace/thumbnails.cljs`)
- **`clear-thumbnail`** no longer sends an RPC immediately. Instead it adds the object-id to a new `::thumbnails-deletion-queue` in application state and starts a 200ms debounce timer.
- **`clear-thumbnail-batch`** (internal event) drains the queue, partitions object-ids into chunks of up to 200, revokes blob URIs for the cleared thumbnails, and sends one `:delete-file-object-thumbnails` RPC per chunk.
- **`remove-from-deletion-queue`** allows downstream events (e.g. `update-thumbnail`) to cancel a pending batched delete for a specific object-id before creating a new thumbnail.
- `update-thumbnail` emits `remove-from-deletion-queue` alongside `assoc-thumbnail` to avoid race conditions.
- Uses `rx/observe-on :async` to ensure each `clear-thumbnail` event lands in its own macrotask, preventing debounce timer races.

### Backend (`backend/src/app/rpc/commands/files_thumbnails.clj`)
- New **`delete-file-object-thumbnails`** RPC method (batch variant of the existing `delete-file-object-thumbnail`):
  - Accepts a vector of up to 200 object-ids.
  - Extracts unique file-ids from the object-ids and checks edit permissions for each file.
  - Runs a single SQL `UPDATE ... SET deleted_at = now() WHERE object_id = ANY(?) AND deleted_at IS NULL RETURNING media_id` — soft-deletes all matching thumbnails in one statement.
  - Touches each returned `media_id` via the storage layer so affected media objects are picked up by garbage collection.

### Tests
- **Backend** (`backend/test/backend_tests/rpc_file_thumbnails_test.clj`): 12 new test cases covering basic deletion, empty/non-existent/mixed batches, already-deleted thumbnails, unauthorized access (single file + cross-file), media touch, max batch size (200), single-item batch, duplicate object-ids in batch, and cross-file isolation.
- **Frontend** (`frontend/test/frontend_tests/data/workspace_thumbnails_test.cljs`): 14 new test cases covering queue state management (`clear-thumbnail` adds to queue, accumulates, removes entries, cancels pending deletes), `clear-thumbnail-batch` drains the queue, `assoc-thumbnail` / `duplicate-thumbnail`, and async watch tests verifying correct RPC calls, batch partitioning, blob URI revocation, empty-queue no-op, and debounced batch emission.

Closes #9944

### DIFF
--- a/backend/src/app/rpc/commands/files_thumbnails.clj
+++ b/backend/src/app/rpc/commands/files_thumbnails.clj
@@ -314,7 +314,7 @@
 (defn- delete-file-object-thumbnails!
   "Soft-deletes multiple object thumbnails in a single UPDATE statement
    with RETURNING, then touches all returned media objects."
-  [{:keys [::db/conn ::sto/storage] :as cfg} object-ids]
+  [{:keys [::db/conn ::sto/storage]} object-ids]
   (let [ids  (db/create-array conn "text" (seq object-ids))
         sql  (str/concat
               "UPDATE file_tagged_object_thumbnail"
@@ -351,6 +351,8 @@
 (sv/defmethod ::delete-file-object-thumbnails
   {::doc/added "1.19"
    ::doc/module :files
+   ::climit/id [[:file-thumbnail-ops/by-profile ::rpc/profile-id]
+                [:file-thumbnail-ops/global]]
    ::sm/params schema:delete-file-object-thumbnails
    ::audit/skip true}
   [cfg {:keys [::rpc/profile-id object-ids]}]
--- a/backend/src/app/rpc/commands/files_thumbnails.clj
+++ b/backend/src/app/rpc/commands/files_thumbnails.clj
@@ -311,10 +311,29 @@
                  :object-id object-id
                  :tag tag})))
 
+(defn- delete-file-object-thumbnails!
+  "Soft-deletes multiple object thumbnails in a single UPDATE statement
+   with RETURNING, then touches all returned media objects."
+  [{:keys [::db/conn ::sto/storage] :as cfg} object-ids]
+  (let [ids  (db/create-array conn "text" (seq object-ids))
+        sql  (str/concat
+              "UPDATE file_tagged_object_thumbnail"
+              " SET deleted_at = now()"
+              " WHERE object_id = ANY(?)"
+              "   AND deleted_at IS NULL"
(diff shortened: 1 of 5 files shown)


luisddm:
@niwinz Followed the "How to Test" directions in the issue. The RPC call to `delete-file-object-thumbnails` actually works, and visually it seems to work as expected. But looking at the aplication state, the `thumbnails-deletion-queue` key is not empty. Shouldn't it get empty after each cleaning burst?

niwinz:
> @niwinz Followed the "How to Test" directions in the issue. The RPC call to `delete-file-object-thumbnails` actually works, and visually it seems to work as expected. But looking at the aplication state, the `thumbnails-deletion-queue` key is not empty. Shouldn't it get empty after each cleaning burst?

I think yes...

luisddm (approved):

PR #9959, opened 2026-06-02
:sparkles: Avoid deduplicating temporary export files

### LINKED ISSUE #9970: Temporary export files may return stale or unavailable content
### Description

When exporting content from Penpot (e.g., images, designs), the temporary export file may return stale content from a previous export instead of the freshly generated file. In some cases, the temporary file URL may be entirely unavailable because the backing storage object was garbage-collected after a prior export reused it.

### Expected behavior

Every export should always return a fresh, immediately downloadable temporary file. Reusing a previous storage object for the same exported bytes should not happen — temporary exports need a fresh object each time to guarantee availability.

### Affected versions

2.17.0

yong2bba:
**Note:** This PR was created with AI assistance as part of the Penpot MCP self-improvement initiative.

## What changed

This PR prevents temporary export uploads from reusing an existing deduplicated `tempfile` storage object.

Temporary export files are short-lived resources addressed through `/assets/by-id/<storage-object-id>`. Reusing a previous storage object for the same exported bytes can return an old temporary asset URL instead of a fresh one, which is risky when the object is eligible for temporary-file garbage collection or has stale backing storage.

The patch makes this safer in two places:

- `upload-tempfile` now explicitly stores export results with `::sto/deduplicate? false`.
- the storage layer refuses hash-based deduplication for the `tempfile` bucket even if a caller accidentally requests it.

## Why

Exported images are disposable download artifacts, not durable media assets. For `tempfile`, freshness and availability are more important than content-addressed reuse.

## Testing

Run in Docker with PostgreSQL and Redis test services:

- `clojure -M:dev:test --focus backend-tests.storage-test/tempfile-objects-are-not-deduplicated`
- `clojure -M:dev:test --focus backend-tests.rpc-management-test/upload-tempfile-returns-fresh-object-for-same-content`
- `clojure -M:dev -m cljfmt.main check src/app/storage.clj src/app/rpc/management/exporter.clj test/backend_tests/storage_test.clj test/backend_tests/rpc_management_test.clj`
- `clojure -M:dev -e "(require 'app.storage 'app.rpc.management.exporter 'backend-tests.storage-test 'backend-tests.rpc-management-test) (println :loaded)"`

## Residual risk

Full backend suite was not run locally; the focused tests cover the changed storage and management-RPC behavior.

Closes #9970


madalenapmelo-kp:

niwinz:
hey @yong2bba thanks for the contribution.
It looks nice. I have removed the changelog entry because it will generate automatically, and also categorize this as enhancement

PR #9963, opened 2026-06-02
:bug: Preload fill images for bitmap exports

### LINKED ISSUE #10154: Bitmap exports show blank or missing image fill regions
### Description

Bitmap exports (PNG/JPEG/WebP) of shapes that use image fills stored inside the `:fills` vector show blank, dark, or missing image-filled regions. SVG exports for the same shapes still contain the expected image references, but the bitmap exporter takes the screenshot before the embedded image data is fully loaded.

### Steps to reproduce

1. Create a shape with an image fill stored in the `:fills` vector.
2. Export the shape as a bitmap (PNG, JPEG, or WebP).
3. Observe that the exported image is missing or shows blank/dark regions where the image fill should appear.

### Expected behavior

The bitmap export should correctly render the image fill, matching the visual appearance of the shape in the workspace and the SVG export.

### Affected versions

Self-hosted and cloud versions prior to 2.17.0.

yong2bba:
## Summary

Fix bitmap exports for shapes that use image fills stored inside the `:fills` vector.

This patch:

- includes nested `:fill-image` entries when populating the frontend render image cache
- allows the `render.html?route=objects` path used by the exporter to request embedded image data
- passes `render-embed=true` from the bitmap exporter, matching the SVG exporter path
- waits for image resources in the exporter before taking Playwright screenshots
- avoids emitting an extra legacy padding `<image>` without a concrete `href` for nested image fills
- adds a frontend regression test for collecting image fills from `:fills`

## Why

Image fills can be present in the fills vector rather than only as top-level `:fill-image`/`:metadata` shape data. In that case static object rendering may produce SVG markup that references images but the bitmap exporter can screenshot before those images are loaded/embedded. This results in blank, dark, or missing image-filled regions in PNG/JPEG/WebP exports while SVG output still contains the expected image references.

## Verification

- `git diff --check`
- static secret-literal scan on the diff
- independent code review of the patch
- frontend regression test added and wired into `frontend-tests.runner`
- reproduced and smoke-tested on a self-hosted Penpot 2.15.3 instance: individual image-fill shape PNG export renders the expected image after the patch

Note: local `cljfmt`/full frontend test execution was not available in my environment (`cljfmt` command not installed), so I added the focused regression test and verified the patch statically plus with runtime smoke on the self-hosted instance.

Closes #10154

### DIFF
--- a/exporter/src/app/browser.cljs
+++ b/exporter/src/app/browser.cljs
@@ -60,6 +60,35 @@
                           :cause (ex-message cause))
                   (p/resolved nil))))))
 
+(defn wait-for-images
+  "Wait until HTML and SVG images currently present in the page are loadable."
+  ([page] (wait-for-images page nil))
+  ([page {:keys [timeout] :or {timeout 15000}}]
+   (-> (.evaluate ^js page
+                  (js* "() => {
+  const hrefOf = (node) => {
+    if (node.currentSrc) return node.currentSrc;
+    if (node.src) return node.src;
+    const href = node.getAttribute && (node.getAttribute('href') || node.getAttribute('xlink:href'));
+    return href || '';
+  };
+  const urls = Array.from(document.querySelectorAll('img, image'))
+    .map(hrefOf)
+    .filter((href) => href && !href.startsWith('data:'));
+  return Promise.all(urls.map((url) => new Promise((resolve) => {
+    const img = new Image();
+    img.crossOrigin = 'anonymous';
+    img.onload = () => resolve(true);
+    img.onerror = () => resolve(false);
+    img.src = url;
+  }))).then(() => new Promise((resolve) => requestAnimationFrame(() => requestAnimationFrame(resolve))));
+}"))
+       (p/timeout timeout)
+       (p/catch (fn [cause]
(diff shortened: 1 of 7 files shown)


madalenapmelo-kp:

niwinz:
Just a question: why the `passes render-embed=true from the bitmap exporter, matching the SVG exporter path` ? the non-svg render does not requiere this overhead

niwinz:
We are closing this in favor of #11396, which targets the same fix with your commits preserved but applies some adjustments while reviewing:

- It reuses the already supported `embed` parameter of the render entrypoint instead of introducing the new `render-embed` one (both were semantically equivalent knobs, and `render-embed` had been passed inertly by the SVG exporter since the exporter rewrite).
- The images-cache pre-population now only runs when embedding is active, so non-embedding consumers of the render route don't pay upfront image downloads.
- As a consequence of activating embedding for these routes, SVG exports of shapes with image fills now contain embedded data URIs instead of URL references; that output change is now explicit and documented.

Thanks @yong2bba for the diagnosis and the initial patch — the fix direction, the legacy href-less `<image>` removal and the `wait-for-images` backstop all carry over.

niwinz:
Superseded by #11396

PR #9964, opened 2026-06-02
:bug: Improve export shape diagnostics

yong2bba:
## Summary

Improve the MCP `export_shape` tool when Penpot returns an error page instead of image bytes.

This patch:

- adds an optional `scale` argument to `export_shape`
- forwards the scale option through the Penpot plugin `exportImage` helper for shape exports
- detects HTML payloads before passing bytes to `sharp`
- returns a targeted diagnostic message when the export result is an HTML error page instead of PNG/JPEG/WebP/SVG data
- escapes generated plugin code arguments with `JSON.stringify`

## Why

When a self-hosted Penpot instance has a broken `/assets/by-id/...` or `/assets/by-file-media-id/...` route, `shape.export()` can return an HTML error page. Previously the MCP server passed those bytes to `sharp`, which surfaced a generic error such as `Input buffer contains unsupported image format`. That hides the actionable cause.

With this patch, users see a direct message that Penpot returned HTML instead of image bytes, including a short body prefix for diagnosis.

## Verification

- `pnpm run types:check` in `mcp/packages/server`
- `git diff --check`
- static secret-literal scan on the diff
- independent code review of the patch

Note: I also drafted a local diagnostic script, but did not include it because this local environment cannot load the `sharp` optional native dependency for `darwin-arm64`, so the script is not a reliable regression test here.

### DIFF
--- a/mcp/packages/plugin/src/PenpotUtils.ts
+++ b/mcp/packages/plugin/src/PenpotUtils.ts
@@ -419,7 +419,12 @@ export class PenpotUtils {
      *   - For mode="shape", it will be PNG or SVG data depending on the value of `asSVG`.
      *   - For mode="fill", it will be whatever format the fill image is stored in.
      */
-    public static async exportImage(shape: Shape, mode: "shape" | "fill", asSVG: boolean): Promise<Uint8Array> {
+    public static async exportImage(
+        shape: Shape,
+        mode: "shape" | "fill",
+        asSVG: boolean,
+        options: { scale?: number } = {}
+    ): Promise<Uint8Array> {
         // Updates are asynchronous in Penpot, so wait a tick to ensure any pending updates are applied before export.
         // The constant wait time is a temporary workardound until a better solution for penpot/penpot-mcp#27
         // is implemented.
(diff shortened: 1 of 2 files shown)


madalenapmelo-kp:

opcode81 (changes requested):
* This PR conflates unrelated changes: The addition of the `scale` parameter is not related to the diagnostics issue.
* The diagnostics issue is being solved at the wrong level.  
  Note that we observed export failures on several occasions, and HTTP 5XX errors were always reliably surfaced, triggering a corresponding error response in the MCP tool call immediately.
    * If other HTTP errors (e.g. 4XX) are not being surfaced correctly by the underlying implementation of `ShapeBase.export`, then this is where the error handling should live.
    * If, in your deployment, the endpoint did not return an error but a success response (200) which, however, did not contain a valid image, then this is purely a problem with the respective deployment.
    
For the above reasons, I propose to reject this PR.

niwinz:
> * This PR conflates unrelated changes: The addition of the `scale` parameter is not related to the diagnostics issue.
> * The diagnostics issue is being solved at the wrong level.
>   Note that we observed export failures on several occasions, and HTTP 5XX errors were always reliably surfaced, triggering a corresponding error response in the MCP tool call immediately.
>   
>   * If other HTTP errors (e.g. 4XX) are not being surfaced correctly by the underlying implementation of `ShapeBase.export`, then this is where the error handling should live.
>   * If, in your deployment, the endpoint did not return an error but a success response (200) which, however, did not contain a valid image, then this is purely a problem with the respective deployment.
> 
> For the above reasons, I propose to reject this PR.

Thanks for the detaled review @opcode81

At is suggested, closing this PR. In any case, thanks for your time.

PR #10013, opened 2026-06-03
:tada: Enable multi-instance horizontal scaling for MCP server

### LINKED ISSUE #10000: Enable multi-instance horizontal scaling for MCP server
## Problem

The MCP server (`mcp/packages/server`) cannot run with multiple instances. All state is kept in-memory on a single process heap, making it a single point of failure and a scalability bottleneck.

The MCP server acts as a gateway between an LLM client (Cursor, Claude, opencode-cli) and the Penpot frontend via an integrated plugin. It maintains two persistent connections per user:

1. **Plugin WebSocket** (port 4402) — the Penpot plugin connects and stays connected
2. **MCP Streamable HTTP** (port 4401) — the LLM client sends tool calls

Both connections for a given user **must** reach the same server instance today. If the plugin connects to Instance A and the LLM client connects to Instance B, tool calls fail because Instance B cannot see the plugin's WebSocket connection (it lives in Instance A's in-memory `PluginBridge` maps).

### Root cause

Two categories of in-memory state prevent horizontal scaling:

| Category | Location | Issue |
|---|---|---|
| Plugin connection & task state | `PluginBridge.clientsByToken`, `pendingTasks` (in-memory `Map`s) | Tool dispatch looks up WebSocket by `userToken` in local map; task responses are correlated via in-memory Promises |
| MCP session identity | `PenpotMcpServer.streamableTransports`, SDK transport's `_initialized`/`sessionId` | A POST with an unknown `mcp-session-id` is rejected as "Session not found"; the SDK transport's session state is bound to a single process |

### Why sticky sessions are not a good solution

The classical approach would be sticky sessions — configure the load balancer to route all connections from the same user to the same instance. This is not a good fit here because:

- **Uneven load distribution.** Sticky sessions bind users to instances regardless of load. A user with a long-running plugin connection occupies capacity on one instance even when idle, while other instances may be underutilized.
- **Operational complexity.** Scaling up or down requires draining sticky sessions from instances before removing them, complicating deployments and auto-scaling.
- **Additional complexity.** Penpot has no sticky sessions for any other service, so this would be a special case. We have different layers of proxies and load balancers — introducing stickiness would make the AWS ELB configuration more complicated. Changes at that layer are outside the product team's control and are always more complex to manage.

## Proposed solution

Route all task requests through Redis pub/sub keyed by user token, and store session identity in Redis. This eliminates the need for both connections to land on the same instance.

### Mechanism 1 — Redis Session Store

Session identity (`sessionId → userToken`) is stored in Redis with TTL. When a POST arrives with an `mcp-session-id` header that doesn't exist locally, the server looks it up in Redis. If found, a fresh SDK transport is created and pre-initialized with the session state (the SDK's `_initialized` and `sessionId` are plain JS properties that can be set directly). Any instance can handle any session.

### Mechanism 2 — Redis Pub/Sub Task Routing

All tool-to-plugin communication goes through Redis pub/sub keyed by user token — always, even when both connections are on the same instance. When a plugin connects, the server subscribes to `penpot.mcp.task.req.<token>`. When a tool executes, the server publishes to that channel and waits for a response on `penpot.mcp.task.res.<taskId>`. The instance with the plugin's WebSocket picks up the request, executes it, and publishes the response back.

This eliminates instance tracking, plugin registries, routing logic, and the fast/slow path distinction entirely.

## Acceptance criteria

- [ ] Two MCP server instances can run simultaneously behind a load balancer with round-robin routing
- [ ] Plugin connects to Instance A, LLM client connects to Instance B — tool calls succeed
- [ ] Session created on Instance A, subsequent tool call on Instance B — session is recovered
- [ ] Plugin disconnects from Instance A and reconnects to Instance B — next tool call routes to B automatically
- [ ] Redis unavailable at startup — server degrades to single-instance mode (existing behavior preserved)
- [ ] Single-user mode without Redis — no changes to existing behavior
- [ ] Legacy SSE endpoints (`/sse`, `/messages`) are out of scope (should be removed)

opcode81:
### Related Ticket

* #10000

### Summary

Allow the MCP server to run as multiple instances behind a plain
round-robin load balancer, removing the previous requirement that a
user's plugin WebSocket and MCP client connection terminate on the same
instance. Behaviour is unchanged when run as a single instance or
without Redis.

Cross-instance MCP sessions: when a request arrives with an
mcp-session-id that was initialised on another instance, the session is
adopted locally instead of rejected. The user token is read from the
query parameter (present on every request, as the configured endpoint
URL is never rewritten), so no shared session store is needed; the
transport is pre-initialised so the SDK's validateSession() accepts it.

Cross-instance task routing: when a Redis URI is configured in
multi-user mode, plugin task requests are routed via Redis pub/sub keyed
by user token. The instance holding a plugin's WebSocket subscribes to
that token's request channel; any instance handling a tool call
publishes the request and awaits the response on a per-request channel.
RedisBridge is a pure transport for the existing serialised
`PluginTaskRequest`/`Response` objects. `PluginTask` is split into an abstract
base plus a local (promise-backed) `PluginTask` and a `RemotePluginTask`
whose resolve/reject publish the outcome back over Redis, so the
existing local dispatch and response-correlation paths are reused
unchanged on the executing instance.

### Design Notes

The design leverages an underlying abstraction, where 

* a directly processed task (WebSocket connection owned locally),
* a task dispatched via Redis to be handled by another instance, and
* a forwarded task from another instance (obtained from Redis)

are all tasks that share the property that they are ultimately resolved (either successfully or with an error) when a response is delivered. The only difference is 

* the channel over which the request is sent/the response is received and 
* what it means to resolve a task.

For forwarded tasks from Redis, the latter is handled through newly introduced class `RemotePluginTask`, which passes results back via Redis.  

Beyond this, all task handling uses the same principles already established in `PluginBridge`, i.e. the map of pending tasks, timeouts, etc.

### Testing

In the agentic devenv, the Redis-based communication can be tested (for a single instance that forwards to itself) by
1. terminating the MCP server (Ctrl+C in tmux console 4)
2. restarting the MCP server with this command:
    ```
    PENPOT_MCP_LOG_LEVEL=debug PENPOT_MCP_SERVER_HOST=0.0.0.0 PENPOT_MCP_REMOTE_MODE=true PENPOT_MCP_REDIS_URI=redis://redis:6379/0 pnpm run bootstrap:multi-user
    ```
 
 I tested it successfully, but a test in an actual multi-instance scenario would be desirable. Does any other instance besides PROD use load balancing as well, allowing a straightforward test?

### Specification Notes

The implementation differs from the original proposal in several ways, but the core idea of Mechanism 2 remains.
Mechanism 1 was deemed unnecessary as stated in the ticket.

> Legacy SSE endpoints (/sse, /messages) are out of scope (should be removed)

While the legacy endpoints are not important to support, Mechanism 1 being dropped means that removal is not necessary.

> Redis unavailable at startup — server degrades to single-instance mode (existing behavior preserved)

The "Redis-backed multi-instance mode" is enabled by setting the environment variable `PENPOT_MCP_REDIS_URI`. If it is set (and we are in multi-user mode), then the Redis-based communication will be used.
If it is unset, existing behaviour is unchanged.

### Potential Follow-Up

Even though the Redis indirection is not a problem, I would still propose to avoid the indirection if the WebSocket connection is owned by the original MCP server instance. It is a very small change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,6 +93,7 @@
 /.pnpm-store
 /.vscode
 /.idea
+*.iml
 /.claude
 /.playwright-mcp
 /tools/__pycache__
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -273,6 +273,7 @@ The Penpot MCP server can be configured using environment variables.
 | `PENPOT_MCP_REMOTE_MODE`                         | Enable remote mode (disables file system access). Set to `true` to enable. | `false`        |
 | `PENPOT_MCP_DEVENV`                              | Enable Penpot development environment tools. Set to `true` to enable.      | `false`        |
 | `PENPOT_MCP_EXPORT_SHAPE_MAX_PARALLEL_REQUESTS`  | Maximum number of parallel export shape requests (multi-user mode only).   | `0` (no limit) |
+| `PENPOT_MCP_REDIS_URI`                           | Redis connection URI (e.g. `redis://host:6379`) enabling multi-instance horizontal scaling via Redis pub/sub task routing (multi-user mode only). When unset, the server runs in single-instance mode, requiring the plugin and MCP client to connect to the same instance. | (unset)        |
 
 ### Logging Configuration
 
--- a/mcp/packages/server/package.json
+++ b/mcp/packages/server/package.json
@@ -5,7 +5,7 @@
   "type": "module",
   "main": "dist/index.js",
   "scripts": {
-    "build:server": "esbuild src/index.ts --bundle --platform=node --target=node18 --format=esm --outfile=dist/index.js --external:@modelcontextprotocol/* --external:ws --external:express --external:class-transformer --external:class-validator --external:reflect-metadata --external:pino --external:pino-pretty --external:pino-loki --external:js-yaml --external:sharp --external:nrepl-client",
+    "build:server": "esbuild src/index.ts --bundle --platform=node --target=node18 --format=esm --outfile=dist/index.js --external:@modelcontextprotocol/* --external:ws --external:express --external:class-transformer --external:class-validator --external:reflect-metadata --external:pino --external:pino-pretty --external:pino-loki --external:js-yaml --external:sharp --external:nrepl-client --external:ioredis",
     "build": "pnpm run build:server && node scripts/copy-resources.js",
     "build:types": "tsc --emitDeclarationOnly --outDir dist",
     "start": "node dist/index.js",
@@ -29,6 +29,7 @@
     "class-transformer": "^0.5.1",
     "class-validator": "^0.14.3",
     "express": "^5.1.0",
+    "ioredis": "^5.6.0",
     "js-yaml": "^4.1.1",
     "nrepl-client": "^0.3.0",
     "penpot-mcp": "file:..",
--- a/mcp/packages/server/src/PenpotMcpServer.ts
+++ b/mcp/packages/server/src/PenpotMcpServer.ts
@@ -4,6 +4,7 @@ import { SSEServerTransport } from "@modelcontextprotocol/sdk/server/sse.js";
 import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
 import { ExecuteCodeTool } from "./tools/ExecuteCodeTool";
 import { PluginBridge } from "./PluginBridge";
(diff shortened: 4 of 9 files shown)


niwinz:
It looks good at first glance. will look on detail tomorrow. Thanks

Issue #9079, opened 2026-04-21
:books: Improve MCP usage documentation based on beta feedback


opcode81:
The MCP beta-test showed that the documentation could be improved in several ways:

* ~~Add more extensive documentation on how to connect the MCP server to all common clients (as in the [documentation of other MCP servers](file:///C:/Users/domin/Dev/serena/docs/_build/02-usage/030_clients.html)).  
  While this is mostly repeating documentation of other applications (and therefore isn't particularly maintainable), it is the most user-friendly thing to do.~~ (This is now handled via #10355)
* Make the documentation concise and modular in order for it to be agent-friendly.  
   Nowadays, documentation is no longer primarily read by humans. Therefore, it is best to strongly structure documentation into several files that an agent can retrieve separately, avoiding context bloat.
* Make clear that the use of a VLM is an absolute necessity.  
  A dissatisfied user was working with vision capabilities disabled, but poor performance is ultimately to be expected in this case. Designing while being blind to how things actually look just does not work.

PR #10117, opened 2026-06-10
:sparkles: MCP installer subcommands with multi-client auto-config and doctor

### LINKED ISSUE #10179: Add MCP installer subcommands to auto-register Penpot MCP with supported clients
### Description

Add MCP installer subcommands — `install`, `uninstall`, `doctor`, and `help` — to the Penpot MCP package. The installer auto-registers Penpot MCP with supported clients so users no longer have to hand-edit each client's config file. An interactive arrow-key picker is shown when `--client` is omitted, and a non-interactive flag is available for scripting.

### Supported clients

`claude-code`, `claude-desktop`, `cursor`, `windsurf`, `cline`, `opencode`, `gemini`, `codex`, `antigravity`, `antigravity-cli`, `generic-json`

### Use case

Setting up Penpot MCP currently requires editing each client's config file by hand. The new `install` subcommand removes that friction and streamlines the setup process, bringing it closer to the pattern used by other MCP servers in the ecosystem.

### Affected versions

2.17.0

mulhamna:
## Summary

Adds MCP installer subcommands — `install` / `uninstall` / `doctor` / `help` (plus `--help` / `-h`) — to the Penpot MCP package. The installer auto-registers Penpot MCP with the supported clients so users no longer have to hand-edit each client's config file. An interactive arrow-key picker is shown when `--client` is omitted, and a non-interactive flag is available for scripting.

Smart-dispatch in `bin/mcp-local.js` preserves the existing `npx -y @penpot/mcp` bootstrap flow — running with no arguments still boots the local server, only the new subcommands are forwarded to the server dist.

### Supported clients

`claude-code`, `claude-desktop`, `cursor`, `windsurf`, `cline`, `opencode`, `gemini`, `codex`, `antigravity`, `antigravity-cli`, `generic-json` (snippet only — for clients not on this list).

### Tested coverage

End-to-end tested on macOS for `claude-code`, `claude-desktop`, and `cursor`. The remaining adapters are wired according to each client's documented configuration shape but have not yet been individually validated on this codebase — issues from users on other clients / platforms are welcome.

## Why

Setting up Penpot MCP currently requires editing each client's config file by hand. The new `install` subcommand removes that friction and brings the setup story closer to the pattern used by other MCP servers in the ecosystem.

## What changed since the initial PR

Following @opcode81's review, the original PR included two additional changes — a PAT-based REST client with a `list_penpot_files` tool, and a `config` subcommand to persist the PAT and base URL — both of which have been reverted:

- The PAT/REST path conflicts with the Penpot MCP server's intentional Plugin-API-only design. The file enumeration gap that motivated the tool should be closed by extending the Plugin API instead; a follow-up ticket will be opened for that.
- With the PAT path removed, the `config` subcommand no longer has a consumer and is removed too.

The branch now contains only the installer plus a docs commit covering tested-client coverage and how to use the installer from a local checkout while the npm artifact has not yet been republished.

## Refs

- Refs #9079 — improvements to MCP usage docs (covered by the installer + docs commits)

## Commit layout

| Commit | Scope |
|---|---|
| `:sparkles: Add MCP installer with multi-client auto-config and doctor` | Installer module, 11 client adapters, interactive picker, doctor, bin smart-dispatch, initial README section |
| `:sparkles: Add PAT-based Penpot REST client and list_penpot_files tool` | (reverted, see below) |
| `:sparkles: Add config subcommand to persist Penpot PAT and base URL` | (reverted, see below) |
| `:rewind: Drop config subcommand after PAT mode removal` | Revert of the config subcommand |
| `:rewind: Drop PAT-based Penpot REST client and list_penpot_files tool` | Revert of the REST/PAT tool per review |
| `:books: Document tested client coverage and npm artifact usage` | README notes on tested clients and local-checkout usage |

Each commit independently passes `pnpm run fmt:check`, `pnpm -r run types:check` and `pnpm -r run build`, so any commit can be reverted on its own without breaking the rest.

## Test plan

- [x] `penpot-mcp --help` shows all subcommands.
- [x] `penpot-mcp install` opens the arrow-key picker with the supported entries.
- [x] `penpot-mcp install --client cursor --dry-run` prints the snippet without modifying `~/.cursor/mcp.json`.
- [x] `penpot-mcp install --client cursor` writes `mcpServers.penpot` to `~/.cursor/mcp.json` and creates a `.bak-<ts>` backup if a file existed.
- [x] `penpot-mcp uninstall --client cursor` removes the entry; running it again reports "entry not present".
- [x] `penpot-mcp doctor` reports the state of ports 4400 / 4401 / 4402 and the plugin manifest.
- [x] `penpot-mcp install --client claude-code` end-to-end (registers via `claude mcp add` when the CLI is on PATH, falls back to `~/.claude.json` otherwise).
- [x] `penpot-mcp install --client claude-desktop` writes the stdio + `mcp-remote` snippet to the macOS config path.

Closes #10179


madalenapmelo-kp:

opcode81 (changes requested):
This PR conflates two unrelated changes:

   1. It adds a tool for interacting with the Penpot backend API, specifically for retrieving the list of available Penpot files.

      This change is not in line with the spirit of the Penpot MCP server, which uses the plugin API as the single interaction mechanism, excluding the backend API intentionally. The plugin API currently does not provide a way to enumerate available files (or switch to another file), so there is currently a functionality gap, but if this gap is deemed relevant, then it should be closed by extending the plugin API rather than by establishing a second interaction mechanism (which has its own set of preconditions, etc.). I would recommend that you add a ticket requesting that the plugin API be extended instead.

      Furthermore, note that other MCP servers exposing the Penpot backend API already exist, so you also have the option of adding such an MCP server to your toolset in order to close the gap in the meantime.

   2. It adds CLI commands for configuring MCP clients.

      While the Penpot MCP server is fairly easy to configure (single URL only), some beta users did remark that the client configuration was a bit cumbersome for them. However, 
      * the new CLI will only be helpful to users who use the MCP server locally or who use it directly through the artefact released on npm. If we include this, we should update the documentation and instruct remote MCP users to use it through npx (once it is released).
      * it appears that the CLI has only been tested for Cursor - presumably on a single platform. I have had a quick look exemplarily, and the implementation does appear to consider the different platforms and client idiosynchrasies correctly, but if it hasn't been tested, we cannot be certain that all the implemented cases actually work.

**Conclusion**

@niwinz I think we should reject (1), noting that the alternative is the extend the plugin API, but (2) is something we could consider.

mulhamna:
Thanks for taking the time on this @opcode81 , the philosophy context on (1) is really helpful — that part wasn't obvious from the code.

Totally get the call on the PAT/REST tool. Happy to drop that commit and open a separate ticket asking for file enumeration in the Plugin API instead. If @niwinz lands on the same conclusion just say the word and I'll force-push without it. If a Plugin API proposal would help move that forward, I'm glad to take a stab at drafting one too.

On the CLI side — you're right, I mostly tested on macOS with Cursor and Claude Code. The other client adapters were modeled after the [jira-commands](https://github.com/mulhamna/jira-commands) installer, which has good coverage across those targets. If you've got a sense of which clients Penpot users actually reach for, I can narrow the initial set to the ones I've confirmed and add the rest as people ask. Or I can do another testing pass — whichever you prefer.

On the docs front, fair, the README assumes local right now. Easy fix to add a section for the npx artifact path once it's released. Want that folded into this PR or kept for a follow-up?

No strong preference on the final shape from my side — just let me know what works best for you all.

mulhamna:
Thanks for taking the time on this @opcode81, the philosophy context on (1) is really helpful — that part wasn't obvious from the code.

Totally get the call on the PAT/REST tool. I've reverted that commit along with the `config` subcommand that depended on it, so the branch is back to just the installer pieces. Happy to open a separate ticket asking for file enumeration in the Plugin API instead — if a Plugin API proposal would help move that forward, I'm glad to take a stab at drafting one too.

On the CLI side — fair callouts. I went back and tested `claude-code`, `claude-desktop`, and `cursor` end-to-end on macOS, and added a short note in the README listing those as the verified set. The other adapters are still wired up but flagged as not individually validated yet, with a request for users to open issues if something doesn't work for their client / platform. If you'd prefer to narrow the initial set to just the three I've verified and add the rest as follow-ups based on user requests, that's also fine — let me know which way you'd rather go.

For docs, I added a section covering how to use the installer from a local checkout while the npm artifact has not yet been republished, and a note for users targeting a remote MCP server.

@niwinz no strong preference on the final shape from my side — happy to follow whichever direction works best for the project. Let me know what'd be most helpful.

mcp/packages/server/src/installer/clients/opencode.ts
```diff
@@ -0,0 +1,59 @@
+import { ClientInstaller, InstallOptions, InstallResult } from "../types";
+import { expandHome, readJsonFile, writeJsonFile, setNested, deleteNested, backupIfExists } from "../configWriter";
+
+const CONFIG_PATH = "~/.config/opencode/opencode.jsonc";
```
niwinz: This checks always for opencode.jsonc but users can have the `opencode.json` (without `c`)
My suggestion is properly handle the both.

niwinz:
@opcode81 Thanks

I'm OK with @opcode81 suggestions, and the current changes looks ok for me (you have removed the (1) :D)
I will put this on my todo list for test it.

mcp/packages/server/src/installer/clients/opencode.ts
mulhamna: Yeah you're right, missed that. Just pushed 9f8b84485 — checks for both opencode.jsonc and opencode.json, sticks with whichever the user already has, and only falls back to creating opencode.jsonc when neither one exists.

opcode81:
It is worth noting that the only thing that is really Penpot-specific about this installation tool is that it will use the URL `http://localhost:4401/mcp` by default. Remote MCP server users would always have to specify the URL, making the tool fully unspecific.
  
Since installing an HTTP-based MCP server is a global concern, I would propose to turn this into an independent tool (and, perhaps, a library) published to npm. We - and others - could then instruct our users to use the tool in order to simplify installation (or integrate the library to make it specific to our server).
I believe there could be some broader community interest, and it just makes more sense from a reusability perspective.

opcode81:
Having searched the web, I found that the tool actually already exists:
https://github.com/neon-solutions/add-mcp

So I think this should probably be closed. The existing tool is, most likely, all that is needed - and they also provide instructions on how to provide one's own adapted installation command.

PS: There is also Smithery, which also supports installation (but add-mcp is probably the way to go):
https://smithery.ai/docs/concepts/cli

mulhamna:
Thanks for digging those up — both add-mcp and Smithery are good references, hadn't come across either before.

A couple of thoughts before deciding to close, just so we're weighing it together:

- Extra hop for users. With a bundled installer, the setup is one tool — npx -y @penpot/mcp install. With an external one, users have to learn that two separate npm packages cooperate (npx -y add-mcp ... plus @penpot/mcp for the server), and the Penpot docs need to keep both versions in sync. For people unfamiliar with the MCP ecosystem, that's a bit more cognitive overhead.
- Trust surface. Recommending a third-party tool means Penpot's setup story now depends on someone else's release cadence, security posture and continued maintenance. add-mcp is currently a single-contributor project (last commit ~weeks ago when I checked), and Smithery requires registry registration, which adds its own constraints.
- Coverage uncertainty. I'd want to actually verify add-mcp covers the clients we wire up here (Cursor, Claude Code, Claude Desktop, Windsurf, Cline, OpenCode, Gemini, Codex, Antigravity, Antigravity-cli, plus the generic-json fallback). If gaps exist, users still hit the manual-edit problem this PR is trying to remove.
- Maintenance footprint. The installer in this PR is ~600 lines, single dependency-free module. It's not a big surface to keep working as part of the Penpot MCP package.

That said, your point on reusability is valid, and I'm not attached to keeping this inside Penpot if there's a better home. A couple of middle-ground options if you'd prefer not to merge as-is:
1. Keep a thin Penpot-specific wrapper that delegates to add-mcp when the external tool covers our clients — fall back to bundled adapters otherwise.
2. Extract the installer module into its own npm package (e.g. @mulhamna/mcp-install or similar), publish it, then have @penpot/mcp depend on it. PR gets slimmed down, the generic logic lives independently, and Penpot keeps a one-command UX.
3. Close this PR and contribute the missing client adapters back to add-mcp upstream, then update Penpot docs to point at it. Higher effort but cleanest split of concerns.

@niwinz @opcode81 — happy to follow whichever direction makes most sense. If you'd like me to actually verify add-mcp's client coverage first before deciding, I can do that and report back here.

niwinz:
Hey @mulhamna 

I will ask it internally because i'm personally agree with you. I'm not very convinced on depending on external tool for something that is directly related to setup penpot. But i don't have strong opinion on it, so i prefer ask for feedback internally.

opcode81:
@mulhamna having discussed this internally, we decided to simplify installation by drawing upon add-mcp (or similar), as described in https://github.com/penpot/penpot/issues/10355.

PR #10141, opened 2026-06-11
:sparkles: Add nitrate sso wards to organization navigation

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot-nitrate/us/13048

### Summary

This MR adds organization-level SSO authorization guards before entering dashboard/workspace routes and wires the OIDC flow for Nitrate-managed org SSO.

Main changes:

* Frontend route guard now checks team SSO status before navigation.
* New backend auth endpoint for SSO pre-check and auth URL generation.
* OIDC callback now supports org SSO flow and stores short-lived SSO authorization in HTTP session props.
* Nitrate client now fetches org SSO config by team and validates session authorization with positive-result caching.
* Organization type now includes sso-active.
* Database migration adds props JSONB column to http_session_v2 for SSO state storage.

### Checklist

### DIFF
--- a/backend/src/app/auth/oidc.clj
+++ b/backend/src/app/auth/oidc.clj
@@ -24,6 +24,7 @@
    [app.http.errors :as errors]
    [app.http.session :as session]
    [app.loggers.audit :as audit]
+   [app.nitrate :as nitrate]
    [app.rpc.commands.profile :as profile]
    [app.setup :as-alias setup]
    [app.tokens :as tokens]
@@ -42,9 +43,9 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (defn- discover-oidc-config
-  [cfg {:keys [base-uri] :as provider}]
+  [cfg {:keys [base-uri skip-ssrf-check?] :as provider}]
   (let [uri (u/join base-uri ".well-known/openid-configuration")
-        rsp (http/req cfg {:method :get :uri (dm/str uri)})]
+        rsp (http/req cfg {:method :get :uri (dm/str uri)} {:skip-ssrf-check? skip-ssrf-check?})]
 
     (if (= 200 (:status rsp))
       (let [data       (-> rsp :body json/decode)
@@ -105,8 +106,8 @@
           keys))
 
 (defn- fetch-oidc-jwks
-  [cfg jwks-uri]
-  (let [{:keys [status body]} (http/req cfg {:method :get :uri jwks-uri})]
+  [cfg jwks-uri {:keys [skip-ssrf-check?]}]
+  (let [{:keys [status body]} (http/req cfg {:method :get :uri jwks-uri} {:skip-ssrf-check? skip-ssrf-check?})]
     (if (= 200 status)
       (-> body json/decode :keys process-oidc-jwks)
       (ex/raise :type ::internal
@@ -118,7 +119,8 @@
   "Fetch and Add (if possible) JWK's to the OIDC provider"
   [cfg provider]
   (try
-    (if-let [jwks (some->> (:jwks-uri provider) (fetch-oidc-jwks cfg))]
+    (if-let [jwks (when-let [jwks-uri (:jwks-uri provider)]
+                    (fetch-oidc-jwks cfg jwks-uri {:skip-ssrf-check? (:skip-ssrf-check? provider)}))]
       (assoc provider :jwks jwks)
       provider)
     (catch Throwable cause
@@ -409,9 +411,9 @@
 (defn- build-redirect-uri
   []
   (let [public (u/uri (cf/get :public-uri))]
-    (str (assoc public :path (str "/api/auth/oidc/callback")))))
+    (str (assoc public :path "/api/auth/oidc/callback"))))
 
-(defn- build-auth-redirect-uri
+(defn build-auth-redirect-uri
   [provider token]
   (let [params {:client_id (:client-id provider)
                 :redirect_uri (build-redirect-uri)
@@ -454,7 +456,7 @@
            :grant-type (:grant_type params)
            :redirect-uri (:redirect_uri params))
 
-    (let [{:keys [status body]} (http/req cfg req)]
+    (let [{:keys [status body]} (http/req cfg req {:skip-ssrf-check? (:skip-ssrf-check? provider)})]
       (if (= status 200)
         (let [data (json/decode body)
               data {:token/access (get data :access_token)
@@ -509,7 +511,7 @@
                   :headers {"Authorization" (str (:token/type tdata) " " (:token/access tdata))}
                   :timeout 6000
                   :method :get}
-        response (http/req cfg params)]
+        response (http/req cfg params {:skip-ssrf-check? (:skip-ssrf-check? provider)})]
 
     (l/trc :hint "user info response"
            :status (:status response)
@@ -755,6 +757,25 @@
         (assoc profile :props props'))
       profile)))
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; ORG SSO HELPERS
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn prepare-org-sso-provider
+  "Build an OIDC provider map dynamically from the Nitrate org SSO config.
+  Uses OIDC discovery via :base-url (or :issuer as fallback) when
+  token/auth/user URIs are absent."
+  [cfg {:keys [client-id client-secret base-url issuer scopes]}]
+  (prepare-oidc-provider cfg
+                         {:type             "oidc"
+                          :client-id        client-id
+                          :client-secret    client-secret
+                          :base-uri         (some-> (or base-url issuer)
+                                                    (str/rtrim "/")
+                                                    (str "/"))
+                          :scopes           (into default-oidc-scopes (or scopes #{}))
+                          :skip-ssrf-check? true}))
+
 (defn- auth-handler
   [cfg {:keys [params] :as request}]
   (let [provider (resolve-provider cfg params)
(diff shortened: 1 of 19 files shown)


backend/src/app/auth/oidc.clj
```diff
@@ -42,9 +43,9 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (defn- discover-oidc-config
-  [cfg {:keys [base-uri] :as provider}]
+  [cfg {:keys [base-uri skip-ssrf-check?] :as provider}]
   (let [uri (u/join base-uri ".well-known/openid-configuration")
-        rsp (http/req cfg {:method :get :uri (dm/str uri)})]
+        rsp (http/req cfg {:method :get :uri (dm/str uri)} {:skip-ssrf-check? (boolean skip-ssrf-check?)})]
```
niwinz: Minor: the (boolean ...) wrapper is not necessary here

backend/src/app/auth/oidc.clj
```diff
@@ -105,8 +106,8 @@
           keys))
 
 (defn- fetch-oidc-jwks
-  [cfg jwks-uri]
-  (let [{:keys [status body]} (http/req cfg {:method :get :uri jwks-uri})]
+  [cfg jwks-uri {:keys [skip-ssrf-check?]}]
+  (let [{:keys [status body]} (http/req cfg {:method :get :uri jwks-uri} {:skip-ssrf-check? (boolean skip-ssrf-check?)})]
```
niwinz: The same, boolean 🔥

backend/src/app/auth/oidc.clj
```diff
@@ -755,6 +757,26 @@
...
+  [cfg {:keys [client-id client-secret base-url issuer scopes]}]
+  (prepare-oidc-provider cfg
+                         {:type             "oidc"
+                          :client-id        client-id
+                          :client-secret    client-secret
+                          :base-uri         (some-> (or base-url issuer)
+                                                    (str/rtrim "/")
+                                                    (str "/"))
+                          :scopes           (into default-oidc-scopes
+                                                  (when scopes (str/split scopes " ")))
```
niwinz: I suggest using something like:

```clojure
(let [v (str/split v #"[\s,]+")]
  (into empty-set app.common.schema/xf:filter-word-strings v))
```

Maybe we can extract this as a helper under app.common.schema or under app.common.data

backend/src/app/auth/oidc.clj
```diff
@@ -779,64 +801,85 @@
...
+          (let [team-id         (:team-id state)
+                organization-id (:organization-id state)
+                sso             (nitrate/call cfg :get-org-sso-by-team {:team-id team-id})
+                provider        (prepare-org-sso-provider cfg sso)
+                ;; verify token or throw error
+                _info           (get-info cfg provider state code)
+                session         (session/get-session request)
+                exp             (ct/in-future {:hours 48})]
+            (when (and session organization-id)
+              (let [props   (or (some-> (:props session) db/decode-transit-pgobject) {})
```
niwinz: The decode operation should be already handled on session, decoding internal storage format of the session props is not responsibility of the oidc module.

backend/src/app/auth/oidc.clj
```diff
@@ -779,64 +801,85 @@
...
+                    sso-map (get props :sso {})
+                    props'  (assoc props :sso (assoc sso-map organization-id exp))]
```
niwinz: This can be done in a single line:

```
props (update props :sso assoc organization-id exp)
```
And since the original `props` is no longer needed here, just use `props` again instead defining new binding `props'`

backend/src/app/auth/oidc.clj
```diff
@@ -779,64 +801,85 @@
...
+                (db/update! cfg :http-session-v2
+                            {:props (db/tjson props')}
+                            {:id (:id session)}
+                            {::db/return-keys false})))
```
niwinz: I think we should use session API for modify the session instead of direct db/update

backend/src/app/rpc/commands/nitrate.clj
```diff
@@ -617,3 +620,36 @@
...
+              organization-id (:organization-id sso)
+              state-token     (tokens/generate cfg {:iss             "oidc"
+                                                    :dest-url        url
+                                                    :team-id         team-id
+                                                    :organization-id organization-id
+                                                    :issuer          issuer
+                                                    :exp             (ct/in-future "4h")})
+              auth-url        (oidc/build-auth-redirect-uri oidc-provider state-token)]
+          {:authorized false
+           :auth-url   auth-url})
```
niwinz: Internally we use `:redirect-uri` for cases we want to handle the redirect, i suggest to use the same pattern.

backend/src/app/migrations.clj
```diff
@@ -484,7 +484,10 @@
     :fn (mg/resource "app/migrations/sql/0148-add-variant-name-team-font-variant.sql")}
 
    {:name "0149-mod-file-library-rel-synced-at"
-    :fn (mg/resource "app/migrations/sql/0149-mod-file-library-rel-synced-at.sql")}])
+    :fn (mg/resource "app/migrations/sql/0149-mod-file-library-rel-synced-at.sql")}
+
+   {:name "0150-mod-http_session_v2"
+    :fn (mg/resource "app/migrations/sql/0150-mod-http_session_v2.sql")}])
```
niwinz: The migration name should not follow the same naming so no `_` should be used.

backend/src/app/nitrate.clj
```diff
@@ -435,6 +438,28 @@
...
+(def ^:private schema:nitrate-sso
+  [:map
+   [:organization-id ::sm/uuid]
+   [:active [:maybe :boolean]]
+   [:provider [:maybe :string]]
+   [:client-id [:maybe :string]]
+   [:base-url [:maybe :string]]
+   [:client-secret [:maybe :string]]
+   [:issuer [:maybe :string]]
+   [:scopes [:maybe :string]]])
```
niwinz: The comma/space separated items can use the `[::sm/set ::sm/text]` and if you use the schema decoding abilties this will parse and convert the scopes directly into a set

backend/src/app/nitrate.clj
```diff
@@ -485,6 +511,41 @@
...
+(defonce ^:private sso-auth-cache
+  (cache/create :expire "1h" :max-size 1024))
+
+(defn- compute-sso-authorization
+  [cfg team-id session]
+  (let [sso (call cfg :get-org-sso-by-team {:team-id team-id})]
+    (if-not (:active sso)
+      {:authorized true :sso sso}
+      (if (or (:issuer sso) (:base-url sso))
+        (let [props           (some-> (:props session) db/decode-transit-pgobject)
```
niwinz: The same, session should not expose not decoded fields

backend/src/app/nitrate.clj
```diff
@@ -485,6 +511,41 @@
...
+              sso-map         (get props :sso {})
+              organization-id (:organization-id sso)
+              exp             (get sso-map organization-id)
+              now             (ct/now)
+              authorized      (boolean (and (ct/inst? exp)
```
niwinz: Why the boolean wrapper?

backend/src/app/auth/oidc.clj
```diff
@@ -887,6 +887,7 @@
    ::http/client
    ::setup/props
    ::db/pool
+   [:app.nitrate/client {:optional true} [:maybe :map]]
```
niwinz: I think this is not optional, but can be nil.

backend/src/app/util/cache.clj
```diff
@@ -24,7 +24,8 @@
 
 (defprotocol ICache
   (get [_ k] [_ k load-fn] "get cache entry")
-  (invalidate! [_] [_ k] "invalidate cache"))
+  (invalidate! [_] [_ k] "invalidate cache")
+  (invalidate-if! [_ pred] "invalidate all entries whose value satisfies pred"))
```
niwinz: Don't put '!` on the protocol methods

backend/src/app/rpc.clj
```diff
@@ -208,6 +211,78 @@
...
+
+(defonce ^:private org-sso-auth-cache
+  (cache/create :expire "15m" :max-size 1024))
+
+(defn invalidate-org-sso-cache-by-org!
+  "Invalidates all org-SSO authorization cache entries for the given organization-id."
+  [organization-id]
+  (cache/invalidate-if! org-sso-auth-cache #(= (:organization-id %) organization-id)))
+
+(defn- coerce-uuid [v]
```
niwinz: This can be put under app.common.uuid as `coerce` helper

backend/src/app/rpc.clj
```diff
@@ -208,6 +211,78 @@
...
+   Only activates when:
+   - Nitrate flag is enabled
+   - Endpoint requires authentication (::auth true by default)
+   - Endpoint is not marked with ::nitrate/org-sso false
+
+   Raises :org-sso-required error if user is not authorized in the org."
+  [_ f mdata]
+  (if (and (contains? cf/flags :nitrate)
+           (::auth mdata true) ;; only for endpoints that needs auth
+           (::nitrate/org-sso mdata true))
```
niwinz: My suggestion is use `::nitrate/sso` instead of `::nitrate/org-sso`

backend/src/app/rpc.clj
```diff
@@ -208,6 +211,78 @@
...
+                                   {:keys [authorized sso]} (nitrate/sso-session-authorized? cfg team-id request)
+                                   entry                    {:authorized      authorized
+                                                             :organization-id (:organization-id sso)}]
+                               (when authorized
+                                 (cache/get org-sso-auth-cache cache-key (constantly entry)))
+                               entry))]
+            (if (:authorized result)
+              (f cfg params)
+              (ex/raise :type :authentication
+                        :code :org-sso-required
```
niwinz: Maybe I'm wrong, but i think we should use `organization` instead of `org` as a cononical term. Do we have normalized it already or we sill mix org and organization on the code?

niwinz: In any case, on this case, i suggest use `:nitrate-sso-required` instead

backend/src/app/rpc.clj
```diff
@@ -220,7 +295,8 @@
     (wrap-audit cfg $ mdata)
     (wrap-spec-conform cfg $ mdata)
     (wrap-params-validation cfg $ mdata)
-    (wrap-authentication cfg $ mdata)))
+    (wrap-authentication cfg $ mdata)
+    (wrap-org-sso cfg $ mdata)))
```
niwinz: Here also, i think this should be named `wrap-nitrate-sso`

frontend/src/app/main/data/dashboard.cljs
```diff
@@ -756,15 +757,35 @@
            (when fetch? ;; If the user belonged to the org
              (rx/of (dtm/fetch-teams)))))))))
 
+(defn- handle-organization-change-sso
```
niwinz: On this case i also suggest replace organization with nitrate

frontend/src/app/main/data/dashboard.cljs
```diff
@@ -756,15 +757,35 @@
...
+  (ptk/reify ::handle-organization-change-sso
+    ptk/WatchEvent
+    (watch [_ state _]
+      (when (contains? cf/flags :nitrate)
+        (let [team-id (:current-team-id state)
+              team    (dm/get-in state [:teams team-id])
+              org-id  (dm/get-in team [:organization :id])]
+          (when (= organization-id org-id)
+            (let [url (rt/get-current-href)]
+              (->> (rp/cmd! :auth-sso {:team-id team-id :url url})
```
niwinz: Looking this rpc method in real use, I think `:auth-sso` is a bit generic method for something very specific to nitrate.  We need to think on it.

backend/src/app/util/cache.clj
pabloalba: But we have already an invalidate!

PR #10150, opened 2026-06-11
:zap: Replace toBlob to capture snapshots without blocking the main thread

### LINKED ISSUE #10045: Image rendering is slow in certain files
### Steps To Reproduce

1. Upload the attached penpot file
2. Zoom, pan and drag

### Expected behavior

User interaction should be fast, and the quality while rendering should be better

### Actual behavior

In the attached file, interaction is slow and the quality is not very good

### Screenshots or video

_No response_

_No response_

_No response_

PRO

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

[penpot_file.zip](https://github.com/user-attachments/files/28639532/4095.4097.4092_.New.Image.Resizing.zip)
[Stroke To Path Slow.zip](https://github.com/user-attachments/files/28718820/Stroke.To.Path.Slow.zip)

elenatorro:
### Related Ticket

https://github.com/penpot/penpot/issues/10045

### Summary

We observed that `canvas.toBlob`  was being executed constantly and blocking the main thread.

The viewport snapshot used on page switches and WebGL context loss event used this method on each `tiles-complete` event, which was synchronous and the PNG encoding run on the main thread, freezing the UI during these captures.

Instead, this PR uses `createImageBitmap`, making the capture async, no encoding, so it prevents blocking the main thread. Alos, instead of creating an image that points to the blob url, it paints now the bitmap on a canvas.

As a note: we close snapshots when the overlay is not showing them.

There's a caveat here: the snapshot on context lost is no longer blurred. Page transition still applies the blur (as it's applied in webgl via skia). We can discuss if we want to have this back, maybe we can apply the blur just with CSS.

### Steps to reproduce

1. Capture a performance trace and check the `toBlob` method is no longer blocking the main thread
2. Check that on page switch, there's still a transition using a blurred snapshot
3.  Check that on context loss event, the canvas is there but frozen.

```
const canvas = document.querySelector('#render');
const gl = canvas.getContext('webgl2');
const ext = gl.getExtension('WEBGL_lose_context');
ext.loseContext();
```

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/sitemap.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/sitemap.cljs
@@ -88,7 +88,7 @@
                      (p/resolved nil)
                      ;; Blur with Skia, then capture the already-blurred frame.
                      (do (wasm.api/render-blurred-snapshot!)
-                         (wasm.api/capture-canvas-snapshot-url)))
+                         (wasm.api/capture-canvas-snapshot)))
                    (p/finally
                      (fn []
                        (wasm.api/apply-canvas-blur)
--- a/frontend/src/app/main/ui/workspace/viewport_wasm.cljs
+++ b/frontend/src/app/main/ui/workspace/viewport_wasm.cljs
@@ -117,6 +117,44 @@
       (uuid? outlined-frame-id) (conj outlined-frame-id)
       (uuid? edition) (disj edition))))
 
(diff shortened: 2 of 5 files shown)


niwinz:
Is this safe change? i mean, if it is a sustantial improvement and is safe, we can consider backport it to 2.16.1

Issue #10095, opened 2026-06-10
Multiple nested flex layouts with texts are too slow


elenatorro:
## Summary

The WebGL render has performance issues rendering deeply nested frames that use flex layout.

## Steps to reproduce

- Open the files [on this project](https://hourly.penpot.dev/#/dashboard/files?team-id=3ded6912-328f-8067-8006-63a8e2767365&project-id=52ae8599-904d-8095-8008-276e4721840e) (ask for an invitation if needed)
- Try to interact (zoom/pan/hover/drag) with the elements of the file

Initial render is too slow, interactions are also slow, and mouse hovering has a lot of lag.

## Expected

- File should render smoothly
- Mouse hovering should be fast

PR #10151, opened 2026-06-11
:zap: Improve tile performance on nested frames with texts

elenatorro:
### Related Ticket

https://github.com/penpot/penpot/issues/10095

### Checklist

### DIFF
--- a/render-wasm/src/render/surfaces.rs
+++ b/render-wasm/src/render/surfaces.rs
@@ -11,7 +11,6 @@ use crate::math::Point;
 use base64::{engine::general_purpose, Engine as _};
 use std::collections::{HashMap, HashSet};
 
-const TEXTURES_CACHE_CAPACITY: usize = 1024;
 const TEXTURES_BATCH_DELETE: usize = 256;
 
 // This is the amount of extra space we're going to give to all the surfaces to render shapes.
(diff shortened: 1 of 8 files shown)


render-wasm/src/render/surfaces.rs
```diff
@@ -525,19 +560,32 @@ impl Surfaces {
...
-            &atlas_image,
-            &self.tiles.transforms,
-            &self.tiles.textures,
-            None,
-            skia::BlendMode::SrcOver,
-            self.atlas_sampling_options,
-            None,
-            None,
-        );
+        // Draw each tile via Surface::draw (clip + offset) instead of
```
AzazelN28: Removing `draw_atlas` is going to impact very negatively on rendering, because `draw_atlas` uses full GPU render parallelization.

Just for reference (this is from the fable report):
> `draw_tile_atlas_to_backbuffer` (`surfaces.rs:521-541`) already shows the right pattern: one
full-atlas snapshot + `draw_atlas`/`draw_image_rect` with source rects. The per-tile path should
do the same (snapshot the atlas once per frame, draw each tile with a `src` rect).

render-wasm/src/render/surfaces.rs
```diff
@@ -1319,11 +1366,30 @@ impl Surfaces {
     }
 
     pub fn draw_cached_tile_into_backbuffer(&mut self, tile: Tile, rect: &Rect) {
```
AzazelN28, line 1310: This is a function that we should remove (in favor of using `draw_atlas` for everything tile related)

Issue #4584, opened 2024-05-14
feature: Stable links (URLs) to shared designs


garrett:
### Is your feature request related to a problem? Please describe.

When sharing to others and then editing the files, moving boards around between pages, and so on, the links I previously shared all break.

This makes it impossible to share either a page (as often the complete view of a page is what I want to share) or a board (as the links can change) in issue trackers. (Sadly, for this particular team and project, we're using Jira instead of Taiga.)

### Describe the solution you'd like.

I'd like a link to persist. It's fine if it redirects, as long as it gets resolved to the proper location. I'd like to be able to move a board across pages and have the previously shared URLs still work somehow.

I'd like to be able to link to a board for an individual piece of the design or to a complete page as an overview (which may have multiple boards and often text outside of the board with notes).

It would be extra wonderful if the link would be human-friendly too, and not just a hash. That is, if we could have links with board and page names (where applicable), then that'd be lovely. Unfortunately, of course, names can chance so that'd break the link. (I guess it could have a unique ID and a friendly name in the URL so people know what it is by looking at the URL and the link won't break?)

### Describe alternatives you've considered.

Instead of just linking (which we have tried), I have to export PNGs and constantly update the PNGs, which can get stale and requires more back and forth work.

The problem is: Nobody knows if something is an old mockup or a new one without asking me specifically. When they don't ask, they sometimes even implement some very old preliminary design. This happens more often than we'd like, as I'm the only design on my team and I have time off or holidays from time to time while the rest of the team continues to work.

### Additional context

As always: Thanks! Penpot is great! :+1:

madalenapmelo-kp:
Hi @garrett,

I've created an enhancement in our backlog so that we can look further into this, you can find the details here: https://tree.taiga.io/project/penpot/issue/7801

garrett:
:tada: Thanks for adding it! Looking forward to it.

filipsajdak:
I'd like to help move this forward and have scoped a minimal first step that I believe is mostly already supported by the codebase.

**Observation:** the `:viewer` route already accepts a `frame-id` UUID query param, and `bundle-fetched` in `frontend/src/app/main/data/viewer.cljs` already resolves that UUID to a frame index on load. Frame UUIDs are stable across renames and reordering within a page. So a link of the form:

```
/#/view?file-id=<uuid>&page-id=<uuid>&frame-id=<uuid>
```

[...]

PR #10170, opened 2026-06-14
:sparkles: Copy stable deep link to selected boards from context menu

filipsajdak:

## What & why

The workspace **Copy link** context-menu action (and its `copy-link` shortcut) copied the raw current location, which only carries `file-id` and `page-id`. The resulting link did not point at any specific board, so it could not be used to reference "this board" from an issue/MR.

This builds the link from the current route plus a `board-id` query param derived from the current selection:

- a selected board contributes its own id;
- a selected shape contributes the id of the board it belongs to.

`board-id` is an existing workspace route param (`zoom-to-frame` already selects and zooms to it on navigation), and board ids are stable across renames and reordering, so the link keeps resolving to the same board. When the selection contains no board, it falls back to the previous behavior (current location).

## Scope

Editor-side deep links only, kept deliberately minimal. Making the **viewer** share link frame-precise (so shared/public links can target a specific board) is a natural follow-up; noted on the issue.

## Related

Relates to #4584 (and the discussion comment there).

## Testing

- `clj-kondo` lint of the changed namespace: 0 errors, 0 warnings.
- Behavior verified by code-path inspection: `app.main.data.workspace/zoom-to-frame` already consumes the `board-id` query param (string or vector), maps it through `uuid/uuid`, and runs `select-shapes` + `zoom-to-selected-shape`, so a generated `board-id` link resolves to the board on load.

I have not built the full devenv in this environment, so I did not exercise it end-to-end in a running instance. Happy to add a Playwright case and/or run the devenv verification if you'd like.

## Risk

Low: single function, no API/DB/schema change, no new dependencies, reuses an existing route param.

### DIFF
--- a/frontend/src/app/main/data/workspace/clipboard.cljs
+++ b/frontend/src/app/main/data/workspace/clipboard.cljs
@@ -1132,11 +1132,35 @@
         (rx/of (dwm/upload-media-workspace params))))))
 
 (defn copy-link-to-clipboard
+  "Copy a link to the clipboard. When one or more boards are part of the
+  current selection, the link deep-links to those boards through their
+  ids (`board-id` route param), which are stable across renames and
+  reordering. Otherwise it falls back to the current location."
   []
   (ptk/reify ::copy-link-to-clipboard
     ptk/WatchEvent
-    (watch [_ _ _]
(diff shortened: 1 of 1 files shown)


madalenapmelo-kp:

Issue #10169, opened 2026-06-13
Plugin API: targeted RPC read of plugin-data without full file download


filipsajdak:
### Summary

The Plugin API can already **write** arbitrary key/value `pluginData` onto shapes and the file root (`setPluginData` / `setSharedPluginData`), and it is persisted inside the file data. However, there is no way to **read** that data back via the RPC API without downloading and decoding the entire file. I'd like to propose a small, read-only RPC command to fetch plugin-data for a file (optionally a specific object / namespace / key).

### Motivation

`pluginData` is a natural place to attach stable external identifiers — e.g. a requirement ID, a GitLab issue/epic reference, or a git commit SHA — onto a file or a board. That enables bidirectional traceability between designs and an external tracker. Today an external CI job or MCP/agent that wants to answer *"which board carries `req-id=REQ-123`?"* must pull the whole file blob per query, which is expensive and awkward. A targeted read closes that gap and complements the existing write API.

### Current state (as I understand it)

- Schema: `common/src/app/common/types/plugins.cljc` (`schema:plugin-data`).
- Stored on shapes (`common/src/app/common/types/shape.cljc`) and the file (`common/src/app/common/types/file.cljc`); namespaces are `:plugin/<id>` (private) and `:shared/<ns>` (shared).
- Written via the `:set-plugin-data` change op; read only through the in-memory frontend plugin runtime (`frontend/src/app/plugins/{shape,file}.cljs`) — no RPC read path.

### Proposed design (read-only, no data-model change)

A new RPC command `get-plugin-data`, modelled on the existing `get-file-stats` pattern in `backend/src/app/rpc/commands/files.clj` (decode → extract one dimension → return a minimal map), gated by `check-read-permissions!`. Rough shape:

- params: `file-id` (required); optional `object-id` + `page-id`; optional `namespace` / `key` filter.
- returns: just the matching plugin-data entries, not the full file.

No new change type, no migration, no new dependency. If you'd also like the read exposed through the plugin runtime for parity (so plugins can read data for a not-currently-loaded context), I can add that as a separate follow-up with the `plugins/CHANGELOG.md` entry — but I'd keep the first PR backend-only for reviewability.

### Questions before I implement

1. Is a dedicated `get-plugin-data` command the shape you'd prefer, or would you rather extend an existing command (e.g. an opt-in projection on `get-file`)?
[...]

PR #10171, opened 2026-06-14
:sparkles: Add get-file-plugin-data RPC to read file plugin data

filipsajdak:

## What & why

The Plugin API can write arbitrary key/value **plugin data** onto the file root (`setPluginData` / `setSharedPluginData`), and it is persisted inside the file data. But there was no way to read it back over RPC without downloading and decoding the whole file.

This adds a `get-file-plugin-data` read command that returns the file-level plugin data map, optionally narrowed to a single `namespace`. It lets external tooling (CI, an MCP/agent) cheaply fetch metadata such as issue/commit/requirement references attached to a file. This is the read counterpart to the existing write API and the primitive behind the traceability use case in #10169.

## Design

- New `::get-file-plugin-data` command in `backend/src/app/rpc/commands/files.clj`, modelled on `::get-file-stats`: `check-read-permissions!`, `bfc/get-file`, return a minimal map.
- Result schema reuses `app.common.types.plugins/schema:plugin-data`.
- Read-only: **no** data-model change, **no** new change type, **no** migration, **no** new dependency. `files` is already in `resolve-methods`, so no `rpc.clj` change.

## Scope

File-root plugin data only, kept deliberately small. Object/page/color/etc. level reads are a natural follow-up.

While scoping the page case I noticed an inconsistency that I asked about in #10169: `process-change` writes page plugin data to `[:pages-index id :plugin-data ns key]`, whereas `changes-builder/set-plugin-data` reads the previous value from `[:pages-index id :options :plugin-data ns key]` (extra `:options`). Confirming the canonical page path is a prerequisite for an object/page-level read, so I left it out here.

## Related

Relates to #10169 (and the discussion there).

## Testing

- `clj-kondo` (repo config): 0 errors; the command file is warning-clean (the test file's pre-existing warnings are untouched).
- Added `get-file-plugin-data` and `get-file-plugin-data-forbidden` tests in `backend/test/backend_tests/rpc_file_test.clj` (empty file -> `{}`; set file-level data via a `:set-plugin-data` change -> full map; namespace filter; absent namespace -> `{}`; non-member -> `:not-found`), mirroring the existing `get-file-stats` tests.

I did not build the backend devenv (no test DB in this environment), so I did not execute the suite here. Happy to run it / adjust once a maintainer confirms the direction.

## Risk

Low: a single read-only command reusing existing helpers and an existing schema.

### DIFF
--- a/backend/src/app/rpc/commands/files.clj
+++ b/backend/src/app/rpc/commands/files.clj
@@ -21,6 +21,7 @@
    [app.common.transit :as t]
    [app.common.types.components-list :as ctkl]
    [app.common.types.file :as ctf]
+   [app.common.types.plugins :as ctpg]
    [app.common.uri :as uri]
    [app.config :as cf]
    [app.db :as db]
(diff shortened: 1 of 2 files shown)


madalenapmelo-kp:

Issue #7091, opened 2025-08-07
feature(design-tokens): Option to export Tokens to Git repository


SethFalco:
### Is your feature request related to a problem? Please describe.

I would like to streamline the process for the designer to push/propose changes in design tokens to the project.

This way, when there are styles only changes or when new tokens are introduced which are not breaking changes, it can have a dedicated pull request attributed to them. Then a developer can review, merge, and use them as necessary.

A developer could then take it upon themselves to configure their CI environment to do staging deploys to see a preview with the new generated CSS.

### Describe the solution you'd like.

Could Penpot have an option to export design tokens to a Git repository?

I would expect this to work similarly to [Weblate](https://weblate.org/en/).

I can see Penpot itself uses Weblate to manage localization, so I'll just give an overview. Weblate is a localization platform with great Git integration. You specify your VCS URL, and what the relevant file to update is, and it can commit or open merge requests itself.

In the case of Penpot, what I'd hope for is:

* Specify the VCS URL, configure the branch, location of `tokens.json`, etc.
* The designer can Export to Git, which just uses a bot user / access token to open a pull request with the changes in design tokens.
* The pull request should use proper Git authoring to attribute the email address associated with the commit on Penpot. ¹

Optional:

* If you would see any value in it, this could even include VCS statuses like removed, added, or changed, along with actions like revert, or to allow a toggleable diff view of the design itself so we can easily identify what's affected.

¹ Here's a demonstration of why preserving authorship is important in my opinion. Many cloud-based tools that integrate with Git "steal" attribution. However, Weblate does this correctly. Despite the translator using a cloud platform, the translator is still the author despite not making the commit.

<img width="1300" height="177" alt="Image" src="https://github.com/user-attachments/assets/1502b998-d379-4853-bf76-ffe29dd9f387" />

### Describe alternatives you've considered.

I currently just export a `tokens.json` and put it in my repo and commit that to Git myself.

I'm not an expert, let me know if that's a weird workflow. 🤔

Reference: https://github.com/svg/svgo.dev/pull/41

### Additional context

N/A

myfunnyandy:
Hey! GitHub sync is in our plans. This is the [user story](https://tree.taiga.io/project/penpot/us/9968). I've added this request there so this proposal and the ideas behind it can be taken into account when we start working on it. 

Also, if you're curious about more plans related to tokens, you can take a look at [this post](https://community.penpot.app/t/what-s-next-for-penpot-design-tokens).

Thank you very much!

[...]

PR #10172, opened 2026-06-14
:sparkles: Add export-file-tokens RPC to push tokens via webhook

filipsajdak:

## What & why

Issue #7091 requests exporting design tokens to a Git repository (with the designer as Git author). Rather than embedding a Git client and credential storage in Penpot, this takes the webhook route proposed in the issue discussion:

A new `export-file-tokens` RPC command serializes the file's design tokens to a W3C DTCG `tokens.json` and emits them as a **webhook event**. The team's existing webhook + its CI (GitLab CI, GitHub Actions, ...) receives the payload, commits `tokens.json`, and opens a PR attributing the designer.

## Design

- New `::export-file-tokens` command in `backend/src/app/rpc/commands/files.clj`: `check-read-permissions!`, read tokens via `bfc/get-file`, serialize with the existing `app.common.types.tokens-lib/export-dtcg-json` (same output as the manual export), raise `:validation`/`:no-tokens-in-file` when the file has no tokens.
- Marked `::webhooks/event? true`; the DTCG JSON and attribution (`file-id`, `file-name`, `project-id`, `exported-by`, `tokens-json`) ride in `::audit/props`, which the existing dispatcher (`app.loggers.webhooks`) delivers to the team's webhook.
- **The tokens are delivered as a pre-serialized JSON string**, on purpose: the webhook dispatcher camel-cases every payload key (`json-write-opts {:key-fn str/camel}`), which would otherwise corrupt DTCG token keys. Keeping `tokens.json` as an opaque string sidesteps that; the receiver does one `JSON.parse`.
- **No** new dependency, **no** credential storage, **no** DB migration, **no** `common/` change. Reuses the whole existing webhook stack.

## Scope / follow-up

This PR is the backend mechanism. The **frontend "Export to Git" tab** in the token export modal is a deliberate follow-up: it needs new UI strings, and Penpot's gettext `.po` translations are regenerated by the `pnpm run translations` extractor across ~30 locales, which I could not run in this environment. I did not want to hand-edit `.po` files blind and risk a broken i18n state. Happy to add the tab (and run the extractor) as a follow-up once the approach is confirmed.

## Related

Relates to #7091 (and the discussion comment there asking whether this webhook approach is preferred over a built-in Git client).

## Testing

- `clj-kondo` (repo config): 0 errors; `files.clj` is warning-clean (only the test file's pre-existing warnings remain).
- Added tests in `backend/test/backend_tests/rpc_file_test.clj`: no-tokens file -> `:validation`/`:no-tokens-in-file`; file with a tokens lib (seeded via a `:set-tokens-lib` change) -> `{:exported true}`; non-member -> `:not-found`.

I did not build the backend devenv (no test DB in this environment), so I did not execute the suite or a live webhook delivery here. Happy to run both once a maintainer confirms the direction.

## Risk

Moderate: one read-only-ish command reusing existing webhook/audit machinery. The webhook only fires when the team has an active webhook and the `:webhooks` flag is enabled; otherwise the command is a no-op beyond returning `{:exported true}`.

### DIFF
--- a/backend/src/app/rpc/commands/files.clj
+++ b/backend/src/app/rpc/commands/files.clj
@@ -14,6 +14,7 @@
    [app.common.files.helpers :as cfh]
    [app.common.files.migrations :as fmg]
    [app.common.files.stats :as cfs]
+   [app.common.json :as json]
    [app.common.logging :as l]
    [app.common.schema :as sm]
    [app.common.schema.desc-js-like :as-alias smdj]
@@ -21,6 +22,7 @@
    [app.common.transit :as t]
    [app.common.types.components-list :as ctkl]
    [app.common.types.file :as ctf]
+   [app.common.types.tokens-lib :as ctob]
    [app.common.uri :as uri]
    [app.config :as cf]
    [app.db :as db]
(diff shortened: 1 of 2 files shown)


madalenapmelo-kp:

niwinz:
I think the approach is in wrong direction, we should maybe improve the file-update webhook event to collect the high level changes. And expose a handler for get the file-tokens without obtaining all the file.

The current approach is just a hook request for emit a webhook, and it may works but the design is strange. Because the name says export but it does not exports.

Additionally, we are working right now in a generalization of exportation process where we can add webhooks, so i will take a note on this issue in order to design correctly that export and import operations.

niwinz:
From my point of view i'm open to a method for obtain tokens instead of the whole file. But the current approach as is  will not be merged on penpot.
In any case thanks for all the value and info.

filipsajdak:
Thanks @niwinz, agreed: the name promises an export and the command only fires a webhook.
I'll withdraw this approach and keep the PR open only until we agree on the shape of what replaces it.

**The tokens reader.**
I'd like to propose `get-file-tokens {:file-id}`:

- same read-permission check as `get-file`;
- returns the file's tokens-lib, or `nil` when the file has no tokens (a value, not an error);
- typed with the existing tokens-lib schema, so JSON clients get DTCG through its `:encode/json` encoder and transit clients get the native value (checked on `develop`: `get-file` over JSON already returns `tokensLib` as DTCG, with set and token names like `brand-core` / `primary-base` left as is);
- reads `get-tokens-lib` the same way `calculate-library-summary` already does.

A CI job holding an access token can then fetch just the tokens after a webhook fires.
Does that name and shape work for you? I'll open a fresh PR against #7091 once you confirm.

**A caveat for the `update-file` webhook summary, as input for your export work.**
`update-file` webhooks are batched per profile and file over 2 minutes.
Batching removes the pending task and queues the new one, so only the last call's props are delivered.
I checked this on `develop` (abb4e00746): a token edit followed 10 s later by a page rename left one queued `process-webhook-event` task, and the single delivery carried only the rename call's event.
A per-call change summary would therefore drop "tokens changed" whenever a later non-token edit lands in the same window.
One way round it: when deduping, merge the pending task's change categories into the new one (for example `:changes #{:tokens :components :objects}`), derived from the `token-change-types` and `library-change-types` sets that `files_update.clj` already has.

For completeness, #10171 (`get-file-plugin-data`) is the same "read one slice without the whole file" pattern, if you want to judge the two together.

PR #10190, opened 2026-06-15
:bug: Fix stroke to path extra points

### LINKED ISSUE #10002: "Stroke to path" generates extra nodes, producing dirty paths and visual artifacts
https://tree.taiga.io/project/penpot/issue/14294

### Description

When converting a stroke to a path, the resulting path contains a large number of extra, unnecessary nodes. This makes the generated SVG harder to work with and can produce visual artifacts such as unexpected gaps or holes in the shape. Competing tools generate a much cleaner path with only the nodes needed to represent the shape accurately.

### Environment

*   hourly (stroke-to-path config flag)
*   Vector editing / Stroke to path

### Steps to reproduce

1.  Draw a path with curves and angles and apply a stroke to it.
2.  Right-click and select **Stroke to path.**
3.  Inspect the resulting path's nodes on the canvas.

Image

### Current result

The generated path contains many extra nodes beyond what is needed to represent the shape. This can cause visual artifacts such as holes or gaps in the fill, and makes the SVG output unnecessarily complex. See attached Image:

Image

### Expected result

The generated path should be as clean as possible, containing only the nodes necessary to represent the stroke geometry accurately — consistent with the output of other design tools. See attached Image 1 for reference:

Image

Design: 
External source: 
External source 2:

AzazelN28:
### Related Ticket

[GitHub Issue 10002](https://github.com/penpot/penpot/issues/10002)

stroke-to-path

### Summary

When converting a stroke to a path, the resulting path contains a large number of extra, unnecessary nodes. This makes the generated SVG harder to work with and can produce visual artifacts such as unexpected gaps or holes in the shape. Competing tools generate a much cleaner path with only the nodes needed to represent the shape accurately.

### Steps to reproduce 

1. Draw a path with curves and angles and apply a stroke to it.
2. Right-click and select Stroke to path.
3. Inspect the resulting path's nodes on the canvas.

### Checklist

### DIFF
--- a/render-wasm/src/shapes/stroke_paths.rs
+++ b/render-wasm/src/shapes/stroke_paths.rs
@@ -58,27 +58,24 @@ pub fn stroke_to_path(
     // For inner/outer strokes, use boolean ops to clip
     // the 2×-width stroke outline to the correct region.
     // Set EvenOdd to preserve the annular ring's inner hole,
-    // then as_winding() on the result fixes contour winding
-    // for Penpot's NonZero fill rule.
+    // then switch to Winding for Penpot's NonZero fill rule.
+    // Use set_fill_type instead of as_winding() because as_winding()
+    // decomposes self-intersecting geometry, which removes points
+    // at intersections of straight lines in closed paths.
+    // Center strokes skip the conversion: fill_path_with_paint
+    // already produces correctly-wound contours.
     let final_path = match render_kind {
-        StrokeKind::Inner => {
-            stroke_outline.set_fill_type(skia::PathFillType::EvenOdd);
-            let inner = stroke_outline
-                .op(&transformed_shape_path, skia::PathOp::Intersect)
-                .unwrap_or(stroke_outline);
-            inner.as_winding().unwrap_or(inner)
-        }
-        StrokeKind::Outer => {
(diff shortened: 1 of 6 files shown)


elenatorro:
These changes are ok! But I still can reproduce one of the problems explained in the issue, this is how I reproduce it:

https://github.com/user-attachments/assets/549d2d98-9df4-4beb-ab05-d857b4494356

AzazelN28:
> These changes are ok! But I still can reproduce one of the problems explained in the issue, this is how I reproduce it:
> 
>  New.File.9.-.Penpot.mp4

Now it is fixed.

render-wasm/src/shapes/stroke_paths.rs
```diff
@@ -58,27 +58,24 @@ pub fn stroke_to_path(
     // For inner/outer strokes, use boolean ops to clip
     // the 2×-width stroke outline to the correct region.
     // Set EvenOdd to preserve the annular ring's inner hole,
```
elenatorro, line 60: This comment is not entirely right, we're not using EventOdd as fill type, right?

elenatorro (approved):

render-wasm/src/shapes/stroke_paths.rs
AzazelN28: Right.

PR #10250, opened 2026-06-17
:bug: Surface real plugin errors instead of bare "Value not valid"

### LINKED ISSUE #9692: Plugin API: handle-error swallows plain JS errors + empty malli explains → bare "Value not valid. Code: :error"
## Summary

`app.plugins.utils/handle-error` in `frontend/src/app/plugins/utils.cljs` swallows two cause shapes:

1. **Plain JS `Error`** (e.g. `TypeError`) — has no `IExceptionInfo`, so `ex-data` returns `nil` → `(get (ex-data cause) :app.common.schema/explain)` is `nil` → message falls to `(ex-data cause)` which is also `nil` → `not-valid` is called with `value=nil` → user sees `[PENPOT PLUGIN] Value not valid. Code: :error` with **no clue what threw**.
2. **CLJS ExceptionInfo with `::sm/explain`** where `error-messages` returns an empty string — same outcome, bare error with no schema details.

Affects all plugin proxy methods (every `applyToken`, `addTheme`, etc. that has a `try/catch → handle-error` wrapper).

## Reproduction

```js
// Trigger any plain JS error inside a plugin proxy method body. E.g. invoke a
// method whose CLJS schema fn unintentionally references a missing arity
// dispatch:  $cljs$core$first$$.$cljs$core$IFn$_invoke$arity$1$(x)
// The Closure-compiled `cljs.core/first` is single-arity, so .IFn_invoke_arity_1$
// is undefined. The TypeError surfaces inside the try{} and goes to handle-error,
// which strips it and re-throws via not-valid with `value=nil`.

try {
  penpot.library.local.tokens.addTheme({ name: "X", group: "Y" });
} catch (e) {
  console.log(e.message);
  // => "[PENPOT PLUGIN] Value not valid. Code: :error"
  // The real TypeError is logged to console.log via (.-stack cause), but
  // structured plugin output (the value the user catches) carries nothing.
}
```

The TypeError is logged via `(js/console.log (.-stack cause))` so it's visible in the host-page DevTools, but the plugin sandbox can't reach the host page console. Tools like MCP `execute_code` only see the value the user catches — which is the bare string.

## Real-world impact

We chased this on a self-hosted 2.15.2 instance for 5+ rounds of cross-Claude debugging. The bare error masked a hand-port bug (`first.IFn$_invoke$arity$1$ is not a function`) for hours of static-analysis. A surface-level message — either the cause's `.-message` or `pr-str` of the explain map — would have collapsed the search to one comment.

Related fixes already merged:
- #9486 `:bug: Fix plugin API error message for nested malli validation paths` — addresses the `interpret-schema-problem` nested-path destructure
- #9632 `:bug: Fix plugins schema validation error` — proxy preservation in `frontend/src/app/util/object.cljc`

Both address the *content* of validation errors when they're well-formed. Neither addresses the **empty/missing cause** path that still results in bare `"Value not valid. Code: :error"`.

## Proposed fix

`frontend/src/app/plugins/utils.cljs` — add a 2-layer fallback chain to `handle-error`:

```diff
 (defn handle-error
   "Function to be used in plugin proxies methods to handle errors and print a readable
    message to the console."
   [plugin-id]
   (fn [cause]
     (let [message
           (if-let [explain (-> cause ex-data ::sm/explain)]
             (do
               (js/console.error (sm/humanize-explain explain))
-              (error-messages explain))
-            (ex-data cause))]
+              (or (error-messages explain)
+                  (try (sm/humanize-explain explain) (catch :default _ nil))
+                  (pr-str explain)))
+            (or (ex-data cause)
+                (.-message cause)
+                (str cause)))]
       (js/console.log (.-stack cause))
       (not-valid plugin-id :error message))))
```

### Why each fallback layer

**A-branch (explain path):**
- `error-messages explain` — preferred i18n path (translated, user-friendly)
- `sm/humanize-explain explain` — malli's built-in formatter (falls back when interpret-schema-problem doesn't recognize the entries)
- `pr-str explain` — EDN dump of the raw explain map (last resort; always non-empty; exposes `:errors`, `:schema`, `:value`)

**B-branch (no-explain path):**
- `ex-data cause` — original behavior, preserves any CLJS error-data map
- `.-message cause` — JS error message (catches TypeError, RangeError, ReferenceError from inside fn bodies)
- `str cause` — final stringification (covers exotic throws like `throw "literal"`)

Both branches now guarantee a non-nil `message` → `not-valid` always produces an actionable user-facing string.

## Verification

After patching, the addTheme-with-broken-fn-body reproducer above produces:

```
[PENPOT PLUGIN] Value not valid: TypeError: $APP.$cljs$core$first$$.$cljs$core$IFn$_invoke$arity$1$ is not a function. Code: :error
```

…which is what we needed all along.

## Notes

- We deployed this exact patch on self-hosted 2.15.2 (compiled-bundle byte-level edit) and ran it in production for ~3h before opening this issue. Caught no false positives, surfaced the underlying bug immediately on the first plugin failure that hit the empty-cause path.
- I'm happy to open the PR myself if you'd prefer — let me know. The fix is two `let`-binding sites, no new dependencies.
- Cross-referenced from cross-Claude debugging session: https://github.com/tknatwork/claude-bus/issues/3 (closed, full retrospective)

filipsajdak:

### Related Ticket

Fixes [#9692](https://github.com/penpot/penpot/issues/9692).

### Summary

The plugin proxy error handler (`obj/reify` `:on-error` -> `handle-error` in `frontend/src/app/plugins/utils.cljs`) only produced a useful message for malli `ExceptionInfo` carrying `::sm/explain`. Two paths fell through to the bare `"[PENPOT PLUGIN] Value not valid. Code: :error"`, hiding the real cause:

1. **Plain JS errors** (e.g. a `TypeError` thrown inside a proxy method) are not `ExceptionInfo`, so `(ex-data cause)` is `nil`; the message bound to `nil` and the real `.-message` only reached the host-page console (invisible to the plugin sandbox / MCP `execute_code`).
2. **Empty/unmappable malli explains** made `error-messages` return `""`, which rendered as `"Value not valid: . Code: :error"`.

Changes (frontend-only, two adjacent fns):
- `error-messages` returns `nil` instead of `""` when there is nothing to render, so the caller can fall back.
- `handle-error` adds fallback chains: explain branch `(or (error-messages explain) (pr-str explain))`; non-explain branch `(or (ex-data cause) (ex-message cause) (str cause))`.

Working cases (well-formed explain producing a non-empty message) are unchanged; this only improves the previously-bare path. It complements #9486/#9632, which fixed the *content* of well-formed explains, not the empty/missing-cause path.

From a plugin:
```js
// Before: "[PENPOT PLUGIN] Value not valid. Code: :error"
// After:  the real error message, e.g. "first is not a function"
penpot.createBoard().nonExistentMethodThatThrows();
```
Any proxy method that throws a plain JS error, or fails malli validation with an explain that doesn't map to a field message, now surfaces a useful message.

### Checklist

### Notes

- Per `CONTRIBUTING.md`, the changelog is generated at release time, so this PR intentionally does not modify `CHANGES.md`. No plugin-types surface changes, so no `plugins/CHANGELOG.md` entry either.

- Tests added to `frontend/test/frontend_tests/plugins/utils_test.cljs` (plain-JS-error path, empty-explain path, `error-messages` nil contract).
- `clj-kondo` and `cljfmt` are clean on the changed files.

### DIFF
--- a/frontend/src/app/plugins/utils.cljs
+++ b/frontend/src/app/plugins/utils.cljs
@@ -304,12 +304,16 @@
 
 (defn error-messages
   [explain]
-  (->> (:errors explain)
-       (reduce csm/interpret-schema-problem {})
-       (flatten-error-map)
-       (map (fn [[field message]]
-              (tr "plugins.validation.message" field message)))
-       (str/join ". ")))
+  (let [msg (->> (:errors explain)
+                 (reduce csm/interpret-schema-problem {})
+                 (flatten-error-map)
+                 (map (fn [[field message]]
+                        (tr "plugins.validation.message" field message)))
+                 (str/join ". "))]
+    ;; Return nil (not "") when the explain has no mappable errors, so
+    ;; `handle-error` can fall back to a non-empty message instead of
+    ;; surfacing a bare "Value not valid. Code: :error" (#9692).
+    (when-not (str/blank? msg) msg)))
 
 (defn handle-error
   "Function to be used in plugin proxies methods to handle errors and print a readable
@@ -320,8 +324,13 @@
           (if-let [explain (-> cause ex-data ::sm/explain)]
             (do
               (js/console.error (sm/humanize-explain explain))
-              (error-messages explain))
-            (ex-data cause))]
+              ;; A well-formed explain renders a message; an empty or
+              ;; unmappable one falls back to the raw explain so we never
+              ;; surface a bare "Value not valid" (#9692).
+              (or (error-messages explain) (pr-str explain)))
+            ;; Non-malli throwables (plain js/Error, TypeError, …) carry no
+            ;; ::sm/explain, so surface their own message instead of nil (#9692).
+            (or (ex-data cause) (ex-message cause) (str cause)))]
       (js/console.log (.-stack cause))
       (not-valid plugin-id :error message))))
 
--- a/frontend/test/frontend_tests/plugins/utils_test.cljs
+++ b/frontend/test/frontend_tests/plugins/utils_test.cljs
@@ -54,3 +54,36 @@
   ;; No validation errors -> no output (callers join with ". " and would
   ;; otherwise emit an empty string, which is fine).
   (t/is (empty? (flatten-error-map {}))))
+
+;; ---------------------------------------------------------------------
+;; Issue #9692 — `handle-error` must surface a useful message instead of a
+;; bare "Value not valid. Code: :error".
+;;
+;; `not-valid` is redefined to capture the rendered message directly, so the
+;; assertions don't depend on `st/state` or the console.
+
+(t/deftest test-handle-error-plain-js-error
+  ;; A plain JS error has no `::sm/explain` and CLJS `ex-data` returns nil, so
+  ;; the handler must fall back to the error's own message rather than nil.
+  (let [captured (atom ::none)]
+    (with-redefs [plugins.utils/not-valid (fn [_plugin-id _code value]
+                                            (reset! captured value) nil)]
+      ((plugins.utils/handle-error #uuid "00000000-0000-0000-0000-000000000000")
+       (js/Error. "boom: not a function")))
+    (t/is (= "boom: not a function" @captured))))
+
+(t/deftest test-handle-error-empty-explain
+  ;; An explain whose errors don't render any message must not produce an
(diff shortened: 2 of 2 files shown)


Alotor:
Thanks for your contribution!

PR #10261, opened 2026-06-17
:bug: Fix plugin API board.guides invalid-schema on set/clear

### LINKED ISSUE #9773: Setting or clearing `board.guides` from a plugin throws a malli invalid-schema error
**Describe the bug**

Assigning a value to `board.guides` from a plugin (Plugin API) always fails. Both
setting guides and clearing them (`board.guides = []`) throw the same error:

```
{:type :malli.core/invalid-schema,
 :data {:schema :app.common.types.grid/grid, :form :app.common.types.grid/grid}}
```

There are two underlying causes:

1. **Unregistered schema reference.** The `guides` setter in
   `frontend/src/app/plugins/shape.cljs` validates against the registered-keyword
   reference `::ctg/grid` (`:app.common.types.grid/grid`):

   ```clojure
   (not (sm/validate [:vector ::ctg/grid] value))
   ```

   That keyword is no longer registered. Commit `575342b3bb`
   (":recycle: Use direct schemas instead of references") removed
   `(sm/register! ::grid schema:grid)` from
   `common/src/app/common/types/grid.cljc` but did not update this plugin call
   site. Because malli can't resolve the reference, validation fails for **any**
   value — including an empty vector — which is why both setting and clearing
   guides break. Every other setter in `shape.cljs` already uses direct schema
   vars (e.g. `cts/schema:stroke`, `types.fills/schema:fill`); this is the only
   remaining keyword reference.

2. **`parse-frame-guide` returns the parser function instead of calling it.** In
   `frontend/src/app/plugins/parser.cljs`, the `column` and `row` branches yield
   the parser *function* rather than its result:

   ```clojure
   (case (obj/get guide "type")
     "column" parse-frame-guide-column      ;; returns the fn, not (… guide)
     "row"    parse-frame-guide-row          ;; returns the fn, not (… guide)
     "square" (parse-frame-guide-square guide))
   ```

   So even with cause 1. fixed, `column`/`row` guides would still fail validation
   (the vector contains a function instead of a guide map). Only `square` is
   invoked correctly.

   (Minor: `parse-frame-guide-square` also calls `parse-frame-guide-column-params`
   instead of the dedicated, otherwise-dead `parse-frame-guide-square-params`.)

**To Reproduce**

1. Open a Penpot file and create/select a board.
2. From a plugin (with `content:write` permission) run:
   ```js
   board.guides = [{
     type: "column",
     display: true,
     params: {
       color: { color: "#DE4762", opacity: 0.2 },
       type: "stretch", size: 12, gutter: 16, margin: 16
     }
   }];
   ```
3. See the error. The same happens for `type: "row"`, and even for
   `board.guides = []` (clearing guides).

**Expected behavior**

Assigning a valid array of guides applies them to the board, reading
`board.guides` back returns the same guides, and `board.guides = []` clears them
— without any error.

**Actual behavior**

Validation throws `:malli.core/invalid-schema` for
`:app.common.types.grid/grid` and the assignment is rejected, regardless of the
value passed (column, row, square, or empty array).

**Screen recordings and screenshots**

N/A (reproducible from the plugin console).

**Desktop (please complete the following information):**
 - OS: macOS (Darwin 25.5.0)
 - Browser: any
 - Version: any

**Additional context**

- Penpot version: `2.15.4-RC1-1261-g565ed042bc` (commit `565ed042bc`)
- Regression introduced by commit `575342b3bb` ("Use direct schemas instead of
  references").
- Affected files:
  - `frontend/src/app/plugins/shape.cljs` (`guides` setter, ~line 1532)
  - `frontend/src/app/plugins/parser.cljs` (`parse-frame-guide`, `parse-frame-guide-square`)
- Suggested fix:
  - Replace `::ctg/grid` with the direct schema var `ctg/schema:grid`.
  - Call the column/row parsers: `(parse-frame-guide-column guide)` /
    `(parse-frame-guide-row guide)`.
  - Use `parse-frame-guide-square-params` in `parse-frame-guide-square`.

filipsajdak:

### Related Ticket

Fixes [#9773](https://github.com/penpot/penpot/issues/9773).

### Summary

Setting or clearing `board.guides` from a plugin threw `:malli.core/invalid-schema` for every value (including `[]`). Three causes:

1. `frontend/src/app/plugins/shape.cljs` validated against `[:vector ::ctg/grid]`, but the `:app.common.types.grid/grid` reference is no longer registered (the namespaced-keyword registrations were replaced by direct schemas). malli then throws for any value. Fix: use the direct var `ctg/schema:grid` (already required as `ctg`), matching every other setter.
2. `frontend/src/app/plugins/parser.cljs` `parse-frame-guide` returned the `parse-frame-guide-column` / `parse-frame-guide-row` fns instead of calling them with `guide`, so column/row guides parsed to a vector containing a function — failing validation even after (1).
3. `parse-frame-guide-square` used `parse-frame-guide-column-params` instead of the dedicated `parse-frame-guide-square-params`.

```js
const [board] = penpot.currentPage.findShapes({ type: "board" });
board.guides = [{ type: "square", display: true,
                  params: { color: { color: "#DE4762", opacity: 0.2 }, size: 8 } }];
board.guides = [];   // clearing
// before: "Value not valid" (invalid-schema thrown); after: both succeed
```

### Checklist

### Notes

- Per `CONTRIBUTING.md`, the changelog is generated at release time, so this PR intentionally does not modify `CHANGES.md`. No plugin-types surface changes, so no `plugins/CHANGELOG.md` entry either.

- Added a regression test in `frontend/test/frontend_tests/plugins/parser_test.cljs` that parses column/square guides and validates the result (and an empty vector) against `ctg/schema:grid`.
- `clj-kondo` and `cljfmt` clean on the changed files.

### DIFF
--- a/frontend/src/app/plugins/parser.cljs
+++ b/frontend/src/app/plugins/parser.cljs
@@ -336,17 +336,17 @@
     (d/without-nils
      {:type (-> (obj/get guide "type") parse-keyword)
       :display (obj/get guide "display")
-      :params (-> (obj/get guide "params") parse-frame-guide-column-params)})))
+      :params (-> (obj/get guide "params") parse-frame-guide-square-params)})))
 
 (defn parse-frame-guide
   [^js guide]
   (when (some? guide)
     (case (obj/get guide "type")
       "column"
-      parse-frame-guide-column
+      (parse-frame-guide-column guide)
 
       "row"
-      parse-frame-guide-row
+      (parse-frame-guide-row guide)
 
       "square"
       (parse-frame-guide-square guide))))
--- a/frontend/src/app/plugins/shape.cljs
+++ b/frontend/src/app/plugins/shape.cljs
@@ -1562,7 +1562,7 @@
                         (let [id (obj/get self "$id")
                               value (parser/parse-frame-guides value)]
                           (cond
-                            (not (sm/validate [:vector ::ctg/grid] value))
+                            (not (sm/validate [:vector ctg/schema:grid] value))
                             (u/not-valid plugin-id :guides value)
 
                             (not (r/check-permission plugin-id "content:write"))
--- a/frontend/test/frontend_tests/plugins/parser_test.cljs
+++ b/frontend/test/frontend_tests/plugins/parser_test.cljs
@@ -7,6 +7,8 @@
 (ns frontend-tests.plugins.parser-test
   (:require
    [app.common.geom.point :as gpt]
+   [app.common.schema :as sm]
+   [app.common.types.grid :as ctg]
    [app.plugins.parser :as parser]
    [cljs.test :as t :include-macros true]))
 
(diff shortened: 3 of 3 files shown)


Alotor:
Thanks for your contribution!

PR #10262, opened 2026-06-17
:bug: Let plugin addToken create tokens in an inactive set

### LINKED ISSUE #10070: Plugin API reference token creation fails when the target token set is inactive
### Steps To Reproduce

1. Create a `primitives` token set.
2. Add a primitive color token such as `color.gray.50` to that set.
3. Create a second token set, for example `semantic`.
4. Without activating the referenced set, try to add a reference token in the semantic set:

```js
sem.addToken({
  type: "color",
  name: "color.bg.default",
  value: "{color.gray.50}"
});
```

5. Activate the involved sets with `toggleActive()` and repeat the same `addToken` call.

### Expected behavior

Creating a reference token should either be allowed when the referenced token exists structurally, even if its set is inactive, or the API should return a specific actionable error explaining that the referenced token set is inactive.

The preferred behavior would be to allow unresolved-but-valid references at creation time and resolve them once the relevant set is activated.

### Actual behavior

The first `addToken` call fails with a generic validation error even though the referenced token exists. After activating the referenced set, the exact same call succeeds.

This makes the failure look like invalid reference syntax or a missing token, when the actual problem is token set activation.

### Screenshots or video

_No response_

_No response_

_No response_

- MCP server: Penpot remote MCP.
- Tools used: `high_level_overview`, `penpot_api_info`, `execute_code`, `export_shape`.
- Method: read-only API documentation review followed by live smoke tests against a connected plugin instance.
- Scope: Penpot MCP server and Penpot Plugin API behavior observed while building and validating an AI-agent-driven kit.

### Frontend Stack Trace

```console
[PENPOT PLUGIN] Value not valid. Code: :error
```

### Backend Stack Trace

```console
No backend stack trace captured.
```

### Additional context

This blocks the natural agent flow: create primitives set, create semantic set, then add alias/reference tokens. The first alias fails without enough information for an agent or developer to self-correct.

Suggested fixes:
- Allow reference tokens whose target exists but is currently inactive.
- Or return an error such as: `reference "{color.gray.50}" cannot be resolved: token set "primitives" is inactive`.
- Document that new token sets are inactive and that reference resolution requires the referenced set to be active.

filipsajdak:

### Related Ticket

Fixes [#10070](https://github.com/penpot/penpot/issues/10070).

### Summary

The plugin `addToken` (`frontend/src/app/plugins/tokens.cljs`) resolved the new token against `ctob/get-tokens-in-active-sets`, then refused creation when `resolved-value` was nil. Creating a token into a set that exists but isn't in the active themes (or a reference to a token in that same set) therefore failed with a generic error — even though the product UI permits it: `sidebar.cljs` resolves with `get-tokens-in-active-sets-force` and the selected set id.

Fix: resolve against `ctob/get-tokens-in-active-sets-force tokens-lib id`, forcing the target set active during resolution, mirroring the UI.

Scope note: fully resolving references that point into *other* inactive sets ("allow unresolved-but-valid references") is a separate, feature-shaped change (deferred resolution) and is intentionally out of scope here.

From a plugin, with a set that is not enabled in the active theme:
```js
const set = penpot.library.local.tokens.addSet({ name: "core" }); // inactive by default
set.addToken({ type: "color", name: "color.gray.50", value: "#808080" });
// before: rejected with a generic "Value not valid"; after: token created
```

### Checklist

- [x] Update the `CHANGES.md` file, referencing the related GitHub issue.

### Notes

- Added tests in `frontend/test/frontend_tests/plugins/tokens_test.cljs` pinning the resolution premise (forced active-set tree surfaces the target set's tokens; plain active-set tree does not).
- `clj-kondo` and `cljfmt` clean on the changed files.
- This touches the same file as open PR #9564 (different defects); flagging for coordination to avoid merge conflicts.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,7 @@
 
 ### :bug: Bugs fixed
 
+- Fix plugin API `addToken` refusing to create a token in a set that exists but isn't active, by forcing the target set active during resolution like the UI does (by @filipsajdak) [Github #10070](https://github.com/penpot/penpot/issues/10070)
 - Fix plugin API `fileVersion.restore()` promise hanging indefinitely on restore failure [Github #9092](https://github.com/penpot/penpot/issues/9092)
 - Fix LDAP provider params schema typo (`bind-passwor` → `bind-password`) introduced during the `clojure.spec` → `malli` migration; the schema slot now matches the runtime key actually read by `prepare-params` (`:password (:bind-password cfg)`) and `try-connectivity` (`(:bind-password cfg)`), so a wrong type for the password no longer slips through unvalidated
 - Fix `login-with-ldap` silently dropping its error message on the `ldap-not-initialized` restriction (typo `:hide` → `:hint`); the message `"ldap auth provider is not initialized"` now actually surfaces in logs and error responses instead of being discarded into an unread key
--- a/frontend/src/app/plugins/tokens.cljs
+++ b/frontend/src/app/plugins/tokens.cljs
@@ -328,7 +328,11 @@
       :fn (fn [attrs]
             (let [tokens-lib (u/locate-tokens-lib file-id)
                   token (ctob/make-token attrs)
-                  tokens-tree (-> (ctob/get-tokens-in-active-sets tokens-lib)
+                  ;; Force the target set active during resolution, mirroring the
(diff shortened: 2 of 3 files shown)


Alotor:
Thanks for the contribution but we'll fix this internaly

PR #10293, opened 2026-06-18
:sparkles: Add sync with design tokens in external libraries

### LINKED ISSUE #9334: Add support for internal libraries and file sync for Design Tokens
## Is your feature request related to a problem? Please describe.

When I work in a file and link a library that contains tokens, I wan to be able to use the library tokens directly in the file, without the need of importing them. And when the tokens in the libary are modified, they must be synchronized into the file, like it's ocurring with the assets.

## Current state of tokens and libraries

Currently,  tokens live in a specific file and are not syncronized to any other file, even if this file is a library.

This means that, in case the user is attaching a library with tokens, they cannot use the tokens from the library as a real source (like components or styles).

The only data that is transferred **via components** is the **applied token name**. This information gets transferred to component copies since they are part of the shapes data. The token value is transferred as a regular raw value.

Tokens list, themes or sets are not transferred in any way to the file that is using the library.

The only way of consume the same tokens in both files is exporting from the file that cointains the json file and importing it into the other file. Since tokens are the same at first instance, everything looks normal, but if the user modifies the tokens in the library or the file, inconsistencies will start to appear. For example, if the user applies a different token to a component copy, or change the value of a token, it will be considered as an override. The user needs to update both files manually in order to keep consistency and manage unresolved tokens.

So in this given file, if the user imports a new JSON with different tokens, if there are a token from the same type and with the same name, the values will get overrided in the component copies.

#### Penpot's native syncronization (libraries and files)

* How tokens are synced through libraries
* What information is synced
* Consistent flows compatible with future external sync

Use case 1

"As a design team lead, I want to maintain our design tokens in a central library so that all product files stay consistent. Designers should be able to apply tokens but not edit them, ensuring our design system remains the single source of truth."

## Internal sync

By now, we will allow only one collection of tokens per file, no matter how many libraries with tokens are attached.

The user will be able to select wich tokens source they want to use at any file - local tokens, attached library tokens, external tokens (git sync). This selection can be modified at any time, but only one collection of tokens can be active. By default, the local tokens will be used.

This is a conservative approach to sync between libraries. It does not involve teams or projects, only files or libraries (and external sync, when ready). Just like components and styles.

If user want to use the same collection of tokens in different files, they could add those tokens to a shared library. External sync will cover parallel syncronizations without libraries involved.

**Not from this US:** in the future there will be mechanisms to extend existing token trees.

#### Basic feature

* A user create a File A and import/add tokens to it.
* The file is published as a Library A
* The user attaches Library A to a new File B.
* **If the File B does not have tokens/sets/themes:**
  * Library A tokens will be used as token tree source
  * There will be a visual indicator to inform the tokens are stored in another file.
* The user **won't be able to create / modify new sets or themes in the file, but can activate/deacivate them**.
  * To edit them, they will be redirected to the original library that stores them.
* The user will be able to **read-only and apply those tokens, but not to add / modify o delete them in this file**.
  * Import feature will be disabled (🔴 take into account the new import button in library modal).
* **If File B already has tokens**
  * Once attached, the user is asked wich tokens source they want to use: Library A tokens or File B tokens (local)
  * If File B tokens are selected:
    * The user can create, edit and delete tokens, sets and themes, since it is a local token tree
  * Library A tokens are not connected in any way to File B tokens (but can be imported manually from the library modal
  * This selection can be changed at any moment in the file settings, so the existing collection of tokens are not deleted, but inactive in the background.
  * Import option is enabled.

##### Multiple libraries with tokens

Like in files, when a library is attached to another library, the user should select the source of tokens they want to use.

#### Migration

**Migration:** What happens with existing files that has a library with tokens and has already a json imported (so two tokens json will “exists” in the same file)?

By default, if tokens exist locally in a file, these tokens will be set as the default tokens source. If the file has one or more libraries with tokens attached, a visual indicator can be displayed to notify the user that an action is requiered or recomended → to choose the tokens source.

#### foundations library + components library

As a team, we have a foundations library with tokens and then different libraries for different touchpoints: App, Web, IOs, etc.. so we can have specific Design Systems libraries with the same foundations.

How it would work with this basic approach:

* The user stores all the tokens in a unique file. In this approach one of the libraries will contain all the primitive and semantic tokens.
* Alternatively, the team would be able to connect each file to an external source through [github sync](https://tree.taiga.io/project/penpot/us/9968) - **(not from this us)**
* Tokens will be stored in the foundations files and would be used in other libraries and files that have this library attached.
* If both libraries and files have tokens - since we allow that right now -, the user would need to decide which is the one they want to use.
* In the future we can allow things like merge them in one single token setup or similar - **(not from this us)**

#### Other use cases & corner cases

* A file has tokens → The user attaches a library with a different set of tokens → they would need to specify which source they want to use.
* A library has tokens → The user attaches another library to it → they would need to specify which source they want to use.
* A file has an attached Library with tokens → the user imports a new set of tokens to the file → the user needs to decide wich token source they want to use.
* No matter if the tokens are the same, the user should need to decide with file is the source of truth.
* A library with tokens is detached → applied tokens will dislpay an unresolved status. if the library is attached again o tokens are imported, everything will work again.
* The user wants to modify a token → they should go to original source, just like in styles.
* A inport imports a set of tokens. Then, attaches a library.
* if the library has token → the user gets a modal to decide which one to use - not from this US
* if the library does not have tokens. The user goes to the library and import a set of tokens → a modal appears - not from this US

See [US #12015 in Taiga](https://tree.taiga.io/project/penpot/us/12015?milestone=513226) for more future use cases, not implemented in this issue.

## Component copies and overrides connections

What happens with components and component copies:

* In components: We shouldn't consider the value changes as an override if the tokens has the same name than the original component.
* if a user activates / deativates sets, the resolved value changes, but the applied token is the same, so this won't be considered an override
* If a user applies a different token to a shape in a component copy, this would be considered an override.

### Design:

https://design.penpot.dev/#/workspace?team-id=9cece348-eca8-815b-8005-3cd37c8a648f\&file-id=dc84c731-c73c-80c4-8007-be079a965e23\&page-id=27f61532-c195-8039-8006-fb7a2b3293e2

##### Prerequisites

* Two Penpot files in the same team:
  * A **Library** file with a Tokens set (a few sets/themes with distinct values, e.g. a color token and a spacing token).
  * A **Consumer** file with no tokens that has the Library linked as a shared library (Assets tab → Libraries).
     

### 1. Default tokens source

* Open the Consumer file and go to the Tokens tab.
* Verify the consumer file is set to Local file
* Add tokens to the file and verify they work as usual (create/apply)

###

### 2. Choose the tokens source for a file

* Go to tokens tab and open the library modal clicking on the button near tokens source
* Confirm that both "File Library" and "Library A" display the number of sets/themes and tokens they have
* Confirm that "File library" is set as Tokens source and "Library A" has a button to set as tokens source
* Switch the source to "Library A" and confirm the tokens panel updates accordingly.
* Confirm that the themes and sets state is transferred to the consumer file this first time.

###

### 3. Access library tokens without importing

* Open the Consumer file and go to the Tokens tab.
* Verify the tokens defined in the linked Library are visible/usable directly in the tokens panel, without any separate "import" step.
* Apply one of the library's tokens (e.g. a color token) to a shape in the Consumer file and confirm it renders correctly.

###

### 4. Library tokens are read-only in the consumer file

* In the Consumer file, confirm tokens coming from the external library cannot be edited or deleted from within the Consumer file (read-only), only from the source Library file itself.

#### 4.a Open connected library through tokens tab

* Confirm that you can access to "library A":
* There is a button near tokens source that allows you to open the library in a new tab
* Themes dropdown: open the **Themes** dropdown and confirm that clicking on "Edit themes" option opens the library in a new tab

###

### 5. Sync when the library token values change

* In the Library file, modify a token's value (e.g. change a color token's hex value) and save/publish the change.
* In the Consumer file, wait for the library update/sync modal and confirm and verify that:
  * The token's value updates.
  * Any shapes using that token visually update to reflect the new value.

###

### 6. Independent activation status of themes/sets

* In the Library file, activate/deactivate a specific theme or token set.
* Go to the Consumer file:
  * Verify no library update is triggered
  * Verify the active theme/set status can differ from the Library file (i.e., activation status is per-file, not shared).
  * Verify no changes are applied to canvas elements

###

### 7. Use current file tokens resolved value

* In the consumer file, change theme or set (so that resolved values differs from the source file)
* In the assets tab, drag a component that use tokens to the canvas
* Verify that the component in the canvas is using current file resolved values, not library resolved values
* go to library and make a change that triggers a library update but does not involve changing tokens (for example, change a component name)
* go back to consumer file and verify that:
  * the modal update appears and the changes are applied
  * no changes involving tokens are triggered

### 8. Sidebar / summary display

* Go to Assets tab and confirm there is an indicator of which one is the tokens source
* Go to library modal and confirm that the modal correctly show info about tokens/themes/sets count.

###

### 9. Regression checks

* Confirm existing local (non-library) token creation, editing, and deletion still work as before in a file with no linked libraries.
* Confirm normal asset sync (colors, components, etc.) still works unaffected by these token changes.
* Test with multiple linked libraries (if supported) to confirm tokens from each are distinguishable and don't collide unexpectedly.

###

### 10. Edge cases

* Unlink the library from the Consumer file and verify library-sourced tokens are removed/handled gracefully (no crashes, no orphaned references).
* Token name collisions between a local token and a library token — confirm defined/expected resolution behavior.

hirunatan:
### Related Ticket

Closes https://github.com/penpot/penpot/issues/9334

### Summary

This PR is working in two things:
 - Allowing to use tokens from an external library directly in a file.
 - Separate the activation status of themes and sets from the tokens library. This way there may be a different status in the library and the file.

I am taking advantage of the change to implement stricter principles in the code architecture and data validation:
- Strictly comply with the data abstraction levels in https://help.penpot.app/technical-guide/developer/abstraction-levels/
- This means that data handling is pushed to the lowe levels, and functions in common.changes, common.logic and data events are thin wrappers on functions in common.files and common.types.
- Always use getter functions to read properties of abstract data types.
- Always use semantic mutators to modify them (trying not to do simple setters, but functions with a business logic meaning).
- In UI components, instead of reading data in container components and passing them down via properties, have a TokensLib and a TokensStatus in the sidebar component (and the modal roots) and pass them via context to where they are needed. The lower level components can directly call them to get the data they need, and if caching is necessary it may be added to TokensLib or TokensStatus instances.
  * This means that any change to tokens library or tokens status repaints the entire sidebar or modal. But this is already happening anyway, since library changes almost always involve recalculating all tokens and redrawing trees or lists. And these types of changes are always initiated by the user; there are no automated processes that modify the library in bulk.

### Checklist


niwinz:
**IMPORTANT** We should revisit the backward compatibility issue before this pr can be merged

PR #10341, opened 2026-06-19
:bug: Raise validation error for incompatible binfile export options

### LINKED ISSUE #7649: Empty response when exporting bin using API
### Steps To Reproduce

Call POST https://design.penpot.app/api/rpc/command/export-binfile
body: {
    "fileId":"<UUID String here>",
    "includeLibraries": true,
    "embedAssets": true
}

### Expected behavior

I Expect the zip file, with the penpot contents, to be returned.

### Actual behavior

Returns an empty zip file.

### Screenshots or video

_No response_

_No response_

_No response_

HOST: https://design.penpot.app

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

For other combinations of includeLibraries and embedAssets the endpoint works fine. Only when both are true, the endpoint fails

filipsajdak:

### Related Ticket

Fixes [#7649](https://github.com/penpot/penpot/issues/7649).

### Summary

`export-binfile` (the `::export-binfile` RPC) calls into the v1/v3 binfile writers, which guard against requesting both `include-libraries` *and* `embed-assets`. That guard threw a raw `IllegalArgumentException`, so the RPC layer reported it to clients as a generic `:server-error` / `:unexpected` (HTTP 500) — the opaque failure described in #7649 (perceived as an empty/failed export when both flags were sent).

Those two options are mutually exclusive by design, i.e. it's invalid input, not a server fault. This PR:
- replaces the raw `throw (IllegalArgumentException. …)` in `backend/src/app/binfile/v3.clj` and `backend/src/app/binfile/v1.clj` with `ex/raise :type :validation :code :incompatible-options`, matching how the rest of the binfile code reports bad input;
- fixes the "mutally" → "mutually" typo in the message;
- adds a regression test.

Reproduced and verified end-to-end on a local devenv:

```
# both options -> before: {:type :server-error :code :unexpected} (HTTP 500)
#                 after:  {:type :validation    :code :incompatible-options}
POST /api/rpc/command/export-binfile
  {"file-id": "<id>", "include-libraries": true, "embed-assets": true}

# valid single-option exports are unaffected (still produce the file):
  {"file-id": "<id>", "include-libraries": false, "embed-assets": true}  -> ok
```

Note: the original "empty zip" symptom no longer occurs for valid parameter combinations on `develop`; the remaining defect was purely the error classification for the mutually-exclusive case.

### Checklist

### Notes

- Regression test `export-binfile-v3-rejects-incompatible-options` added to `backend/test/backend_tests/binfile_test.clj`; run locally via kaocha against the test DB — `1 tests, 5 assertions, 0 failures`.
- Per `CONTRIBUTING.md`, the changelog is generated at release time, so this PR does not modify `CHANGES.md`.

### DIFF
--- a/backend/src/app/binfile/v1.clj
+++ b/backend/src/app/binfile/v1.clj
@@ -305,8 +305,9 @@
 (defn write-export!
   [{:keys [::bfc/include-libraries ::bfc/embed-assets] :as cfg}]
   (when (and include-libraries embed-assets)
-    (throw (IllegalArgumentException.
-            "the `include-libraries` and `embed-assets` are mutally excluding options")))
+    (ex/raise :type :validation
+              :code :incompatible-options
+              :hint "the `include-libraries` and `embed-assets` are mutually exclusive options"))
 
   (write-export cfg))
 
--- a/backend/src/app/binfile/v3.clj
+++ b/backend/src/app/binfile/v3.clj
@@ -221,8 +221,9 @@
   [{:keys [::bfc/embed-assets ::bfc/include-libraries] :as cfg} file-id]
 
   (when (and include-libraries embed-assets)
-    (throw (IllegalArgumentException.
-            "the `include-libraries` and `embed-assets` are mutally excluding options")))
+    (ex/raise :type :validation
+              :code :incompatible-options
+              :hint "the `include-libraries` and `embed-assets` are mutually exclusive options"))
 
   (let [detach? (and (not embed-assets) (not include-libraries))]
     (db/tx-run! cfg (fn [cfg]
--- a/backend/test/backend_tests/binfile_test.clj
+++ b/backend/test/backend_tests/binfile_test.clj
@@ -207,6 +207,27 @@
       (t/is (= (count result) 1))
       (t/is (every? uuid? result)))))
 
+(t/deftest export-binfile-v3-rejects-incompatible-options
+  ;; Regression for #7649: requesting both `include-libraries` and
+  ;; `embed-assets` must fail with a proper `:validation` error (clear code)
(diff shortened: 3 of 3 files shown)


filipsajdak:
Marking this ready for review.

`export-binfile` called with both `include-libraries` and `embed-assets` threw a raw `IllegalArgumentException`, which reached the client as `{:type :server-error :code :unexpected}` (HTTP 500) rather than a validation error. This raises `:validation` / `:incompatible-options` instead - in `binfile/v3.clj` (`get-file`) and `binfile/v1.clj` (`write-export!`) - and corrects a "mutally" -> "mutually" typo in the existing message.

Verified on a local devenv: the RPC now returns a `:validation` error for the incompatible combination, and valid single-option exports are unaffected. Regression test `export-binfile-v3-rejects-incompatible-options` added in `binfile_test.clj` and passing.

Relates to #7649.

filipsajdak:
Retargeted this from `develop` to `staging`, and rebased accordingly (single commit, same change).

@luisddm mentioned on #11168 that you are in the feature freeze for 2.18 and asked for that fix to go to `staging`, so I assumed the same applies to this one since it is also a bug fix. Happy to move it back to `develop` if you would rather it waited for the next cycle — just say so.

filipsajdak:
Small status question on this one: it is still mergeable and merges clean against `staging`, with the regression test passing.

It has no milestone or labels, unlike its siblings, so I suspect it slipped through triage rather than being deliberately parked. Would you like it milestoned for 2.18 where it currently targets `staging`, or should I retarget it to `develop` for 2.19? Happy to move it either way — just let me know which you prefer.

niwinz:
It is ok for 2.18.1, i will merge it as CI pases.

Thanks :heart:

PR #10372, opened 2026-06-22
:bug: Accept fractional gap and padding in plugin API layout setters

filipsajdak:
> **Note:** This PR was created with AI assistance as part of the Penpot MCP self-improvement initiative.

### Summary

The flex/grid layout proxies validate `rowGap`, `columnGap`, and the four padding setters with `sm/valid-safe-int?`, so **fractional values are rejected** — e.g. `flex.rowGap = 10.5` fails with "Value not valid", even though:

- the data model types these as `::sm/safe-number` — `common/src/app/common/types/shape/layout.cljc`: `:row-gap`/`:column-gap` and `:p1`-`:p4` are all `safe-number`; and
- the sidebar gap/padding inputs accept decimals.

So the plugin API is stricter than the model — the same class of defect as the merged #9780 (negative `letterSpacing`).

Fix: switch the 16 gap/padding guards (8 in `flex.cljs`, 8 in `grid.cljs`) to `sm/valid-safe-number?` — matching the model and the predicate **already used by the flex-element setters in the same file** (so this is making validation consistent, not loosening it). Integer-only setters are intentionally left as `valid-safe-int?`: `zIndex`, and the grid track indices/counts and cell positions/spans. Also fixes a `:righPadding` typo in two grid `rightPadding` error branches.

### Steps to reproduce / verify

Verified end-to-end on a local devenv via the plugin API:

```js
const b = penpot.createBoard(); const flex = b.addFlexLayout();
flex.rowGap = 10;    // applied
flex.rowGap = 10.5;  // before: rejected (stays 10);  after: 10.5
flex.verticalPadding = 4.5;  // before: rejected;  after: 4.5
// same for grid via b.addGridLayout()
```

### Checklist

- [x] Use `develop` as the target branch.
- [x] Brief summary of the changes.
- [x] How to reproduce / verify.
- [x] Add or modify integration tests.

### Notes

- Regression test added to `frontend/test/frontend_tests/plugins/page_active_validation_test.cljs` (reuses that file's flex/grid proxy harness): with `throwValidationErrors` on, fractional gap/padding values must be accepted for both flex and grid.
- `clj-kondo` (0 errors) and `cljfmt` clean on the changed files.
- Per `CONTRIBUTING.md` the changelog is release-generated, so `CHANGES.md` is not modified.
- No existing tracking issue; filing the fix directly per the bug-fix path in `CONTRIBUTING.md`.

Issue #9780, opened 2026-05-20
Plugin API rejects negative `letterSpacing` values


girafic:
**Describe the bug**

Setting a negative `letterSpacing` on a text shape (or text range) from a plugin
is rejected with:

```
[PENPOT PLUGIN] Value not valid: -0.56. Code: :letterSpacing
```

Negative letter spacing (tighter tracking) is valid in Penpot — the product's own
typography input accepts `-200` to `200` (`:min -200 :max 200 :step 0.1` in
`frontend/src/app/main/ui/workspace/sidebar/options/menus/typography.cljs`). The
Plugin API should accept the same range.

There are two underlying causes, both in `frontend/src/app/plugins/text.cljs`:

1. **The validation regex excludes negatives.** `letter-spacing-re`
   (`frontend/src/app/plugins/text.cljs`) is `#"^\d*\.?\d*$"`, which has no
   provision for a leading minus sign, so `"-0.56"` fails `re-matches` and the
   `letterSpacing` setter reports `not-valid`.

2. **The `letterSpacing` text-range setter has an inverted condition.** It uses
   `(or (empty? value) (re-matches letter-spacing-re value))` to decide
   *invalid*, i.e. it rejects values that **match** the regex and lets
   non-matching input through — the opposite of intent. The sibling `lineHeight`
   range setter correctly uses `(not (re-matches ...))`.

**To Reproduce**

1. Open a Penpot file and select a text shape.
2. From a plugin (with `content:write` permission), run:
   ```js
   text.letterSpacing = "-0.56";
   ```
3. See the error `Value not valid: -0.56. Code: :letterSpacing`. The value is not
   applied.

**Expected behavior**

`text.letterSpacing = "-0.56"` applies the negative letter spacing to the text,
matching the range allowed by the Penpot UI (`-200` to `200`). Reading
`text.letterSpacing` back returns `-0.56`. Positive and zero values keep working,
and non-numeric input is still rejected.

**Screen recordings and screenshots**

N/A (reproducible from the plugin console).

**Desktop (please complete the following information):**
 - OS: macOS (Darwin 25.5.0)
 - Browser: any
 - Version: any

**Additional context**

- Penpot version: `2.15.4-RC1-1261-g565ed042bc` (commit `565ed042bc`)
- Affected file: `frontend/src/app/plugins/text.cljs`
  - `letter-spacing-re` definition
  - `letterSpacing` text-range setter
[...]

PR #10373, opened 2026-06-22
:bug: Accept fractional border radius in plugin API shape setters

filipsajdak:

### Summary

The ShapeProxy `borderRadius` setters (`borderRadius` plus the four per-corner variants) validate with `sm/valid-safe-int?`, so **a fractional radius is rejected** — e.g. `shape.borderRadius = 7.5` fails with "Value not valid", even though:

- the model types `:r1`-`:r4` as `::sm/safe-number` (`common/src/app/common/types/shape.cljc`);
- the radius sidebar input only constrains `:min 0` (it is not integer-only); and
- `ctsr/set-radius-*` store the value verbatim.

So the plugin API is stricter than the model — the same class of defect as the merged #9780 (negative `letterSpacing`) and the sibling layout-gap fix.

Fix: switch the 5 guards to `sm/valid-safe-number?`, keeping the existing non-negative guard on the all-corners `borderRadius` setter. Integer-only setters in the file (`getRange` bounds, `setParentIndex`) are intentionally left as `valid-safe-int?`.

Verified end-to-end on a local devenv via the plugin API:

```js
r.borderRadius = 7;    // applied
r.borderRadius = 7.5;  // before: rejected;  after: 7.5
r.borderRadiusTopLeft = 2.5;  // before: rejected;  after: 2.5
```

### Checklist

- [x] Use `develop` as the target branch.
- [x] Brief summary of the changes.
- [x] How to reproduce / verify.
- [x] Add or modify integration tests.

### Notes

- Regression test added to `frontend/test/frontend_tests/plugins/page_active_validation_test.cljs` (reuses its proxy harness): with `throwValidationErrors` on, fractional radius values must be accepted.
- Per `CONTRIBUTING.md` the changelog is release-generated, so `CHANGES.md` is not modified.
- No existing tracking issue; filing directly per the bug-fix path in `CONTRIBUTING.md`. Companion to the layout gap/padding fix (#10372).


madalenapmelo-kp:

Alotor:
Thanks for your contribution!

PR #10374, opened 2026-06-22
:bug: Accept delay 0 in plugin API interaction setter

filipsajdak:

### Summary

The `InteractionProxy.delay` setter (`frontend/src/app/plugins/shape.cljs`) rejected `0`:

```clojure
(or (not (number? value)) (not (pos? value)))  ; rejects delay = 0
```

But the model allows it — `set-delay` (`common/src/app/common/types/shape/interactions.cljc`) only asserts `(sm/check-safe-int delay)`, with no positivity requirement, so `0` (an immediate after-delay interaction) is valid.

Fix: `(or (not (sm/valid-safe-int? value)) (neg? value))` — allows 0 and positive integers, still rejects negatives, and additionally rejects non-integers up front (the old `number?` check let a fractional delay through, which then tripped the model's `check-safe-int` assertion downstream).

This is the same "plugin stricter than the model" family as the merged #9780, the layout-gap fix (#10372), and the border-radius fix (#10373).

```js
const i = board.addInteraction("after-delay", { type: "navigate-to", destination: other });
i.delay = 0;   // before: rejected ("Value not valid");  after: accepted
```

### Checklist

- [x] Use `develop` as the target branch.
- [x] Brief summary of the changes.
- [x] How to reproduce / verify.
- [x] Add or modify integration tests.

### Notes

- Regression test added to `frontend/test/frontend_tests/plugins/interactions_test.cljs` (sets a positive delay, then 0, asserting it takes effect).
- Per `CONTRIBUTING.md` the changelog is release-generated, so `CHANGES.md` is not modified.
- No existing tracking issue; filing directly per the bug-fix path in `CONTRIBUTING.md`.


madalenapmelo-kp:

Alotor:
Thanks for your contribution!

PR #10376, opened 2026-06-23
fix(rpc): handle nil profile-id in get-profile, prevent exception swallowing

### LINKED ISSUE #9276: bug: `get-profile` RPC swallows all exceptions despite the documented contract

singlaamitesh:
## Summary

Fixes #9276. The `::get-profile` RPC method should only return anonymous profile for two cases:
1. No `profile-id` in session
2. DB call raises `:not-found`

Previously, when `profile-id` was nil, the code attempted a DB call that could throw unrelated exceptions. While the catch block technically only returned anonymous for `:not-found` exceptions, the nil profile-id case was not handled explicitly.

## Changes

- Added explicit nil check for `profile-id` before DB call
- Returns anonymous profile immediately when no session profile-id
- Catch block unchanged — only `:not-found` exceptions return anonymous

## Testing

```bash
clj -M:test app.rpc.commands.profile
```


madalenapmelo-kp:

niwinz:
Duplicate of #9254 

Thanks for the contribution in any case :heart:

PR #10392, opened 2026-06-23
:sparkles: Improve MCP handling when tab is frozen in the browser

### LINKED ISSUE #10323: MCP integration hangs when the Penpot tab is backgrounded/frozen by the browser
## Steps To Reproduce

1. Connect the Penpot MCP integration (plugin connected to the MCP server, e.g. driving Penpot from an external AI client/IDE).
2. Switch away from the Penpot browser tab and keep working in the AI client for several minutes, so the Penpot tab stays backgrounded (and/or the browser's Memory Saver / Energy Saver kicks in — more aggressive on battery).
3. From the AI client, issue an MCP tool call (e.g. execute code) while the Penpot tab is still in the background.

## Expected behavior

The MCP tool call either succeeds, or fails quickly with a clear, actionable error explaining that the Penpot tab is suspended and needs to be re-focused. The connection should recover automatically once the tab is active again.

## Actual behavior

The MCP call appears to "hang" and only fails after the full 30s task timeout with a generic error. Root cause: the MCP plugin runs entirely inside the Penpot browser tab and holds a WebSocket to the MCP server. When the browser throttles/freezes/discards the backgrounded tab, the page's event loop is paused, so the plugin cannot process the task — yet the connection still looks healthy to the server because protocol-level ping/pong frames are answered by the browser's network stack, not by the (frozen) page JavaScript. There was no auto-reconnect, no Page Lifecycle awareness, and no application-level liveness signal, so a frozen/discarded tab produced silent hangs.

## Screenshots or video

N/A

## Desktop (please complete the following information)

- OS: any
- Browser & version: Chromium-based browsers with Memory Saver / Energy Saver / tab freezing (e.g. Chrome, Edge); most pronounced when the device is on battery

## Smartphone (please complete the following information)

N/A

## Environment (please complete the following information)

- Host: any (design.penpot.app or self-hosted) running the MCP integration
- Affected component: `mcp/packages/plugin` (in-tab plugin) and `mcp/packages/server` (`PluginBridge`)

## Frontend Stack Trace

```console
(none — fails silently; the plugin's event loop is paused so no error is thrown client-side)
```

## Backend Stack Trace

```console
Error: Task <id> timed out after 30 seconds
    at PluginBridge.sendPluginTask (mcp/packages/server/src/PluginBridge.ts)
```

## Additional context

The MCP plugin cannot run JavaScript *during* an active browser freeze — that's a browser-imposed constraint, not something the app can override from inside the tab. The fix therefore focuses on resilience and fast, clear failure rather than preventing freezing outright:

- **Plugin** (`main.ts`): application-level heartbeat (every 10s, emitted from the page event loop so its absence reveals a suspended tab); auto-reconnect with exponential backoff; Page Lifecycle awareness (`resume` / `visibilitychange` to recover immediately on wake, and a `freeze` event handler that notifies the server right before the tab sleeps).
- **Server** (`PluginBridge.ts`): per-connection liveness tracking (`lastHeartbeat`) and a `frozen` flag set by the plugin's freeze notice; task dispatch now fails fast with an actionable message ("the Penpot tab has been frozen/suspended — click/focus it and retry") instead of hanging for the full timeout.

Users can also mitigate today by excluding the Penpot origin from the browser's Memory Saver ("Always keep these sites active") and/or keeping the machine on AC power. A fully freeze-proof path would require running Penpot in a desktop/Electron wrapper with `backgroundThrottling: false`.

Alotor:
### Related Ticket

Closes #10323

### Summary

Creates a new disconnection handler in the plugin interface and adds a heartbeat message. This is done so we can check that the connection still alive, the ping/pong were not enough because when the browser freezes a tab the WS connection is still working. 

Also I've added a handler of chrome specific events freeze/resume. This improves the Chrome use case. Firefox don't support this API that's why we have the heartbeat as an alternative.

### Steps to reproduce 

To force Chrome the browser to "freeze" the tab you can access chrome://discards/ where you can force a freeze.

In Firefox the best way is to open the debugger and pause the browser.

### Checklist

### DIFF
--- a/.github/workflows/tests-mcp.yml
+++ b/.github/workflows/tests-mcp.yml
@@ -49,4 +49,4 @@ jobs:
       - name: Tests
         working-directory: ./mcp
         run: |
-          pnpm -r run test;
+          pnpm run test;
--- a/docs/mcp/index.md
+++ b/docs/mcp/index.md
@@ -147,6 +147,27 @@ If you just want to try Penpot AI workflows quickly through the MCP, follow this
 
 Once all five steps are done, your AI client should list Penpot tools.
 
+### Keep the Penpot tab active
+
+The MCP plugin runs inside your Penpot browser tab. If the browser puts that tab to sleep, freezes it, or unloads it to save memory, the MCP server cannot run tasks in Penpot until the tab wakes up again.
+
+When this happens, MCP fails fast instead of waiting for a long task timeout:
+
+* In Chrome and Chromium-based browsers, the plugin can report when the tab is being frozen.
+* In Firefox, Safari, and other browsers that do not expose the same freeze event, MCP uses plugin heartbeats. If the browser stops running the plugin JavaScript, the heartbeat becomes stale and the MCP server reports that the Penpot tab appears to be suspended.
+* If the browser unloads the tab completely, the plugin disconnects and MCP reports that no Penpot plugin instance is connected.
+
+To recover, open or focus the Penpot tab again, wait until MCP reconnects, and retry the prompt.
+
+To reduce the chances of the browser putting Penpot to sleep during long MCP sessions:
+
+| Browser | Recommended setting |
+| --- | --- |
+| Chrome | Go to **Settings → Performance → Always keep these sites active** and add your Penpot site. Pinning the Penpot tab also helps prevent Chrome tab deactivation. |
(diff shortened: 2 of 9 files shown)


niwinz (approved):

opcode81 (approved):
Looks good to me.

PR #10395, opened 2026-06-24
fix: enable left sidebar resizing with hover affordance, reduce right sidebar default width for premium figma feel.

parinith-web:
:bug: Fix left sidebar resize handle and reduce right sidebar width

### Changes

Fixes left sidebar resizing (previously non-functional), adds a hover-visible highlight on the resize handle so users can clearly see where to grab it, and reduces the right sidebar's default width slightly for a more efficient use of screen space and to give a premium vibe like figma.

### Issue

- The left sidebar's resize handle had a CSS bug (zero-width grid column) that made its hit-area unreachable — nothing to hover or click.

- The left sidebar's `width` CSS property was bound to the wrong CSS variable, so even a working drag wouldn't visually resize it.

- Once resizing worked, the handle blended into the canvas/ruler with no visual distinction, making it hard to find without trial and error.

- The right sidebar's default width (318px) was wider than necessary for its content, taking up more canvas space than needed.

### Result

- The left sidebar now resizes smoothly by dragging its edge, within a configured min/max range.

- A thin, theme-colored highlight line appears on hover over the resize handle, making the draggable edge clearly
 distinguishable from the canvas - invisible at rest so it doesn't add visual clutter or compete with the canvas for attention.

- Left sidebar's min/max width range is synced between `constants.cljs` and `sidebar/common/sidebar.scss`, per the existing comment requiring these to stay in sync.

- Right sidebar's default width reduced from 318px to 300px, giving the canvas a bit more room while staying within the minimum width its internal grid layout requires.

- Right sidebar resizing behavior is unchanged (resizable only in Inspect → Code mode, as before).

### Testing

Tested locally via the devenv:

- Dragged the left sidebar across its full min/max range.
- Confirmed hover highlight appears only on the resize handle, in both expanded and default sidebar states.
- Confirmed right sidebar still renders correctly at its new default width with no layout overflow or misalignment.
- Confirmed right sidebar's existing resize behavior in Inspect → Code mode is unaffected.

### Screen Recordings

1. before the fix (handle unresponsive, no resize)
https://github.com/user-attachments/assets/631e9baa-4e90-4acf-bf99-080c43481b3a

3. after the fix (resize working, hover highlight visible)
https://github.com/user-attachments/assets/6e6f9793-143e-4356-8df8-ba1264dfa009

Signed-off-by: 
@parinith-web  <parinithreddymav@gmail.com>

### DIFF
--- a/frontend/src/app/main/constants.cljs
+++ b/frontend/src/app/main/constants.cljs
@@ -18,11 +18,11 @@
 ;; Before changing these values also check:
 ;; frontend/src/app/main/ui/workspace/sidebar/common/sidebar.scss
 
-(def right-sidebar-default-width 318)
-(def right-sidebar-default-max-width 768)
+(def right-sidebar-default-width 300)
+(def right-sidebar-default-max-width 400)
 
-(def left-sidebar-default-width 318)
-(def left-sidebar-default-max-width 500)
+(def left-sidebar-default-width 277)
+(def left-sidebar-default-max-width 350)
 
 (def page-metadata
   "Default data for page metadata."
--- a/frontend/src/app/main/ui/workspace/sidebar.scss
+++ b/frontend/src/app/main/ui/workspace/sidebar.scss
@@ -9,12 +9,12 @@
 
 .left-settings-bar {
   display: grid;
-  grid-template: "header header" deprecated.$s-52 "content resize" 1fr / 1fr 0;
+  grid-template: "header header" deprecated.$s-52 "content resize" 1fr / 1fr deprecated.$s-8;
   position: relative;
   grid-area: left-sidebar;
   min-width: var(--left-sidebar-width);
   max-width: var(--left-sidebar-width-max);
-  width: var(--right-sidebar-width);
+  width: var(--left-sidebar-width);
   background-color: var(--panel-background-color);
   height: 100vh;
   max-height: 100vh;
@@ -47,6 +47,23 @@
   width: deprecated.$s-8;
   cursor: ew-resize;
   height: 100%;
+
+  &::after {
+    content: "";
+    position: absolute;
+    top: 0;
+    left: 50%;
+    transform: translateX(-50%);
+    width: 2px;
+    height: 100%;
+    background-color: var(--color-accent-primary);
+    opacity: 0;
+    transition: opacity 0.1s ease-in-out;
+  }
+
+  &:hover::after {
+    opacity: 1;
+  }
 }
 
 .tab-spacing {
--- a/frontend/src/app/main/ui/workspace/sidebar/common/sidebar.scss
+++ b/frontend/src/app/main/ui/workspace/sidebar/common/sidebar.scss
@@ -20,12 +20,12 @@
 // ============================================================
 //
 // Left sidebar
-$left-sidebar-width: px2rem(318);
-$left-sidebar-width-max: $sz-500;
+$left-sidebar-width: px2rem(277);
+$left-sidebar-width-max: px2rem(350);
 
 // Right sidebar
-$right-sidebar-width: px2rem(318);
-$right-sidebar-width-max: px2rem(768);
+$right-sidebar-width: px2rem(300);
+$right-sidebar-width-max: px2rem(400);
 
 // ============================================================
 // GRID SYSTEM (used inside right sidebar)


madalenapmelo-kp:

frontend/src/app/main/ui/workspace/sidebar/common/sidebar.scss
```diff
@@ -20,12 +20,12 @@
 // ============================================================
 //
 // Left sidebar
-$left-sidebar-width: px2rem(318);
-$left-sidebar-width-max: $sz-500;
+$left-sidebar-width: px2rem(277);
```
EvaMarco, line 23: Thanks for working on this!

We won't be able to merge the sidebar width changes. Reducing the width causes horizontal scrollbars to appear in several situations, and avoiding horizontal scrolling is an important design constraint for the sidebar.

We'd like to keep the current dimensions unchanged.

Image

frontend/src/app/main/ui/workspace/sidebar/common/sidebar.scss
```diff
@@ -20,12 +20,12 @@
...
+$left-sidebar-width-max: px2rem(350);
 
 // Right sidebar
-$right-sidebar-width: px2rem(318);
-$right-sidebar-width-max: px2rem(768);
+$right-sidebar-width: px2rem(300);
+$right-sidebar-width-max: px2rem(400);
```
EvaMarco, line 28: We'd also like to keep the 768 value here.

This size is intentional, as it provides enough room to comfortably inspect the generated code. Reducing it to 400 significantly impacts that workflow.

frontend/src/app/main/ui/workspace/sidebar.scss
```diff
@@ -9,12 +9,12 @@
...
 .left-settings-bar {
   display: grid;
-  grid-template: "header header" deprecated.$s-52 "content resize" 1fr / 1fr 0;
+  grid-template: "header header" deprecated.$s-52 "content resize" 1fr / 1fr deprecated.$s-8;
   position: relative;
   grid-area: left-sidebar;
   min-width: var(--left-sidebar-width);
   max-width: var(--left-sidebar-width-max);
-  width: var(--right-sidebar-width);
+  width: var(--left-sidebar-width);
```
EvaMarco, line 17: Nice catch! This is a valid bug fix. Thanks for spotting and correcting it.

frontend/src/app/main/ui/workspace/sidebar.scss
```diff
@@ -47,6 +47,23 @@
   width: deprecated.$s-8;
   cursor: ew-resize;
   height: 100%;
+
+  &::after {
```
EvaMarco, line 51: We really like this enhancement. Using a pseudo-element to increase the resize hit area makes the interaction much nicer without affecting the visual appearance.

The only change we'd request is to use CSS logical properties instead of the physical ones (top, left, width, height). We try to consistently prefer logical properties throughout the codebase to better support different writing modes and keep the styling consistent.

Could you update this to use logical properties instead?

frontend/src/app/main/constants.cljs
```diff
@@ -18,11 +18,11 @@
 ;; Before changing these values also check:
 ;; frontend/src/app/main/ui/workspace/sidebar/common/sidebar.scss
 
-(def right-sidebar-default-width 318)
-(def right-sidebar-default-max-width 768)
+(def right-sidebar-default-width 300)
```
EvaMarco, line 21: We'd like to keep the current dimensions unchanged.

frontend/src/app/main/ui/workspace/sidebar/common/sidebar.scss
```diff
@@ -20,12 +20,12 @@
 // ============================================================
 //
 // Left sidebar
-$left-sidebar-width: px2rem(318);
-$left-sidebar-width-max: $sz-500;
+$left-sidebar-width: px2rem(277);
+$left-sidebar-width-max: px2rem(350);
```
EvaMarco, line 24: We'd like to keep the current dimensions unchanged.

EvaMarco (changes requested):

niwinz:
closing for inactivity

PR #10410, opened 2026-06-24
:sparkles: Plugin API new test suite

Alotor:
### Related Ticket

Depends on https://github.com/penpot/penpot/pull/10412 to fix the tests. Otherwise the CI of this MR will fail.

### Summary

This MR includes a new Plugin API test suite together with almos 37 bugs that have been discovered through this suite.

### Steps to reproduce 

- Start the new plugin-api-test-suite
- Run the tests

You can also run the suite against hourly but there are multiple errors and some of those crash the application.

### Checklist

- [x] Choose the correct target branch; use `develop` by default.
- [x] Provide a brief summary of the changes introduced.
- [x] Add a detailed explanation of how to reproduce the issue and/or verify the fix, if applicable.
- [x] Include screenshots or videos, if applicable.
- [x] Add or modify existing integration tests in case of bugs or new features, if applicable.
- [x] Refactor any modified SCSS files following the refactor guide.
- [x] Check CI passes successfully.

PR #10412, opened 2026-06-24
:bug: Fix problems with plugins API

### LINKED ISSUE #10394: Plugin API has bugs across metadata, shapes, text, comments, tokens, and interactions with several missing methods
The following issues have been found while developing the new test battery for the plugins API. I report all of them together because they will be fixed in the same PR.

The same PR will have the way of testing them through a new test battery that runs as a plugin.

## Metadata, files, and platform APIs

- Problem with the `shapesColors` metadata key where returned color metadata does not match the `ColorShapeInfo` shape expected by plugin consumers.
- Problem with `FileVersion.createdAt` where the stored JavaScript `Date` value is handled like a Luxon value.
- Problem with library color, typography, and component proxies where `libraryId` is not available through the public Plugin API.
- Problem with `localStorage.removeItem` where it calls `getItem` instead of removing the item.

## Comments

- Problem with `CommentThread.remove` where owners cannot remove their own comment threads.
- Problem with `page.removeCommentThread` where an unnecessary Promise wrapper causes it to throw.

## Text, fonts, and typography

- Problem with plugin font and typography application for both whole text shapes and selected text ranges.
- Problem with `typography.applyToTextRange` where it reads range bounds from unavailable public fields instead of the internal range data.
- Problem with `TextRange.shape` where it returns an internal/raw shape value instead of a shape proxy.
- Problem with font-family token handling where the Plugin API does not accept a string `fontFamilies` token value.
- Problem with typography token `resolvedValue` where it does not return `TokenTypographyValue[]`.
- Problem with shadow token `resolvedValue` where it does not return `TokenShadowValue[]`.
- Problem with font-family token `resolvedValue` where it does not return `string[]`.

## Shapes, geometry, and component operations

- Problem with `utils.geometry.center` argument handling where the public utility call does not match the internal helper signature.
- Problem with `backgroundBlur` where the proxy key casing is wrong on shape proxies.
- Problem with boolean shape type reporting where Plugin API consumers do not receive `"boolean"`.
- Problem with `flatten` where it does not return the resulting paths created by the operation.
- Problem with z-order methods where they act on the current selection or wrong shape context instead of the target shape.
- Problem with `is-variant-container?` where it does not always return a boolean.
- Problem with `combineAsVariants` where it does not use the component ids passed by the plugin caller.
- Problem with `group()` and `centerShapes()` where empty input produces invalid work instead of returning `null`.

## Interactions and prototyping

- Problem with overlay interactions where non-manual overlay types can lack a default position.
- Problem with plugin interaction setters where they pass only a shape id instead of a usable shape object.

## Layout grids and guides

- Problem with `addColumnAtIndex` where valid grid track types are not accepted.
- Problem with `removeRulerGuide` where it does not use the guide argument supplied by the plugin caller.
- Problem with board guide setter schema and parsing for Plugin API guide updates.
- Problem with empty grid-track syncing where it can trigger a zero-byte WASM allocation.
- Problem with grid track indices exposed through the Plugin API where invalid indices are not validated.

## Tokens and library assets

- Problem with `TokenSet.duplicate` where it does not return the duplicated token set.
- Problem with `theme.addSet` and `theme.removeSet` where they use a keyword lookup instead of reading the set name correctly.
- Problem with `LibraryColor` gradient/image setters where mutually exclusive color representations are not cleared when switching representation.

## Type declarations

- Problem with Plugin API types where readonly markers are missing in applicable places and the old `Image` type is not deprecated.

## Other API issues

- Missing `ShapeBase.swapComponent` from the Plugin API.
- Missing `File.revn` from the Plugin API.
- Missing `LibraryTypography.setFont` from the Plugin API.
- Missing `Group.isMask` from the Plugin API.
- Missing plugin API test-suite plugin and CI workflow coverage for the API surface.

Alotor:
### Related Ticket

Closes #10394

### Summary

Several bugfixes found when developing the new Plugins API test suite.

### Steps to reproduce 

The new API suite provides an easy interface to test API related issues:

https://github.com/penpot/penpot/pull/10410

### Checklist

### DIFF
--- a/frontend/src/app/plugins/tokens.cljs
+++ b/frontend/src/app/plugins/tokens.cljs
@@ -110,6 +110,34 @@
               "textCase"       (:text-case m)
               "textDecoration" (:text-decoration m)}]))
 
+(defn- shadow-key->camel
+  "Renames a shadow composite field name (kebab string) to its public camelCase
+   member name. The shadow schema is closed; offset-x/offset-y are its only
+   multi-word fields, so the rest (blur, spread, color, inset) pass through."
+  [k]
+  (case k
+    "offset-x" "offsetX"
+    "offset-y" "offsetY"
+    k))
+
+(defn- shadow-entry->js
+  "Converts one resolved shadow entry (a JS Map of field name -> tokenscript
+   symbol) into a plain JS object using the public member names and the
+   unit-converted values."
+  [^js m]
+  (let [out #js {}]
+    (.forEach m (fn [sym k]
+                  (obj/set! out (shadow-key->camel k)
+                            (ts/tokenscript-symbols->penpot-unit sym))))
+    out))
+
+(defn- shadow-resolved-value->js
+  "Converts a resolved shadow composite (a sequence of shadow entries) into the
+   plugin's `TokenShadowValue[]` shape."
+  [entries]
+  (when (some? entries)
+    (into-array (map shadow-entry->js entries))))
+
 (defn- get-resolved-value
   [token tokens-tree]
   (let [resolved-tokens (ts/resolve-tokens tokens-tree)
@@ -124,6 +152,13 @@
       ;; A typography token resolves to a composite; expose it as the documented
       ;; `TokenTypographyValue[]` rather than the raw tokenscript structure.
       (typography-resolved-value->js
+       (ts/tokenscript-symbols->penpot-unit resolved-value))
+
+      (= :shadow (:type token))
+      ;; A shadow token resolves to a list of composites whose entries the
+      ;; tokenscript unit conversion leaves as raw symbols; expose them as the
+      ;; documented `TokenShadowValue[]`.
+      (shadow-resolved-value->js
        (ts/tokenscript-symbols->penpot-unit resolved-value))
 
       :else
--- a/frontend/src/app/plugins/tokens.cljs
+++ b/frontend/src/app/plugins/tokens.cljs
@@ -463,7 +463,7 @@
            ;; Guard against nil to prevent `enable-set` from conj'ing nil
            ;; into the theme's :sets — which would send `:sets #{nil}` to the
            ;; backend and crash the workspace.
-           (let [set-name (obj/get token-set :name)
+           (let [set-name (obj/get token-set "name")
                  theme    (u/locate-token-theme file-id id)]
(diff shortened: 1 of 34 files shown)


frontend/src/app/plugins/tokens.cljs
```diff
@@ -96,13 +96,82 @@
...
+              "textDecoration" (:text-decoration m)}]))
+
+(defn- structured-key->camel
+  "Renames a tokenscript composite field name (kebab string) to its public
+   camelCase member name; single-word fields pass through unchanged."
+  [k]
+  (case k
+    "offset-x" "offsetX"
+    "offset-y" "offsetY"
+    k))
```
niwinz, line 121: Is this correct? I mean, in general the convention should be use camelCase so using this only for offset- misses all other keys or i'm wrong?

frontend/test/frontend_tests/plugins/shape_bugfixes_test.cljs
```diff
@@ -0,0 +1,201 @@
...
+                                                                :hidden false}})]
+      (let [blur (.-backgroundBlur proxy)]
+        (t/is (= (str blur-id) (aget blur "id")))
+        (t/is (= 12 (aget blur "value")))))))
+
+(t/deftest flatten-returns-proxies-for-converted-shapes
+  ;; `convert-selected-to-path` runs the WASM boolean/path pipeline, so this
+  ;; test stays at the proxy boundary: it verifies `flatten` forwards the
+  ;; selected ids to the conversion and wraps the result back into proxies.
+  (let [file-id  (random-uuid)
```
niwinz: use (uuid/next), will help on debuging on gnerating sequential uuids

niwinz (approved):

frontend/src/app/plugins/tokens.cljs
Alotor: Is correct because shadows only have `:offset-x  :offset-y  :blur  :spread  :color  :inset` fields. Only need to convert the two offsets. I'm going to rename the function so it's more clear that is only for shadows.

frontend/test/frontend_tests/plugins/shape_bugfixes_test.cljs
Alotor: ok

PR #10428, opened 2026-06-25
:sparkles: Add minor changes to the mcp that simplifies its setup

### LINKED ISSUE #10355: Simplify MCP server configuration for common MCP clients
MCP server beta users understandably mentioned that they found the configuration of their MCP client a bit cumbersome. The goal of this ticket is to simplify the client setup by using the library/tool [add-mcp](https://github.com/neon-solutions/add-mcp) as follows:

1. For local MCP server users (who use the npm package), add an installation script which automatically configures the user's preferred client; add-mcp can be used as a library to support this.
2. For remote MCP server users, add an npx command to the documentation which uses add-mcp to simplify installation.

MischaPanch:
### Related Ticket

https://github.com/penpot/penpot/issues/10355

### Summary

Added a wrapper around mcp-add to the penpot/mcp package, extended docs around that and
removed client-specific mcp config instructions (except for Claude Desktop).

### DIFF
--- a/mcp/packages/server/tsconfig.json
+++ b/mcp/packages/server/tsconfig.json
@@ -17,6 +17,9 @@
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true
   },
-  "include": ["src/**/*"],
+  "include": [
+    "src/**/*",
+    "../plugin/src/index.d.ts"
+  ],
   "exclude": ["node_modules", "dist"]
 }
--- a/docs/mcp/index.md
+++ b/docs/mcp/index.md
@@ -69,7 +69,7 @@ There are three key pieces:
 
 ![How the AI client, Penpot MCP server, plugin, and design file connect](/img/mcp/mcp-flow.webp)
 
-### Basic concepts 
+### Basic concepts
 
 Some important concepts for users:
 * **Integrations page**: MCP is configured under **Your account → Integrations → MCP Server**. Here you enable or disable MCP, get the server URL and manage the MCP key.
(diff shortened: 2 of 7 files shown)


niwinz:
The fmt of ts code is failing. Apart of that, for me is ready to merge.

MischaPanch:
Hi @niwinz , thanks for the review. Something went wrong during rebase, but it should be fine now. I formatted and fixed the conflicts in package.json, should be ready for merge

mcp/packages/server/tsconfig.json
```diff
@@ -17,6 +17,9 @@
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true
   },
-  "include": ["src/**/*"],
+  "include": [
+    "src/**/*",
+    "../plugin/src/index.d.ts"
```
MischaPanch, line 22: Hm, I think this should be rolled back. It was for using `McpOptions.getServerUrl` which in the end just complicated things, so we currently use the hardcoded localhost option (although we do read the port from the env var). Since you already reviewed @niwinz , I wanted to ping you before removing. I guess you don't insist on using `McpOptions.getServerUrl` for the url (a minor duplication), right?

opcode81 (changes requested):
This does not actually achieve that the setup command is exposed through the published artifact when running it via npx.

opcode81:
@MischaPanch I dropped the unrelated docs commit. An improved version thereof is now here: #10565

opcode81:
I am taking this over.

opcode81:
Superseded by 
* #10604 
* #10605 (follow-up after npm release)

PR #10439, opened 2026-06-26
:lipstick: Add fade effect to templates carousel near navigation arrows

parinith-web:
### Changes

Adds a fade effect at the edges of the "Libraries & Templates" carousel so cards visually dissolve into the background before reaching the navigation arrows, instead of being abruptly cut off behind them.

### Issue

- The left/right navigation arrows overlapped directly on top of the nearest card when scrolling, cutting off its title text and thumbnail abruptly since the arrow's square bounding box left visible gaps around its circular shape.

### Result

- Added two fade overlays (`content-fade-left` / `content-fade-right`) positioned absolutely within the carousel section, each tied to the same `can-move` state that controls whether the corresponding arrow is shown.
- The fade only appears on the side where scrolling is actually available, so it correctly disappears once you've scrolled all the way in that direction.
- Cards now fade smoothly into the background near the arrows regardless of scroll position, with no overlap or abrupt cutoff.

### Testing

Tested locally via the devenv:

- Confirmed right fade appears on initial load and disappears once scrolled fully right.
- Confirmed left fade appears after scrolling right and disappears once scrolled back to the start.
- Confirmed cards no longer show abrupt text/thumbnail cutoff near either arrow at any scroll position.

### Screen Recordings

1. before the fix (card text/thumbnail cut off abruptly behind the arrow)
https://github.com/user-attachments/assets/5b17e833-7fc0-4389-88cd-abcbf9f0276d

2. after the fix (card fades smoothly into the background near the arrow)
https://github.com/user-attachments/assets/8efde636-8660-43b4-a5ed-1c7bfa7b4f51

Signed-off-by: @Parinith-web  <parinithreddymav@gmail.com>


madalenapmelo-kp:

niwinz:
Marked ready for review, if proceed, would be nice to have an issue created for this PR

frontend/src/app/main/ui/dashboard/templates.scss
```diff
@@ -147,6 +147,25 @@
   margin-inline-end: px2rem(44);
 }
 
+.content-fade {
+  position: absolute;
+  inset-block-start: 0;
+  inset-block-end: 0;
+  width: px2rem(90);
```
EvaMarco, line 154: Please use logical properties.

frontend/src/app/main/ui/dashboard/templates.scss
```diff
@@ -147,6 +147,25 @@
...
+  pointer-events: none;
+  z-index: 1;
```
EvaMarco, line 156: Please use proper z-index variables defined on frontend/src/app/main/ui/ds/z-index.scss

EvaMarco (changes requested):

EvaMarco:
Thanks for the PR!

Our contribution workflow requires opening an issue before submitting a PR for this kind of change. This helps us discuss and validate the proposal before implementation.

Please follow the workflow described here:
https://github.com/penpot/penpot/blob/develop/CONTRIBUTING.md#workflow

Could you open an issue describing the proposed change and link it to this PR? Once that's done, we'll be happy to continue the review. Thanks!

niwinz:
Closing for inactivity

PR #10450, opened 2026-06-27
:bug: Highlight first search result in font picker and fix Enter confirm

### LINKED ISSUE #3204: Highlight first found font at the fonts list when searching

filipsajdak:
> This pull request was prepared with AI assistance (Claude Sonnet 4.6) as part of a Penpot MCP self-improvement initiative. All logic has been reviewed and verified by the human author.

## Related issue

Closes #3204

## Summary

When a user types a search term that filters out the currently-active font, the font picker showed no selection at all and pressing Enter would close without applying any font.

A previous attempt (#9512, merged then reverted) fixed the visual highlight by resetting `selected` state on every search change. That caused every keystroke to immediately live-apply a new font via the `(mf/with-effect [@selected] (on-select @selected))` path -- which is why it was reverted.

This PR takes a different approach:

**No state mutation during search.** A pure render-time binding `effective-selected` is computed from the existing `@selected` and `fonts` values:

```clojure
effective-selected
(if (and (seq fonts) (not (d/seek #(= (:id %) (:id @selected)) fonts)))
  (first fonts)
  @selected)
```

When the current font is absent from the filtered results `effective-selected` is the first filtered font; otherwise it is `@selected`. No atom is mutated, so the `on-select` effect is not triggered while the user types.

**Enter key applies the first result explicitly.** The `on-key-down` handler checks at confirm time whether the current selection is in the filtered list:

```clojure
(kbd/enter? event)
(let [first-result (when-not (d/seek #(= (:id %) (:id @selected)) fonts)
                     (first fonts))]
  (if first-result
    (do (on-select first-result) (on-close))
    (on-close)))
```

Arrow-key live-preview is unchanged (it still mutates `@selected`).

**Deps updated for correctness.** `on-key-down` deps include `on-select` and `on-close`; the global keydown listener effect includes `on-key-down` so it is always reinstalled with the latest callback.

## Behaviour

| Scenario | Before | After |
|---|---|---|
| Search narrows, current font still in results | tick on current font | tick on current font (unchanged) |
| Search narrows, current font filtered out | no tick visible; Enter closes with no change | tick on first result; Enter applies first result |
| Arrow-key navigation | live-preview updates | unchanged |
| Typing (no navigation) | font unchanged until Enter | font unchanged until Enter (no live-apply) |

## Testing

- `clj-kondo --lint ../common/src src/`: 0 errors, 0 warnings
- `cljfmt check src/`: all files formatted correctly
- `clojure -M:dev:shadow-cljs compile test`: build completed, 0 warnings

Manually verified against the devenv: open the font picker on a text element, type a search term that excludes the current font -- the first matching row now shows the selection tick and pressing Enter applies it without flickering through fonts while typing.


madalenapmelo-kp:

PR #10453, opened 2026-06-27
:bug: Center canvas and select layer when navigating search results

### LINKED ISSUE #10422: Canvas does not center on the selected layer when navigating through layer search results
### Description
When using the previous and next buttons to navigate through layer search results, neither the layers panel nor the canvas update correctly. The matching layer is highlighted in the layers panel but the panel does not scroll to bring it into view, and the canvas does not center on it. The navigation buttons appear to have no visible effect beyond advancing the internal search cursor.

### Environment
- Pro
- Layers panel / layer search

### Steps to reproduce
1. In a file with a lot of layers, open the layers panel and use the search field to search for a layer name.
2. Use the **previous** or **next** buttons to navigate between the matching results.
3. Observe the layers panel and the canvas.

### Current result
- The layer is **not selected or highlighted** in the layers panel.
- The layers panel **does not scroll** to bring the matching layer into view.
- The canvas **does not center** on the selected layer.

Image

### Expected result
Each time the selected layer changes via the previous or next navigation buttons:
- The layer should be **selected and highlighted** in the layers panel.
- The layers panel should **scroll** to make the layer visible.
- The canvas should **center on the layer**, making it immediately visible without any manual action.

parinith-web:
### Changes

Fixes layer search navigation in the workspace sidebar so that clicking the previous/next buttons actually selects the matching layer and centers the canvas on it.

### Issue

Closes #10422

When navigating between search results using the previous/next arrows in the Find & Replace panel, only the internal counter was updating. The layer was not being selected and the canvas was not centering on the matching shape.

The root cause was that `navigate-next` and `navigate-prev` only called `swap! state*` to update the index, but `state*` resets on re-render (triggered by the search itself), so the index always read as 0 and the same layer was repeatedly targeted. Additionally, no selection or zoom event was ever emitted.

### Fix

- Added `text-match-ids-ref` and `match-idx-ref` as stable refs that survive re-renders, replacing direct reads from component state.
- Both handlers now emit `dw/select-shape` and `dw/zoom-to-selected-shape` on each navigation step, selecting the layer and centering the canvas.

### Testing

Tested locally via the devenv with a file containing multiple layersmatching the search term:

- Confirmed canvas centers on each matching layer when clicking next/prev.
- Confirmed the correct layer is selected and highlighted in the layers panel.
- Confirmed wrap-around works correctly (last → first and first → last).

### Screen Recordings

1. before the fix
https://github.com/user-attachments/assets/af8fa559-ae42-4f3f-a2ae-5a3f126861fd

2. after the fix
https://github.com/user-attachments/assets/70e1b1bf-f0a4-472d-91a8-762ef24bc417

Signed-off-by: @Parinith-web

 <parinithreddymav@gmail.com>


madalenapmelo-kp:

niwinz:
Asked internally to design/ux team for a feedback for this.
The code looks good, thanks :heart:

PR #10454, opened 2026-06-27
:bug: Position overlays by frame selrect, not filter-inflated bounds

### LINKED ISSUE #9048: bug: Overlay misalignment in some prototypes
### Steps To Reproduce

* Unzip and import: [Clicking Away.penpot.zip](https://github.com/user-attachments/files/26825857/Clicking.Away.penpot.zip)
* Click on the + and - to interact with the prototype

### Expected behavior

I expected the new overlay to show aligned with the top-center as the overlay which invoked the new overlay.

### Actual behavior

The new overlay shows up always a bit to the left.

### Screenshots or video

https://github.com/user-attachments/assets/f746a534-030f-4d36-8ddb-ce433edd06ca

design.penpot.app and self-host

### Additional context

I tried this in a simpler file but it then worked. So I'm not sure what is unique about this situation. It is definitively a bug though.

---

Example of one of the interactions I setup:

Image

Clicking on the `-` to opens the `S` overlay, using the `-` parent (M) as reference. All the overlays (S, M, L) have the same width.

---

The workaround I found was to use manual placement, which is a bit annoying when I have to start creating new variations and want the prototype to keep working.

filipsajdak:

## Related issue

Closes #9048

## Summary

When a prototype interaction opens an overlay positioned relative to a frame (e.g. `top-center` / `center` / `bottom-right`), the overlay can appear shifted (the reporter saw it "always a bit to the left"). It happens whenever the destination overlay frame has effects or overflow (drop shadow, blur, outer stroke, or `show-content` children) that make its rendered object bounds larger than its `selrect`.

## Root cause

`calc-overlay-position` (`common/src/app/common/types/shape/interactions.cljc`) measured the overlay with its full object bounds:

```clojure
overlay-size (gsb/get-object-bounds objects dest-frame)
```

but measured the frame it is positioned relative to with the plain `selrect`. The centering/right/bottom formulas then mix the two, e.g. for `top-center`:

```clojure
x = base-x + (relative-selrect-width - overlay-bounds-width) / 2
```

`get-object-bounds` inflates the box by shadow/blur/stroke padding and overflowing children, so when those exist the visible frame ends up off-center by half that padding.

The sibling helper `calc-overlay-pos-initial` already uses `(:selrect dest-frame)` for exactly this computation, so the two were inconsistent. The viewer separately reserves the full bounds size and re-aligns the selrect (`frontend/src/app/main/ui/viewer.cljs` `calculate-size` / `calculate-delta`), so the position itself should be computed from the visible `selrect`.

## Fix

Use the destination frame `selrect` (the visible frame box) instead of its object bounds when computing the overlay position:

```clojure
overlay-size (:selrect dest-frame)
```

This makes `calc-overlay-position` consistent with `calc-overlay-pos-initial`, and the `gsb` (geom.shapes.bounds) require becomes unused and is removed.

## Reproduction (live, in a devenv)

Using the file attached to #9048 (`Clicking Away.penpot`), in the prototype viewer, opening one overlay from another (same width, `top-center`, relative to the same frame):

- Before: consecutive overlays were 20px apart horizontally (the destination frames render at ~200px wide for a 160px selrect, so the bounds were 40px wider; `(160 - 200) / 2 = -20`).
- After: consecutive overlays share the same left edge (centered as expected).

## Tests

- Added `calc-overlay-position-ignores-filter-bounds` to `common/test/common_tests/types/shape_interactions_test.cljc`. It builds two destination frames with the same selrect (one with a bounds-inflating drop shadow), asserts the shadow really does enlarge `get-object-bounds` beyond the selrect, and then checks that `calc-overlay-position` returns the same position with and without the shadow for `:center`, `:top-center`, `:top-right`, `:bottom-center` and `:bottom-right`.
- `clojure -M:dev:test` (common): 1063 tests, 24452 assertions, 0 failures.
- `clj-kondo` and `cljfmt`: clean for the changed files.


madalenapmelo-kp:

niwinz:
Thanks :heart:

PR #10455, opened 2026-06-27
:lipstick: Make typography assets adapt width on sidebar resize

### LINKED ISSUE #10437: Make typography assets responsive
Unlike other assets, typography assets does not adapt its width when the user change the sidebar size. The rest of the assets do. 

Image

**Solution**

Typography assets should adapt its size.

mvanhorn:
### Related Ticket

The fix is a styling alignment in the typography assets component. Compare the width/layout rules in `typographies.scss` (and the class bindings in `typographies.cljs`) against how a sibling asset group that already resizes correctly is styled, and bring the typography item/group container in line - typically replacing a fixed/min width or absent `width: 100%`/flex sizing with the responsive rule the other asset groups use so the row follows the sidebar's resizable width. Because the issue carries a `needs definition` label, keep the change minimal and visual-only (match existing asset responsive behavior, do not introduce new layout semantics); confirm against the other asset groups' SCSS to avoid over-reaching.

Fixes #10437

### Checklist

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/assets/typographies.scss
+++ b/frontend/src/app/main/ui/workspace/sidebar/assets/typographies.scss
@@ -7,7 +7,14 @@
 @use "refactor/common-refactor.scss" as deprecated;
 
 .assets-list {
+  box-sizing: border-box;
+  width: 100%;
   padding: 0 0 0 deprecated.$s-4;
+}
+
+.typographies-group {
+  min-width: 0;
+  width: 100%;
 }
 
 .drop-space {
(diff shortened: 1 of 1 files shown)


madalenapmelo-kp:

EvaMarco:
Hello @mvanhorn 

Once the conflicts with the main branch are resolved, we can continue with the review.

There is also a layout detail that needs to be adjusted: when this menu is expanded, it should fill the full available width from its parent container, including its children. At the same time, this component is shared with the right sidebar, where it must keep the existing var(--options-width) max-width constraint.

Please make sure the solution handles both contexts correctly instead of applying a global width change that could break the sidebar behavior.

EvaMarco:
Hi! We’re closing this PR for now because the related issue is still in the definition phase and is not ready for implementation yet.

The changes were started before the requirements were fully defined, so we’ll revisit the implementation once the issue is ready.

Thanks for your contribution!

PR #10492, opened 2026-06-30
:bug: Fix SVG raw not being saved

### LINKED ISSUE #10488: Using Cow prevents caching SVGRaw
### Steps To Reproduce

Create a SVG Raw

We do this:

```rust
// We don't want to change the value in the global state
let mut shape: Cow<Shape> = Cow::Borrowed(shape);
let shape = shape.to_mut();
```

before this:
```rust
let font_manager = skia::FontMgr::from(self.fonts().font_provider().clone());
let dom_result = skia::svg::Dom::from_str(&sr.content, font_manager);
match dom_result {
  Ok(dom) => {
    dom.render(self.surfaces.canvas_and_mark_dirty(fills_surface_id));
    shape.set_svg(dom); // <<< 👀 HERE! Clone-on-write mutates the clone, not the original. 
  }
  Err(e) => {
    eprintln!("Error parsing SVG. Error: {}", e);
  }
}
``` 

### Expected behavior

SVG DOM should be available when the SVG Raw is rendered.

### Actual behavior

SVG DOM is created on every `render_shape` call.

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

AzazelN28:
### Checklist

### DIFF
--- a/render-wasm/src/shapes.rs
+++ b/render-wasm/src/shapes.rs
@@ -722,6 +722,23 @@ impl Shape {
         self.invalidate_extrect();
     }
 
+    pub fn update_svg_raw_content(&mut self, font_manager: skia::FontMgr) {
+        match &self.shape_type {
+            Type::SVGRaw(sr) => {
+                let dom_result = skia::svg::Dom::from_str(&sr.content, font_manager);
+                match dom_result {
+                    Ok(dom) => {
+                        self.set_svg(dom);
+                    }
+                    Err(e) => {
+                        eprintln!("Error parsing SVG. Error: {}", e);
+                    }
+                }
+            }
+            _ => panic!("Updating SVG raw content on non SVG Raw shape"),
+        }
+    }
+
     pub fn set_svg_raw_content(&mut self, content: String) {
         self.shape_type = Type::SVGRaw(SVGRaw::from_content(content));
     }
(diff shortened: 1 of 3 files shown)


render-wasm/src/shapes.rs
```diff
@@ -722,6 +722,23 @@ impl Shape {
...
+                match dom_result {
+                    Ok(dom) => {
+                        self.set_svg(dom);
+                    }
+                    Err(e) => {
+                        eprintln!("Error parsing SVG. Error: {}", e);
+                    }
+                }
+            }
+            _ => panic!("Updating SVG raw content on non SVG Raw shape"),
```
belen-albeza, line 738: Shall we return a Result with a critical error instead of just panicking here?

AzazelN28: I think we should always `panic!` in this kind of situation instead of returning a `Result` for two reasons:

1. A `Result` can be ignored (as it already happened with other errors) and this can lead to a silently degraded experience.
2. A `panic!` near the source of error is a better way of handling/detecting errors that should never happen (like this, we should never allow updating svg raw content on non svg raw shapes).

PR #10522, opened 2026-07-01
:bug: Show flex/grid element options for text layers in a layout

### LINKED ISSUE #10514: missing flex element options in text layers inside a layout
### Description
When a text element is placed inside a frame with a flex or grid layout, the flex element options are no longer available in the design panel. This option is present for other element types in the same context and works correctly in production (Penpot.app). The issue affects both render engines and is therefore a regression introduced in a recent version.

### Environment
- PRE
- Context: text element inside a flex or grid layout frame
- Affects: both render engines
- Works correctly in: production (Penpot.app)

### Steps to reproduce
1. Create a frame and apply a **flex** or **grid** layout to it.
2. Place a text element inside the frame.
3. Select the text element and inspect the design panel.
4. Look for the flex element controls.

### Current result
The flex element options are not shown in the design panel for text elements inside a layout frame.

Image

### Expected result
Text elements inside a layout should display the same flex element controls as other element types.

Image

filipsajdak:

## Related issue

Closes #10514

## Summary

When a text element is placed inside a frame with a flex or grid layout, the layout-item ("FLEX ELEMENT" / "GRID ELEMENT") options were missing from the Design panel. The same options render correctly for every other shape type (boards, rectangles, paths, etc.) in the same context, so text was the only affected type -- a regression.

## Root cause

`text.cljs` invoked the shared `layout-item-menu*` component with `[:&` (cljs-map props) instead of `[:>` (JS object props):

```clojure
;; frontend/src/app/main/ui/workspace/sidebar/options/shapes/text.cljs
(when is-layout-child?
  [:& layout-item-menu*        ; <- should be [:>
   {:ids ids
    :type type
    ...}])
```

`layout-item-menu*` is memoized with a custom equality check (`check-layout-item-menu-props`) that reads its props as a JS object via `unchecked-get` with camelCase keys (`"isLayoutChild"`, `"isFlexParent"`, `"isGridParent"`, `"appliedTokens"`, ...). Those keys only exist when the component is invoked with `[:>` (rumext converts the map to a JS object with camelCased keys). With `[:&` the props arrive as a cljs map, the camelCase lookups miss, and the flex/grid element controls do not render for text.

Every other shape option file (`rect.cljs`, `circle.cljs`, `path.cljs`, `group.cljs`, `svg_raw.cljs`) already uses `[:>` -- `text.cljs` was the lone outlier.

## Fix

Invoke `layout-item-menu*` with `[:>` in `text.cljs`, matching the other shapes and the component's prop expectations. One-character change; the passed props are otherwise identical to `rect.cljs`.

## Verification (live, in a devenv)

Made a board a flex container so its text child became a flex item, then selected the text:

- Before: the Design panel showed only a bare "LAYOUT ELEMENT" label with the sizing icons -- no STATIC/ABSOLUTE toggle, Z-index or margin controls.
- After: the text shows the full "FLEX ELEMENT" section (STATIC/ABSOLUTE, Z-index, vertical/horizontal margins, sizing and alignment) -- identical to a board child in the same container.

`clj-kondo` and `cljfmt` are clean for the changed file; the frontend compiles with 0 warnings.

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/options/shapes/text.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/shapes/text.cljs
@@ -171,7 +171,7 @@
          :cell (ctl/get-cell-by-shape-id (first parents) (first ids))}])
 
      (when is-layout-child?
-       [:& layout-item-menu*
+       [:> layout-item-menu*
         {:ids ids
          :type type
          :values layout-item-values


madalenapmelo-kp:

NatachaMenjibar:
Hey @filipsajdak thanks for your contribution, but this issue was already fixed in this PR https://github.com/penpot/penpot/pull/10516 so it's duplicated

Issue #10539, opened 2026-07-02
Not all elements showing up in sidebar with grid layout


ChaoticLeah:
### Steps To Reproduce

1. Create a board with the grid layout
2. Make a ton of rows and columns (In this demo I have an 18 x 4, if that doesnt work try more/refresh the tab or contract and expand the frame as sometimes it does actually work as expected)
3. Put anything inside of it and fill it up. (A bunch of basic rectangles copy pasted to fill it will do)
4. Try selecting the lowest rectangle in the list, it wont be the last one. You can also try selecting a rectangle from the viewport and it wont select any item seen in the sidebar

### Expected behavior

Items should never go missing from the sidebar, they should always be there

### Actual behavior

Items above a certain point dont show up in the sidebar making working on a big project hard

### Screenshots or video

<img width="1125" height="831" alt="Image" src="https://github.com/user-attachments/assets/47e4be0f-ad0d-44e7-8c77-42a796166c8d" />

Another picture of me selecting all the elements on the sidebar only for it to select half of them:

<img width="1143" height="772" alt="Image" src="https://github.com/user-attachments/assets/263edb0b-598f-41d2-827a-13b2115941b9" />

### Desktop (please complete the following information)

Windows, Firefox Version 152.0.3 (64-bit)

Also tried: 
 - The unofficial penpot desktop app which uses electron (chromium based)

### Environment (please complete the following information)

Self hosted penpot 2.16.1

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

Sometimes through some expanding and contracting the child contents it appears to sometimes come back but usually not for long and not often.

This also may not be exclusive to just grids but it appears to work all fine with a normal frame and a bunch of children, the same is true for flex layouts.

All this applies to both the old renderer and the new opengl renderer

EvaMarco:
Hello @ChaoticLeah 

Thanks for the report! We tried to reproduce the issue by following the steps you described, but we haven't been able to reproduce the behavior on our side.

<img width="2923" height="1352" alt="Image" src="https://github.com/user-attachments/assets/b4642a2f-cb1f-4099-a49e-699d31088ffc" />

Could you provide a bit more information? A screen recording showing the issue or the file where it occurs would be especially helpful. If there are any additional steps or specific conditions required to trigger the problem, please let us know as well.

That will help us investigate further. Thanks!

filipsajdak:
[...]

PR #10554, opened 2026-07-03
:bug: Load layers-panel chunk of a selected off-screen child

filipsajdak:

## Related issue

Part of #10539.

## Summary

The workspace layers panel lazy-loads a container's children in chunks of `default-chunk-size` (50): on expand it renders the first chunk and loads more as an `IntersectionObserver` sentinel scrolls into view. There is a dedicated effect that is supposed to *also* load enough chunks to include any currently **selected** child immediately (e.g. when you select a shape in the canvas and it isn't in the first 50 rendered rows). That path was broken, so selecting an off-screen child never revealed/selected it in the layers panel -- matching the "selecting a rectangle from the viewport won't select any item seen in the sidebar" part of #10539.

## Root cause

In `frontend/src/app/main/ui/workspace/sidebar/layer_item.cljs`, the effect computes the render index of a selected child with:

```clojure
(let [idx (.indexOf shapes sel-id)]
  (when (>= idx 0) idx))
```

but `shapes` at that point is a vector of **resolved shape objects**, while `sel-id` is a **uuid**. `.indexOf` compares by identity, so it is always `-1`; `selected-child-render-idx` is always `nil` and the code falls back to `default-chunk-size`. A selected child living beyond the first 50 is therefore never loaded.

## Fix

Match the selected id against each shape's `:id` using a value-based lookup:

```clojure
(some (fn [sel-id]
        (d/index-of-pred shapes #(= (:id %) sel-id)))
      selected)
```

`d/index-of-pred` returns the index (or `nil`), which is exactly what `some` and the surrounding `min-count` math expect.

## Verification (live, in a devenv)

Built a group with 64 children, then reproduced the windowed state (collapse + re-expand the container so only the first 50 render) and scrolled the layers panel to the top so the lazy-load sentinel stays below the fold and does not fire:

- **Before:** selecting a not-yet-rendered child leaves the panel at 50 rendered rows -- the child is never loaded or selected in the sidebar.
- **After:** selecting the same child loads its chunk (panel goes to 64 rendered rows) and the child is revealed/selected.

`clj-kondo` and `cljfmt` are clean for the changed file; the frontend compiles with 0 warnings.

## Note

This fixes the *selected-child reveal* path specifically. The broader "items above a certain point don't show until you scroll" behaviour reported in #10539 is the windowing/sentinel itself (only the first chunk renders until the sentinel intersects); that is a separate, more general question about the layers-panel lazy-loading and is not changed here.


madalenapmelo-kp:

niwinz:
This PR is a comment only, it does not changes any effective code, is expected?

filipsajdak:
Good catch, and thanks for checking - you are right, this PR no longer changes any effective code.

The fix it originally carried is already on `develop`: `layer_item.cljs` now computes the index with `d/index-of-pred`, landed by #10600 ("Fix sidebar scroll to shape"). My branch predated that, so when I rebased, the code hunk was correctly dropped as redundant and only the explanatory comment I had added above the block survived. I should have caught that on the rebase rather than leaving it for you to spot.

Since #10600 covers the case this PR targeted - selecting an off-screen child never loading its chunk, because `(.indexOf shapes sel-id)` compared a uuid against resolved shape objects and so was always `-1` - there is nothing left here worth your review time. Closing it.

Sorry for the noise, and thanks for the ping.

Issue #9947, opened 2026-05-31
bug: @penpot/mcp npx install breaks on repeated runs after npm prunes pnpm virtual store


shbernal:
### Steps To Reproduce

1. Use a current Node/npm setup where `npx` is provided by npm 11.

   In my case:

   ```console
   $ node -v
   v24.11.1
   $ npm -v
   11.6.2
   $ npx --version
   11.6.2
   ```

2. Run the published MCP package:

   ```console
   npx -y @penpot/mcp@stable
   ```

3. If prompted by pnpm 11's build approval gate, approve only the required build scripts:

   ```console
   cd ~/.npm/_npx/<hash>/node_modules/@penpot/mcp
   corepack pnpm approve-builds esbuild sharp
   ```

4. Run the package again:

   ```console
   npx -y @penpot/mcp@stable
   ```

### Expected behavior

The local MCP server should start reliably on repeated `npx -y @penpot/mcp@stable` invocations, as documented. The local plugin server should be available at `http://localhost:4400/manifest.json`, and the MCP endpoint should be available at `http://localhost:4401/mcp`.

### Actual behavior

The first repaired run can start, but a later `npx -y @penpot/mcp@stable` invocation breaks the cached package install. The package's nested pnpm virtual store is removed by npm's `npx` reify step, but pnpm still thinks the workspace is up to date.

The next build then fails because workspace symlinks point to deleted pnpm virtual-store entries:

```console
packages/plugin build: Error: Cannot find module '/home/shb/.npm/_npx/3c82a6444c262e31/node_modules/@penpot/mcp/packages/plugin/node_modules/typescript/bin/tsc'
packages/common build: Error: Cannot find module '/home/shb/.npm/_npx/3c82a6444c262e31/node_modules/@penpot/mcp/packages/common/node_modules/typescript/bin/tsc'
```

I confirmed the failure mode by repairing the nested install manually, then invoking `npx` again:

```console
$ du -sh ~/.npm/_npx/3c82a6444c262e31/node_modules/@penpot/mcp
175M
$ test -e ~/.npm/_npx/3c82a6444c262e31/node_modules/@penpot/mcp/node_modules/.pnpm && echo virtual-store=yes
virtual-store=yes

$ timeout 8s npx -y @penpot/mcp@stable
# fails with missing typescript/bin/tsc

$ du -sh ~/.npm/_npx/3c82a6444c262e31/node_modules/@penpot/mcp
2.8M
[...]

PR #10555, opened 2026-07-04
:bug: Fix MCP local launcher runtime path

Craffo:
_Draft for maintainer feedback. I can adjust the runtime/packaging approach if there is a preferred direction._

### Related Ticket

Related to #9947

### Summary

The published local MCP launcher currently behaves like a development bootstrap on every run:

- installs the pnpm workspace
- rebuilds the server and plugin
- then starts the runtime

That makes `npx -y @penpot/mcp` mutate its installed package directory at launch time, which can fail when npm/npx later prunes or reifies that cached install. It also makes normal startup slower than necessary.

This PR keeps the existing bootstrap flow intact, but adds a runtime path for packaged installs:

- detect existing built server and plugin artifacts
- start `packages/server/dist/index.js` directly
- serve the built plugin via `vite preview`
- fall back to the existing bootstrap command when artifacts are missing

The package script used for publishing now builds before `npm pack`, and npm ignore rules allow the built artifacts to be included in the tarball.

Backward compatibility is preserved because source checkouts without `dist` artifacts still use the current bootstrap path exactly as before. The existing `bootstrap` and `bootstrap:multi-user` scripts are unchanged, so developer workflows remain the same.

### Runtime Behavior

For users installing `@penpot/mcp` from npm, a normal launch should be able to:

1. start immediately when runtime artifacts already exist;
2. avoid package installation during normal startup;
3. avoid rebuilding unless artifacts are missing.

Development bootstrap remains available and unchanged, but packaged npm installs can now use the prebuilt runtime path.

### Steps to reproduce

Before this change:

1. Run `npx -y @penpot/mcp`.
2. The launcher runs `pnpm -r install`, `pnpm run build`, and `pnpm run start`.
3. Repeated `npx` runs can fail when npm prunes the nested pnpm install state.

In environments using pnpm versions with stricter build-script approval behavior, the repeated install path can also fail before the MCP server starts.

### Verification

Verified with:

- `node --check bin/mcp-local.js`
- `pnpm exec prettier --check bin/mcp-local.js package.json`
- `pnpm run test`
- `bash scripts/pack`
- production-only install of the generated tarball with `npm install --omit=dev`
- launch of installed `penpot-mcp`, confirming no pnpm install/build output

### Backwards Compatibility

- Existing `bootstrap` and `bootstrap:multi-user` scripts are unchanged.
- Source checkouts without built artifacts still fall back to the existing bootstrap flow.
- Packaged installs with built artifacts can start directly through the runtime path.
- No Docker or self-host configuration is changed.

### Checklist

### DIFF
--- a/mcp/.npmignore
+++ b/mcp/.npmignore
@@ -1,0 +1,12 @@
+.idea
+.claude
+node_modules
+*.tgz
+*.bak
+*.orig
+temp
+*.tsbuildinfo
+
+# Log files
+logs/
+*.log
--- a/mcp/bin/mcp-local.js
+++ b/mcp/bin/mcp-local.js
@@ -1,31 +1,86 @@
 #!/usr/bin/env node
 
-const { execSync } = require("child_process");
+const { execSync, spawn } = require("child_process");
 const fs = require("fs");
 const path = require("path");
 
 const root = path.resolve(__dirname, "..");
 const pkg = require(path.join(root, "package.json"));
+const serverRoot = path.join(root, "packages/server");
+const pluginRoot = path.join(root, "packages/plugin");
+const serverEntry = path.join(serverRoot, "dist/index.js");
 
 function pnpmVersion() {
-  const match = (pkg.packageManager || "").match(/^pnpm@([^+]+)/);
-  return match ? match[1] : "latest";
+    const match = (pkg.packageManager || "").match(/^pnpm@([^+]+)/);
+    return match ? match[1] : "latest";
 }
 
 function run(command) {
-  execSync(command, { cwd: root, stdio: "inherit" });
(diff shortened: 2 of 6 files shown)


opcode81:
Thanks for the analysis. The diagnosis likely accurate, and shipping prebuilt artifacts is a reasonable direction. The failure depends on newer npm/npx versions actually re-reifying the `_npx` cache on repeated runs (and thereby pruning the nested pnpm store), which is why we never came across it ourselves.

We won't merge this implementation though. Duplicating the server's runtime deps into the root manifest creates two hand-synced lists that will drift. We'll think about what to do.

opcode81:
#10564 is an alternative, minimal fix.

opcode81:
#10564 was merged, so this can be closed.

PR #10557, opened 2026-07-05
:bug: Prevent caching of env-generated config.js in self-hosted Docker

### LINKED ISSUE #10556: Self-host: env-generated /js/config.js is cached 7 days and only build-versioned, so PENPOT_FLAGS changes don't reach already-cached browsers
### Summary

On self-hosted Docker installs, `/js/config.js` is generated from `PENPOT_FLAGS` at container start, but it is served with `Cache-Control: public, max-age=604800` (7 days) and its URL is versioned only by the build (`config.js?version=<APP_BUILD>`). A flags-only change (edit `PENPOT_FLAGS`, restart the frontend) does not change the build version, so the `config.js` URL is unchanged and browsers that previously loaded the app keep serving the pre-change `config.js` from cache for up to 7 days. The flag change is invisible until cache expiry or a manual site-data clear.

Concrete impact: enabling `enable-login-with-google` (plus `PENPOT_GOOGLE_CLIENT_ID` / `PENPOT_GOOGLE_CLIENT_SECRET`) does not show the "Log in with Google" button for returning users, even though the backend initializes the provider and the freshly served `config.js` already contains the flag. `frontend/src/app/main/ui/auth/login.cljs` gates the SSO buttons on `(some (partial contains? cf/flags) [:login-with-google ...])`, and `cf/flags` is populated from the cached `config.js`.

### Environment
- Self-hosted, `penpotapp/frontend` 2.16.2 (the relevant code is also on `main`), Docker.

### Steps to reproduce
1. Run a Penpot instance and open it once in a browser (populates the HTTP cache).
2. Add a flag to `PENPOT_FLAGS` (e.g. `enable-login-with-google`, with the Google client id/secret) and restart the frontend container.
3. In the same browser, reload the login page.

### Observed
- The new behavior (e.g. the Google SSO button) does not appear on reload.
- The frontend serves the updated `config.js` body, but with a long cache:
  ```
  $ curl -sI https://<host>/js/config.js
  cache-control: public, max-age=604800
  ```
- `index.html` loads it as `./js/config.js?version=2.16.2-<build>`; the `?version=` value is unchanged by a flags-only restart, so the browser reuses its cached copy.

### Expected
- After changing `PENPOT_FLAGS` and restarting, a normal reload reflects the new flags without requiring end users to clear their browser cache.

### Root cause
- `docker/images/files/nginx-entrypoint.sh` rewrites config.js from env at boot:
  ```
  update_flags /var/www/app/js/config.js
  update_oidc_name /var/www/app/js/config.js
  ```
- `docker/images/Dockerfile.frontend` ships config.js as a plain, non-content-hashed file:
  ```
  COPY ./files/config.js /var/www/app/js/config.js
  ```
- `docker/images/files/nginx.conf.template` caches it 7 days as a generic `.js` asset:
  ```
  location ~* \.(js|css|jpg|png|svg|gif|ttf|woff|woff2|wasm|map)$ {
      include /etc/nginx/nginx-security-headers.conf;
      add_header Cache-Control "public, max-age=604800" always; # 7 days
  }
  ```
`config.js` is the one asset whose content changes independently of the build, but it is cached like the content-hashed bundles, and its URL query is the build version rather than a content hash, so there is no cache-bust when only the content changes.

### Proposed fix
In `docker/images/files/nginx.conf.template`, exempt config.js from the long-cache rule with an exact-match location, e.g.:
```
location = /js/config.js {
    include /etc/nginx/nginx-security-headers.conf;
    add_header Cache-Control "no-store, no-cache, must-revalidate, max-age=0" always;
}
```
Alternatively, bust the URL when the content changes (write a content hash or boot token into the `config.js?version=` query in index.html) instead of using only the build version.

### Workaround for self-hosters
Mount an nginx override that serves config.js uncached (an exact-match `location = /js/config.js` with `Cache-Control: no-store`), then recreate the frontend. Existing clients still need to clear site data once, since the previously cached copy retains its 7-day lifetime.

NextLevelManagementAdvisors:
Closes #10556

### Problem
On self-hosted Docker, `/js/config.js` is generated from `PENPOT_FLAGS` at container start (`docker/images/files/nginx-entrypoint.sh` → `update_flags /var/www/app/js/config.js`) but is served with `Cache-Control: public, max-age=604800` (7 days) by the generic `.js` rule in `docker/images/files/nginx.conf.template`. Its URL is versioned only by the build (`config.js?version=<APP_BUILD>`), so a flags-only change (edit `PENPOT_FLAGS`, restart) does not change the URL and already-cached browsers keep the pre-change `config.js` for up to 7 days. The flag change is invisible until cache expiry or a manual cache clear — e.g. enabling `enable-login-with-google` does not show the Google button for returning users, since `frontend/src/app/main/ui/auth/login.cljs` gates the SSO buttons on `cf/flags`, which is populated from the cached `config.js`.

config.js is the one shipped asset whose content changes independently of the build (it is copied plain in `docker/images/Dockerfile.frontend`, not content-hashed), so it should not be cached like the content-hashed bundles.

### Change
Add an exact-match `location = /js/config.js` in `docker/images/files/nginx.conf.template` (before the generic `.js|.css|...` rule) that sends `Cache-Control: no-store, no-cache, must-revalidate, max-age=0`, keeping the security-headers include. One file changed, plus a one-line `CHANGES.md` note.

### Verification
Verified the behavior on a running 2.16.2 self-hosted frontend by applying an equivalent exact-match `location = /js/config.js` no-cache override: `nginx -t` passed, and `curl -I /js/config.js` flipped from `public, max-age=604800` to `no-store, no-cache, must-revalidate, max-age=0`, while other `.js` bundles kept `max-age=604800`. After that, a `PENPOT_FLAGS` change was reflected on a normal reload without clearing browser cache.

### Test plan for maintainers
1. Build the frontend image from this branch.
2. `curl -I /js/config.js` → expect `Cache-Control: no-store, no-cache, must-revalidate, max-age=0`.
3. `curl -I /js/<any-bundle>.js` → still `public, max-age=604800`.
4. Change a `PENPOT_FLAGS` value, restart, reload without clearing cache → change is reflected.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -43,6 +43,7 @@
 
 ### :bug: Bugs fixed
 
+- Fix self-hosted Docker serving stale `/js/config.js` for up to 7 days after a `PENPOT_FLAGS`-only change and restart, by exempting it from the generic static-asset `Cache-Control: public, max-age=604800` rule in `nginx.conf.template` [Github #10556](https://github.com/penpot/penpot/issues/10556)
 - Fix LDAP provider params schema typo (`bind-passwor` → `bind-password`) introduced during the `clojure.spec` → `malli` migration; the schema slot now matches the runtime key actually read by `prepare-params` (`:password (:bind-password cfg)`) and `try-connectivity` (`(:bind-password cfg)`), so a wrong type for the password no longer slips through unvalidated
 - Fix `login-with-ldap` silently dropping its error message on the `ldap-not-initialized` restriction (typo `:hide` → `:hint`); the message `"ldap auth provider is not initialized"` now actually surfaces in logs and error responses instead of being discarded into an unread key
 - Fix `get-view-only-bundle` crashing when a share-link viewer encounters a team member whose email lacks `@` (NullPointerException in `obfuscate-email`) or whose domain has no `.` (previously produced a dangling-dot `****@****.`); now the viewer-side obfuscation is nil-safe and omits the trailing dot when the domain has no TLD
--- a/docker/images/files/nginx.conf.template
+++ b/docker/images/files/nginx.conf.template
@@ -163,6 +163,11 @@ http {
         location / {
             include /etc/nginx/overrides/location.d/*.conf;
 
+            location = /js/config.js {
+                include /etc/nginx/nginx-security-headers.conf;
+                add_header Cache-Control "no-store, no-cache, must-revalidate, max-age=0" always;
+            }
+
             location ~* \.(js|css|jpg|png|svg|gif|ttf|woff|woff2|wasm|map)$ {
                 include /etc/nginx/nginx-security-headers.conf;
                 add_header Cache-Control "public, max-age=604800" always; # 7 days


madalenapmelo-kp:

niwinz:
Hey @NextLevelManagementAdvisors 
Please remove the changelog entry.

The change is OK to merge, waiting until changelog entry is removed.

niwinz:
Closing for inactivity, feel free to reopen when ready.

PR #10619, opened 2026-07-09
:bug: Fix overrides lost after switch

### LINKED ISSUE #10588: Overrides lost after switch
### Steps To Reproduce

1. Create a component that has 2 variants (both with a text layer called "label", living in a flex component)
2. Bring in a copy of the component.
3. Change the text of the copy variant
4. Copy paste the copy and switch to another variant, in a way that it retains the text changes
5. Now go to your library file and move the whole component one pixel (the component hasn't changed)
6. Return to your file and see how the first copy that had the override is fine, but the second one, that was switche after the text override happened, has now lost the override.

### Expected behavior

Would like for the override to remain

### Actual behavior

Overrides are lost when a variant is switched after the override happened

### Screenshots or video

https://github.com/user-attachments/assets/dd1b12b6-d85d-4472-915e-03b1a37e99e8

PRO

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context
- **Important**: Test this issue creating the component + variant with & without the new render activated. It behaves different

luisddm:

### Related Ticket

Closes #10588 

### Summary

In `update-attrs-on-switch`, when a component copy's variant is switched and user overrides are "kept" (preserved), the function computes a new :touched set by resolving sync groups of the updated attrs. For text content, this produces `#{:content-group}` but loses the text sub-groups `(:text-content-text, :text-content-attribute, :text-content-structure)` that were in the original shape's `:touched`.

When a subsequent component update triggers `update-attrs → text-change-value`, that function checks for the presence of `:text-content-text` in `:touched`. Since it's missing, it falls through to the :else branch and returns the master's content instead of the user's override.

#### Fix

When `:content-group` is among the updated sync groups, additionally preserve the text sub-groups from the previous shape's `:touched` set in the new :touched.

### DIFF
--- a/common/src/app/common/logic/libraries.cljc
+++ b/common/src/app/common/logic/libraries.cljc
@@ -2345,7 +2345,12 @@
               updated-sync-groups (into #{}
                                         (keep #(ctk/resolve-sync-group (:type previous-shape) %))
                                         updated-attrs)
-              new-touched (set/union (or (:touched current-shape) #{}) updated-sync-groups)
+              text-sub-touched #{:text-content-text :text-content-attribute :text-content-structure}
+              new-touched (set/union (or (:touched current-shape) #{})
+                                     updated-sync-groups
+                                     (when (contains? updated-sync-groups :content-group)
+                                       (set/intersection (or (:touched previous-shape) #{})
+                                                         text-sub-touched)))
               roperations (into [{:type :set-touched :touched new-touched}] roperations)
               uoperations (into (list {:type :set-touched :touched (:touched current-shape)}) uoperations)]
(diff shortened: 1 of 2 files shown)


hirunatan:
See also https://github.com/penpot/penpot/pull/9147, that is when this error was introduced.

luisddm:
Note to self: add a test in `common-tests.logic.variants-switch-test` before merging.

luisddm:
> Note to self: add a test in `common-tests.logic.variants-switch-test` before merging.

Test added 👍🏽

EvaMarco (approved):

PR #10639, opened 2026-07-10
:bug: Fix dashboard user menu submenu not closing on hover out

### LINKED ISSUE #10549: User menu subsections in the dashboard do not close when hovering away from the parent option
### Description
In the dashboard, the floating user menu contains options that expand subsections on hover. When the cursor moves away from the parent option, the subsection remains open instead of closing. This makes the menu feel unresponsive and can leave stale submenus visible while the user navigates elsewhere.

### Environment
- All
- Dashboard / floating user menu

Image

### Steps to reproduce
1. Open the dashboard.
2. Click on the user avatar or name to open the floating user menu.
3. Hover over a menu option that has a subsection — the subsection expands.
4. Move the cursor away from that option (to another option or outside the menu).
5. Observe whether the subsection closes.

### Current result
The subsection remains open after the cursor leaves the parent option. It does not close on mouse out.

### Expected result
The subsection should close automatically when the cursor moves away from the parent option, consistent with standard hover menu behaviour.

AKnassa:
## What

In the dashboard floating user menu, the expandable options (Help & Learning, Community & Contributions, About Penpot) now close their submenu when the cursor moves away from the option, while still keeping it open while the cursor travels from the option into the floating submenu itself.

Closes #10549

## Why

The submenu was only ever *opened* (on `pointerenter`); nothing closed it when the pointer left the option, so a stale submenu stayed visible until the whole menu was dismissed.

## How

Pointer-leave on the expandable options and on the submenus clears the open submenu after a 200 ms grace period, unless the pointer is back over either element by then — the same approach `workspace/context_menu.cljs` already uses for its hover submenus. The grace period matters because the submenu floats beside the menu with a gap the pointer has to cross.

`dropdown-menu*` gained optional `on-pointer-enter` / `on-pointer-leave` props passed through to its `<ul>`; existing callers pass nothing and are unaffected.

Five Playwright regression tests cover: close on hovering a non-expandable sibling, staying open while walking the pointer into the submenu (stepped `mouse.move` — a plain `hover()` teleports and misses this class of bug), close on leaving the menu entirely, switching between expandable options, and opening via keyboard Enter.

## Testing

- `profile-menu.spec.js`: 7/7 (2 pre-existing + 5 new).
- Frontend unit suite: 417 tests, 0 failures. `clj-kondo`, `cljfmt`, `prettier` clean.

### DIFF
--- a/frontend/playwright/ui/specs/profile-menu.spec.js
+++ b/frontend/playwright/ui/specs/profile-menu.spec.js
@@ -28,6 +28,92 @@ test("Navigate to penpot changelog from profile menu", async ({ page }) => {
   );
 });
 
+test("Submenu closes when hovering a menu option without submenu", async ({
+  page,
+}) => {
+  const dashboardPage = new DashboardPage(page);
+  await dashboardPage.goToDashboard();
+
+  await dashboardPage.openProfileMenu();
+  await page.getByText("About Penpot").hover();
+
+  const changelogSubmenuItem = page.getByText("Penpot Changelog");
+  await expect(changelogSubmenuItem).toBeVisible();
+
+  await dashboardPage.userProfileOption.hover();
+  await expect(changelogSubmenuItem).toBeHidden();
+});
+
+test("Submenu stays open while moving the pointer into it", async ({
+  page,
+}) => {
+  const dashboardPage = new DashboardPage(page);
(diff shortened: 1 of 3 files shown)


niwinz:
Looks ok for me.

Thanks :heart:

PR #10645, opened 2026-07-10
:bug: Guard team-container* when team-id is not a uuid

### LINKED ISSUE #10644: Dashboard sends a malformed get-font-variants request when team-id is missing from the URL
### Description

When the dashboard route is loaded without a `:team-id` query parameter (for example `/#/dashboard/recent`), the client fires a `:get-font-variants` RPC with an empty params map `{}`, which the backend rejects with HTTP 400.

### Steps to reproduce

1. Sign in to Penpot.
2. Navigate to `/#/dashboard/recent` (or any dashboard URL without a `:team-id` query param).
3. Open the browser dev tools and inspect the network requests.

### Expected behavior

The dashboard renders without sending a `:get-font-variants` request with empty params.

### Affected versions

main

niwinz:

## What

Fixes a 400 HTTP error from `:get-font-variants` that the client fires when the dashboard route is loaded without a `:team-id` query parameter (e.g. `/#/dashboard/recent`).

## Why

When the URL reached the dashboard without a `:team-id`, `team-container*` was emitting `dtm/initialize-team` with a `nil` team-id, which set `:current-team-id` to `nil` in the application state. The dashboard and workspace initialize events then built `df/fetch-fonts` with a `nil` team-id, producing a `:get-font-variants` RPC with empty params `{}` that the backend rejected with HTTP 400. The `(assert (uuid? team-id) ...)` checks in `dashboard/initialize` and `workspace/initialize-workspace` document the contract, but those asserts are elided in production builds, so the bug only surfaced at runtime.

## How

Guard `team-container*` in `frontend/src/app/main/ui.cljs` so it skips the `dtm/initialize-team` / `dtm/finalize-team` emissions and returns `nil` (no children, no context providers, no React key) when `team-id` is not a uuid. The `with-effect` body and the render are guarded independently, and the cleanup closure captures the same `team-id` as the setup, so transitioning between valid teams still emits the matching initialize/finalize pair. Since `team-container*` is the single wrapper used by both the dashboard and workspace sections in `page*`, one change covers both entry points without touching the event handlers or the RPC layer.

Fixes #10644

### DIFF
--- a/frontend/src/app/main/ui.cljs
+++ b/frontend/src/app/main/ui.cljs
@@ -133,18 +133,21 @@
    ::mf/private true}
   [{:keys [team-id children]}]
   (mf/with-effect [team-id]
-    (st/emit! (dtm/initialize-team team-id))
-    (fn []
-      (st/emit! (dtm/finalize-team team-id))))
+    (when (uuid? team-id)
+      (st/emit! (dtm/initialize-team team-id))
+      (fn []
+        (st/emit! (dtm/finalize-team team-id)))))
 
-  (let [{:keys [permissions] :as team} (mf/deref refs/team)]
-    (when (= team-id (:id team))
-      [:> (mf/provider ctx/current-team-id) {:value team-id}
-       [:> (mf/provider ctx/permissions) {:value permissions}
-        [:> (mf/provider ctx/can-edit?) {:value (:can-edit permissions)}
-         ;; The `:key` is mandatory here because we want to reinitialize
-         ;; all dom tree instead of simple rerender.
-         [:* {:key (str team-id)} children]]]])))
+  (if-not (uuid? team-id)
+    nil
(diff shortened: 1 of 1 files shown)


hirunatan:
With these changes, the error is gone. But the screen goes completely blank because the main dashboard container does not render anything inside. If a user gets here at some moment they can be lost. Perhaps it would be good to show some message explaining what's happening and with a link to the home screen. Or even directly raise a redirect to the root page.

(Also if I remove the `team-id` from the URL without reloading the page, i get a `Rendered fewer hooks than expected` error, because all `mf/provider` hooks in the else part of the if disappear).

niwinz:
The user reaches this because of a race condition, so once race condition is fixed, that should be never a valid state

hirunatan (approved):

PR #10653, opened 2026-07-10
:recycle: Extract GPU-free RenderResources

### LINKED ISSUE #10046: Use WebGL render in exports
### Description

In order to support using WebGL Render through Skia for export rendering, we need to develop a headless path that we can execute from the backend

### Describe the solution you'd like

- When exporting multiple boards to PDF or when exporting multiple images, we should use the WebGL render
- When exporting multiple files, the exporter should handle them using the WebGL render and returning a zip file

Export process is still being defined

elenatorro:
### Related Ticket

https://github.com/penpot/penpot/issues/10046

### Summary

This is a quick refactor for the upcoming multiple pdf pages exports (headless export), in order to separate the changes in smaller PRs.

PDF-related fixes only apply on this PR on single PDF pages (which doesn't go through the exporter, it happens locally).

### Steps to reproduce 

- Check everything renders and works as normal
- Export single frames in PDF format (check emoji, bgblur, etc). Pages should export full-size (frame size), no margins. This needs ` enable-wasm-export` config flag

### Checklist


render-wasm/_build_env
```diff
@@ -41,7 +41,7 @@ export EMCC_CFLAGS="--no-entry \
     -sMAX_WEBGL_VERSION=2 \
     -sEXPORT_NAME=createRustSkiaModule \
     -sEXPORTED_RUNTIME_METHODS=GL,UTF8ToString,stringToUTF8,HEAPU8,HEAP32,HEAPU32,HEAPF32 \
-    -sENVIRONMENT=web \
+    -sENVIRONMENT=web,worker,node \
```
elenatorro, line 44: This would be for the next PR, we can revisit it in there

PR #10656, opened 2026-07-10
:bug: Fix layout padding persisted as string after invalid input

### LINKED ISSUE #10638: Layout padding saved as string after invalid input in multi-selection, causing persistence 500
### Steps To Reproduce

1. Open a file in the workspace editor.
3. Create a frame and add Flex layout.
4. Duplicate the frame
5. Select only one frame.
6. In the Layout section, set Horizontal padding to 33 and press Enter.
7. Without blurring the field Shift+click another frame to multi-select both.
8. In the same field, type an invalid expression (e.g. abc) and press Enter.
10. Wait for autosave

### Expected behavior

- Invalid input is rejected and the field reverts to the previous valid value.
- layout-padding values (:p1–:p4) remain numbers in shape data.
- The file persists without errors.
- Server-side Malli validation passes (::sm/safe-number for padding fields).

### Actual behavior

- The field may appear to revert correctly in the UI.
- Shape data stores padding as strings (e.g. "33" instead of 33).
- Autosave/persistence fails with HTTP 500.
- Server error: invalid shape found '<shape-id>' with Malli validation errors on layout-padding :p2 and :p4 (expected number, got string).
- The canvas may still render correctly because JS coerces numeric strings; the failure only appears on save.

### Screenshots or video

Root cause (suspected):
1. In app.main.ui.ds.controls.numeric-input, a use-effect stores the last value via fmt/format-number, which returns a string in last-value*.
2. On multi-selection, the input value prop becomes nil (mixed values), but last-value* can remain stale (e.g. "49.5").
3. On invalid input, the fallback path calls on-change with that string because (not= fallback-value (str value)) is true when value is nil ((str nil) → "").
4. on-padding-change in layout_container.cljs uses mth/finite, which in ClojureScript accepts numeric strings via js/isFinite but does not coerce them to numbers.
5. The frontend does not validate shapes before persist; the backend does, so the error surfaces only on save.

Relevant files:
1. frontend/src/app/main/ui/ds/controls/numeric_input.cljs (fallback + last-value* effect, ~lines 318–325, 777–788)
2. frontend/src/app/main/ui/workspace/sidebar/options/menus/layout_container.cljs (on-padding-change + mth/finite, ~lines 1204–1217)
3. common/src/app/common/types/shape/layout.cljc (padding schema requires ::sm/safe-number)

Notes:
1. Reproduces with manual UI only; no MCP, plugins, or tokens required.
2. Requires the new numeric input (tokens/numeric-input); the deprecated input does not show this behavior.
3. Same mth/finite pattern exists in on-gap-change and **may affect gap fields or different fields** similarly.

_No response_

_No response_

PRE

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

AKnassa:
## What

Typing an invalid value (e.g. `abc`) into a layout padding or gap field could silently store the value as a **string** instead of a number in the shape data. The canvas kept rendering normally, but the next autosave failed with an HTTP 500 (`::sm/safe-number` Malli validation on the backend), leaving the file unable to persist.

Fixes #10638

## Why

Two defects lined up:

1. The design-system numeric input caches the last committed value in `last-value*` — but a sync effect stored it as a **formatted string** (`fmt/format-number` rounds to two decimals). On invalid input, the fallback re-emitted that cached value through `on-change` whenever it differed from `(str value)`. Two states trigger this without any race: the non-nillable expanded padding inputs on a mixed multi-selection (emit the string `"0"`), and any committed value with more than two decimals, e.g. from a `10/3` expression (emit the rounded string).
2. The padding/gap handlers guarded with `(mth/finite val 0)`, but the CLJS branch of `finite?` intentionally skips the `number?` check, and `js/isFinite` coerces numeric strings — so the string passed straight into `dwsl/update-layout`. The frontend does not validate shapes before persisting; the backend does, so the failure only surfaced on save.

## How

- `numeric-input` now keeps the committed **number** (or `nil`) in `last-value*`; formatted strings exist only in the DOM. The invalid-input fallback restores the display without emitting `on-change` — the shapes still hold their previous values, so there is nothing to commit. This also removes a spurious re-emission on every Enter (number vs string comparison always differed).
- `Esc` now fully discards the typed text (`raw-value*` + dirty flag), so the blur that follows no longer commits the discarded input.
- Token application dedups on the token **name** instead of its resolved value: re-applying the same token stays a no-op (`toggle-token` would unapply it), while applying a token whose value equals the current committed value still emits — previously that case only worked by accident of the string/number mismatch.
- The four padding/gap handlers coerce with `(mth/finite (d/parse-double val) 0)` as defense in depth (numeric strings coerce to their value, garbage/`nil`/`Infinity` fall back to 0).

Covered by Storybook play tests on the component contract (`on-change` never receives strings; Esc revert; mixed-selection unify; token idempotence; precision revert; nillable clear), `parse-value` unit tests, and a Playwright regression that multi-selects two boards, commits invalid text and asserts every `:layout-padding`/`:layout-gap` value in the `update-file` payload is a number (red on the previous code, green now).

Worth flagging for maintainers (out of scope here):

- CLJS `mth/finite?`/`d/num?` accept numeric strings while the JVM branches reject them; fixing it globally is risky — e.g. `app.rasterizer/svg-has-intrinsic-size?` relies on `d/num?` coercing attribute strings.
- Files already corrupted with string paddings have no repair migration (precedent: `0015-fix-text-attrs-blank-strings`).
- Pressing Enter in an untouched mixed nillable padding field emits `nil` and unifies paddings to 0 — pre-existing behavior, unchanged by this PR.


niwinz:
Thanks for you contribution, I'm closing this PR because we have two different PRs fixing the same issue, so I have considered to take the best from both prs and open a single one: https://github.com/penpot/penpot/pull/10758

PR #10657, opened 2026-07-10
Fix internal error when dragging a nested layout with boolean shapes inside

### LINKED ISSUE #10647: Dragging an inner layout + Boolean operations inside throws an internal error
[Codex Icons.zip](https://github.com/user-attachments/files/29879793/Codex.Icons.zip)

### Description
When dragging a frame with a layout (flex or grid) and elements with Boolean operations inside that is nested inside another layout, an internal error is thrown. The interaction causes the application to display an error screen, preventing the drag operation from completing.

### Environment
- PRO
- Wasm 

### Steps to reproduce
1. open the attached file, select an inner element and drag it outside (check video)

### Current result
An internal error is thrown and the application displays an error screen. The drag operation does not complete successfully.

https://github.com/user-attachments/assets/2ebef0ba-1a00-4762-9596-9fdd3a714a7e

Log

[report(5).txt](https://github.com/user-attachments/files/29880769/report.5.txt)

### Expected result
Dragging an inner layout frame should work without errors, repositioning the element correctly within or outside its parent layout.

AKnassa:
### Related Ticket

Fixes #10647

### Summary

Dragging a frame with a layout that contains boolean shapes, when it is nested inside another layout, crashed the workspace with an internal error (`WASM Error (panic)` in `_start_temp_objects`, as in the report attached to the issue).

Two stacked problems caused it:

1. **Stale heap views.** The boolean calculation captured the Emscripten heap view (`HEAPU32`) once and kept using it across calls that can grow WASM memory (allocation, and the calculation itself, which allocates its result). When memory grows, Emscripten replaces the views and the old one goes dead — the child ids were silently written to nowhere and the result was read as garbage. On big files this is near-certain to happen, because `start_temp_objects` deep-cloned the whole shapes pool, doubling memory use on every recalculation.
2. **No cleanup on failure.** When that first failure threw, the temporary shapes pool was never discarded, so *every* later boolean recalculation panicked in `_start_temp_objects` — the user-visible error screen.

The main fix makes `calculate-bool` re-fetch heap views after every allocating call and always discard the temporary pool (and the staging buffer), even when the calculation fails, without masking the original error.

Follow-up hardening in the same area (driven by an adversarial edge-case audit of the fix):

- Bool operands are normalized like the CLJS fallback (`calc-bool-content*`): missing ids, hidden children and svg-raw children no longer participate. Previously a missing base child silently emptied the whole bool, and a hidden child still shaped the geometry.
- Bools with no effective operands short-circuit to empty content (no temp-pool churn, no zero-byte allocation, no `(rseq nil)` crash in release builds).
- While the WASM module is still loading at boot, the calculation transparently falls back to the CLJS implementation instead of crashing on the empty module.
- The same stale-heap-view pattern was fixed in the other serializers on this code path: `set-shape-path-content`, `set-shape-strokes` (which also silently leaked a staging buffer for strokes without fill), `propagate-modifiers`, `get-selection-rect`, `calculate-position-data`, `set-shape-children`, `set-focus-mode` and `set-render-include-filter!`.
- Rust: `State::start/end_temp_objects` now move the shapes pool instead of deep-cloning it (removes the memory doubling that triggered growth, and the unrecoverable state a failed mid-clone allocation produced). Lifecycle unit tests added.
- Nested bools are now recalculated children-first after a drag, so an outer bool reads the freshly recomputed content of the bool inside it instead of stale content (the WASM path uses stored content for nested bools; the ids were previously an unordered set).

### Steps to reproduce

1. Open the file attached to #10647 (Codex Icons) with the WASM renderer enabled.
2. Select an inner layout frame that contains boolean shapes and drag it outside its parent layout.
3. Before: internal error screen (`_start_temp_objects` panic). After: the drag completes normally.

Verification: 25 new Node unit tests (fake WASM module that faithfully simulates Emscripten heap growth and the Rust staging-buffer/temp-pool state machines) — each fix was proven with a failing test first, and the buffer-release assertions were mutation-tested. Full frontend suite: 448 tests / 1836 assertions / 0 failures. Note: the Rust unit tests could not be run locally (the native skia build was not possible on this machine) — they rely on CI.

While auditing this, a pre-existing WASM/CLJS divergence was also found (text and frame children produce different boolean geometry in each engine — glyph outlines/frame rect vs bounding box/children-only). It predates this bug and needs a decision on which behavior is canonical, so it is not touched here; happy to file it as a separate issue.

### Checklist

### DIFF
--- a/frontend/src/app/main/data/workspace/modifiers.cljs
+++ b/frontend/src/app/main/data/workspace/modifiers.cljs
@@ -715,6 +715,24 @@
             (rx/of (set-temporary-selrect selrect)
                    (set-temporary-modifiers modifiers))))))))
 
+(defn bool-ids-children-first
+  "The bool ancestors of `ids`, ordered children before parents.
+
+   The order matters for the WASM path: it calculates a bool from the
+   STORED content of nested bool children (the CLJS fallback recomputes
+   them recursively), and update-shapes folds each update into the
+   objects seen by the next one — so a nested bool must be recomputed
+   before the outer bool reads its content (see issue #10647)."
+  [objects ids]
+  (->> ids
+       (into #{}
+             (comp
+              (mapcat (partial cfh/get-parents-with-self objects))
+              (filter cfh/bool-shape?)
+              (map :id)))
+       (sort-by #(- (count (cfh/get-parent-ids objects %))))
+       (vec)))
+
 (defn propagate-structure-modifiers
   [modif-tree objects]
   (letfn [(propagate-children
@@ -819,12 +837,7 @@
                       (ctm/apply-structure-modifiers modifiers))))
 
               bool-ids
-              (into #{}
-                    (comp
-                     (mapcat (partial cfh/get-parents-with-self objects))
-                     (filter cfh/bool-shape?)
-                     (map :id))
-                    ids)
+              (bool-ids-children-first objects ids)
 
               undo-id (js/Symbol)]
           (rx/concat
--- a/frontend/src/app/render_wasm/api.cljs
+++ b/frontend/src/app/render_wasm/api.cljs
@@ -2467,9 +2467,13 @@
 
 (defn calculate-bool*
   [bool-type ids]
+  ;; NOTE: heap views are (re)fetched after every allocating call:
+  ;; both _alloc_bytes and _calculate_bool (which allocates the result
+  ;; buffer) can grow WASM memory, and growth detaches any previously
+  ;; captured view (see issue #10647).
   (let [size   (mem/get-alloc-size ids UUID-U8-SIZE)
-        heap   (mem/get-heap-u32)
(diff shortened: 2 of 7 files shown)


madalenapmelo-kp:

elenatorro:
Hi, @AKnassa. Thanks for your contribution, but we've already partially solved it in here https://github.com/penpot/penpot/pull/10697 and we're changing part of the related code.

Since the WebGL render is still in beta, we recommend small contributions if possible instead of PRs with many changes at the moment. Once we fully release it we'll write better contribution guidelines. As a rule of thumb: keep contributions to the new render small and precise. Well label the issues that could be good for external contribution.

Also, I'm assuming that most of the PR is AI-developed. AI changes tend to be huge and verbose, so we also recommend reviewing changes carefully before submitting this kind of contributions.

PR #10659, opened 2026-07-12
:bug: Reject invalid formulas on numeric inputs

### LINKED ISSUE #9581: Invalid formulas are accepted on numeric inputs
https://tree.taiga.io/project/penpot/issue/14131

#### Issue

Inputs shouldn’t accept an invalid formula or assume any operator preference.

#### How to reproduce

1.  Create a 100x100px rectangle
2.  Click on the width field
3.  Fill it with “10+\*3”
4.  Press Enter

#### Expected behavior

The invalid value is not saved, retaining the same previous width

#### Actual behavior

A value is saved (310)

#### Screenshots or video

Not required

#### Environment

*   It happens on PRE/PRO

Linux Mint 22.1 (xia) (Ubuntu 24.04 noble)    
Chrome Version 147.0.7727.137 (Official Build) (64-bit)

Design: 
External source: 
External source 2:

AKnassa:
## What this does
Stops the width/height/rotation inputs from silently accepting a nonsense formula.

## Why
Typing `10+*3` into a numeric input saved **310**. The parser quietly ignored the stray operator and picked one for you. The value you got was not the value you typed, and nothing told you so.

## What changed
- `+`, `*` and `/` are no longer accepted where a number should be. `10+*3` is now rejected and the field keeps its original value, which is what the issue asks for.
- Negative numbers are still valid operands. `10 + -3` still evaluates to 7, `10 - -5` to 15, `* -3` still works as a relative operator.

## How to see it
Draw a 100×100 rectangle, type `10+*3` into the width field, press Enter. Previously the width became 310. Now the input is rejected and the width stays 100.

## Testing
New tests pin both halves: invalid operator placement is rejected, and every form of negative operand still evaluates correctly.
Full frontend suite: **433 tests, 0 failures.**

## Note for reviewers
An invalid formula reverts the field silently, with no visible message. That is the existing behaviour for all invalid input and matches the issue's stated expectation ("keeping the original width value"). Adding a visible error affordance would be a product decision and is not attempted here.

Fixes #9581


hirunatan:
Merged. Good work! Thanks very much.

PR #10660, opened 2026-07-12
:bug: Repair variant referential integrity errors

### LINKED ISSUE #10606: Integrity errors related to variants are not repaired
### Steps To Reproduce

- Have a file with referential integrity errors in variants, like the ones described in https://github.com/penpot/penpot/issues/10572
- Execute the `debug.repair()` script in devtools

### Expected behavior

The error should be repaired and the file restored to a consistent state (although possibly some information has been lost).

### Actual behavior

The error is not repaired and the file is still corrupt.

### Additional context

These are the validation codes not repaired:

Code | What it checks | Auto-repair
-- | -- | --
:not-a-variant | Variant-container child is not a variant | No
:invalid-variant-id | variant-id ≠ container id | No
:invalid-variant-properties | Different properties across variants | No
:variant-not-main | Variant is not a main instance | No
:parent-not-variant | Parent is not a variant-container | No
:variant-bad-name | Variant :name ≠ container name | No
:variant-bad-variant-name | :variant-name ≠ derived from properties | No
:variant-component-bad-name | Container name ≠ component merge-path(path, name) | No
:variant-component-bad-id | Component variant-id ≠ shape variant-id | No

AKnassa:
## What this does
Makes variant referential integrity errors actually repairable, instead of silently doing nothing.

## Why
The repair methods for variant integrity errors were empty stubs. Running a repair on a file with variant errors reported success and changed nothing, so affected files stayed broken with no way back.

## What changed
Implements the repairs that can be derived unambiguously from the file itself:
- a shape whose name has drifted from its container's name
- a variant name that no longer matches its properties
- a component name/path that has drifted from its container
- a child carrying the wrong variant id

The repair is idempotent and converges: running it twice changes nothing the second time, and it never rewrites the user's container name.

## How to see it
Take a file with variant integrity errors and run the repair. Previously a no-op; now the derivable errors are fixed and the file validates.

## Testing
New tests in `common/test/common_tests/files/repair_variants_test.cljs`.
Common suite: **1083 tests, 0 failures.**

## Note for reviewers — scope
Four error codes are deliberately **left unrepaired**: `:not-a-variant`, `:parent-not-variant`, `:variant-not-main`, `:invalid-variant-properties`. Repairing those means choosing between destructive options (move the shape out, convert it, or strip its variant-ness), and picking wrong destroys user data. That is a product decision, not something to guess at — happy to implement them once you say which behaviour you want.

Fixes #10606 (partially — see scope note)


hirunatan:
Sorry, I was already working on a solution to this issue https://github.com/penpot/penpot/pull/10768.

I find that your use of the helper function `get-variant-container` is not necessary, because the validation functions have more context and can give the exact value to repair in the error args. This way most errors may be repaired without ambiguity.

And also I have make all errors repairable, even when there is some loss of data. It's better to have some shapes deleted than having the whole file blocked without possibility of using it.

I have added some ideas of you, though, to the other PR, so I've mentioned you there.

Thanks anyway for the contribution, please keep doing it. Its much appreciated.

PR #10661, opened 2026-07-12
:bug: Do not apply the new-password policy to the old-password field

### LINKED ISSUE #10626: Changing password returns "Password should be at least 8 characters" error on the old password field
### Steps To Reproduce
1. Create an account with the CLI tools with short temporary password
2. Change the password in the new account in the user settings
3. Fail to change the password

### Expected behavior

User should be able to change the password whatever the length of the old password is.

### Actual behavior

User can't change the password

### Screenshots or video

Image

- Windows 11
- Chrome

_No response_

- Debian
- Docker version 29.5.3, build d1c06ef
- 2.16.0

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

AKnassa:
## What this does
Lets you change your password even when your current password is shorter than 8 characters.

## Why
The form applied the **new** password rules to your **existing** password. If your current password predates the 8-character policy (for example an account created via `manage.sh`), you were told it was "too short" — on a field you are not allowed to change. There was no way out of the form.

The frontend was the only place enforcing this. The backend accepts any non-empty string up to 500 chars for the old password, and the login form only requires it to be present.

## What changed
- The old-password field now only has to be filled in, capped at 500 characters to match the backend.
- The 8-character policy still applies to the new password being set.

## How to see it
Sign in with an account whose password is under 8 characters, go to **Settings → Password**, and change it. Previously blocked, now works.

## Testing
9 new tests in `frontend/test/frontend_tests/ui/settings_password_schema_test.cljs`, covering the bug, boundaries, unicode, and confirming the 8-char rule still guards the new password.
Full frontend suite: **438 tests, 0 failures.**

## Note for reviewers
Social/OAuth-registered users have no old password at all, and the backend already allows it to be absent. The field is still *required* on the frontend. That looks like a separate defect, so it is left alone here rather than silently loosened.

Fixes #10626


niwinz:
Looks ok. thanks :heart: 
Waiting for CI to merge

PR #10662, opened 2026-07-12
:bug: Guard typography tokens against circular references

### LINKED ISSUE #9580: Internal error on Typo tokens due to circular dependencies on other token types
https://tree.taiga.io/project/penpot/issue/12488

#### Issue

There’s an uncontrolled circular dependency allowed between typography tokens and other tokens that provide, in theory, valid referenced values.

It happens when you reference from the typography token fields other valid tokens (number token, font family token, font weight token, etc), **filling just one field at a time**, and then, from the referenced tokens, reference the typo tokens as well.

#### How to reproduce

1.  Create a Font Family token with a valid font value
2.  Create a Typography token referencing the previous on its Font Family field (fill just this field)
3.  Edit the Font Family token and reference in its value the Typography token (forming the circular dependency)
4.  Press the “Save” button

#### Expected behavior

The same as what happens with other font tokens that can reference other valid basic tokens.    
Show a message on step 3, when a circular resolved dependency is detected: ”Missing token references: number“.

Actual behavior

A warning is displayed in the resolved value: “Resolved value: RangeError: Maximum call stack size exceeded”

An Internal error happens if you press the “Saved” button, with the following [error log file trace](https://media-protected.taiga.io/attachments/9/f/d/c/a07229cdd6ecac6471f78d6db7a9fd7a045be528f404ed0acef8895acb40/report-66.txt?token=aQnL2Q%3AX7GmQbmP6DJ-mr88K6BhtkFo1SuM6lfj57UPiqnqs_dWjqrp3yZ_LQi3r8pqckxxWT2q_YQtdDvhD3yyAp8dXw#_taiga-refresh=issue:4644578):

_Hint: Maximum call stack size exceeded_    
_Prof ID: c6b12730-702d-11eb-b327-eb0147e048f5_    
_Team ID: 5daeec9f-0eaf-80c6-8006-99b799b8eff4_

_Data:_    
_{:hint "Maximum call stack size exceeded"}_

_Trace:_    
_RangeError: Maximum call stack size exceeded_

#### Environment

*   It happens on PRE and PRO

Linux Mint 22.1 (xia) (Ubuntu 24.04 noble)    
Google Chrome Version 141.0.7390.107 (Official Build) (64-bit)

#### URL

Not required
https://github.com/tokens-studio/penpot/issues/155

Design: 
External source: 
External source 2:

AKnassa:
## What this does
Replaces an internal error with a proper validation message when typography tokens reference each other in a circle.

## Why
Creating a circular reference between token types (a typography token pointing at a token that points back) raised an internal error. The user saw a crash instead of being told what was wrong with their token.

## What changed
- Circular references are detected across nested token values, not just top-level ones, and reported as a validation error on the form.
- Reference-finding now recurses into maps and sequences, so composite tokens (typography, shadow) are covered rather than only scalar ones.

## How to see it
Create a typography token whose value references a token that references it back. Previously an internal error; now a validation message on the field.

## Testing
New tests covering circular detection through nested/composite token values.
Full frontend suite: **439 tests, 0 failures.** Common suite: **1069 tests, 0 failures.**

## Note for reviewers
The message surfaced is the existing `errors.tokens.circular-reference`. The issue's text suggests a different wording ("Missing token references: number"); I kept the existing, accurate message rather than inventing new copy. Happy to change the wording if you prefer.

Fixes #9580


niwinz:
Looks nice from a quick view. Thanks for the contribution.
For now i leave it unsigned, ready to be reviewed by someone more aware of tokens.

EvaMarco:
Thanks for the update! Overall, the implementation looks good, but there's one issue that needs to be addressed.

Your changes modify the behavior of the error handling. When a token references itself, the error should be reported as a **self reference**, not as a **circular reference**. This change is causing a large number of tests to fail.
Screenshot from 2026-07-30 09-32-50


You can verify this by running the Playwright tests locally. They should make it clear which cases are affected and help confirm that the expected behavior is preserved.

EvaMarco (changes requested):
Do not mix the self reference error with the circular reference error.

PR #10663, opened 2026-07-12
:bug: Fix clipboard crash when copying as SVG

### LINKED ISSUE #10596: Uncaught DOMException: 'image/svg+xml' not supported for write (clipboard write crash)
### Steps To Reproduce

Exact trigger not yet pinned down - this surfaced in the browser console during a normal editing session (self-hosted Penpot, MCP plugin active, several boards with shapes/text/rectangles created and edited over ~15 minutes) without a specific reproduction sequence captured. Reporting with the evidence available since it's an uncaught exception that visibly breaks the current operation:

```
Uncaught Rejection: Type 'image/svg+xml' not supported for write
Uncaught (in promise) DOMException: Type 'image/svg+xml' not supported for write
```

The error is a native `DOMException` from the browser's async Clipboard API (`navigator.clipboard.write()` with a `ClipboardItem`). Most browsers only support a small allowlist of MIME types for clipboard writes (typically `text/plain`, `text/html`, and raster image types like `image/png`) - `image/svg+xml` is rejected outright by the browser, not by Penpot. This strongly suggests a code path in Penpot (likely a "copy as image"/"copy to clipboard" action on a shape or export) is constructing a `ClipboardItem` with an SVG MIME type without checking browser support or catching the rejection, causing an uncaught promise rejection.

### Expected behavior

Either don't offer SVG as a clipboard-write format (since the Clipboard API can't support it), or catch the rejection and fall back to a supported format (e.g. rasterize to PNG first, matching what browsers actually allow to be copied to the clipboard).

### Actual behavior

An uncaught `DOMException` is thrown, visible in the console as an "Uncaught Rejection" - whatever UI action triggered it fails silently from the user's perspective (no visible error toast was reported), which reads as the app being unresponsive/crashed.

### Environment

Self-hosted Penpot 2.16.2, MCP version 2.16.0 (per the frontend's own console version log line).

### Additional context

If it helps narrow down the trigger: this was captured in a session that had just finished a long sequence of MCP-driven `executeCode` tasks creating/positioning shapes and exporting several shapes as PNG via `penpotUtils.exportImage(shape, "shape", false)` - those PNG exports all completed successfully beforehand. The SVG-specific failure happened afterward, several minutes after the last MCP task, most likely from a manual UI action (e.g. right-click "Copy" on a shape, or a keyboard-shortcut copy) rather than anything MCP-triggered.

AKnassa:
## What this does
Stops "Copy as SVG" from throwing an uncaught error in browsers that will not put SVG on the clipboard.

## Why
Copying a shape as SVG called `clipboard.write()` with an `image/svg+xml` payload. Browsers that do not support writing that MIME type reject the promise with a `DOMException`, and nothing caught it — so it surfaced as an uncaught error and the copy just failed.

## What changed
- When `clipboard.write()` rejects specifically because the SVG type is unsupported, fall back to copying the SVG markup as plain text, which every browser accepts.
- The fallback is matched to that specific failure rather than swallowing every error, so genuine failures (for example a denied clipboard permission) still surface instead of being hidden.

## How to see it
In a browser that does not support SVG clipboard writes, select a shape and use **Copy as SVG**. Previously it threw; now the SVG lands on the clipboard as text.

## Testing
New tests in `frontend/test/frontend_tests/util_clipboard_test.cljs` cover the unsupported-type path, the fallback, and that unrelated errors are not swallowed.
Full frontend suite: **437 tests, 0 failures.**

Fixes #10596


niwinz:
Looks nice. I will test it and merge if not issues.
Thanks

frontend/test/frontend_tests/util_clipboard_test.cljs
```diff
@@ -0,0 +1,205 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC
```
niwinz: ;; Copyright (c) KALEIDOS INC Sucursal en España SL

AKnassa: Done 👍. fixed in 3897f88064.

niwinz:
Ready to merge, lets wait CI.
Thanks :heart:

PR #10664, opened 2026-07-12
:sparkles: Expand the token tree after importing it

### LINKED ISSUE #9819: Display token tree expanded right after importing it
https://tree.taiga.io/project/penpot/issue/14249

Currently, after importing tokens, the token tree appears collapsed.
It should appear expanded so the user can check that everything has been imported correctly without needing to click on every token type to check them.

Design:  
External source: 
External source 2:

AKnassa:
## What this does
Shows the tokens you just imported, instead of leaving them collapsed.

## Why
After importing a token set, the token tree stayed folded shut. You imported tokens and appeared to get nothing — you had to go hunting and expand each type by hand to confirm the import worked at all.

## What changed
- Token types in a newly imported set start expanded, so the imported tokens are visible straight away.

## How to see it
Import a token set (JSON). The token types now appear expanded with their tokens visible, rather than collapsed.

## Testing
New tests in `frontend/test/frontend_tests/tokens/logic/token_actions_test.cljs`.
Full frontend suite: **434 tests, 0 failures.**

## Note for reviewers
The issue's Design field is empty, so "which sets should expand" is a genuine product decision. This expands the token types of every imported set. If you would rather only expand the active set, or only the first, say so and it is a small change.

Fixes #9819


niwinz:
This also, looks ok on quick view, but i prever to have a review from a tokens team.
Leaving unasigned for now marked ready to review.

Thanks for the contribution

PR #10666, opened 2026-07-12
:bug: Fix radial gradient handles on rotated ellipses

### LINKED ISSUE #10069: Radial gradient handles blow out in size when rotated on ellipses
### Steps To Reproduce

- Draw an ellipse (only happens with ellipses)
- Apply a Radial Gradient
- Rotate gradient

### Expected behavior

The handles rotate respecting original sizes.

### Actual behavior

The handle size increases like crazy (it's related to the ellipse ratio).

### Screenshots or video

https://github.com/user-attachments/assets/7c7d12fe-24c0-40a6-8b73-fa9f848cd32a

AKnassa:
## What this does
Fixes the radial gradient handles ballooning in size when you rotate an ellipse.

## Why
The gradient handle positions were composed against the shape's rotation incorrectly, so once an ellipse was rotated the handles were drawn far larger than the gradient they control — making them awkward to grab and misleading about where the gradient actually is.

## What changed
- The handle geometry now composes the shape transform correctly, so handles track the gradient at any rotation.
- The geometry was pulled out into a pure function so it can be tested directly.

## How to see it
Draw an ellipse, apply a radial gradient, rotate it. The handles previously blew out in size; now they stay on the gradient.

## Testing
New tests in `frontend/test/frontend_tests/ui/gradient_handlers_test.cljs`.
Full frontend suite: **431 tests, 0 failures.**

## Note for reviewers — please read
This fixes the **SVG renderer only**. `render-wasm/src/shapes/fills.rs` (`to_radial_shader`) contains the same wrong composition, so with the WebGL renderer enabled the handle will still be wrong. I could not build or test the Rust side on my machine, so I did not want to submit an unverified change to it. Happy to follow up, or for someone with a wasm toolchain to take that half.

Fixes #10069 (partially — see note above)


belen-albeza:
Hi @AKnassa thanks for the PR!

>  I could not build or test the Rust side on my machine

What is your machine? (OS, cpu…)

PR #10667, opened 2026-07-12
:sparkles: Make plugin API validation errors precise and crash-safe

### LINKED ISSUE #10072: Plugin API validation errors are too generic to diagnose the failure
### Steps To Reproduce

1. Trigger different validation failures through the Plugin API. Examples observed during verification:
   - Create a reference token whose referenced set is inactive.
   - Create a numeric token with a number value instead of a string value.
2. Compare the returned error messages.

### Expected behavior

Validation errors should include the failing API call or context, the offending field, the received value and type, the expected value and type, and any relevant invariant.

Example:

```text
addToken: "value" must be a string; got number (16). Use "16".
```

For reference resolution:

```text
addToken: reference "{color.gray.50}" cannot be resolved because token set "primitives" is inactive.
```

### Actual behavior

Distinct failures collapse into near-identical generic strings such as `Value not valid`, which does not provide enough information to decide whether the token, property, value, reference, or set state is wrong.

### Screenshots or video

_No response_

_No response_

_No response_

- MCP server: Penpot remote MCP.
- Tools used: `high_level_overview`, `penpot_api_info`, `execute_code`, `export_shape`.
- Method: read-only API documentation review followed by live smoke tests against a connected plugin instance.
- Scope: Penpot MCP server and Penpot Plugin API behavior observed while building and validating an AI-agent-driven kit.

### Frontend Stack Trace

```console
[PENPOT PLUGIN] Value not valid. Code: :error
[PENPOT PLUGIN] Value not valid: Field value is invalid: Invalid data. Code: :error
```

### Backend Stack Trace

```console
No backend stack trace captured.
```

### Additional context

This is a cross-cutting reliability issue. LLM-driven agents can recover from specific errors, but generic messages cause random retries and incorrect workarounds.

Improving error messages is likely the cheapest high-impact fix because it improves debuggability for every other API limitation.

AKnassa:
## What this does
Makes plugin API validation errors say what actually went wrong, instead of "Invalid data".

## Why
When a plugin passed a bad argument, the API returned a generic message. Plugin authors had no way to tell *which* field was wrong, what was expected, or what they actually sent — so diagnosing a failing call meant guessing.

## What changed
- A failing field now reports the field, the schema it was expected to satisfy, and the value it received (with its type).
- The value is rendered by a bounded, cycle-safe renderer. This matters: plugin arguments arrive as native JS objects, and naively printing a self-referencing object would recurse forever — crashing inside the error handler itself. Output is depth- and length-capped, and unicode is truncated on character boundaries rather than splitting surrogate pairs.

## How to see it
From a plugin, call an API method with a wrong-typed argument. Previously "Invalid data"; now a message naming the field, the expectation and the received value.

## Testing
New tests in `frontend/test/frontend_tests/plugins/utils_test.cljs`, including a self-referencing object and unicode truncation.
Full frontend suite: **450 tests, 0 failures.**

## Note for reviewers — scope
This is deliberately scoped to **schema-validation** messages. It does not add the failing method name as a prefix (`addToken: ...`), and it does not improve the domain-specific token-resolution message in the issue's second example — those route through a different path and improving them well needs a product call on wording. Compound schemas still render as raw EDN, which is truthful but not pretty. Happy to take direction on any of these.

Fixes #10072 (partially — see scope note)

### DIFF
--- a/frontend/src/app/plugins/utils.cljs
+++ b/frontend/src/app/plugins/utils.cljs
@@ -328,10 +328,217 @@
           [[(str/join "." path) (:message v)]])))
     m)))
 
+(def ^:private max-repr-length 100)
+(def ^:private max-repr-depth 2)
+(def ^:private max-repr-items 5)
+
+(defn- abbreviate
+  "Shorten `s` to `max-repr-length` code points. Cutting on a UTF-16 code
+  unit would split surrogate pairs and render astral characters as mojibake."
+  [s]
+  (if (> (count s) max-repr-length)
+    (let [points (js/Array.from s)]
+      (if (> (alength points) max-repr-length)
+        (dm/str (.join (.slice points 0 max-repr-length) "") "…")
+        s))
+    s))
+
+(defn- value->type
+  [value]
+  (cond
+    (string? value)             "string"
+    (boolean? value)            "boolean"
+    (number? value)             "number"
+    (keyword? value)            "keyword"
+    (map? value)                "object"
+    (coll? value)               "array"
+    (array? value)              "array"
+    (fn? value)                 "function"
+    (instance? js/Object value) "object"
+    :else                       "unknown"))
+
+(defn- data-property
+  "Own property `k` of the JS object `o`, or `::skip` when `k` is an accessor.
+  Plugin proxies expose their contents through getters that read the
+  application state and may throw, so the error path must not run them."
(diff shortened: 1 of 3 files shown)


Alotor (approved):

Alotor:
Thanks for you contribution.

PR #10670, opened 2026-07-13
:bug: Fix several issues in RPC command handlers

### LINKED ISSUE #10669: Add missing validation checks to RPC command handlers
### Description

Several RPC command handlers lacked explicit validation checks:

- `link-file-to-library` had no circular-reference detection, allowing transitive library cycles (A→B→C→A)
- `search-files` relied entirely on SQL for authorization scoping with no explicit team permission check
- `search-files` had no max length constraint on `:search-term`
- `get-file` ETag did not include `:deleted-at`, risking stale cached responses for soft-deleted files
- `create-file-thumbnail` performed storage I/O inside the DB transaction

### Affected modules

backend/src/app/rpc/commands/

### Affected versions

2.17.0+

niwinz:

## What

Fixes several validation gaps and missing guards in RPC command handlers found during a security audit of the backend RPC layer.

## Why

- `link-file-to-library` had no circular-reference detection, allowing transitive library cycles
- `search-files` relied entirely on SQL JOINs for authorization with no explicit permission check, and had no max-length constraint on the search term
- `get-file` ETag didn't include `:deleted-at`, risking stale cached responses for soft-deleted files
- `create-file-thumbnail` performed storage I/O (`sto/put-object!`) inside the DB transaction

## How

- Added `bfc/get-libraries` check before linking to detect transitive cycles
- Added `teams/check-read-permissions!` call and 250-char max on `:search-term`
- Added `:deleted-at` to the `get-file-etag` hash
- Moved media validation, file lookup, and `sto/put-object!` outside `db/tx-run!`, following the existing Pattern A from `create-file-object-thumbnail!`

Closes #10669

### DIFF
--- a/backend/src/app/rpc/commands/files.clj
+++ b/backend/src/app/rpc/commands/files.clj
@@ -156,11 +156,13 @@
     (assoc mfile :permissions perms)))
 
 (defn get-file-etag
-  [{:keys [::rpc/profile-id]} {:keys [modified-at revn vern permissions]}]
+  [{:keys [::rpc/profile-id]} {:keys [modified-at revn vern deleted-at permissions]}]
   (str profile-id "/" revn "/" vern "/" (hash fmg/available-migrations) "/"
        (ct/format-inst modified-at :iso)
        "/"
-       (uri/map->query-string permissions)))
+       (uri/map->query-string permissions)
+       "/"
+       (some-> deleted-at (ct/format-inst :iso))))
 
 (sv/defmethod ::get-file
   "Retrieve a file by its ID. Only authenticated users."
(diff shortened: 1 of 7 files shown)


backend/src/app/rpc/commands/files_thumbnails.clj
```diff
@@ -448,12 +393,57 @@
...
+  (let [storage (::sto/storage cfg)
+        file    (bfc/get-file cfg file-id :include-deleted? true :load-data? false)
+        props   (db/tjson (or (:props params) {}))
+        {:keys [path mtype]} (:media params)
+        hash    (sto/calculate-hash path)
+        data    (-> (sto/content path)
+                    (sto/wrap-with-hash hash))
+        tnow    (ct/now)
+
+        media   (sto/put-object! storage
```
Alotor: in the transaction there is a check if the connection is read only that is not made for this put-object that you got out of the transaction, is that ok?

backend/src/app/rpc/commands/files.clj
```diff
@@ -1100,6 +1102,12 @@
               :code :invalid-library
               :hint "A file cannot be linked to itself"))
 
+  (let [transitive-deps (bfc/get-libraries cfg [library-id])]
+    (when (contains? transitive-deps file-id)
```
Alotor, line 1109: i think it's better to check for transitive dependencies **after** the permissions.

niwinz: ack

backend/src/app/rpc/commands/files_thumbnails.clj
niwinz: ack

PR #10677, opened 2026-07-13
:bug: Fix inconsistent theme selection (Penpot itself vs. plugins)

### LINKED ISSUE #10676: Inconsistent theme handling: Penpot vs. plugins
Penpot itself and Penpot plugins use different default theme selection strategies if no theme is explicitly selected in Penpot by the user:
* Penpot itself defaults to dark.
* Plugins default to whatever is configured in the user's system/browser, which could be light.

opcode81:
**Note:** This PR was created with AI assistance as part of the Penpot MCP self-improvement initiative.

## What

Makes the theme reported to plugins match the theme Penpot actually
displays. Previously, when no theme was explicitly selected in the user's
profile, Penpot itself rendered dark while every plugin was told the theme
was light (on systems whose browser preference is light) — so plugin panels
opened light inside a dark Penpot, and stayed that way.

Fixes #10676

## Why

Penpot itself and the plugin runtime used different default theme selection
strategies when no theme is explicitly selected by the user:

- Penpot itself defaults to **dark**.
- The plugin runtime (both `penpot.theme` and the `themechange` event)
  defaulted to the **system/browser preference**, which can be light.

For every explicit choice — light, dark, system, and the legacy "default" —
the two resolutions already agreed; only the unset case diverged. That is
why the bug hides on any profile that has ever picked a theme, and shows up
on fresh profiles. Plugins cannot work around it: the wrong value is
reported both at open time and on every theme change, so a plugin following
the documented pattern (as all bundled plugins do) has no correct signal
available.

## How

Extract the app's own resolution into a single shared function,
`app.util.theme/resolve-theme` ("system" follows the system theme;
"default" and unset mean dark), and use it everywhere a theme is resolved:

- the app's own `use-initialize` (behaviour unchanged — this is the
  resolution it already implemented inline),
- the plugin runtime's `getTheme` (`penpot.theme`),
- the plugin runtime's `themechange` handling, whose now-dead
  default-to-dark remapping is removed.

With one source of truth, the app and the plugin runtime can no longer
drift apart.

Verified manually in the workspace: with no explicit profile theme, plugins
now open dark to match Penpot, and toggling the profile theme switches the
plugin panel live via `themechange`.

### DIFF
--- a/frontend/src/app/plugins/events.cljs
+++ b/frontend/src/app/plugins/events.cljs
@@ -53,16 +53,8 @@
 
 (defn- get-theme
   [state]
-  (let [theme (get-in state [:profile :theme])]
-    (cond
-      (or (not theme) (= theme "system"))
-      (theme/get-system-theme)
-
-      (= theme "default")
-      "dark"
-
-      :else
-      theme)))
+  (theme/resolve-theme (get-in state [:profile :theme])
+                       (theme/get-system-theme)))
 
 (defmethod handle-state-change "themechange"
   [_ _ old-val new-val _]
@@ -70,9 +62,7 @@
         new-theme (get-theme new-val)]
     (if (identical? old-theme new-theme)
       ::not-changed
-      (if (= new-theme "default")
-        "dark"
-        new-theme))))
+      new-theme)))
 
 (defmethod handle-state-change "shapechange"
   [_ plugin-id old-val new-val props]
--- a/frontend/src/app/plugins/api.cljs
+++ b/frontend/src/app/plugins/api.cljs
@@ -237,15 +237,7 @@
     :getTheme
     (fn []
       (let [theme (get-in @st/state [:profile :theme])]
-        (cond
-          (or (not theme) (= theme "system"))
-          (theme/get-system-theme)
-
-          (= theme "default")
-          "dark"
-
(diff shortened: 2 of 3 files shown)


frontend/src/app/plugins/events.cljs
```diff
@@ -53,26 +53,16 @@
...
+  (theme/resolve-theme (get-in state [:profile :theme])
+                       (theme/get-system-theme)))
 
 (defmethod handle-state-change "themechange"
   [_ _ old-val new-val _]
   (let [old-theme (get-theme old-val)
         new-theme (get-theme new-val)]
     (if (identical? old-theme new-theme)
       ::not-changed
-      (if (= new-theme "default")
```
Alotor, line 73: This was alright, we don't want the "default" value to arrive to the user (this will be returned in the event) that's why we were changing default to "dark". Maybe here we can also call to `theme/resolve-theme`

opcode81: Since `old-theme` and `new-theme` are already resolved through `get-theme`, which now calls `resolve-theme`, this change should be fine.

Alotor: i see, ok

PR #10678, opened 2026-07-13
:sparkles: Improve workspace debug sidebar workflow

pabloalba:
### Summary

This PR improves the workspace debugging workflow by:
* Moving shape debug information into a dedicated Debug tab in the right sidebar
* Adding a quick action to copy a shape id
* Making the shape-panel, show-ids and show-touched debug toggles update the UI without reloading the page.
* Saving the debug options on local storage, so they are preserved on page reload

### Checklist

### DIFF
--- a/frontend/src/app/main/data/workspace/clipboard.cljs
+++ b/frontend/src/app/main/data/workspace/clipboard.cljs
@@ -1138,6 +1138,13 @@
     (watch [_ _ _]
       (clipboard/to-clipboard (rt/get-current-href)))))
 
+(defn copy-id-to-clipboard
+  [id]
+  (ptk/reify ::copy-id-to-clipboard
+    ptk/WatchEvent
+    (watch [_ _ _]
+      (clipboard/to-clipboard id))))
+
 (defn copy-as-image
   []
   (ptk/reify ::copy-as-image
--- a/frontend/src/app/main/data/workspace/clipboard.cljs
+++ b/frontend/src/app/main/data/workspace/clipboard.cljs
@@ -1143,7 +1143,9 @@
   (ptk/reify ::copy-id-to-clipboard
     ptk/WatchEvent
     (watch [_ _ _]
-      (clipboard/to-clipboard id))))
+      (->> (rx/from (clipboard/to-clipboard id))
+           (rx/map (fn [_]
+                     (ntf/info "The id has been copied to the clipboard")))))))
 
 (defn copy-as-image
   []
--- a/frontend/src/app/main/ui/workspace/context_menu.cljs
+++ b/frontend/src/app/main/ui/workspace/context_menu.cljs
@@ -149,9 +149,11 @@
   {::mf/private true}
   [{:keys [shapes]}]
   (let [multiple?         (> (count shapes) 1)
+        layout            (mf/deref refs/workspace-layout)
 
         do-copy           #(st/emit! (dw/copy-selected))
         do-copy-link      #(st/emit! (dw/copy-link-to-clipboard))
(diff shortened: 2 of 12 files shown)


frontend/src/app/main/ui/workspace/context_menu.cljs
```diff
@@ -200,6 +202,10 @@
                    (reset! enabled-paste-props* false))))))]
 
     [:*
+     (when (and (not multiple?)
+                (contains? layout :debug-panel))
```
hirunatan: This does not work for me; it seems that the layout does not contain `debug` at this point.

Anyway, I think this would better linked to `:show-ids` debug state, not the shapes panel, and checked with `(dbg/enabled? :show-ids)`.

frontend/src/app/main/data/workspace/clipboard.cljs
```diff
@@ -1138,6 +1138,13 @@
     (watch [_ _ _]
       (clipboard/to-clipboard (rt/get-current-href)))))
 
+(defn copy-id-to-clipboard
+  [id]
+  (ptk/reify ::copy-id-to-clipboard
+    ptk/WatchEvent
+    (watch [_ _ _]
+      (clipboard/to-clipboard id))))
```
hirunatan: Perhaps it would be good to show a notification here. If not, the user is not sure if something has happened.

frontend/src/app/main/ui/workspace/sidebar/options.cljs
```diff
@@ -255,7 +280,10 @@
                             :objects objects
                             :page-id page-id
                             :file-id file-id
-                            :shapes shapes}])]
+                            :shapes shapes}]
+
+          :debug
+          [:> debug-shape-info*])]
```
hirunatan, line 286: When this tab is open the sidebar should be expanded by default (or at least being able to be expanded), because the default width is quite narrow.

Also the debug content should adjust to the size, with horizontal scroll or text wrap. Now the excess text is clipped to the right.

frontend/src/app/main/ui/workspace/sidebar/debug.cljs
```diff
@@ -18,15 +18,25 @@
...
+
+        on-toggle-enabled
         (mf/use-fn
          (fn [event option]
            (dom/prevent-default event)
            (dom/stop-propagation event)
            (dbg/toggle! option)
-           (js* "app.main.reinit(true)")))
+           (when-not (contains? no-reload-options option)
+             (js* "app.main.reinit(true)"))))
```
hirunatan, line 42: The `panel-title` (in line 47 below) is no longer necessary, because now we have the tab title.

I think we don't need a close button here, too. If we want to disable the panel we can use again the debug options button.

frontend/src/app/main/ui/workspace/context_menu.cljs
pabloalba: done

frontend/src/app/main/data/workspace/clipboard.cljs
pabloalba: done

frontend/src/app/main/ui/workspace/sidebar/options.cljs
pabloalba: done

pabloalba: I've make it resizable, so the horizontal scroll is not needed

frontend/src/app/main/ui/workspace/sidebar/debug.cljs
pabloalba: done

hirunatan (approved):

PR #10685, opened 2026-07-14
:tada: Add experimental Penpot desktop app

webar4i:

## What

Add an experimental Linux desktop distribution for self-hosted Penpot. The Electron launcher starts the official Docker Compose stack, reports startup progress and diagnostics, and opens Penpot in a dedicated desktop window. A release workflow produces downloadable AppImage and DEB packages.

## Why

Running self-hosted Penpot currently requires users to manage the Compose stack and browser entry point separately. This draft proposes a lower-friction desktop experience while keeping Docker as an explicit dependency and preserving the existing self-hosted data model.

## How

The launcher uses Electron sandboxing, context isolation, disabled Node.js integration, exact-origin navigation checks, and a narrow preload API. It reuses the repository's official Compose file, waits for the local service to become ready, and only uses `docker compose stop` for user-requested shutdowns so volumes are never deleted. The Linux workflow runs the focused tests, builds AppImage and DEB artifacts, and publishes tagged `desktop-v*` releases.

This is intentionally Linux-first and unsigned. Windows/macOS packaging, code signing, automatic updates, and the long-term release cadence remain decisions for maintainers. There is no migration or breaking change for existing deployments.

Related issue: none; this draft is intended to start maintainer discussion.

Launcher design reference: [Lazyweb report](https://www.lazyweb.com/report/lazyweb/30ee0f58-6f75-44ca-8e6e-3e41c0a0cb29/?source=create).

Validated with four passing unit tests, JavaScript syntax checks, workflow YAML parsing, clean AppImage/DEB builds, package-content inspection, and a smoke launch against a healthy local Penpot 2.16 instance.


niwinz:
Hey @webar4i sorry for late response.

What is the intention of this PR?
I mean, we already have an non-official desktop https://github.com/author-more/penpot-desktop
For now we are not in disposition to maintain ourselves this. So this PR will not be merged.

Thanks for your effort but i suggest you joining forces with existing project or go forward with your own project.

PR #10687, opened 2026-07-14
:sparkles: Add prune-unrelated-items debug utility

pabloalba:
### Summary

Adds a new debug.prune-unrelated-items() console function to help isolate bugs by stripping unrelated content from a file.

Behaviour:

Takes the current selection as the seed set.
Expands it iteratively to a fixpoint by adding, for each shape: its ancestors, its full subtree, and — if it belongs to a component copy from the current file — the main instance (with its own ancestors and subtree). Each shape is tracked as a [page-id shape-id] pair to handle identical ids across pages.
Deletes all pages that contain no related shapes, and all unrelated shapes from the remaining pages.
Before applying, prints a console.table of shapes to be deleted (name, id, page-id) and shows a confirmation dialog with page and item counts.
All deletions are wrapped in a single undo transaction so the whole operation can be undone in one step.</s>

### Checklist

### DIFF
--- a/frontend/src/debug.cljs
+++ b/frontend/src/debug.cljs
@@ -9,12 +9,16 @@
    [app.common.data :as d]
    [app.common.data.macros :as dm]
    [app.common.exceptions :as ex]
+   [app.common.files.helpers :as cfh]
    [app.common.files.repair :as cfr]
    [app.common.files.validate :as cfv]
    [app.common.json :as json]
    [app.common.logging :as l]
    [app.common.pprint :as pp]
    [app.common.transit :as t]
+   [app.common.types.component :as ctk]
+   [app.common.types.components-list :as ctkl]
+   [app.common.types.container :as ctn]
    [app.common.types.file :as ctf]
    [app.common.uuid :as uuid]
    [app.main.data.changes :as dwc]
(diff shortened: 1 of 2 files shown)


frontend/src/debug.cljs
```diff
@@ -528,3 +533,112 @@
   [o]
   (app.common.pprint/pprint o {:level 100 :length 100}))
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; PRUNE UNRELATED ITEMS
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn- get-related-ids
```
hirunatan, line 540: I have a mini rant about abbreviating variable names (nothing personal, good vibes) 🙏 

Memory is cheap and we have autocomplete keyboards... When I see `pindex` I haven't a clue of what is it until I look at the code. Instead, if I'd seen `pages-index` it would be easy to think that it's the pages index inside file data. Also if the function was called `get-related-shape-ids` it would be immediately obvious that it only finds shapes, and not pages or components, for example...

About `fdata`, I'd replace it for `file-data` too, but it's true that this abbreviation is so used in our code that has becomed obvious, so well, let it be.

hirunatan (approved):

PR #10696, opened 2026-07-15
:bug: Fix numeric input propagating strings through finite? guard

### LINKED ISSUE #10638: Layout padding saved as string after invalid input in multi-selection, causing persistence 500

umi008:
Closes #10638

## What

Entering an invalid value in a layout padding numeric field (e.g. "abc", "16px", or clearing when nillable is false) causes the fallback value to be persisted as a string instead of a number. The backend Malli schema rejects the string, returning a 500 error from persistence.

## Why

Four-link contamination chain:

1. `fmt/format-number` always returns a string (by design — it's for display).
2. The `last-value*` ref in `numeric-input*` is set directly from the format result, contaminating it with a string.
3. `mth/finite?` on CLJS uses `js/isFinite` without a `number?` guard. `js/isFinite("16")` returns `true`, letting strings pass through sanitization.
4. The backend Malli schema (`::sm/safe-number`) rejects the string → 500.

## How

Two defensive fixes, ~4 lines each:

**Fix A — `numeric_input.cljs`**: Coerce `last-value*` with `d/parse-double` in two places so the ref always holds a number or nil, never a string. `d/parse-double` returns nil for unparseable input (`""`, `nil`, `"abc"`), which is safe.

**Fix B — `math.cljc`**: Add `(number? v)` to the CLJS branch of `finite?`, aligning it with the CLJ implementation which already has the guard. This prevents any future contamination from slipping through `js/isFinite` string coercion.

Both fixes are additive and backward-compatible. All ~30 call sites of `mth/finite` and `mth/finite?` were audited — none intentionally pass strings expecting coercion.

- Created `math_test.cljc` with 8 test cases for number/string/nil/NaN/other types
- Added 6 roundtrip tests to `ds_controls_numeric_input_test.cljs` covering the full contamination chain

### DIFF
--- a/common/src/app/common/math.cljc
+++ b/common/src/app/common/math.cljc
@@ -34,12 +34,14 @@
   #?(:cljs (js/isNaN v)
      :clj (Double/isNaN v)))
 
-;; NOTE: on cljs we don't need to check for `number?` so we explicitly
-;; ommit it for performance reasons.
+;; NOTE: we need `number?` guard on cljs because `js/isFinite` coerces
+;; strings to numbers, accepting "16" as finite when it shouldn't.
+;; This caused a bug where string values from format-number were
+;; propagated through the system until Malli rejected them (issue #10638).
 
 (defn finite?
   [v]
-  #?(:cljs (and (not (nil? v)) (js/isFinite v))
+  #?(:cljs (and (not (nil? v)) (number? v) (js/isFinite v))
      :clj (and (not (nil? v)) (number? v) (Double/isFinite v))))
 
 (defn finite
--- a/common/test/common_tests/math_test.cljc
+++ b/common/test/common_tests/math_test.cljc
@@ -1,0 +1,59 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS INC Sucursal en España SL
+
+(ns common-tests.math-test
+  (:require
+   [app.common.math :as mth]
+   [clojure.test :as t]))
+
+(t/deftest finite?-number-test
+  (t/testing "finite? returns true for positive integer"
+    (t/is (true? (mth/finite? 16))))
+
+  (t/testing "finite? returns true for zero"
+    (t/is (true? (mth/finite? 0))))
+
+  (t/testing "finite? returns true for negative float"
+    (t/is (true? (mth/finite? -42.5))))
+
+  (t/testing "finite? returns true for very large number"
+    (t/is (true? (mth/finite? 1e308)))))
+
+(t/deftest finite?-string-test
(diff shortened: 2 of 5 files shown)


madalenapmelo-kp:

niwinz:
Looks Ok. Thanks.

Lets pass the CI (looks like Linter is failing)

umi008:
done - cljfmt fixed the indent and require ordering. thanks @niwinz

niwinz:
Thanks for you contribution, I'm closing this PR because we have two different PRs fixing the same issue, so I have considered to take the best from both prs and open a single one: https://github.com/penpot/penpot/pull/10758

PR #10702, opened 2026-07-15
:bug: Fix system crash on grid element deletion of a component

### LINKED ISSUE #10101: Crash with referencial integrity error deleting a component inside a grid
### Steps To Reproduce

Steps to reproduce:
1. Import [this file](https://drive.google.com/file/d/1cfrp3Sj6P7KpjgC1CLOlZsx8gs5uYzB8/view?usp=drive_link).
2. Open **Add Fill types (simplified)**,
3. Select any component inside the 3x3 grid (all with name Board).
4. Delete it

### Expected behavior

It gets deleted an nothing else.

### Actual behavior

When the file is persisted in backend, an internal error occurs, with several referential integrity errors of type `:missing-slot` and message `Shape has been swapped, should have swap slot`.

### Screenshots or video

Image

n/a

n/a

PRO, PRE

### Frontend Stack Trace

```console
====================
Data:
--------------------
{:type :validation,
 :code :referential-integrity,
 :hint "error on validating file referential integrity",
 :file-id #uuid "4509f150-a6e5-8109-8008-2677761e5657",
 :details
 [{:code :missing-slot,
   :hint "Shape has been swapped, should have swap slot",
   :shape
   {:id #uuid "58851279-9232-80ae-8008-26535f68454f",
    :type :frame,
    :y 15507.001576007993,
    :hide-fill-on-export false,
    :layout-gap-type :multiple,
    :layout-padding {#, #, #, #},
    :transform #app.common.geom.matrix/Matrix{#, #, #, #, #, #},
    :rotation 0,
```

### Backend Stack Trace

```console

```

### Additional context

_No response_

MischaPanch:
### Related Ticket

https://github.com/penpot/penpot/issues/10101

#### By Human (Mischa)

The issue was a crash due to referential integrity violation when a part of a component was deleted. It turned out that the issue is not deletion but an automatic reordering of subelements in a grid that happens when an element is hidden through the delete. Generally, due to the reordering, subelements of the master and a component instance can become out of sync and fail validation (causing the crash) before an eventual sync would fix this. So the issue is larger than just the manifestation in the linked ticket.

New regression tests were added, both in clojure and in the component testing plugin. Turns out that this issue is not  frontend related, so clojure tests were the most appropriate vehicle for reproducing and analyzing this. I am not fully confident that the new tests in the component testing plugin were catching the crash, but as regression tests they don't hurt anyway.

While this started as a story under the component testing umbrella, and therefore builds on that branch, the solution and clojure tests can be merged independently if so desired. From what I understood about the issue and the solution, it might be a bit urgent to fix this - components with subelements going out of sync with the main component could happen quite often, I imagine.

@opcode81 FYI

#### By Fable
What was going wrong

The setup. In Penpot you can turn a group of shapes into a component — a reusable master element. The original is called the main; every time
you place it somewhere, you get a copy that stays linked to the main so edits flow from main to copy. Mains can contain other components inside
them (nested instances), and copies mirror that inner structure.

The fragile assumption. A copy's inner pieces are linked to the main's inner pieces by position: "the copy's 3rd child corresponds to the main's
3rd child." That correspondence isn't stored anywhere explicit — it's implied purely by the order of the children lists. Penpot's file
validator enforces it: if a copy child at position 3 doesn't point at the main child at position 3, the validator assumes it was deliberately
swapped for something else and demands a marker recording that swap. No marker → the file is declared corrupt → the app crashes with an
"integrity" error.

The bug family. Several parts of Penpot reorder ONE side's children without the other side following, silently breaking that positional pairing:

1. Your crash: "Deleting" a child inside a copy doesn't really delete it, it just hides it (so it can be restored later). Your copy was a grid
layout (children auto-arranged in cells). On the next layout recalculation, the hidden child, having no grid cell anymore, got moved to the
front of the children list. Inserting one element at the front pushes every sibling one position over, so every child now mismatched its
counterpart in the main → crash. (Deleting the first child masked the bug, which is why existing tests missed it, moving the first child to the
front changes nothing.)
2. Related case: reordering or deleting a child inside the main left copies mismatched the same way, at least until a background sync caught up and validation could run before it did.

How it was fixed

Four complementary changes:

1. Smarter validation. The validator now distinguishes a reorder from a swap: if a copy child still points at one of the main's children (just
at a different position), that's harmless (the sync will realign it) so no crash. Only a child pointing at something genuinely foreign is
flagged.
2. Stable grid reordering. When a grid layout re-sorts its children, children without a cell (hidden ones) now stay where they were instead of jumping to the front.
3. A general safety rule. Copies' internal structure is only allowed to be changed by the component-sync machinery. The kind of change that scrambled your copy (a wholesale rewrite of a children list) is now rejected outright when aimed at a copy — closing the loophole for any code path, not just this one.
4. Delete propagation. Deleting a piece inside a main now also deletes the corresponding pieces in all its copies in the same file, immediately, so no copy is ever left pointing at something that no longer exists.

All verified: the previously failing regression tests now pass, the full test suite is green (1076 tests), and the end-to-end test that used to
hang the live app now completes and passes.

### Steps to reproduce 

See ticket

### DIFF
--- a/.serena/memories/common/component-data-model.md
+++ b/.serena/memories/common/component-data-model.md
@@ -24,6 +24,12 @@ Variant masters are main instances and component roots. Their descendants may th
 
 Masters are not normally touched through `set-shape-attr`, but touched flags can appear on master shapes through cloning/duplication paths. `add-touched-from-ref-chain` in `app.common.logic.variants` unions touched flags from ancestors into the copy being processed, so upstream/master touched state can affect downstream switch behavior.
 
+## Swap slots and positional matching
+
+- A swap slot (stored via `ctk/set-swap-slot`, a `:touched` group `swap-slot-<uuid>`) marks a copy sub-head that was SWAPPED to another component; `compare-children` then pairs it to the main child by slot instead of by `shape-ref`.
+- Copy sub-heads without a slot are paired to main children by `shape-ref` (seek, not index). `find-near-match` (positional) is only a validator/repair heuristic; validity requires membership of the ref among the near-main parent's children, not index equality (`mem:common/file-change-validation-migration-subtleties`).
+- Copy child ORDER converges to the main's via the async sync (`moved` branch of `compare-children`); local code must never reorder copy children directly (guards in `:mov-objects`/`:reorder-children`).
+
 ## Cloning paths
 
 `make-component-instance` in `app.common.types.container` produces a clean component copy through `update-new-shape`, dissociating attrs such as `:touched`, `:variant-id`, and `:variant-name` on cloned shapes.
--- a/.serena/memories/common/file-change-validation-migration-subtleties.md
+++ b/.serena/memories/common/file-change-validation-migration-subtleties.md
@@ -7,6 +7,8 @@
 - `set-shape-attr` treats `:position-data` as derived and never touched. Geometry/content-path changes use approximate equality; geometry differences under about 1px can be ignored for touched purposes.
 - Width/height are excluded from the `is-geometry?` branch in `set-shape-attr`; do not assume all geometry-group attrs follow identical ignore-geometry behavior.
 - `process-touched-change` marks the owning component modified when a touched shape belongs to a main instance; component-data changes can come from shape ops through this second pass.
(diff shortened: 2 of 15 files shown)


MischaPanch:
Should be merged after #10679

opcode81:
Since #10679 is now merged, I have rebased the branch onto develop.

Alotor:
I'm closing this MR because I open another one with some adjustments

https://github.com/penpot/penpot/pull/10956

PR #10731, opened 2026-07-16
:bug: Repair text shapes with empty content to prevent workspace update failures

### LINKED ISSUE #10725: Text shape with empty content breaks workspace updates
### Description

In workspaces that contain a text shape whose `:content` is structurally incomplete (a root with an empty `:children` vector, e.g. `{:type "root" :vertical-align "top" :children []}`), any subsequent `/api/main/methods/update-file` call that contains even a single mod-obj for that shape fails the backend `validate-shape` check and rolls back the entire batch — including unrelated geometry changes to other shapes on other pages in the same file.

The user observes this as "my edits don't save" / Sentry alert:

```
data assertion error: invalid shape found '9a0ed99f-4b8b-5e12-a951-e21adba846d4'
    at app.common.files.changes$validate_shape (changes.cljc:475)
    at app.common.files.changes$process_operations (changes.cljc:623)
```

The structurally incomplete `:content` is most often produced by the inspector `update-root-attrs` path when a text shape has no prior `:content` (e.g. a brand-new text shape that has never been edited) and the user changes a root-level attribute like vertical-align.

### Steps to reproduce

1. Create a new text shape (or have one whose `:content` is `nil` / `{:type "root"}`).
2. Without first typing into the shape, change a root-level attribute in the inspector (e.g. vertical-align from "top" to "center").
3. Continue editing any other shape and try to save.
4. Observe that `update-file` returns a 500 with `code: data-validation, hint: invalid shape found '<shape id>'`.

The same shape then blocks any later `update-file` that targets it (or contains a mod-obj whose ops touch any of its attrs), even pure-geometry changes like `:points`, `:selrect`, `:width`, `:height`, `:flip-x`, `:flip-y`.

### Expected behavior

- The frontend must not be able to produce a text shape `:content` that lacks a paragraph-set (the canonical root → paragraph-set → paragraph → span tree).
- A text shape with already-broken content must not be able to take down an otherwise valid batch of geometry edits to other shapes.
- Files that already contain broken text content must self-heal on next load.

### Affected versions

2.17.0-RC4 (and presumably earlier 2.17.x).

niwinz:

## What

A text shape with structurally incomplete `:content` (root with empty `:children` vector, e.g. `{:type "root" :vertical-align "top" :children []}`) used to slip past the frontend editor but fail the backend `validate-shape` schema. Because `process-operations` runs validation on every `mod-obj`, a single broken text shape anywhere in the file rolled back the entire `update-file` batch — including unrelated geometry edits to other shapes on other pages — so the user observed "my edits don't save".

## Why

The frontend's `update-root-attrs` path produced the broken content when the inspector set a root attribute (e.g. `vertical-align`) on a brand-new text shape that had no prior `:content` (the seeded branch was `(assoc shape :content (d/txt-merge {:type "root"} attrs))`, which writes a root node with no paragraph-set). The backend schema was permissive (`{:children {:optional true} [:maybe …]}`) and never failed on the producing runtime, so the broken content could persist across saves and only surface as a hard validation error on the next unrelated edit.

## How

Five orthogonal changes, all the way from root cause to comprehensive repair:

- **Frontend root cause** — `update-root-attrs` now seeds new text content from `tc/v2-default-text-content` so the canonical `root → paragraph-set → paragraph → span` tree is always present.
- **Frontend defensive** — new `ensure-valid-text-content` helper normalizes any root with empty/missing `:children` at the entry of `v2-update-text-shape-content`.
- **Frontend optimization** — with a well-formed baseline, the new-shape path no longer churns redundant `:content` set operations in the changes-builder diff.
- **Schema tightening** — `:children` is no longer `:optional` and no longer wrapped in `[:maybe ...]`, so any new bad content fails immediately on the producing runtime.
- **Comprehensive migration** — `0025-repair-empty-text-content` walks `:pages-index` and `:components` and repairs broken content at all three levels (root, paragraph-set, paragraph), handling all corner cases: nil/empty/non-vector children, non-map items, wrong types. Preserves root-level attrs (`:vertical-align`, etc.) and is idempotent on healthy content.

## Testing

- `common`: full suite via `clojure -M:test` — 1102 tests, 24,554 assertions, 0 failures.
- `frontend`: full suite via `node target/tests/test.js` — 403 tests, 0 failures.
- `clj-kondo` and `cljfmt` clean on both `common/` and `frontend/`.
- 33 exhaustive tests for migration-0025 in dedicated namespace `common-tests.files-migrations-0025-test`, covering shape guards, all three content levels, preservation, multi-item scenarios, container coverage, and idempotency.

Fixes #10725

### DIFF
--- a/common/src/app/common/files/migrations.cljc
+++ b/common/src/app/common/files/migrations.cljc
@@ -1874,6 +1874,110 @@
         (update :pages-index d/update-vals update-container)
         (d/update-when :components d/update-vals update-container))))
 
+(defmethod migrate-data "0025-repair-empty-text-content"
+  ;; Repair text shapes whose :content tree has empty/missing :children
+  ;; at any of the three levels:
+  ;;   Level 1: root with no paragraph-set
+  ;;   Level 2: paragraph-set with no paragraph
+  ;;   Level 3: paragraph with no span
+  ;; Such shapes fail the backend `validate-shape` schema and would also
+  ;; break the v2 editor's `cljs->dom` roundtrip. Re-seed the canonical
+  ;; root -> paragraph-set -> paragraph -> span tree, preserving the
+  ;; original root-level attributes (e.g. :vertical-align) when present.
+  ;; Idempotent on healthy content.
+  [data _]
+  (let [default-span {:text "" :fills types.text/default-text-fills}
+        default-paragraph {:type "paragraph" :children [default-span]}
+        default-paragraph-set {:type "paragraph-set" :children [default-paragraph]}
+
+        ;; Level 3: repair paragraph with empty/missing children
+        repair-span (fn [span]
+                      (if (and (map? span)
+                               (string? (:text span)))
+                        span
+                        default-span))
+
+        repair-paragraph (fn [paragraph]
+                           (if (and (map? paragraph)
+                                    (= "paragraph" (:type paragraph)))
+                             (cond
+                               ;; Children is nil or empty vector - seed with default span
+                               (or (nil? (:children paragraph))
+                                   (and (vector? (:children paragraph))
+                                        (empty? (:children paragraph))))
+                               (assoc paragraph :children [default-span])
+
+                               ;; Children is a vector - repair any invalid spans
+                               (vector? (:children paragraph))
+                               (update paragraph :children
+                                       (fn [children]
+                                         (mapv repair-span children)))
+
+                               ;; Children is not a vector - replace with default
+                               :else
+                               (assoc paragraph :children [default-span]))
+                             default-paragraph))
+
+        ;; Level 2: repair paragraph-set with empty/missing children
+        repair-paragraph-set (fn [paragraph-set]
+                               (if (and (map? paragraph-set)
+                                        (= "paragraph-set" (:type paragraph-set)))
+                                 (cond
+                                   ;; Children is nil or empty vector - seed with default paragraph
+                                   (or (nil? (:children paragraph-set))
+                                       (and (vector? (:children paragraph-set))
+                                            (empty? (:children paragraph-set))))
+                                   (assoc paragraph-set :children [default-paragraph])
+
+                                   ;; Children is a vector - repair any invalid paragraphs
+                                   (vector? (:children paragraph-set))
+                                   (update paragraph-set :children
+                                           (fn [children]
+                                             (mapv repair-paragraph children)))
+
+                                   ;; Children is not a vector - replace with default
+                                   :else
+                                   (assoc paragraph-set :children [default-paragraph]))
+                                 default-paragraph-set))
+
+        ;; Repair content at all levels, handling all edge cases
+        repair-content (fn [content]
+                         (cond
+                           ;; Content is not a valid root map - create default
+                           ;; Preserve root-level attrs if content is a map
+                           (or (nil? content)
+                               (not (map? content))
+                               (not= "root" (:type content)))
+                           (merge types.text/default-root-attrs
+                                  {:type "root"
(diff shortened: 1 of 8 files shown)


common/src/app/common/files/changes.cljc
```diff
@@ -619,8 +642,15 @@
       (when (and *state* page-id)
         (swap! *state* collect-shape-media-refs shape page-id))
 
-      ;; NOTE: we only perform hard validation on backend
-      #?(:clj (validate-shape shape page-id))
+      ;; NOTE: we only perform hard validation on backend. Pure-geometry
```
Alotor: I'm not sure about this, for example x, y, width, height can break the schema when they have incorrect values (are not safe-int for example, or even 0 for width/height). Selrect and points also is something that i've seen failing the schema.

common/src/app/common/files/changes.cljc
```diff
@@ -619,8 +642,15 @@
...
+      ;; mod-objs cannot change shape schema validity, so skip
+      ;; validation there: a single broken shape elsewhere in the file
+      ;; must not be able to roll back an unrelated batch of geometry
+      ;; edits. Validation is still enforced for any op that touches
+      ;; content-affecting attributes (`:content`, `:position-data`,
+      ;; `:fills`, `:strokes`, fonts, etc.).
+      #?(:clj (when-not (geometry-only-mod-obj? change)
+                (validate-shape shape page-id)))
```
Alotor: Instead of filtering by geometry you can just not validate when the shape was already invalid:

```clojure
#?(:clj
   (when-not (cts/valid-shape? shape)
     ;; A shape that was already invalid before this change must not block
     ;; unrelated edits in the same batch; anything this change itself broke
     ;; is still rejected.
     (when (cts/valid-shape? (get objects id))
       (validate-shape shape page-id))))
```

common/src/app/common/files/migrations.cljc
```diff
@@ -1874,6 +1874,52 @@
...
+  ;; Re-seed the canonical root -> paragraph-set -> paragraph -> span
+  ;; tree, preserving the original root-level attributes (e.g.
+  ;; :vertical-align) when present. Idempotent on healthy content.
+  [data _]
+  (let [default-root [{:type "paragraph-set"
+                       :children [{:type "paragraph"
+                                   :children [{:text ""
+                                               :fills types.text/default-text-fills}]}]}]
+
+        needs-repair? (fn [content]
```
Alotor: this method only check if the root has empty children, but the schema has tree levels:
```
root            :children  → [:vector {:min 1}  paragraph-set ]   ← level 1
  paragraph-set :children  → [:vector {:min 1}  paragraph     ]   ← level 2
    paragraph   :children  → [:vector {:min 1}  span          ]   ← level 3
```
i think you can check the levels 2 and 3 from here and mirgrate them if they are broken

common/src/app/common/files/changes.cljc
niwinz: agree

common/src/app/common/files/migrations.cljc
niwinz: Should be fixed

common/src/app/common/files/changes.cljc
niwinz: I just reverted this for now

common/src/app/common/files/changes.cljc
niwinz: reverted this for now

common/src/app/common/files/migrations.cljc
niwinz: Fixed, and added exhaustive test suite for the migration, covering several corner cases

PR #10734, opened 2026-07-16
:whale: Migrate remaining Penpot images to DHI [3/3]

### LINKED ISSUE #10720: Migrate Docker images to Docker Hardened Images (DHI)

Our Docker images (docker/imagemagick, docker/devenv, and the five under
docker/images/) are built on ubuntu:26.04 and nginxinc/nginx-unprivileged.
Keeping these patched against CVEs is entirely on us: there's no SBOM, no
provenance attestation, and no guarantee of how current the base image's
packages are at any given build. I'm frustrated that our security posture
for these images depends on us remembering to bump base image versions and
manually auditing what's actually inside them, rather than on a
continuously-maintained upstream.

Migrate all 7 Dockerfiles to Docker Hardened Images (DHI) -- Docker's line
of minimal, continuously CVE-patched base images with SBOM and SLSA Build
Level 3 provenance attestations, built on the same Debian/Alpine
foundations we already use (no exotic distro migration required).

Proposed as 3 separate PRs to keep review manageable:

1. Build tooling (manage.sh): adds --tag/--push to every image build
   command. build-devenv and build-imagemagick-docker-image become
   local-only by default (--push now required to publish -- today they
   push by default unless --local is passed, which I'd like to flip while
   touching this file anyway). Independent of the base-image changes
   below.
2. docker/imagemagick/Dockerfile + docker/devenv/Dockerfile.
3. docker/images/Dockerfile.{frontend,backend,exporter,mcp,storybook},
   plus a consolidated notes doc under docs/technical-guide/developer/.

Important tradeoff to flag: DHI ships a full "-dev" tag (shell + package
manager, root) and a minimal non-dev runtime tag (no shell, non-root) per
image. imagemagick, storybook, and mcp move to the true minimal runtime.
devenv, frontend, backend, and exporter keep the "-dev" tag as their final
image, because each needs a shell and/or package manager at container
*runtime*, not just build time (devenv is an interactive dev container by
design; frontend templates its nginx config via bash/sed/awk/envsubst at
startup; backend needs fontforge/python3 at runtime; exporter runs a full
headless-browser/Playwright stack). We still get DHI's patched base +
SBOM/provenance on all 7 either way -- just not the fully-hardened
shell-less runtime on those four. 

> Workflow note: DHI has been free and open source (Apache 2.0) since
December 2025, no subscription needed -- but pulling from dhi.io does
require an authenticated `docker login` with a (free) Docker ID, unlike
our current anonymously-pullable base images. This means CI needs a Docker
ID configured as a credential, and contributors need to log in once before
building locally.

- Do nothing / keep manually bumping ubuntu:26.04 and auditing packages
  ourselves. Doesn't solve the underlying problem (no SBOM/provenance,
  patching still depends on us remembering).
- Move to plain Alpine or Debian slim images without DHI. Gets smaller
  images, but without the signed SBOM, SLSA provenance, or Docker's
  patching commitment -- we'd be back to auditing CVEs ourselves.
- A different hardened-image provider (e.g. Chainguard Images), which
  offers a broadly similar free tier. Didn't pursue this since DHI builds
  on the same Debian/Alpine userland we already use, minimizing the diff;
  switching provider is still on the table if there's a reason to prefer
  it.

### Additional context

Per-file summary of the proposed migration:

| File | New base | Runtime strategy |
|---|---|---|
| docker/imagemagick/Dockerfile | dhi.io/debian-base:trixie-debian13 | True non-dev runtime, libs vendored via ldd |
| docker/devenv/Dockerfile | dhi.io/debian-base:trixie-debian13-dev | "-dev" kept as final |
| docker/images/Dockerfile.storybook | dhi.io/nginx:1.31.1-alpine3.22 | True non-dev runtime |
| docker/images/Dockerfile.frontend | dhi.io/nginx:1.31.1-debian13-dev | "-dev" kept as final |
| docker/images/Dockerfile.mcp | dhi.io/node:24.18.0-debian13 | True non-dev runtime |
| docker/images/Dockerfile.backend | dhi.io/debian-base:trixie-debian13-dev | "-dev" kept as final |
| docker/images/Dockerfile.exporter | dhi.io/node:24.18.0-debian13-dev | "-dev" kept as final |

Detailed notes about the process: [dhi-migration-notes.md](https://github.com/user-attachments/files/30084844/dhi-migration-notes.md)

Testing performed: all 7 images built and ran individually (docker
build/docker run); the full stack (frontend + backend + exporter + mcp +
postgres + valkey) was validated together via
docker/images/docker-compose.yaml, including an actual file export through
the UI.

Full troubleshooting history and package-name gotchas found along the way
(Ubuntu -> Debian package renames, a PAM configuration issue on minimal
images, dpkg conffile prompts, etc.) will be included in the individual
PR descriptions rather than as a separate doc in the repo.

References:
- https://docs.docker.com/dhi/
- https://www.docker.com/press-release/docker-makes-hardened-images-free-open-and-transparent-for-everyone/

### Extra: adding the same attestations to our own images

**Attestations** are metadata generated at build time by BuildKit and
attached to the image itself, describing how it was built and what it
contains. They travel with the image (as part of its manifest), so anyone
pulling it can inspect this information without needing a separate report
or portal. There are two types:

- **SBOM** (Software Bill of Materials): a full inventory of every
  package/library the image contains, or that was used to build it --
  names, versions, and licenses. Think of it as an ingredients list for
  the image.
- **Provenance**: metadata about *how* the image was built -- which
  source repo/commit, which build steps, which builder. This is what lets
  you verify an image was actually built from the source you expect,
  rather than tampered with somewhere in the pipeline.

Both matter for the same reason as the rest of this migration: supply
chain security. If a vulnerability is announced in some library tomorrow,
an SBOM lets you immediately check which of our images are affected,
instead of manually auditing each Dockerfile. Provenance lets anyone
(including automated policy tools) verify an image actually came from our
CI, not from somewhere else.

This is exactly what DHI's own base images already give us for free (SBOM
+ SLSA Build Level 3 provenance on every hardened image) -- this change
extends the same idea to the images *we* build and publish.

So...

While at it, `docker/build-push-action` (already v7 in our workflows)
supports build attestations natively since v4 -- no extra tooling needed.
Added `provenance: mode=max` + `sbom: true` to every image build in CI, and
the equivalent `--provenance=mode=max --sbom=true` flags to manage.sh's
`--push` builds (local-only builds can't carry attestations -- the local
image store doesn't support them, only registry pushes do).

This means every image we publish now carries its own signed SBOM and
build provenance, the same kind of supply-chain metadata DHI's base images
already give us for free -- now for our own image layers too.

Ref: https://docs.docker.com/build/metadata/attestations/

bameda:
> [!CAUTION]
> Merged after https://github.com/penpot/penpot/pull/10733

### Related Ticket

https://github.com/penpot/penpot/issues/10720

### Summary

- Move Dockerfile.frontend, .backend, .exporter, .mcp and .storybook under docker/images/ from ubuntu:26.04 / nginx-unprivileged / a manual Node tarball install to Docker Hardened Images (Debian 13, or Alpine for storybook). 
- storybook and mcp get a true non-dev runtime; frontend, backend and exporter keep the -dev tag as their final image, since each needs a shell and/or package manager at container runtime (nginx templating, fontforge/python3, and a headless-browser stack, respectively).
- Added `provenance: mode=max` and `sbom: true` to every
`docker/build-push-action` step in the workflows. These already push to a registry
(`push: true`), which is a requirement for attestations to attach at all.
`docker/build-push-action` has supported this since v4; we're already on
v7, so no action version bump needed.

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -107,6 +107,19 @@ jobs:
           username: ${{ secrets.PUB_DOCKER_USERNAME }}
           password: ${{ secrets.PUB_DOCKER_PASSWORD }}
 
+      # Frontend, backend, exporter, storybook and mcp now build FROM
+      # Docker Hardened Images (dhi.io). DHI is free (Apache 2.0, no
+      # subscription), but pulling from it still requires an
+      # authenticated login -- a separate `docker login` against a
+      # different registry host, even though it reuses the same
+      # PUB_DOCKER_* credentials as the DockerHub login above.
+      - name: Login to Docker Hardened Images registry (base image pull)
+        uses: docker/login-action@v4
+        with:
+          registry: dhi.io
+          username: ${{ secrets.PUB_DOCKER_USERNAME }}
+          password: ${{ secrets.PUB_DOCKER_PASSWORD }}
+
       # Every build pushes an immutable `build-<key>` tag; if it already
       # exists in the registry, this exact image was already built and the
       # whole build can be skipped. Moving tags are handled by `promote`.
@@ -164,6 +177,8 @@ jobs:
           file: ./docker/images/Dockerfile.${{ matrix.image }}
           platforms: linux/amd64,linux/arm64
           push: true
+          provenance: mode=max
+          sbom: true
           # Immutable tag only; branch tags are moved atomically for the
           # whole image set by the `promote` job.
           tags: ${{ secrets.DOCKER_REGISTRY }}/${{ matrix.image }}:build-${{ needs.prepare.outputs.build_key }}
--- a/docker/images/Dockerfile.backend
+++ b/docker/images/Dockerfile.backend
@@ -1,4 +1,4 @@
-FROM ubuntu:26.04 AS build
+FROM dhi.io/debian-base:trixie-debian13-dev AS build
 LABEL maintainer="Penpot <docker@penpot.app>"
 
 ENV LANG='C.UTF-8' \
@@ -47,7 +47,8 @@ RUN set -eux; \
        --add-modules java.base,jdk.net,jdk.management.agent,java.se,jdk.compiler,jdk.javadoc,jdk.attach,jdk.unsupported,jdk.jfr,jdk.jcmd \
        --output /opt/jre;
 
-FROM ubuntu:26.04 AS image
+
+FROM dhi.io/debian-base:trixie-debian13-dev AS image
 LABEL maintainer="Penpot <docker@penpot.app>"
 
 ENV LANG='C.UTF-8' \
@@ -55,11 +56,13 @@ ENV LANG='C.UTF-8' \
     JAVA_HOME="/opt/jre" \
     PATH=/opt/jre/bin:/opt/imagick/bin:$PATH \
     DEBIAN_FRONTEND=noninteractive \
-    TZ=Etc/UTC
+    TZ=Etc/UTC \
+    LD_LIBRARY_PATH=/opt/imagick/lib/deps
 
 RUN set -ex; \
-    useradd -U -M -u 1001 -s /bin/false -d /opt/penpot penpot; \
     apt-get -qq update; \
+    apt-get -qqy --no-install-recommends install passwd; \
+    useradd -U -M -u 1001 -s /bin/false -d /opt/penpot penpot; \
     apt-get -qq dist-upgrade; \
     apt-get -qqy --no-install-recommends install \
         ca-certificates \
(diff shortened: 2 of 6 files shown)


.github/workflows/build-docker.yml
```diff
@@ -85,6 +85,19 @@ jobs:
           username: ${{ secrets.PUB_DOCKER_USERNAME }}
           password: ${{ secrets.PUB_DOCKER_PASSWORD }}
 
+      # Frontend, backend, exporter and mcp now build FROM Docker Hardened
```
yamila-moreno: and storybook

yamila-moreno (approved):

.github/workflows/build-docker.yml
bameda: Fixed

PR #10774, opened 2026-07-22
:bug: Fix tick icons are not aligned in font selector

### LINKED ISSUE #10597: Tick icons are not aligned in font selector
Tick icons are not aligned in font selector

Image

how to reproduce
- create a text layer
- change the font family with the font selector
- close the selector
- open again the font selector and search the same font family
- check the tick icons

luisddm:
### Related Ticket

Closes #10597


EvaMarco:
This is fixed only when there is not scrollbar on the dropdown.
Screenshot from 2026-07-24 11-42-03

But when we do not filter the fonts and we have the scrollbar we have the icons not alligned again.
Screenshot from 2026-07-24 11-41-32

PR #10823, opened 2026-07-24
Update MCP configuration for Claude Code

acontreras89:
The Claude Code configuration expects a `type=http` property (rather than `transport`).

Alternatively we could use the CLI, like so:

```sh
claude mcp add --transport http penpot $REMOTE_OR_LOCAL_URL
```


madalenapmelo-kp:

opcode81:
The documentation was already changed on branch `develop`, rendering this obsolete.

We could consider accelerating the documentation update by cherry-picking 51c0129485b7c4cce9662f746768966a54ec9b07.

PR #10824, opened 2026-07-24
:bug: Fix wasm context restore guards

### LINKED ISSUE #10810: WASM renderer panics when WebGL context is restored mid-reload
### Description

After a WebGL context loss/restore, the workspace can call into the WASM renderer while `reload-renderer!` is still between teardown (`clear-canvas`) and re-init. That window produces WASM panics (reported as `:panic` / `:wasm-critical`) instead of a safe no-op or a completed reload.

Production error reports show a clear sequence: `context-lost` → `context-restored` → viewport/DPR resize → panic in `_resize_viewbox` (seen on large screens in 2.17.0-RC5). A related unclassified failure in `_render_from_cache` (`:wasm-critical`, error code 0) was also reported in 2.17.1-RC1 after SVG upload + transforms; it may share the same GPU-pressure / mid-teardown family.

Root cause (frontend bridge): `on-webgl-context-restored` cleared lost state / emitted `context-restored` before the async reload finished, which flipped the UI out of inspect/read-only and allowed resize, modifiers, and other APIs to hit WASM while the GL context was not ready. Several public APIs also lacked readiness guards (or returned `nil` where callers expected vectors), so mid-reload interaction could assert or panic (e.g. drag → `set-modifiers` with non-vector modifiers).

### Steps to reproduce

1. Open a workspace with `render-wasm/v1` enabled and the CLJS REPL connected.
2. Simulate mid-reload teardown:

```clojure
(require '[app.render-wasm.api :as api])
(api/clear-canvas {:lose-browser-context? false})
```

3. Trigger any of:
   - `(api/resize-viewbox 1920 1080)` — previously panicked
   - create a shape and drag it — previously asserted in `set-modifiers` (`expected a vector`)
   - viewport resize / DPR change / pan-zoom while the context is still down
4. Optionally recover with `reload-renderer!` (canvas `#render` + payload) and confirm the workspace becomes usable again without a full page reload.

Natural (harder) path: force a real `webglcontextlost` / `webglcontextrestored` on a large viewport and interact (resize window, change DPR, drag shapes) during recovery.

### Expected behavior

While the renderer is lost or mid-reload, application WASM calls no-op safely (or wait until reload completes). `context-restored` is emitted only after reload succeeds. After recovery, viewport sync, rendering, and transforms work without panics or asserts. A failed reload keeps the UI in a lost/safe state.

### Affected versions

2.17.0-RC5, 2.17.1-RC1 (and current develop until the readiness/reload guards land)

superalex:
### Related Ticket

Closes #10810 

### Setup
1. Start the frontend watch as usual.
2. Open a workspace in the browser with **render-wasm** enabled (`render-wasm/v1`).
3. From `frontend/`, open a CLJS REPL connected to the browser:
```bash
npx shadow-cljs cljs-repl :main
```
Wait until you get cljs.user=> 

**1) Simulate mid-teardown**
```
(require '[app.render-wasm.api :as api]
         '[app.render-wasm.wasm :as wasm]
         '[app.main.data.render-wasm :as drw]
         '[app.main.store :as st])
;; Keep the canvas before clear (clear-canvas sets wasm/canvas to nil)
(def canvas (or wasm/canvas (js/document.getElementById "render")))
(api/clear-canvas {:lose-browser-context? false})
```

Check state:
```
[(wasm/live?) (wasm/ready?) wasm/context-initialized?]
```

**2) Exercise calls that used to hit a half-torn-down renderer**
In the workspace: create a shape and drag it (selection rect may stay on SVG), delete some shapes, zoom in/out, panning...

3) Recover without a full page reload
```
(set! wasm/canvas canvas)
(-> (api/reload-renderer! (#'api/build-reload-payload))
    (.then (fn [_]
             (reset! wasm/reloading? false)
             (st/emit! (drw/context-restored))
             (js/console.log "wasm ready?" (wasm/ready?))))
    (.catch (fn [e]
              (reset! wasm/reloading? false)
              (js/console.error e))))
```
Then check:

- Promise resolves (or log the rejection).
- Browser console: wasm ready? …
- Canvas paints again.
- Zoom, pan, create/drag still work.

https://github.com/user-attachments/assets/e67141a2-b416-456a-bc74-68a6de111b71

### Checklist

### DIFF
--- a/frontend/src/app/render_wasm/api.cljs
+++ b/frontend/src/app/render_wasm/api.cljs
@@ -2587,6 +2587,12 @@
         (throw cause)))))
 
 (defn calculate-bool
+  "WASM implementation of `path/calc-bool-content`.
+
+  It must always return content: the result is stored as the bool shape's `:content`
+  and persisted as a file change, so returning `nil` would destroy the path.
+  When the render context is unusable, we fall back to the pure CLJS calculation,
+  which needs no render context."
   [shape objects]
 
   ;; We need to be able to calculate the boolean data but we cannot
@@ -2595,24 +2601,27 @@
   ;; temporary and then we serialize the objects needed to calculate the
   ;; boolean object.
   ;; After the content is returned we discard that temporary context
-  (h/call wasm/internal-module "_start_temp_objects")
+  (if-not (initialized?)
+    (path/calc-bool-content shape objects)
+    (do
+      (h/call wasm/internal-module "_start_temp_objects")
 
-  (try
-    (let [bool-type (get shape :bool-type)
-          ids (get shape :shapes)
-          all-children
-          (->> ids
-               (mapcat #(cfh/get-children-with-self objects %)))]
+      (try
+        (let [bool-type (get shape :bool-type)
+              ids (get shape :shapes)
+              all-children
+              (->> ids
+                   (mapcat #(cfh/get-children-with-self objects %)))]
 
-      (h/call wasm/internal-module "_init_shapes_pool" (count all-children))
-      (run! set-object all-children)
+          (h/call wasm/internal-module "_init_shapes_pool" (count all-children))
+          (run! set-object all-children)
 
-      (-> (calculate-bool* bool-type ids)
-          (path.impl/path-data)))
-    (finally
-      ;; Always restore the main shapes pool: leaving the temp pool
-      ;; active would make the next `_start_temp_objects` panic.
-      (h/call wasm/internal-module "_end_temp_objects"))))
+          (-> (calculate-bool* bool-type ids)
+              (path.impl/path-data)))
+        (finally
+          ;; Always restore the main shapes pool: leaving the temp pool
+          ;; active would make the next `_start_temp_objects` panic.
+          (h/call wasm/internal-module "_end_temp_objects"))))))
 
 (def POSITION-DATA-U8-SIZE 36)
 (def POSITION-DATA-U32-SIZE (/ POSITION-DATA-U8-SIZE 4))
--- a/frontend/src/app/render_wasm/api.cljs
+++ b/frontend/src/app/render_wasm/api.cljs
@@ -93,11 +93,14 @@
 
 
 (defn initialized?
-  "True when the WASM render context is ready to receive design-state
-  operations. Use it to skip WASM work during transient states (e.g. while
-  switching renderer with a text shape being edited)."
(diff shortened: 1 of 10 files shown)


frontend/src/app/render_wasm/api.cljs
```diff
@@ -2404,91 +2462,97 @@
...
-        (mem/free)
-        content)
-      (catch :default cause
-        (mem/free)
-        (throw cause)))))
+        (catch :default cause
+          (mem/free)
+          (throw cause))))))
 
 (defn calculate-bool
```
elenatorro, line 2589: avoid returning nil on calculate-bool:

```suggestion
(defn calculate-bool
  "WASM implementation of `path/calc-bool-content`.

  It must always return content: the result is stored as the bool shape's `:content`
  and persisted as a file change, so returning `nil` would destroy the path.
  When the render context is unusable, we fall back to the pure CLJS calculation,
  which needs no render context."
  [shape objects]

  ;; We need to be able to calculate the boolean data but we cannot
  ;; depend on the serialization flow.
  ;; start_temp_object / end_temp_object create a new shapes_pool
  ;; temporary and then we serialize the objects needed to calculate the
  ;; boolean object.
  ;; After the content is returned we discard that temporary context
  (if-not (initialized?)
    (path/calc-bool-content shape objects)
    (do
      (h/call wasm/internal-module "_start_temp_objects")

      (try
        (let [bool-type (get shape :bool-type)
              ids (get shape :shapes)
              all-children
              (->> ids
                   (mapcat #(cfh/get-children-with-self objects %)))]

          (h/call wasm/internal-module "_init_shapes_pool" (count all-children))
          (run! set-object all-children)

          (-> (calculate-bool* bool-type ids)
              (path.impl/path-data)))
        (finally
          ;; Always restore the main shapes pool: leaving the temp pool
          ;; active would make the next `_start_temp_objects` panic.
          (h/call wasm/internal-module "_end_temp_objects"))))))
```

elenatorro (changes requested):

frontend/src/app/render_wasm/api.cljs
superalex: Done in 1c7fa3bb3c — `calculate-bool` now falls back to `path/calc-bool-content` when the WASM context is not ready, so we never persist `nil` content.

elenatorro (approved):

PR #10832, opened 2026-07-24
:bug: Fix selrect misplacement on grid and flex layouts

elenatorro:
### Related Ticket

No reference ticket

### Summary

DOM selection overlay was displaced after leaving a grid/flex layout. workspace-selrect, wasm-modifiers are defonce behavior-subjects set by set-wasm-modifiers but cleared only by finish-transform. If a control unmounted on the interaction, the stale selrect still displaced the next selection

Example:

[screen-recorder-fri-jul-24-2026-13-59-09.webm](https://github.com/user-attachments/assets/58a8217b-4a56-4a76-bb35-81c6c8c815c1)

Fix: Clear both streams on stop-grid-layout-editing (grid) and on flex-control unmount while resizing via `fcc/clear-transform-preview! (padding/margin/gap)`. Only affects render-wasm.

Note: when moving selectors to WASM, this issue should not happen either

### Steps to reproduce

Sometimes is not trivial to reproduce, but doing this several times should trigger it (check the video from above):

1. Select a frame with grid layout and enter grid edit mode.
2. Start a gesture on it by grabbing a grid track or cell.
3. Before releasing cleanly, select the frame above.
4. The new selection's DOM selectors are displaced by the old frame's offset.

### DIFF
--- a/frontend/src/app/main/data/workspace/grid_layout/editor.cljs
+++ b/frontend/src/app/main/data/workspace/grid_layout/editor.cljs
@@ -11,6 +11,8 @@
    [app.common.types.shape.layout :as ctl]
    [app.main.data.helpers :as dsh]
    [app.main.data.workspace.viewport-wasm :as dwvw]
+   [app.main.streams :as ms]
+   [beicon.v2.core :as rx]
    [potok.v2.core :as ptk]))
 
 (defn hover-grid-cell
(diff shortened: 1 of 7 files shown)


frontend/src/app/main/data/workspace/grid_layout/editor.cljs
```diff
@@ -88,7 +90,12 @@
   (ptk/reify ::stop-grid-layout-editing
     ptk/UpdateEvent
     (update [_ state]
-      (update state :workspace-grid-edition dissoc grid-id))))
+      (update state :workspace-grid-edition dissoc grid-id))
+
+    ptk/EffectEvent
+    (effect [_ _ _]
+      (rx/push! ms/wasm-modifiers nil)
+      (rx/push! ms/workspace-selrect nil))))
```
niwinz: This should be using `clear-transform-preview!` ?

elenatorro: thanks, I took advantage and move the clear-transform-preview to be reused as this is also being called from different places (from ui and data dirs)

PR #10837, opened 2026-07-25
:bug: Keep svg-raw children as uuids on binfile import

### LINKED ISSUE #10496: Moving a file with custom fonts between teams breaks it with a referential integrity (child-not-found) error
### Steps To Reproduce

1. In team A, use a file that uses **custom fonts** (not the default ones) and also contains imported SVG content (shapes of type `svg-raw`, e.g. `svg-defs` / `svg-text`). The file where this was observed is: https://design.penpot.dev/#/workspace?team-id=928d26ac-13f6-8105-8004-e199c3e8da42&file-id=e41ba4c1-9e85-816e-8008-41440d51a51d&page-id=e41ba4c1-9e85-816e-8008-41440d51a51e
2. Export the file from team A.
3. Import it into team B (a different team that does **not** have those custom fonts installed).
4. Open the imported file (or run `debug.validate()` in the console).
5. See the file fail to validate / save with a referential integrity error.

Additional notes:
- The same breakage can be reproduced **within the same team** using export → import → `debug.validate()` in the console.
- It does **not** reproduce with a minimal file that only contains a single text using a custom font, so the custom font alone is not enough to trigger it — the presence of imported `svg-raw` content (and likely the font reference inside it) appears to be required.
- The custom fonts used by the file are **not carried over** when the file is moved between teams, which seems to be the root of the problem.

### Expected behavior

Moving a file between teams (export/import) should keep the file valid and openable, even when the destination team is missing the custom fonts used by the file. Missing fonts should degrade gracefully (e.g. fall back to a default font and/or surface a warning), never leave the file in an inconsistent state that fails referential integrity validation and blocks saving.

### Actual behavior

After importing the file into a team that lacks its custom fonts, the file is left in an inconsistent state: several `svg-raw` shapes reference children that do not exist. On the next persisted commit the `update-file` request is rejected with HTTP 400 and a `:referential-integrity` validation error (`:child-not-found`), effectively breaking the file. Some users cannot even open it.

Looking at the committed changes around the failure, the commit starts with a batch of `mod-typography` changes followed by many `mod-obj` changes — consistent with a font/typography reconciliation running on open against a team that is missing the referenced custom fonts.

### Screenshots or video

[report (19).txt](https://github.com/user-attachments/files/29504203/report.19.txt)

[SVGRaw.zip](https://github.com/user-attachments/files/29504226/SVGRaw.zip)

The console shows the validation error (`:child-not-found` on a `:svg-raw` shape) and the commit containing `mod-typography` + `mod-obj` changes.

_No response_

_No response_

- Host: PRE — https://design.penpot.dev
- Version: 2.17.0-RC1-49-gf993f203bd

### Frontend Stack Trace

```console
Context:
--------------------
Hint:     error on validating file referential integrity
Prof ID:  5ff7a650-fbfb-11ea-9439-46c5bd3a6919
Team ID:  28f17539-42ad-80d6-8003-baa3ebf56ae4
File ID:  edd29975-84f7-8153-8008-414e83a45ec2
Version:  2.17.0-RC1-49-gf993f203bd
HREF:     https://design.penpot.dev/#/dashboard/recent?team-id=28f17539-42ad-80d6-8003-baa3ebf56ae4

====================
Data:
--------------------
{:type :validation,
 :code :referential-integrity,
 :hint "error on validating file referential integrity",
 :file-id #uuid "edd29975-84f7-8153-8008-414e83a45ec2",
 :details
 [{:code :child-not-found,
   :hint "Child 0f8e724d-a7ed-8073-8008-4148995d652b not found in parent 0f8e724d-a7ed-8073-8008-4148995d652a",
   :shape
   {:id #uuid "0f8e724d-a7ed-8073-8008-4148995d652a",
    :type :svg-raw,
    :name "svg-defs",
    ...},
   :file-id #uuid "edd29975-84f7-8153-8008-414e83a45ec2",
   :page-id #uuid "e41ba4c1-9e85-816e-8008-41440d51a51e",
   :shape-id #uuid "0f8e724d-a7ed-8073-8008-4148995d652a",
   :args {:parent-id #uuid "0f8e724d-a7ed-8073-8008-4148995d652a", :child-id "0f8e724d-a7ed-8073-8008-4148995d652b"}}
  {:code :child-not-found,
   :hint "Child 0f8e724d-a7ed-8073-8008-41489958036e not found in parent 0f8e724d-a7ed-8073-8008-41489958036d",
   :shape {:id #uuid "0f8e724d-a7ed-8073-8008-41489958036d", :type :svg-raw, :name "svg-defs", ...},
   :shape-id #uuid "0f8e724d-a7ed-8073-8008-41489958036d"}
  {:code :child-not-found,
   :hint "Child 0f8e724d-a7ed-8073-8008-414899527d9f not found in parent 0f8e724d-a7ed-8073-8008-414899527d9e",
   :shape {:id #uuid "0f8e724d-a7ed-8073-8008-414899527d9e", :type :svg-raw, :name "svg-text", ...},
   :shape-id #uuid "0f8e724d-a7ed-8073-8008-414899527d9e"}],
 :uri "https://design.penpot.dev/api/main/methods/update-file?id=edd29975-84f7-8153-8008-414e83a45ec2",
 :status 400}

Trace:
--------------------
Error: http error
  at new $APP.$cljs$core$ExceptionInfo$$ (.../js/shared.js?version=2.17.0-RC1-49-gf993f203bd:1415:120)
  at $APP.$cljs$core$ex_info$cljs$0core$0IFn$0_invoke$0arity$03$$ (.../js/shared.js:1416:358)
  at $APP.$cljs$core$ex_info$cljs$0core$0IFn$0_invoke$0arity$02$$ (.../js/shared.js:1416:131)
  at $APP.$app$main$repo$handle_response$$ (.../js/shared.js:7221:287)
  ...
====================

Last events (most relevant):
--------------------
[":app.main.data.changes/commit"
 ":app.main.data.changes/commit-changes"
 ":app.main.data.workspace/update-page-position-data"
 ...
 ":app.main.data.fonts/fetch-fonts"
 ":app.main.data.fonts/fonts-loaded"
 ":app.main.data.persistence/error"]
```

### Backend Stack Trace

```console
None observed (the failure is surfaced as a `:referential-integrity` validation error returned by `update-file` with HTTP 400).
```

### Additional context

- The original file (team A): https://design.penpot.dev/#/workspace?team-id=928d26ac-13f6-8105-8004-e199c3e8da42&file-id=e41ba4c1-9e85-816e-8008-41440d51a51d&page-id=e41ba4c1-9e85-816e-8008-41440d51a51e
- The file behaves normally in its original team: you can create layers, pages, move shapes, add new shapes, and enter/leave the workspace without issues. The breakage only appears after moving it to another team (export/import).
- Initialization looks normal (e.g. components reported as v1; `enabled="fdata/path-data plugins/runtime design-tokens/v1 variants/v1 layout/grid styles/v2 fdata/objects-map text-editor/v2 tokens/numeric-input render-wasm/v1 components/v2 fdata/shape-data-type"`).
- This was initially suspected to be a render issue, but it is unrelated to render. It was also raised that files created via the MCP might end up invalid; however, the file is valid and editable in its original team, so the trigger appears to be the cross-team export/import of custom fonts rather than MCP creation.
- The failing details consistently point to `svg-raw` shapes (`svg-defs`, `svg-text`) whose declared children are missing after the move.
- Hypothesis: moving a file between teams does not bring along the custom fonts it uses; the resulting font/typography reconciliation on open (the `mod-typography` + `mod-obj` commit) leaves `svg-raw` subtrees with dangling child references, which then fail referential integrity validation. Moving files that use custom fonts across teams should be handled carefully (fonts should travel with the file or be remapped/fallback) so the file is never left invalid.

filipsajdak:

Fixes #10496.

## Root cause

The custom fonts turn out to be a **red herring** - this reproduces with no custom fonts involved at all, and in a single team. What actually breaks the file is the binfile round trip itself.

An `svg-raw` shape can be a container: importing an SVG builds a tree of `svg-raw` shapes, and `cfh/group-like-shape?` (`common/src/app/common/files/helpers.cljc`) explicitly treats an `svg-raw` *with children* as group-like. But `schema:svg-raw-attrs` was an empty map:

```clojure
(def ^:private schema:svg-raw-attrs
  [:map {:title "SvgRawAttrs"}])
```

`:frame`, `:group` and `:bool` all declare `[:shapes [:vector ::sm/uuid]]`; `:svg-raw` did not. So on import the JSON decoder (`decode-shape` in `backend/src/app/binfile/v3.clj`) had no type information for those child ids and left them as **plain strings**:

```
svg-root    children= ["00000000-0000-0003-0000-000000000002"]   <- string
svg-text    (present in :objects, id #uuid "...0003-...0002")
Root Frame  children= [#uuid "...0003-...0001"]                  <- correct
```

The child shape is not lost - the reference just stops resolving against the objects map. `validate-file` then reports `:child-not-found`, and the next persisted change touching that page fails `validate-file-affected!` with HTTP 400 `:referential-integrity`.

That also explains the two things the reporter noticed: a minimal text-only file does not reproduce (nothing is `svg-raw`), and the failing commit looks font-related only because `fix-deleted-fonts` happens to be the *first* write after opening a moved file - any change touching that page would have triggered the same validation.

## Fix

Declare `:shapes` on `schema:svg-raw-attrs`, optional because a leaf `svg-raw` shape has no children.

## Verification

Regression test `import-binfile-v3-preserves-svg-raw-children` in `backend/test/backend_tests/binfile_test.clj`: it builds a file with an `svg-raw` parent and an `svg-raw` child, exports it with `v3/export-files!`, re-imports it, and asserts the child ids come back as uuids and that the imported file passes `validate-file`.

Confirmed it is not a vacuous test - reverting only the schema change makes it fail with exactly the reported error:

```
FAIL import-binfile-v3-preserves-svg-raw-children
expected: (every? uuid? (:shapes root))
  actual: (not (every? uuid? ["00000000-0000-0003-0000-000000000002"]))
...
:code :child-not-found, :hint "Child 00000000-0000-0003-0000-000000000002 not found in parent 00000000-0000-0003-0000-000000000001"
```

With the fix: `backend-tests.binfile-test` 2 tests / 10 assertions / 0 failures, and the full `common` suite 1069 tests / 24453 assertions / 0 failures. `clj-kondo` 0 errors, `cljfmt` clean.

## Scope / follow-ups

- This stops files from being **corrupted on import**. Files already imported before this change still hold string child ids and would need a repair pass - note that the existing `:child-not-found` handler in `repair.cljc` fixes the error by *removing* the child from the parent, which for these files would discard intact content rather than restore the reference.
- The issue also asks that custom fonts travel with a file across teams. That is a separate behaviour and is not addressed here; the graceful-degradation path (`fix-deleted-fonts`) already exists, and with this fix the file stays valid when the destination team lacks the fonts.


madalenapmelo-kp:

niwinz:
Merge to `develop` but direct backport to `staging`.
Thanks :heart:

PR #10875, opened 2026-07-28
:wrench: Add exporter headless backend

### LINKED ISSUE #10966: Add headless render-wasm export path for PDF exports
### Summary

We need to split shared code between the frontend render and the export render, to use in the export build only the necessary code.

We also need to decide how the exporter would be deployed to include this new path. In the beginning, all of this will be under a feature flag.

elenatorro:
### Related Ticket

  https://github.com/penpot/penpot/issues/10046

  ### Summary

  https://github.com/user-attachments/assets/cff6fc5d-0be5-46ef-b05d-4212502ced70

  * Adds an option to the exporter that renders `:is-wasm` exports in-process with render-wasm instead of driving a headless browser. It works for PNG, JPEG, WEBP and PDF (including multi-page PDF from frame exports).
  * The current export path remains the default and is completely untouched unless the new flag is enabled.
  * Extracts the CLJS side of the render-wasm into `common`, so the exporter can drive WASM without pulling in the frontend app.
  * Splits the render-wasm build in two: one tuned for the frontend renderer, one tuned for the exporter.

  #### How it works:

  - The render-wasm build targets `-sENVIRONMENT=web,worker,node`, so the exporter (Node) loads a `render-wasm.js/render-wasm.wasm` built from the same Rust source the browser uses.

  - Request pipeline:
     - renderer wasm fetches the page's objects via get-page
     - serialize the scene into WASM
     - provision resources (team font variants, google fonts through the local proxy, emoji/noto fallback fonts, image fills fetched once and cached across requests)
     - recompute text layout with the real fonts
     - render each object to PNG/JPEG/WEBP/PDF bytes
     - creates a temp file -> the same on-object contract the current exporter uses
     - after each request, the image cache is trimmed to a byte budget (via LRU eviction)

  #### Shared render-wasm module (the big rename in this diff)

  - The exporter needs the byte layouts, serializers and font catalog the workspace uses, but not React/DOM/store. Those namespaces moved out of `frontend/src/app/render_wasm` into `common`: `common/src/app/common/render_wasm`, namespaces `wasm.*` ->
  `app.common.render-wasm.*`.
  - The google fonts catalog moved to `common/resources/fonts`, so both hosts bake the same `gfonts.json`. **Note:** we'll review if we can serve fonts directly from the BE instead of using the file.
  - `shared.js` (enum generated by the `ToCljs` derive) deliberately stays **out** of `common`: it is a per-build artifact, so each host gets the copy from its own render-wasm build (`frontend/src/app/render_wasm/api/shared.js`,
  `exporter/src/app/wasm/shared.js`). The key -> export-name table still lives in `common`.

  #### Separate wasm builds

  The same Rust source now produces two artifacts, differing only in compiler options:

  | Target | Tuned for | Cargo profile | Consumed by |
  | --- | --- | --- | --- |
  | `frontend` | speed | `release` (`-O3`) | `frontend/resources/public/js` |
  | `export` | size | `size` (`-Oz`) | `exporter/resources/wasm` |

  ```sh
  ./build               # both targets, frontend first
  ./build frontend   # workspace / viewer renderer
  ./build export        # headless exporter renderer
```

  - Each target keeps its own CARGO_TARGET_DIR (target/<target>), so switching between them does not invalidate the other's cache.
  - The exporter ships its own artifact instead of reading frontend/resources/public/js, so neither module reaches into the other's build output.

  Important:

  - :svg stays on the browser path even when the flag is enabled, as it needs real svg markup.
  - Concurrency: there is a single shared wasm instance and it's serialized through a promise queue. We're working on a follow-up PR.

  Steps to reproduce

  #### How to activate it

  Note: it's deactivated by default.

The flag `enable-wasm-export` should be active both as penpot flag (`_env`) and as a config flag (resources/public/js/config.js)

  #### How to test it locally

  - Build the exporter's wasm artifact: cd render-wasm && ./build export
  - Important: restart the exporter to apply env variables
  - Export multiple shapes as PNGs -> check a .zip is downloaded
  - Export multiple PDF pages -> check it's generated properly, in order
  - Export text using a google font -> check the glyphs render instead of falling back
  - Export emoji too
  - You can try out new features such as background blur, which should be mostly supported. If there's something that is not supported, we should fix it in a different PR.
  - svg-raw shapes are not yet supported.

  To confirm it works, check the exporter logs and look for backend="wasm":

  `INF [app.renderer] hint="render", type=:pdf, wasm-export=true, backend="wasm"`

### CODE REVIEW

File: exporter/scripts/build [hunk 1]
```diff
@@ -8,6 +8,25 @@ export NODE_ENV=production;
...
+# in the bundle, so build it here rather than depending on some other module's
+# build having run first — same as `frontend/scripts/build` does.
+pnpm run build:wasm;
+
+# Both are produced by `render-wasm/build` and are git-ignored. Verify they
+# arrived where this build expects: `shared.js` is a compile dependency, so a
+# path drift would surface as an opaque dependency trace from `pnpm run build`
+# below rather than the message here.
+WASM_SRC="../frontend/resources/public/js";
+WASM_SHARED="../frontend/common/src/app/render_wasm_common/api/shared.js";
```
elenatorro: I have some doubts about the bulid process here. Maybe we can tell the exporter where to place them when building for deploying, and place them within the exporter directory 🤔

backend/scripts/_env
```diff
@@ -73,6 +73,21 @@ export PENPOT_USER_FEEDBACK_DESTINATION="support@example.com"
 
 export PENPOT_NITRATE_BACKEND_URI=http://localhost:3000/admin-console
 
+# Headless WASM export: render `:is-wasm` exports via the in-process Skia/WASM
+# pipeline (no browser). Reads the render-wasm artifact from PENPOT_WASM_DIR,
+# defaulting to ../frontend/resources/public/js (render-wasm.js/.wasm) — which
+# is where render-wasm/build leaves it in devenv. The docker image ships the
+# artifact inside the exporter bundle and points PENPOT_WASM_DIR at it.
+#
+# PENPOT_INTERNAL_URI must be set with it: the devenv public-uri is https with
```
elenatorro: Note: this is only for the devenv

backend/scripts/_env
```diff
@@ -12,6 +12,10 @@ export PENPOT_MANAGEMENT_API_KEY=super-secret-management-api-key
 # PENPOT_DATABASE_*, PENPOT_REDIS_URI, PENPOT_OBJECTS_STORAGE_*, AWS_*) is owned by
 # docker/devenv/defaults.env and injected via the main service's env block.
 
+if [ -f /home/selfsigned.crt ]; then
+    export NODE_EXTRA_CA_CERTS=/home/selfsigned.crt;
```
elenatorro, line 17: Just for local development, to allow exporter requests

exporter/src/app/renderer/wasm.cljs
```diff
@@ -0,0 +1,451 @@
...
+;; --- shape bundle fetch (backend RPC)
+
+(defn- fetch-objects
+  "Fetches the page's `objects` map from the backend via the `get-page` RPC,
+  using the same auth the exporter uses elsewhere (management key + bearer)."
+  [{:keys [file-id page-id share-id token]}]
+  (let [headers (rpc-headers token)
+        ;; share-id is an OPTIONAL uuid on the backend; it must be omitted when
+        ;; absent, not sent as nil (nil fails the uuid schema).
+        body    (t/encode-str (cond-> {:file-id file-id
```
superalex, line 141: fetch-objects calls get-page with only file-id / page-id / share-id, so the whole page is loaded, serialized, and provisioned (images, fallback fonts, text relayout) even when the export targets a small subset.

get-page already accepts optional :object-id as uuid | #{uuid} and prunes to each root + children — the same contract the browser render path uses. params already has the export roots in :objects.

Could we pass them through here?

PR #10891, opened 2026-07-28
:sparkles: Add basic WASM SVG export path

### LINKED ISSUE #11377: Add basic single-file SVG export via WASM render
### Description

Ship the first-cut single-file SVG export path through the WASM/WebGL renderer under the `wasm-export` flag.

Covers:

- End-to-end wiring (`_render_shape_svg` + frontend `export-svg`)
- Solid leaf fills for basic shapes
- Group/frame traversal
- Native SVG wrappers for opacity and mix-blend-mode
- Frame clip-content via `<clipPath>`
- Correct page size for unclipped frames (extended bounds include overflowing children)
- Headless snapshot tests and the `preview-snapshots` visual gallery

Out of scope for this issue: shadows, layer blur, masks, image fills/strokes, text, and special stroke kinds that need `save_layer` workarounds.

### Use case

Designers can export a selected object or frame to a real vector SVG from the WASM renderer for the basic cases, keeping export consistent with the canvas when WebGL is active.

### Affected versions

develop / upcoming with `wasm-export` and `render-wasm/v1`

superalex:
Closes: #11377 

## What

- Adds a first-cut client-side SVG export path through render-wasm.
- Covers the basic cases only: solid leaf shapes, group/frame traversal, native SVG wrappers for opacity and mix-blend-mode, and frame `clip content` via `<clipPath>`.
- Leaves shadows, layer blur, masks, and save_layer stroke/fill workarounds for follow-up PRs.

## Why

Skia's SVG backend drops `save_layer` content, so a faithful vector SVG export needs its own compositor. Landing the thin end-to-end path first (FFI + frontend + basic markup) keeps review small and lets later PRs add effect re-emission incrementally.

## How

- **SVG compositor:** short-lived Skia SVG fragments for leaf geometry; native `<g>` wrappers for opacity/blend; frame clip as `<clipPath>`.
- **vector.rs:** only expose `render_leaf_content` / `draw_shape_geometry` as `pub(super)` — no SVG-specific defer logic in the shared renderer.
- **Wiring:** `_render_shape_svg` + `export-svg` on the existing wasm export path; headless `insta` snapshots plus `./preview-snapshots` for visual review.

### DIFF
--- a/frontend/src/app/render_wasm/api.cljs
+++ b/frontend/src/app/render_wasm/api.cljs
@@ -2870,6 +2870,25 @@
       (mem/free)
       result)))
 
+(defn render-shape-svg
+  [shape-id scale]
+  (when (initialized?)
+    (let [buffer (uuid/get-u32 shape-id)
+          offset
+          (h/call wasm/internal-module "_render_shape_svg"
+                  (aget buffer 0)
+                  (aget buffer 1)
+                  (aget buffer 2)
+                  (aget buffer 3)
+                  scale)
+
+          heap (mem/get-heap-u8)
+          heapu32 (mem/get-heap-u32)
+          length (aget heapu32 (mem/->offset-32 offset))
+          result (dr/read-image-bytes heap (+ offset 4) length)]
+      (mem/free)
+      result)))
+
 (defn init-wasm-module
   [module]
   (let [default-fn (unchecked-get module "default")
--- a/frontend/src/app/render_wasm/api.cljs
+++ b/frontend/src/app/render_wasm/api.cljs
@@ -2688,6 +2688,24 @@
     (mem/free)
     result))
 
+(defn render-shape-svg
+  [shape-id scale]
+  (let [buffer (uuid/get-u32 shape-id)
+        offset
+        (h/call wasm/internal-module "_render_shape_svg"
+                (aget buffer 0)
+                (aget buffer 1)
+                (aget buffer 2)
+                (aget buffer 3)
+                scale)
+
+        heap (mem/get-heap-u8)
+        heapu32 (mem/get-heap-u32)
+        length (aget heapu32 (mem/->offset-32 offset))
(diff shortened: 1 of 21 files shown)


frontend/src/app/render_wasm/api.cljs
```diff
@@ -2870,6 +2870,24 @@
       (mem/free)
       result)))
 
+(defn render-shape-svg
+  [shape-id scale]
```
elenatorro, line 2874: I think we could check `initialize?` here as we do in other render paths

render-wasm/src/render/svg/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+    let page_w = bounds.width() * scale;
+    let page_h = bounds.height() * scale;
+    let rect = skia::Rect::from_xywh(0., 0., page_w, page_h);
+
+    let (defs, body) = render_body(shared, id, tree, scale, rect, -bounds.left(), -bounds.top())?;
+
+    let mut out = String::with_capacity(body.len() + defs.len() + 256);
+    out.push_str("<?xml version=\"1.0\" encoding=\"utf-8\" ?>\n");
+    out.push_str(&format!(
+        "<svg xmlns=\"http://www.w3.org/2000/svg\" xmlns:xlink=\"http://www.w3.org/1999/xlink\" \
```
elenatorro, line 56: should we add viewBox property here? (I'm asking cause I'm not 100% sure of it)

elenatorro (approved):

frontend/src/app/render_wasm/api.cljs
superalex: yes!, thanks!

render-wasm/src/render/svg/mod.rs
superalex: absolutely! :)

elenatorro (approved):

PR #10896, opened 2026-07-28
:bug: Fix crash when a nil shape id reaches WASM modifiers

### LINKED ISSUE #10894: Workspace crashes with "Cannot read properties of null (reading '__u32_buffer')" when applying transform modifiers in the WASM renderer
### Description

The workspace crashes with `TypeError: Cannot read properties of null (reading '__u32_buffer')` when a transform is applied through the WASM renderer (`render-wasm/v1`) and the modifier tree contains an entry with a missing (nil) shape id. The crash happens in `app.common.uuid/get-u32` while `app.render-wasm.mem/write-uuid` writes the shape id to the WASM heap during `propagate-modifiers`.

Observed in production via an error report during a variants flow (duplicate/rename a variant followed by repositioning and resizing of the variant container). In production builds the asserts that would catch the nil id upstream are elided, so a nil id (e.g. `(:parent-id shape)` of a shape that is not present in the page objects, used by the variant reposition/resize step of `add-new-variant`) reaches the WASM heap write and crashes the workspace.

Stack trace (top):

```
TypeError: Cannot read properties of null (reading '__u32_buffer')
  at app$common$uuid$get_u32
  at app$render_wasm$mem$heap32$write_uuid
  at app$render_wasm$api$propagate_modifiers
  at app$main$data$workspace$modifiers$apply_wasm_modifiers (WatchEvent watch)
```

### Steps to reproduce

Exact reproduction is not known (taken from a production error report). Suspected sequence:

1. Open a file with the WASM renderer (`render-wasm/v1`) active.
2. Add or duplicate a variant so that the variant reposition/resize step runs when the newly created variant shape is not present in the page objects (or otherwise trigger `update-dimensions` / `update-position` with a nil shape id).
3. The workspace crashes with an Internal Error page.

### Expected behavior

Modifier entries with missing/nil shape ids are ignored (never reach the WASM heap write) and the workspace does not crash.

### Affected versions

2.17.1-RC1 (design.penpot.app)

niwinz:

## What

Fixes a production workspace crash `TypeError: Cannot read properties of null (reading '__u32_buffer')` that happened when applying transform modifiers through the WASM renderer.

Fixes #10894

## Why

A modifier tree containing a `nil` shape id reached `wasm.api/propagate-modifiers` / `wasm.api/set-structure-modifiers`, and `mem.h32/write-uuid` crashed calling `uuid/get-u32` on nil while writing to the WASM heap. Production builds elide the upstream asserts that would catch this, so a nil id (e.g. `(:parent-id shape)` of a missing shape in the variants reposition/resize flow) sailed through.

## How

- `apply-wasm-modifiers` and `set-wasm-modifiers` sanitize the modif-tree, dropping nil-keyed entries (with a warning log) before anything crosses the WASM boundary.
- `parse-geometry-modifiers` skips nil ids and `parse-structure-modifiers` filters entries with nil `:id`/`:parent`, as defense in depth.
- The variant events `resposition-and-resize-variant` and `add-new-variant` bail out early when the involved shape/container is missing (the likely nil source).
- Adds a reproduction test (`frontend-tests.logic.wasm-modifiers-nil-id-test`) that failed before the fix, asserting no nil id crosses the WASM boundary while valid shapes in the same modif-tree are still processed.

Closes #10894

### DIFF
--- a/frontend/src/app/main/data/workspace/modifiers.cljs
+++ b/frontend/src/app/main/data/workspace/modifiers.cljs
@@ -15,6 +15,7 @@
    [app.common.geom.point :as gpt]
    [app.common.geom.rect :as grc]
    [app.common.geom.shapes :as gsh]
+   [app.common.logging :as log]
    [app.common.math :as mth]
    [app.common.types.component :as ctk]
    [app.common.types.container :as ctn]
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/data/workspace/modifiers.cljs
```diff
@@ -885,29 +901,29 @@
...
+              ;; needs to have the updated children info.
+              ;; `update-layout? false`: recalculating a bool path can never change
+              ;; `:hidden`, and the layout check would recompute the whole boolean
+              ;; path in WASM once per bool shape just to find that out.
+              (dwsh/update-shapes bool-ids path/update-bool-shape (assoc options
+                                                                         :with-objects? true
+                                                                         :update-layout? false)))
+
+             (if undo-transation?
+               (rx/of (dwu/commit-undo-transaction undo-id))
```
EvaMarco, line 925: This undo-id is not defined.

hirunatan (approved):

PR #10907, opened 2026-07-29
:bug: Set correct upstream Host header for internal proxy_pass locations

### LINKED ISSUE #10835: NGinX Frontend forwards the client's Host header to backend/exporter, breaking Istio STRICT mTLS routing (falls back to PassthroughCluster)
### Steps To Reproduce

1. Deploy Penpot via the official Helm chart in a Kubernetes cluster with Istio service mesh enabled (sidecar injection on) and PeerAuthentication set to STRICT (either mesh-wide or namespace-wide).
2. Access the Penpot instance through the frontend (nginx), e.g. GET /api/main/methods/get-profile.
3. Observe that requests proxied by nginx to the backend (/api, /assets, /readyz, /ws/notifications) or to the exporter (/api/export) fail intermittently or systematically, while the frontend static assets and pages load fine.
4. From inside the frontend pod, run `curl http://penpot-backend:6060/api/main/methods/get-profile` (using the Service DNS name directly) — this succeeds.
5. Run `curl http://localhost:8080/api/main/methods/get-profile` (going through nginx) — this fails with `upstream connect error or disconnect/reset before headers. reset reason: connection termination` (or `connection timeout`, or `filter_chain_not_found` in the backend's istio-proxy logs, depending on the mTLS mode being toggled during testing).

### Expected behavior

Requests proxied by nginx to the backend/exporter should reach their destination reliably, regardless of the Host header sent by the end client (browser), when running behind a service mesh like Istio with strict mTLS enabled.

### Actual behavior

The nginx.conf.template sets a single global directive at the http{} level:

    proxy_set_header Host $http_host;

This forwards the original client-facing Host header (e.g. "penpot-preprod.example.com") on every internal proxy_pass call, including the ones targeting the internal backend/exporter services (/api, /assets, /readyz, /ws/notifications, /api/export), instead of the actual upstream service hostname declared in $PENPOT_BACKEND_URI / $PENPOT_EXPORTER_URI.

Under a plain Kubernetes Service (no mesh), this has no effect because routing is purely IP/port based. But with Istio (and mTLS in STRICT mode), Envoy's sidecar needs the outbound HTTP request's Host/:authority (or SNI) to match a known cluster/service host to route the connection through the correct mTLS outbound cluster. Since nginx sends the external Host instead of the internal one, Envoy cannot match it to any known cluster and falls back to "PassthroughCluster", sending the request as plaintext. The backend's inbound listener, configured for STRICT mTLS, then rejects/hangs on this plaintext connection (filter_chain_not_found / connection_termination / connection timeout depending on the exact PeerAuthentication mode).

We confirmed this is the root cause: setting `proxy_set_header Host $proxy_host;` (or the destination host explicitly) in each of the relevant locations (/api, /assets, /readyz, /ws/notifications, /api/export) resolves the issue completely.

This affects any service-mesh setup with mTLS/host-based routing (Istio, and likely Linkerd or others with similar behavior), not just our specific environment.

### Screenshots or video

(none — this is a backend/networking issue, not visible in the UI)

_No response_

_No response_

- Host: self-hosted (internal preprod instance)
- Deployment method: official Penpot Helm chart (penpot-1.7.0 / app version 2.17.0)
- Kubernetes: OVHcloud managed Kubernetes
- Service mesh: Istio 1.30.3, PeerAuthentication mode STRICT
- Relevant nginx.conf.template snippet (frontend container):

    http {
        ...
        proxy_set_header Host $http_host;
        ...
        location /api {
            proxy_pass $PENPOT_BACKEND_URI/api;
            proxy_buffering off;
        }
        location /assets {
            proxy_pass $PENPOT_BACKEND_URI/assets;
            ...
        }
        location /api/export {
            proxy_pass $PENPOT_EXPORTER_URI;
        }
        location /readyz {
            proxy_pass $PENPOT_BACKEND_URI$request_uri;
        }
        location /ws/notifications {
            proxy_set_header Connection 'upgrade';
            proxy_pass $PENPOT_BACKEND_URI/ws/notifications;
        }
    }

### Frontend Stack Trace

```console
(none — no JS error, the issue happens at the reverse-proxy/network layer)
```

### Backend Stack Trace

```console
istio-proxy (backend sidecar) access log:
[timestamp] "GET /api/main/methods/get-profile HTTP/1.1" 503 UC upstream_reset_before_response_started{connection_termination} ... PassthroughCluster ...

istio-proxy (backend sidecar), filter_chain_not_found case:
[timestamp] "- - -" 0 NR filter_chain_not_found - "-" 0 0 0 - "-" "-" "-" "-" "-" - - <backend_pod_ip>:6060 <frontend_pod_ip>:<port> - -

Browser-visible error:
upstream connect error or disconnect/reset before headers. reset reason: connection termination
```

### Additional context

Root cause confirmed step by step:
1. Ruled out PostgreSQL, DNS caching, sidecar UID exclusion, and stale mTLS certificates as causes.
2. Confirmed via istioctl proxy-config that the outbound cluster for penpot-backend is correctly configured with proper SPIFFE SAN validation.
3. Captured Envoy debug logs (connection/filter/http) on the backend sidecar and confirmed the request from curl (with an explicit Host header matching the service name) succeeds and is correctly routed as `outbound|6060||penpot-backend...`, while the same request without an explicit/correct Host header falls back to PassthroughCluster.
4. Verified in nginx.conf that the only relevant difference between a working manual curl and the actual nginx-proxied request is the Host header sent upstream ($http_host vs the internal service name).

Suggested fix: explicitly override the Host header in the locations that proxy to internal services (backend, exporter), e.g.:

    location /api {
        proxy_set_header Host $proxy_host;
        proxy_pass $PENPOT_BACKEND_URI/api;
        proxy_buffering off;
    }

(and similarly for /assets, /readyz, /ws/notifications, /api/export)

Happy to submit a PR with this change if maintainers agree with the approach.

smalot:
## Description

Fixes #10835

The nginx frontend template (`nginx.conf.template`) sets a single global directive at the `http {}` level:

```nginx
proxy_set_header Host $http_host;
```

This forwards the client-facing Host header (e.g. the public domain the browser used to reach Penpot) on **every** `proxy_pass`, including the ones targeting internal services: backend (`/api`, `/assets`, `/readyz`, `/ws/notifications`) and exporter (`/api/export`).

This works fine on a plain Kubernetes Service, since routing there is purely IP/port based and the Host header is irrelevant to it. However, in a service-mesh setup with mTLS enforced (e.g. Istio with `PeerAuthentication` in `STRICT` mode), the sidecar proxy relies on the outbound request's Host/`:authority` to match it to the correct upstream cluster/service and originate mTLS accordingly. Since nginx sends the *external* Host instead of the internal service hostname, the mesh sidecar cannot match the request to a known cluster, and the connection falls back to a generic passthrough path — sent in plaintext. The backend's inbound listener, expecting strict mTLS, then rejects or hangs on that plaintext connection, resulting in random-looking `502`/`503`/connection-reset errors on `/api`, `/assets`, `/ws/notifications`, etc.

## Change

Explicitly override the `Host` header in each location that proxies to an internal service, using nginx's `$proxy_host` variable (which always reflects the host:port declared in that location's `proxy_pass`, regardless of the incoming request's Host):

```nginx
location /api {
    proxy_set_header Host $proxy_host;
    proxy_pass $PENPOT_BACKEND_URI/api;
    proxy_buffering off;
}
```

Applied to:
- `/api`
- `/assets`
- `/api/export`
- `/readyz`
- `/ws/notifications`

`/admin-console` was intentionally left untouched since it already sets its own explicit `Host` header, and `@handle_redirect` was left untouched since it proxies to external storage (S3-like), where the original Host override logic still applies.

## Why this doesn't break non-mesh deployments

`$proxy_host` simply resolves to the same value nginx would use internally to establish the upstream connection (host:port from `proxy_pass`). For plain Kubernetes/Docker-compose deployments without a service mesh, this is functionally equivalent to what the backend/exporter already expect (and is in fact closer to standard reverse-proxy best practice for internal service calls than forwarding the client's original Host).

## Testing

Reproduced and validated the fix in a Kubernetes cluster running Istio 1.30.3 with `PeerAuthentication` set to `STRICT`:
- **Before**: `curl http://localhost:8080/api/main/methods/get-profile` (through nginx) failed with `upstream connect error or disconnect/reset before headers. reset reason: connection termination`, while calling the backend service directly (bypassing nginx) succeeded.
- **After**: the same request through nginx returns `200 OK`, and Envoy logs confirm the request is now correctly routed through the mesh's outbound cluster (`outbound|6060||penpot-backend...`) instead of falling back to `PassthroughCluster`.

Also verified that WebSocket upgrade (`/ws/notifications`) and the exporter path (`/api/export`) continue to work as expected.

## Related issue

Full root-cause investigation and diagnostic steps are documented in #10835.

### DIFF
--- a/docker/images/files/nginx.conf.template
+++ b/docker/images/files/nginx.conf.template
@@ -111,6 +111,7 @@ http {
         }
 
         location /assets {
+            proxy_set_header Host $proxy_host;
             proxy_pass $PENPOT_BACKEND_URI/assets;
             recursive_error_pages on;
             proxy_intercept_errors on;
@@ -127,10 +128,12 @@ http {
         }
 
         location /api/export {
+            proxy_set_header Host $proxy_host;
             proxy_pass $PENPOT_EXPORTER_URI;
         }
 
         location /api {
+            proxy_set_header Host $proxy_host;
             proxy_pass $PENPOT_BACKEND_URI/api;
             proxy_buffering off;
         }
@@ -142,10 +145,12 @@ http {
 
         location /readyz {
             access_log off;
+            proxy_set_header Host $proxy_host;
             proxy_pass $PENPOT_BACKEND_URI$request_uri;
         }
 
         location /ws/notifications {
+            proxy_set_header Host $proxy_host;
             proxy_set_header Upgrade $http_upgrade;
             proxy_set_header Connection 'upgrade';
             proxy_pass $PENPOT_BACKEND_URI/ws/notifications;


madalenapmelo-kp:

yamila-moreno:
Hi @smalot, thanks for your contribution. We've checked the change and we'd like to merge it. In order to do this properly, I'd like to suggest a couple of changes:
- change the target so it's `penpot:staging`; this file had some changes and rebase is advised
- update the commit message to fit the contribution guidelines

Cheers!

yamila-moreno:
I'm taking this [PR here](https://github.com/penpot/penpot/pull/11233) so I can change the target and the commit message. Cheers!

PR #10947, opened 2026-07-30
:bug: Fix stale DNS caching in frontend nginx MCP proxy

### LINKED ISSUE #10946: MCP WebSocket proxy fails after penpot-mcp container restarts independently (stale nginx DNS resolution)
## Description

In multi-container deployments (Docker Compose or similar), `wss://<host>/mcp/ws?userToken=...` fails to connect from the browser after the `penpot-mcp` container has restarted or been recreated independently of `penpot-frontend` — e.g. after bumping the `penpotapp/mcp` image tag, an OOM restart, or an orchestrator rescheduling just that container.

## Root cause

`docker/images/files/nginx-mcp-locations.conf.template` uses a plain `proxy_pass` target:

```
location /mcp/ws {
    ...
    proxy_pass $PENPOT_MCP_URI_WS;
    proxy_http_version 1.1;
}
```

`$PENPOT_MCP_URI_WS` / `$PENPOT_MCP_URI` here are **shell** variables substituted once by `envsubst` in `nginx-entrypoint.sh` at container startup — by the time nginx reads the generated `/etc/nginx/overrides/server.d/mcp-locations.conf`, this is a plain string literal (e.g. `proxy_pass http://penpot-mcp:4402;`).

nginx resolves a literal `proxy_pass` hostname exactly once, when the config is loaded, and never re-checks it — regardless of the `resolver 127.0.0.11 valid=10s;` directive already present in `overrides/http.d/resolvers.conf` (generated in the same entrypoint script). That resolver directive only takes effect for **nginx-level variables** evaluated per-request; it currently has no effect on these three `/mcp/*` locations.

So if `penpot-mcp` gets a new container IP after `penpot-frontend`'s nginx has already started and cached the old one, nginx keeps forwarding to the dead address until `penpot-frontend` itself is restarted. This surfaces in `docker logs penpot-frontend` as:

```
connect() failed (111: Connection refused) while connecting to upstream, ... upstream: "http://<stale-ip>:4402/mcp/ws?...", host: "..."
```

and to the end user as a browser console error like:

```
Firefox can't establish a connection to the server at wss://<host>/mcp/ws?userToken=...
```

## Steps to reproduce

1. Run Penpot via Docker Compose with `enable-mcp` and separate `penpot-frontend`/`penpot-mcp` containers on a shared network.
2. Confirm MCP works (enable it in Your account → Integrations → MCP Server, open a project).
3. Recreate/restart only `penpot-mcp` such that it receives a different container IP (e.g. another container temporarily occupies its old IP, or in a larger network with IP churn this happens naturally over time).
4. Reload the Penpot workspace and try to (re)connect MCP — the WebSocket fails, and `docker logs penpot-frontend` shows `connect() failed (111: Connection refused)` pointing at the old IP.

## Suggested fix

Route each `/mcp/*` location's `proxy_pass` through a real nginx variable via `set`, so the existing `resolver` directive actually re-resolves the upstream within its TTL instead of caching it for the container's lifetime:

```
location /mcp/ws {
    proxy_set_header Connection 'upgrade';
    set $mcp_ws_backend $PENPOT_MCP_URI_WS;
    proxy_pass $mcp_ws_backend;
    proxy_http_version 1.1;
}
```

(same pattern for `/mcp/stream` and `/mcp/sse`). I've verified this fix locally: force-recreated the `penpot-mcp` container onto a different IP while leaving `penpot-frontend` untouched, and the `/mcp/ws` WebSocket upgrade kept returning `101 Switching Protocols` both immediately and after the resolver's TTL window, with no frontend restart needed.

I'll open a PR with this fix.

## Environment

- Penpot `2.17.0` (Docker images)
- Docker Compose, `penpot-frontend`/`penpot-mcp`/`penpot-backend`/`penpot-exporter` as separate containers on a shared bridge network

780Farva:
## Summary

Closes #10946.

`docker/images/files/nginx-mcp-locations.conf.template` proxies each `/mcp/*` location to a plain string (`proxy_pass http://penpot-mcp:4402;` after `envsubst` bakes in the shell variable at container startup). nginx resolves a literal `proxy_pass` hostname once, at config load, and never re-checks it — so the `resolver 127.0.0.11 valid=10s;` directive already generated in `overrides/http.d/resolvers.conf` has no effect here (it only re-resolves nginx-level variables evaluated per-request).

In multi-container deployments, if `penpot-mcp` restarts/is recreated independently of `penpot-frontend` (image bump, OOM, orchestrator reschedule) and gets a new IP, `penpot-frontend`'s nginx keeps forwarding to the dead old address until the frontend container itself is restarted. Users see `wss://<host>/mcp/ws` fail to connect from the browser, with `connect() failed (111: Connection refused)` in the frontend's nginx logs.

- Route each `/mcp/ws`, `/mcp/stream`, `/mcp/sse` location's `proxy_pass` through a `set $var ...;` first, so it's a real nginx variable — this lets the pre-existing `resolver` directive re-resolve `penpot-mcp` within its 10s TTL instead of caching the address for the container's lifetime.
- No other files changed; `nginx-entrypoint.sh`'s `envsubst` call and variable names are untouched.

## Testing notes

Reproduced against a live Docker Compose deployment (`penpot-frontend`/`penpot-mcp`/`penpot-backend`/`penpot-exporter` on a shared bridge network, Penpot 2.17.0):

1. Confirmed the bug: `docker logs penpot-frontend` showed `connect() failed (111: Connection refused)` pointing at a stale `penpot-mcp` IP after the mcp container had been recreated ~5 minutes after the frontend started.
2. Applied this template change, recreated `penpot-frontend` to regenerate the config, verified with `nginx -t` that the generated config is still valid, and confirmed `docker exec penpot-frontend cat /etc/nginx/overrides/server.d/mcp-locations.conf` shows the `set`/`proxy_pass` pair with real values substituted.
3. Forced a genuine IP change on `penpot-mcp` (a throwaway container temporarily occupied its old address) **without restarting `penpot-frontend`**, and confirmed the `/mcp/ws` WebSocket upgrade kept returning `101 Switching Protocols` both immediately and after the resolver's 10s TTL window.

No UI-visible change; this only affects internal nginx→penpot-mcp proxying.

### DIFF
--- a/docker/images/files/nginx-mcp-locations.conf.template
+++ b/docker/images/files/nginx-mcp-locations.conf.template
@@ -1,16 +1,19 @@
 location /mcp/ws {
     proxy_set_header Upgrade $http_upgrade;
     proxy_set_header Connection 'upgrade';
-    proxy_pass $PENPOT_MCP_URI_WS;
+    set $mcp_ws_backend $PENPOT_MCP_URI_WS;
+    proxy_pass $mcp_ws_backend;
     proxy_http_version 1.1;
 }
 
 location /mcp/stream {
-    proxy_pass $PENPOT_MCP_URI/mcp;
+    set $mcp_stream_backend $PENPOT_MCP_URI/mcp$is_args$args;
+    proxy_pass $mcp_stream_backend;
     proxy_http_version 1.1;
 }
 
 location /mcp/sse {
-    proxy_pass $PENPOT_MCP_URI/sse;
+    set $mcp_sse_backend $PENPOT_MCP_URI/sse$is_args$args;
+    proxy_pass $mcp_sse_backend;
     proxy_http_version 1.1;
 }


madalenapmelo-kp:

PR #10957, opened 2026-07-30
:bug: Add dedicated RPC methods for plugin registry operations

### LINKED ISSUE #10952: Add dedicated RPC methods for plugin registry operations with permission validation
### Description

The plugin system currently uses the generic `update-profile-props` RPC method to persist plugin registry data. This endpoint accepts arbitrary prop updates, allowing injection of plugin entries.

niwinz:
## What

Adds dedicated RPC methods for plugin management and validates plugin integrity on open:

1. **Closed permission schema** - Plugin permissions are now restricted to a fixed set of valid values
2. **Dedicated RPC methods** - New `add-profile-plugin` and `remove-profile-plugin` methods replace the generic profile update for plugin operations
3. **Integrity validation on open** - When opening a plugin, the manifest is re-fetched and re-confirmation is required if it differs from the stored version

## Why

The plugin system previously used a generic profile update method to persist plugin registry data. This approach had several issues:

- Plugin permissions were stored as arbitrary strings without validation
- Plugin operations were not atomic (full registry overwrite)
- No integrity check when opening plugins with remote manifests

The new approach ensures:
- Only valid permissions can be stored (schema validation)
- Plugin operations are atomic and validated (dedicated RPC methods)
- Plugins with changed manifests require explicit re-confirmation

## How

**Backend**
- New `add-profile-plugin` RPC method with atomic upsert and permission validation
- New `remove-profile-plugin` RPC method with atomic deletion
- Registered in RPC method resolution

**Common**
- Added `valid-permissions` set with 11 valid permission strings
- Created closed permission schema using enum to restrict values

**Frontend**
- Migrated from full registry overwrite to granular RPC calls
- Optimistic updates with rollback on failure
- Modified plugin opening to require re-confirmation when manifest differs
- On fetch error, show warning instead of loading with stale data
- Bundled plugins (no URL) skip validation

Closes #10952

### DIFF
--- a/backend/src/app/rpc.clj
+++ b/backend/src/app/rpc.clj
@@ -400,6 +400,7 @@
           'app.rpc.commands.management
           'app.rpc.commands.media
           'app.rpc.commands.nitrate
+          'app.rpc.commands.plugins
           'app.rpc.commands.profile
           'app.rpc.commands.projects
           'app.rpc.commands.search
--- a/backend/src/app/rpc.clj
+++ b/backend/src/app/rpc.clj
@@ -388,6 +388,7 @@
           'app.rpc.commands.management
           'app.rpc.commands.media
           'app.rpc.commands.nitrate
+          'app.rpc.commands.plugins
           'app.rpc.commands.profile
           'app.rpc.commands.projects
           'app.rpc.commands.search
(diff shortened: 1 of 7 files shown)


hirunatan (approved):

niwinz:
I think i will retarget this to 2.19.0

Issue #10753, opened 2026-07-17
Remote MCP: websocket connects but live browser commands never arrive (execute_code times out)


lucaspbdesign-source:
## Summary

Remote MCP server connects and responds to stateless calls, and the browser plugin establishes a healthy websocket after connecting via **File → MCP Server → Connect**. But commands that require a round-trip into the browser (`execute_code` in the MCP tool spec) never arrive on that websocket — confirmed via the Network tab, zero frames in either direction — and every such call times out after 30s. This reproduces consistently, including after regenerating the MCP key and reconnecting the plugin fresh.

## Environment

- Client: Claude Code (VS Code extension)
- MCP config: remote HTTP transport, per the documented Claude Code format:
  ```json
  "penpot": {
    "transport": "http",
    "url": "https://design.penpot.app/mcp/stream?userToken=..."
  }
  ```
- Browser: (fill in — Chrome/Firefox/Safari/etc., version)
- OS: macOS (Darwin)
- Penpot file: single file, single tab open

## Steps to reproduce

1. Enable MCP Server under Account → Integrations, generate a key.
2. Add the server to Claude Code config as above, restart the client.
3. Open a Penpot file, connect via **File → MCP Server → Connect** (menu shows "Manage (Status: enabled)" with a green dot afterward).
4. From the MCP client, call a tool that requires live browser execution (e.g. run arbitrary plugin-API JS and return a trivial value like `1`).
5. Call times out after 30s.

## What we checked

- **Static/stateless MCP tool calls succeed instantly** (e.g. API doc lookups served directly by the MCP server) — confirms the MCP server itself is reachable and the config/token are valid.
- **Websocket connects successfully**: DevTools → Network → filtering for `ws` shows a request to `design.penpot.app` with status `101`, type `websocket`, sitting in the expected "Pending" (i.e. open) state.
- **Zero messages ever cross the socket**: opened the websocket's Messages/Frames panel in DevTools, retried the live command — no inbound or outbound frame appeared during the entire 30s timeout window.
- **Browser console stays silent** during the failed call — no errors, warnings, or logs of any kind.
- **Ruled out stale token**: regenerated the MCP key from Integrations settings, updated the client config, restarted the client, reconnected the plugin — identical failure.
- **Ruled out stale plugin session**: explicitly clicked Disconnect then Connect again in the file's MCP server menu immediately before retrying — identical failure.
[...]

PR #10967, opened 2026-07-30
:bug: Fix MCP issues pertaining to connection handling and task dispatching via Redis

### LINKED ISSUE #10953: MCP tool call timeout is too low for some operations
The current timeout of 30 seconds is not sufficient for some calls, especially in larger files. We should make the timeout a configuration option.

### LINKED ISSUE #10958: After leaving a file in Penpot, MCP requests run into timeouts
### Steps To Reproduce

1. With MCP integration enabled, open a file in Penpot.
2. Check that MCP server can be used to access the file.
3. In Penpot, leave the file by going back to the dashboard.
4. Issue new request via the MCP server. 
5. Observe the request timing out.

### Expected behavior

After leaving the file, the MCP server should immediately report that there is no connection to Penpot.

### Actual behavior

Requests time out, wasting time.

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

### LINKED ISSUE #10961: Duplicate WebSocket MCP connection attempts deregistered the original connection's routing entries
### Description

In multi-user mode, the MCP server allows only one plugin WebSocket connection per user token and rejects any further connection attempt using the same token. However, the cleanup performed while rejecting the newcomer deregisters the wrong connection: it deletes the token's routing entry (`clientsByToken`) and the token's Redis request-channel subscription, both of which belong to the established, healthy connection.

The impact is that a single rejected duplicate attempt silently breaks the working MCP integration for that user. The established plugin connection remains open and keeps heartbeating, and the Penpot UI continues to show it as connected, but the server can no longer route any task to it: every MCP tool call for that user fails (previously with a 30-second timeout per request; since the fail-fast fix, with an immediate "No Penpot instance connected for user token" error), even though a valid, live plugin connection exists. Because duplicate attempts occur in normal operation (a second browser tab, or the plugin's automatic reconnect while the old socket is still open), users hit this without doing anything unusual, and there is no recovery other than reconnecting. Recovery is racy as well: after the erroneous eviction the token appears free, so whichever side reconnects first claims it — if a stale plugin instance wins, tasks are routed to an instance that cannot execute them and every request times out.

### Steps to reproduce

1. Run the MCP server in multi-user mode and connect the Penpot MCP plugin with a valid user token.
2. Open a second plugin connection with the same token (e.g. open the same Penpot account in a second tab, or trigger a reconnect while the first socket is still open).
3. The second connection is rejected as a duplicate (expected). Now issue any MCP tool call for that user.

### Expected behavior

Rejecting a duplicate connection must not affect the established connection: the existing plugin connection stays registered for task routing and MCP tool calls continue to work.

### Affected versions

develop (MCP server, multi-user mode)

opcode81:
# Summary

This PR resolves several issues in the MCP server:

* #10958  
  In multi-instance mode (where tasks are dispatched via Redis), the case where a connection from Penpot was no longer present was not correctly handled, resulting in a timeout instead of an immediate failure indicating no connected instances.
* #10961  
  Rejection of multiple connections for the same token was incorrectly handled: Removal of the duplicate connection also incorrectly removed entries for the original connection (token association, Redis association).
* #10953  
  Tool executions via the plugin connection used a timeout of 30 seconds, which was insufficient for some longer-running tasks. The timeout is now configurable via an env. variable and the default was raised to 120 seconds.
  
Additional issue which is likely fixed by these changes: #10753

NOTE: These are 3 logically separate changes, so squashing not recommended.

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -272,6 +272,7 @@ The Penpot MCP server can be configured using environment variables.
 | `PENPOT_MCP_REPL_PORT`                           | Port for the REPL server (development/debugging)                           | `4403`         |
 | `PENPOT_MCP_REMOTE_MODE`                         | Enable remote mode (disables file system access). Set to `true` to enable. | `false`        |
 | `PENPOT_MCP_DEVENV`                              | Enable Penpot development environment tools. Set to `true` to enable.      | `false`        |
+| `PENPOT_MCP_TOOL_TIMEOUT_S`                      | Timeout, in seconds, for tool calls dispatched to the Penpot plugin        | `120`          |
 | `PENPOT_MCP_EXPORT_SHAPE_MAX_PARALLEL_REQUESTS`  | Maximum number of parallel export shape requests (multi-user mode only).   | `0` (no limit) |
 | `PENPOT_MCP_REDIS_URI`                           | Redis connection URI (e.g. `redis://host:6379`) enabling multi-instance horizontal scaling via Redis pub/sub task routing (multi-user mode only). When unset, the server runs in single-instance mode, requiring the plugin and MCP client to connect to the same instance. | (unset)        |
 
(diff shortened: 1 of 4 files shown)


opcode81:
@niwinz the issues fixed here are likely to negatively affect quite a few users of the remote MCP server, so we should consider bringing the changes to PROD as soon as possible.

niwinz:
> @niwinz the issues fixed here are likely to negatively affect quite a few users of the remote MCP server, so we should consider bringing the changes to PROD as soon as possible.

I'll put them on my high priority task list. I  will look on it tomorrow and once merged probably tag the next RC with this changes and other fixes merged today (realated to reports).

niwinz:
After looking the code, i will merge it right now :D

PR #10996, opened 2026-07-31
:bug: Restrict webhook creation to team editors

### LINKED ISSUE #10995: Improve webhooks permission checks
### Description

The webhook creation endpoint does not verify that the requesting user has editor permissions on the team. This allows non-editor users (such as viewers) to create webhooks, which should be restricted to team editors.

### Steps to reproduce

1. Authenticate as a user with viewer permissions on a team
2. Call the webhook creation endpoint with valid parameters
3. Observe that the webhook is created successfully

### Expected behavior

The webhook creation endpoint should verify that the requesting user has editor permissions on the team. Users without editor permissions should receive a permission denied error.

### Affected versions

Current development branch

niwinz:

## What

The webhook creation endpoint did not verify that the requesting user has editor permissions on the team. This allowed non-editor users (such as viewers) to create webhooks, which should be restricted to team editors.

## Why

The endpoint lacked a permission check to ensure only team editors can create webhooks. Without this check, users with lower permission levels could perform actions that should be restricted to editors.

## How

- Added permission verification to the webhook creation endpoint
- The endpoint now checks that the requesting user has editor permissions on the team
- Users without editor permissions receive a permission denied error

Closes #10995

### DIFF
--- a/backend/src/app/rpc/commands/webhooks.clj
+++ b/backend/src/app/rpc/commands/webhooks.clj
@@ -120,7 +120,7 @@
   {::doc/added "1.17"
    ::sm/params schema:create-webhook}
   [{:keys [::db/pool] :as cfg} {:keys [::rpc/profile-id team-id] :as params}]
-  (check-webhook-edition-permissions! pool profile-id team-id profile-id)
+  (t/check-edition-permissions! pool profile-id team-id)
   (validate-quotes! cfg params)
   (validate-webhook! cfg nil params)
   (insert-webhook! cfg params))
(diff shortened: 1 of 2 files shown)


luisddm:
@niwinz Shouldn't we apply the same restrictions also to update/delete webhooks?

niwinz:
> @niwinz Shouldn't we apply the same restrictions also to update/delete webhooks?

I will check it

niwinz:
Superseded by #11029 which consolidates webhook permission fixes (creation, edit, and delete) into a single atomic PR.

PR #10998, opened 2026-08-01
:bug: Fix selection stolen by trailing click after a move drag

wi-fa:
### Related Ticket

No existing issue found for this — happy to open one if preferred.

### Summary

After a move drag ends, the browser fires a compatibility `click` on `viewport-controls` (pointer-down and pointer-up land on the same element, and left-clicks are not prevented). `on-click` (`viewport/actions.cljs`) only guards against the rubber-band flow (`(nil? selrect)`), so it reaches `(dw/select-shape (:id @hover))` — and `@hover` at that moment can point at the WRONG shape:

- hover keeps updating during a drag (`hooks.cljs` freezes only `hover-ids`, `@hover` stays live for the ALT-drag distance measure), and
- hover hit-tests resolve against **committed, pre-reflow geometry** (the worker index only sees committed changes, never live modifiers).

So when a flex-layout child is drag-reordered, the release point sits over a **sibling's old bounds** — the trailing click then moves the selection from the dragged shape to that sibling.

The area-select flow already guards this exact hazard: `data/workspace/selection.cljs` clears `selrect` **asynchronously** specifically so the trailing `on-click` doesn't misfire. This PR brings the move flow in line, using the same pattern as the existing `disable-paste-ref`:

- `on-pointer-up` sets a `just-finished-move-ref` when it terminates a `:move` transform (with a scheduled reset for releases that produce no click),
- `on-click` skips the hover reselect for that single trailing click and clears the flag.

Applied to both the classic and wasm viewports. `clj-kondo` passes with no new findings.

### Steps to reproduce

1. Create a board with a flex layout and four children `[A, B, C, D]`.
2. Press on **A** and drag it rightwards past **B** and **C** (the live reorder swaps it along), release while inside the layout.
3. The reorder commits correctly, but the selection/bounding box lands on **C** (the sibling immediately before A's new slot) instead of staying on the dragged **A**.

The same mechanism can steal selection on any move drag that releases over another shape's pre-drag bounds (not layout-specific).

### Checklist

- [ ] Add or modify existing integration tests in case of bugs or new features, if applicable. *(A Playwright regression needs a get-file fixture containing a flex board — I can add one if a maintainer points me at the preferred way to generate fixtures.)*
- [x] Refactor any modified SCSS files following the refactor guide. *(No SCSS touched.)*


madalenapmelo-kp:

wi-fa:
just wanna show you what i found. imo the bounding box shouldnt land on the green box but on the grey box when im letting it go?

https://github.com/user-attachments/assets/7891c1a8-ffba-4bc4-845b-15b252e01581

niwinz:
Would be nice to open an issue for this in any case.  I triaged it and leave it to be ready to be reviewed. 

Thanks.

PR #11014, opened 2026-08-03
:bug: Validate font-id team ownership in create-font-variant

### LINKED ISSUE #11013: Validate font-id team ownership in create-font-variant
### Description

The create-font-variant endpoint does not validate that the font-id belongs to the same team when creating a new variant. This allows cross-team font injection where a user with team edit permissions could create a font variant referencing a font-id from another team.

### Steps to reproduce

1. Create a font in Team A with at least one variant
2. As a user with edit permissions in Team B, call create-font-variant with the font-id from Team A
3. The endpoint accepts the request and creates a variant referencing Team A's font

### Expected behavior

The endpoint should validate that the font-id belongs to the same team and reject the request if it doesn't.

### Affected versions

Current main branch

niwinz:

## What

- The `create-font-variant` endpoint now validates that when a `font-id` already has variants, they belong to the same team
- This prevents cross-team font injection where a user with team edit permissions could create a font variant referencing a font-id from another team

## Why

The endpoint was not checking team ownership of existing font variants, allowing a BOLA (Broken Object Level Authorization) vulnerability where users could reference fonts from other teams.

## How

- Added validation in `create-font-variant` to check that existing variants of the specified font-id belong to the same team
- The check only applies when the font-id already has variants (new fonts are not affected)

Closes #11013

### DIFF
--- a/backend/src/app/rpc/commands/fonts.clj
+++ b/backend/src/app/rpc/commands/fonts.clj
@@ -93,6 +93,18 @@
 
 (declare create-font-variant)
 
+(defn- check-font-team-ownership!
+  "When font-id already has variants belonging to a different team,
+  raises :not-found to prevent cross-team font injection."
+  [conn team-id font-id]
+  (let [row (db/get* conn :team-font-variant
+                     {:font-id font-id}
+                     {::db/columns [:team-id]})]
+    (when (and row (not= (:team-id row) team-id))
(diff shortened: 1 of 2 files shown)


luisddm (approved):

niwinz:
We have removed the font-id, i need to review if this ISSUE still makes sense

PR #11038, opened 2026-08-04
:bug: Add accumulated storage byte quota for media uploads

### LINKED ISSUE #11037: Media uploads have no accumulated storage byte quota
### Description

Media upload endpoints do not enforce an accumulated storage byte quota, allowing unbounded storage consumption by a single team or account.

### Steps to reproduce

1. Upload multiple large media files to a team
2. Observe that there is no cumulative storage limit enforced
3. Storage usage grows without bound

### Expected behavior

Media upload endpoints must enforce an accumulated storage byte quota to prevent unbounded storage consumption.

### Affected versions

All versions prior to the fix.

niwinz:
## What

Add accumulated storage byte quota for media uploads to prevent unbounded storage consumption.

## Why

Media upload endpoints did not enforce a cumulative storage byte quota, allowing a single team or account to consume unlimited storage through repeated uploads.

## How

- Add `media-storage-bytes-per-team` quota that sums `storage_object` sizes from both `file_media_object` (media + thumbnails) and `team_font_variant` (otf/ttf/woff1/woff2)
- Default limit is 20 GiB per team, configurable via `PENPOT_QUOTES_MEDIA_STORAGE_BYTES_PER_TEAM`
- Invoke check in `upload-file-media-object` before processing, looking up team-id via file -> project -> team_id join

Closes #11037

Note: This is an internal quota enforcement that does not affect normal usage patterns.

### DIFF
--- a/backend/src/app/config.clj
+++ b/backend/src/app/config.clj
@@ -194,6 +194,7 @@
     [:quotes-team-access-requests-per-requester {:optional true} ::sm/int]
     [:quotes-upload-sessions-per-profile {:optional true} ::sm/int]
     [:quotes-upload-chunks-per-session {:optional true} ::sm/int]
+    [:quotes-media-storage-bytes-per-team {:optional true} ::sm/int]
 
     [:auth-token-cookie-name {:optional true} :string]
     [:auth-token-cookie-max-age {:optional true} ::ct/duration]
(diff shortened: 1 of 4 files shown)


niwinz:
Before merge i need to check the performance of the new query (large one)

niwinz:
I have tested the performance, it is not the best but for now it can work. We need to thing a bit better approach to have storage usage precalculated and/or cached.

luisddm (approved):

PR #11059, opened 2026-08-04
:bug: Add backend password validation with complexity rules and dictionary check

### LINKED ISSUE #11055: Backend accepts passwords without complexity validation
### Description

Backend does not enforce password complexity rules or check against common passwords, allowing users to set weak passwords that are vulnerable to brute force attacks.

### Steps to reproduce

1. Attempt to set a weak password (e.g., "password123")
2. Observe that the backend accepts the password without validation
3. Weak passwords are stored without complexity requirements

### Expected behavior

Backend must enforce password complexity rules (minimum length, character requirements) and reject common passwords to prevent weak password usage.

### Affected versions

All versions prior to the fix.

niwinz:
## What

Add backend password validation with complexity rules and dictionary check.

## Why

Backend did not enforce password complexity rules or check against common passwords, allowing users to set weak passwords that are vulnerable to brute force attacks.

## How

- Enforce minimum 8-character password length
- Require at least 1 lowercase letter, 1 uppercase letter, 1 digit, and 1 special character
- Reject common passwords using Passay library with bundled wordlist
- Apply validation during registration and password change flows

Closes #11055

Note: This is an internal security enhancement that does not affect existing user accounts.


luisddm:
@niwinz The validation works as expected, but I think the feedback that the user receives when they enter an invalid password ('_Something wrong has happened_') is quite poor and generic. Instead, I suggest showing something like a '_Weak password_' message in the toast, or even better, translating the contents of the details field in the RPC response to provide more information. This applies when changing password, too.

create-account

create-account-json

niwinz:
> @niwinz The validation works as expected, but I think the feedback that the user receives when they enter an invalid password ('_Something wrong has happened_') is quite poor and generic. Instead, I suggest showing something like a '_Weak password_' message in the toast, or even better, translating the contents of the details field in the RPC response to provide more information. This applies when changing password, too.

ACK

luisddm:
The password validation experience has been improved by making error messages specific and actionable. Before, a user would just see a generic "password is too weak" error. Now, the backend returns structured details about exactly which rules failed (missing uppercase, too short, etc.) and the frontend translates those into readable hints displayed to the user. This was applied across registration, password change, and password recovery flows.

The common password dictionary check has been removed because it was redundant — every password in the 10,000-word list would already fail the character composition rules (requiring uppercase, lowercase, digits, and special characters), so the dictionary check could never reject anything that passed those rules. The file was deleted and validation simplified to just length and character composition.

PR #11075, opened 2026-08-04
:sparkles: Put the graph subsystem behind a flag, off by default

alvorithm:
## Why

`app.graph.ladybug` imports `com.ladybugdb.*` at namespace load, and both namespaces that reach the subsystem required it at the top level: `app.http.debug`, which registers the `/dbg` routes on every boot, and `app.srepl.main`, which loads with the REPL server. Every backend therefore linked the Ladybug native library into the JVM at boot. Nothing on the request path uses the graph, so for any deployment that never opens the console that is startup time, resident memory, and the crash surface of a native library, all for nothing.

## What

`82d771976c` adds a `:graph` flag to `flags/varia` and leaves it out of `flags/default`, so it ships off. Both require sites now go through `requiring-resolve` at call time. The nine graph routes (`/dbg/graph` plus eight `/dbg/actions/graph-*`) are registered only when the flag is on, and the export fieldset in `debug.tmpl` is guarded by the same flag. With the flag on, `ig/init-key ::routes` requires `app.graph.debug` and `app.graph.ingest` eagerly, so a missing or unusable native library fails the boot rather than the first console request.

The `/dbg` admin gate is untouched. The flag decides which routes exist, not who may reach them.

No tracked file turns the flag on. `backend/scripts/_env` leaves it out, so a devenv boots with the subsystem off exactly as a released build does. `docker/images/docker-compose.yaml` is untouched, because its `PENPOT_FLAGS` lists only overrides of `flags/default` and `:graph` already defaults off.

## Turning it on and off

The flag is read once, at boot. Switching it means editing `PENPOT_FLAGS` and restarting the backend. The "Feature flags for team" form on `/dbg` does not reach it, because that form operates on `app.common.features/supported-features`, which is a different mechanism.

On: add `enable-graph` to `PENPOT_FLAGS`. Off: list neither `enable-graph` nor `disable-graph`, since off is the default.

The devenv is off as well, so reviewing the on path takes one gitignored file. `backend/scripts/_env.local` is sourced right after `_env` by every backend and exporter dev script, `start-dev` and `repl` among them:

```sh
echo 'export PENPOT_FLAGS="$PENPOT_FLAGS enable-graph"' > backend/scripts/_env.local
```

`flags/parse` applies flags left to right and the last one wins, so the trailing `enable-graph` turns the subsystem on for that one checkout. Delete the file to go back. Restart the backend in either direction.

What a reviewer sees on a restarted backend, holding an admin session:

| | flag off | flag on |
| --- | --- | --- |
| `GET /dbg/graph` | 404 | the graph console |
| the eight `/dbg/actions/graph-*` endpoints | 404 | their handlers |
| "Export graph (Ladybug)" fieldset on `/dbg` | absent | present |
| `(find-ns 'app.graph.ladybug)` in the REPL | `nil` | the namespace |

Without an admin session the difference still shows: a registered route answers 401 and an unregistered one answers 404.

## Verification

`-verbose:class` over a boot's namespace load plus `ig/init-key ::routes`: 9 `com.ladybugdb` classes before this change with no flag set, 0 after it with the flag off, 9 with `enable-graph`.

Backend suite: 497 tests, 3851 assertions, 0 failures. `backend_tests.helpers` pins `app.config/flags` to a list that omits `:graph`, so the suite now runs without loading Ladybug at all. No existing test covers `app.http.debug`, before or after this change.

### DIFF
--- a/backend/resources/app/templates/debug.tmpl
+++ b/backend/resources/app/templates/debug.tmpl
@@ -222,6 +222,7 @@ Debug Main Page
         </div>
       </form>
     </fieldset>
+    {% if graph-enabled %}
     <fieldset>
       <legend>Export graph (Ladybug):</legend>
       <desc>Given a FILE-ID, builds the graph projection and downloads
@@ -237,6 +238,7 @@ Debug Main Page
         </div>
       </form>
     </fieldset>
+    {% endif %}
     <fieldset>
       <legend>Import binfile:</legend>
       <desc>Import penpot file in binary format.</desc>
--- a/backend/src/app/http/debug.clj
+++ b/backend/src/app/http/debug.clj
@@ -22,8 +22,6 @@
    [app.config :as cf]
    [app.db :as db]
    [app.features.file-migrations :as feat.fmig]
-   [app.graph.debug :as graph.debug]
-   [app.graph.ingest :as graph.ingest]
    [app.http.session :as session]
    [app.rpc.commands.auth :as auth]
    [app.rpc.commands.files-create :refer [create-file]]
@@ -61,6 +59,7 @@
      ::yres/body    (-> (io/resource "app/templates/debug.tmpl")
                         (tmpl/render {:version (:full cf/version)
                                       :profile profile
+                                      :graph-enabled (contains? cf/flags :graph)
                                       :current-clock  ct/*clock*
                                       :current-offset (if offset
                                                         (ct/format-duration offset)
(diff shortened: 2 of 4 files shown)


alvorithm:
Update: found it, and it is not the pin on its own. `frontend/scripts/setup` on `graph-backend` still runs `pnpx playwright install chromium`. `pnpx` is `pnpm dlx`, so it resolves the *latest* Playwright rather than the pinned one: the job installs the browsers for 1.62.1 (chromium revision 1234), then runs the tests with the pinned 1.61.1, which looks for revision 1228 and finds nothing. `aad9bc8f65` on develop (":paperclip: Fix playwright version", 27 July) replaced `pnpx` with `pnpm exec`, which is exactly the fix. `graph-backend` forked before it.

Reproduced in the devenv against this branch's frontend tree, with an isolated `PLAYWRIGHT_BROWSERS_PATH` so the real cache stays untouched:

```
pnpx playwright install chromium      -> chromium_headless_shell-1234
chromium.launch()                     -> Executable doesn't exist at .../chromium_headless_shell-1228/...
pnpm exec playwright install chromium -> chromium_headless_shell-1228
chromium.launch()                     -> LAUNCH OK
```

NOTE: below is outdated; had to close that PR (fumbled it).

~~So I have opened #11079, which merges develop into graph-backend. On that PR **Frontend Tests pass**, along with Backend, Common, Library, the bundle build, and both commit checks. Two jobs are red there and both are red on develop itself: "Plugins Runtime Linter & Tests" fails a `prettier --check` on `plugins/apps/composable-test-suite/package.json`, a file the merge takes from develop byte for byte, and Integration Tests fails real assertions in `tokens/apply.spec.js` and `toolbar.spec.js`. The change that matters there is that Integration now runs the suite instead of dying in 49 seconds with 337 of 338 tests reported as "did not run".~~

~~@superalex could you review #11079? Once it lands I will push a merge of the updated graph-backend into `alvorithm-graph-flag`, which re-runs this PR's checks against a base that has the fix, and they should clear.~~

PR #11079, opened 2026-08-05
Merge develop into graph-backend

alvorithm:
## Context

`graph-backend` forked from `develop` at `a006a12ab6` and has not taken it since. 291 commits have landed on `develop` in the meantime, so every PR against `graph-backend` is reviewed and tested against a base that is months old. [#11075](https://github.com/penpot/penpot/pull/11075) is the immediate case: its Frontend and Integration jobs fail for a reason that has nothing to do with its change.

Those two jobs fail because `frontend/scripts/setup` on `graph-backend` still runs `pnpx playwright install chromium`. `pnpx` resolves the latest Playwright rather than the pinned one, so the job installs the browsers for `1.62.1` (revision 1234) and then runs the tests with the pinned `1.61.1`, which looks for revision 1228 and finds nothing. `aad9bc8f65` on `develop` fixed exactly this by replacing `pnpx` with `pnpm exec`, and this merge brings it.

Reproduced in the devenv, on the current `graph-backend` frontend tree, with an isolated `PLAYWRIGHT_BROWSERS_PATH`:

```
pnpx playwright install chromium      -> chromium_headless_shell-1234
node: chromium.launch()               -> Executable doesn't exist at .../chromium_headless_shell-1228/...
pnpm exec playwright install chromium -> chromium_headless_shell-1228
node: chromium.launch()               -> LAUNCH OK
```

The error text matches the CI failure verbatim.

## Changes

- `2b23a89950` merges `origin/develop` (`34702fd46b`) into `graph-backend`.

Two resolutions, one textual and one semantic:

- `backend/deps.edn` conflicted. `develop` bumps `software.amazon.awssdk/s3` and `sts` to `2.50.1`, `graph-backend` adds `com.ladybugdb/lbug` after them. Resolved by taking both.
- `backend/src/app/srepl/main.clj` merged cleanly and was still wrong. `develop` moved the system value from `app.main/system` to `app.system/system` and dropped the `app.main` require, while `graph-backend` added `ingest-file-to-graph!`, which still called `main/system`. Git kept both halves. Changed the call to `sys/system`; `clj-kondo --parallel --lint ../common/src/ src/` is clean.

## Verification

The merge is textual except for that one conflict. `git merge-tree --write-tree` of `alvorithm-graph-flag` against this branch reports no conflicts, so #11075 still applies once this lands. The merged `app.http.debug` keeps the graph routes and their flag guard.

CI on this PR: Backend, Common, Library, Frontend Tests, the bundle build, and both commit checks pass. Integration Tests runs 290 cases and fails one, `toolbar.spec.js` › `User creates a text shape with the toolbar text tool`, which is one of the two that fail on `develop` itself. "Plugins Runtime Linter & Tests" fails a `prettier --check` on `plugins/apps/composable-test-suite/package.json`, a file this merge takes from `develop` byte for byte, and is red on `develop` too.

Not verified locally: the backend and common suites against the merged tree. `develop` reworked `app.common.types.shape`, `app.common.logic.shapes`, `app.common.logic.libraries`, and `app.common.types.file`, which the graph projection reads, so CI on this PR is the first real check of that surface.

## Deferred

Re-auditing the graph projection against the reworked shape and library model. It is a `graph-parity` concern, not a blocker for this merge.


alvorithm:
Superseded. `graph-backend` is now rebased onto `develop` (`2a7e00c063`, 53 commits ahead) instead of taking it by merge, per @superalex: merges into public branches, rebase on the rest.

The rebase produced the same single `backend/deps.edn` conflict this PR resolved, plus one commit for `develop`'s `app.main/system` to `app.system/system` rename, and its tree is byte-identical to `2b23a89950`, the commit CI ran green here. Against the rebased base this PR now lists the 59 pre-rebase commits, so merging it would only re-add duplicates.

PR #11101, opened 2026-08-06
:sparkles: Add the graph subsystem and graph visualization console to the backend

alvorithm:
_This work was produced with the help of language models._

@niwinz I am adding the additional commits we discussed yesterday, motivation follows. The branch is also rebased on current `develop`, so the diff shows only its own work. Everything below is updated. The two new sections are **What the additional commits add** and **The 81 lines in `common/`, and why they are not risky**.

### Related Ticket

Internal discussion; I can write one for project management purposes, if desired.

### Summary

Adds capabilities to Penpot, currently exposed via `dbg`, to
* project a Penpot file into a graph,
* draw it with selectable detail,
* store it in a queryable graph database,
* stream file edits into the graph

**An internal inspection tool.**  These functions provide a structural view into the Penpot data model, and a useful substrate for programmatic manipulation. They are not meant to be user-facing.

**Off by default.**  The `:graph` flag gates the feature. Unless enabled,
 `app.graph.*` is not required,  the `com.ladybugdb` class does not enter the JVM, and the `/dbg` graph routes remain unregistered.

**Self-contained** ~3.8k lines under `backend/src/app/graph/`, ~450 lines of tests, one console template and two dependencies: a graph database for storage ([LadybugDB](https://github.com/LadybugDB/ladybug), C++) and a graph layout library for visualization ([G6](https://github.com/antvis/G6), JS).

**Infrastructure-focused** This PR lays the groundwork, but the Penpot file [closure] to database mapping is minimalistic. Mapping every Penpot design entity as a graph node, and relations between them as edges, is ongoing work.

### What the additional commits add

Sixteen commits on top of what you saw. Four themes.

**Bulk load through in-memory Arrow.** Rows are built as Arrow vectors, handed to Ladybug as a virtual table and `COPY`d in. No file is written and no value is rendered as text for the engine to re-parse, which removes a whole defect class: three of this branch's four earlier defects were escaping bugs in a CSV writer. `app.graph.bulk` is deleted. Marginal ingest went from 4.0 to 1.21 ms/shape.

**Tests, where there were none.** Two namespaces, 17 tests, 55 assertions, both running against in-process `:memory:` databases with no Postgres and no session.

- `graph_binder_gate_test` prepares every Cypher statement the sync path emits against the live schema without executing it, so a renamed column or a reserved word emitted unquoted fails here rather than in a live session. It was not green when written: it caught a statement writing a column that no longer existed.
- `graph_sync_parity_test` projects a file into one database, applies a change list to that database and the same list to the file data, projects the result into a second database, and diffs the two by row and column. Cold projection and incremental sync are two implementations of one mapping and nothing checked they agreed. They did not, in four places, all fixed here. The most consequential: a container's stored `:shapes` list runs bottom to top while `IsChildOf.position` numbers children in z-order, so any container edited live carried its children in the opposite order to a rebuild.

**The console is read-only.** `/dbg/actions/graph-query` binds each statement against the live schema before running it. A statement that does not bind reports the binder's own message and executes nothing, which also turns a misspelt table into an error instead of an empty result. A statement that binds runs only if the engine's own read/write analysis calls it read-only. A session graph is rebuilt from the file by Reload, so a mutation from the console would produce a graph no rebuild reproduces.

**Column types.** Ladybug fixes a column's type at table creation with no widening afterwards, so the Malli-to-Ladybug mapping is the whole of the graph's typing. `selrect` is `DOUBLE[4]`, four numbers a consumer reads directly, rather than JSON somebody has to parse. Two mapping defects fell out: `::sm/set` and `::sm/one-of` were unmapped and fell through to `STRING`.

### One prerequisite left, and its commit is still in this branch

**#11124, `:bug: Unbreak the backend CI gates for create-font-variant`, has merged.** Its two commits are on `develop`, so they are no longer part of this diff.

**#11125, `:bug: Declare the shape attributes stored files carry`, is open and ready for review.** Its commit stays here for now, deliberately, so that this pull request's CI exercises the tree it proposes rather than a tree that cannot build. It drops out on the next rebase once #11125 lands, and this branch shrinks by that much. The commit here is patch-identical to the one on `fix/shape-attrs-declaration`, so the drop is clean.

What this pull request needs from it: the graph derives its columns from Penpot's own Malli schemas, so `app.graph.schema.projection` reads the entries of `schema:shape-attrs` and each entry becomes a column. A key the schema does not declare becomes *no column at all*, which is why **`rotation` was missing from every shape node in every graph produced**, along with `flip-x`, `flip-y`, `hide-in-viewer` on anything but a frame, the four `svg-*` keys, `rx`/`ry`, `content` on svg-raw, and the geometry of bools and paths. The argument for the change is in #11125, where it can be judged on its own merit rather than as a graph dependency.

The graph's own side of that change is now a separate commit on this branch, `:sparkles: Add the file-level graph columns and tighten the svg ones`: the `DOUBLE[4]` and `DOUBLE[6]` column types for `svg_viewbox` and `svg_transform`, and the four file-level attributes declared as a projection `:extra` because `ctf/schema:file` cannot carry them.

Reviewing #11125 first still makes this one smaller.

### Impact when disabled (default)

The new flag `:graph` is in **`common/src/app/common/flags.cljc`**, **not** in `default`, which means the feature is off by default. If the feature is merged, even users who keep it off see these changes:

- **`backend/deps.edn`** gains `com.ladybugdb/lbug {:mvn/version "0.19.1"}` - on the classpath and in the built jar (more below).
- **`backend/src/app/main.clj`** gets `::mbus/msgbus (ig/ref ::mbus/msgbus)` in the `::http.debug/routes` config. `msgbus` exists for RPC anyways but this line adds one edge to the Integrant graph and injects an existing component that the console's live feed reads. With the flag off that value is never used.

That's it.

Instead of just hiding the routes, we made `app.http.debug` and `app.srepl.main` reach `app.graph.*` through `requiring-resolve` at call time rather than requiring it at the top. This was to avoid the cost of class loading and extra crash surface of the native library. You still need admin for `/dbg`.

Regarding the G6 JS library, used by the graph view, it’s loaded from jsDelivr at a pinned version and inaccessible with the flag off, so it doesn’t generate any requests.

### Steps to test in devenv

**Caveat**: the "Feature flags for team" form on `/dbg` does not apply here; that form operates on `app.common.features/supported-features`

**Activating.**  export the flag and restart (the flag is only read at boot):

```sh
PENPOT_FLAGS=enable-graph ./manage.sh run-devenv
```

[]`backend/scripts/_env` builds `PENPOT_FLAGS` starting from the inherited value, and `flags/parse` takes the last occurrence, so the exported flag survives.  Verify what happens looking at the backend's own `hint="welcome to penpot", flags="…"`.

When the flag is present, the admin gate refuses the request (**401**); when it's absent , the route is simply not registered (**404**).

| `graph` in the boot `flags=` line | absent | present |
| --- | --- | --- |
| `GET /dbg` | 401 | 401 |
| `GET /dbg/graph` | **404** | **401** |
| `GET /dbg/actions/graph-export` | **404** | **401** |

Use of the feature (flag on, admin session):

* navigate to `/dbg/`. The **fieldset** "Export graph (Ladybug)" (appears only when the flag is on) allows you to convert the Penpot file into a `.lbug` file for download. This can be used for exchange with visualization tools, local [LadyBugDB](https://github.com/LadybugDB/ladybug) queries with its REPL or feeding into beadpot.
* navigate to `/dbg/graph`. The new **graph console** offers a file load dialog (your teams are presented in a folded tree widget for point-and-click loading), a  canvas to draw the selected file's graph, a widget to send Cypher queries (with table output and canvas refresh), a changes monitor, and more. 

Demo video [here](https://drive.google.com/file/d/1KHN33xeYSaGYT4Uru1lh4VdLb1hjychf/view) (internal).

### Security and supply chain

We introduce two dependencies, so we should discuss what this means for Penpot.

#### In-process graph database: LadybugDB
In my opinion, this is the key impact you should consider when accepting the PR: **an off flag does not hide the dependency from CVE scanners.** Trivy, Grype, and friends read the classpath, not `PENPOT_FLAGS`; any future CVE in `lbug` (none are currently known) will be reported against the Penpot backend image.

- **Licence**: MIT (unproblematic).
- **Origin**: Maven Central. No 3rd-party repo in `deps.edn`.
- **Size**: 27 MB, because the jar carries native libraries for the supported platforms.
- **Runtime exposure with the flag off**: none, (verified by class loading). `-verbose:class` over a boot's namespace load plus `ig/init-key ::routes` counts 9 `com.ladybugdb` classes before this change, 0 after it with the flag off, and 9 with `enable-graph`.
- **The graph engine's provenance**: [Ladybug](https://github.com/LadybugDB/ladybug) is the MIT fork of KuzuDB, a solid graph DB system which was discontinued due to an Apple team adqui-hire. Ladybug picks it up and is actively developed, but this fork is carried on by a small community with a current bus factor of one.

**Question about [LadyBugDB](https://github.com/LadybugDB/ladybug).** `lbug` is currently an unconditional dependency. We can put it behind a `deps.edn` alias instead, so it is absent from the default build and the scanner question goes away. This would need 1) a change in the build and a 2) second setting for the flag-on path (that requires manual sync). Is it something to consider for this PR?

### Graph visualization library: G6

The second external dependency is **[G6](https://github.com/antvis/G6)**, the JavaScript library the console draws with. We currently load it from jsDelivr, version pinned to avoid remote dependency updates. It is client-side and only fetched with the console open.

**Question.** Vendoring G6 into `backend/resources` would put it in the jar, and declaring it in `frontend/package.json` would be the natural move if the console leaves `/dbg`. Do you want any of these to happen in this PR?

### Verifications

Routes as discussed above (table). Class loading: the `-verbose:class` counts above.

| check | result |
|---|---|
| `cljfmt check --parallel=true src/ test/` | all source files formatted correctly |
| `clj-kondo --parallel --lint ../common/src/ src/` | 0 errors, 0 warnings |
| common suite | 1142 tests, 24702 assertions, 0 failures |
| backend suite | 649 tests, 4359 assertions, 0 failures |
| `graph_binder_gate_test` + `graph_sync_parity_test` | 17 tests, 55 assertions, 0 failures |

- The existing suites cover the flag-off path and run without loading [Ladybug](https://github.com/LadybugDB/ladybug).
- The two graph namespaces above are the first tests of `app.graph`. Coverage of the read path and the asset kinds is still to come.
- The graph console was exercised by hand against a real file: load, query, a refused mutation, a reported binder error, reload, unload.

The backend numbers are measured with #11124's two fixes, which are now on `develop`. Unmodified `develop` @ `88697794ce`, before they landed, was 649 tests, 4355 assertions, 3 failures.

### Out of scope

**Slimming down the additional 27MB in the penpot.jar.** We could envisage contributing to [LadybugDB](https://github.com/LadybugDB/ladybug) so that they start publishing platform-specific artifacts. Keping Penpot linux arm64 and amd64 targets would bring the .jar down to ~17MB, keeping one to ~9MB.

**Moving the console out of `/dbg`**, raised in conversation and not attempted here. It would move the template, and probably the G6 question with it.

### Checklist

- [x] Add or modify existing integration tests in case of bugs or new features, if applicable. *(two new backend test namespaces. No integration tests: the feature has no UI outside `/dbg`)*
- [x] Check CI passes successfully. *(every job green, Integration Tests included)*

AI-assisted-by: mixed models


alvorithm:
Two prerequisites split out of this branch, so they can be judged on their own and this one shrinks:

- #11124 `:bug: Unbreak the backend CI gates for create-font-variant` — nothing to do with this feature. An unused binding fails the Lint step, which runs before the tests, and a test still sends a param removed in 2.18. Both arrived with #11014 after the last Backend run on `develop`, so nothing had reported them.
- #11125 `:bug: Declare the shape attributes stored files carry` — the 81 lines in `common/`, with the argument for them standing on its own.

Both commits stay in this branch for now, deliberately, so CI here exercises the tree this pull request proposes rather than one that cannot build. They drop out on the next rebase once the prerequisites land.

Reviewing #11124, then #11125, then this one is the cheapest order.

alvorithm:
Rebased onto `develop` @ `43b12bc4b9`, and the #11125 commit here is now the reviewed version.

Three things changed, nothing else:

- **#11124 dropped out.** It merged, so its two commits are on `develop` and no longer in this diff.
- **The #11125 commit was replaced** with the one @niwinz's review produced. `git range-diff` reports it patch-identical to `fix/shape-attrs-declaration`, so it drops out cleanly when #11125 lands. The substance: `{:optional true}` and `[:maybe …]` are no longer both applied to the same key. The record decides which one, because `app.common.record/defrecord` cannot remove a base field, so `rx`, `ry`, the `svg-*` group and the rest lose the `[:maybe]`, while the bool and path geometry loses the `{:optional true}`.
- **The graph's own half of that commit is now its own commit**, `:sparkles: Add the file-level graph columns and tighten the svg ones`. The old commit bundled `common/` and `backend/`, which is what kept #11125 from being a clean subset.

Diffing the new tip against the old one shows exactly two things: the updated `shape.cljc`, and `develop`'s own render-wasm commits. `backend/` is byte-identical.

Local checks on the rebased tree: `clj-kondo` 0 errors 0 warnings, `cljfmt` clean, common suite 1142 tests / 24702 assertions / 0 failures, `graph_binder_gate_test` + `graph_sync_parity_test` 17 tests / 55 assertions / 0 failures.

niwinz:
After a quick look on code i have no red flags. So looks almost ready to merge.
Tomorrow i will try to take tome and with help of AI agent to undertand the flow of data and go a bit deeper on understanding.

Once this merged, i have several ideas, one bigger than i need to open as a task for make this prototype more reliable (per example: a good way to preserve and store changes instead of discarding them and also a better way to consume them in a stream processing way, but this is just an example) i'm also want to integrate this into penpot frontend application to make it available to be used outside the dbg interface. We need more pieweeks xD

niwinz:
> Question. Vendoring G6 into backend/resources would put it in the jar, and declaring it in frontend/package.json would be the natural move if the console leaves /dbg. Do you want any of these to happen in this PR?

No is ok for now, the PR is self-contained and exposes good prototype under dbg, putting it under frontend/package.json will be probably a task part of the "refactort" on trying to move this under frontend application. No changes nededed on this PR

PR #11122, opened 2026-08-06
:bug: Use constant-time comparison for shared key authentication

### LINKED ISSUE #11121: Shared key authentication uses timing-vulnerable comparison
## Description

The `wrap-shared-key-auth` middleware compares shared keys using the standard `=` operator, which terminates on the first mismatch. This is vulnerable to timing attacks where an attacker could infer the correct key byte-by-byte by measuring response times.

## Impact

While exploitation requires measuring response times with high precision and many samples, this is a well-documented vulnerability class. The fix is straightforward: use constant-time comparison.

## Solution

Replace `=` with `MessageDigest/isEqual` from the JDK, which compares all bytes regardless of where the first mismatch occurs.

niwinz:
## What

- Replace timing-vulnerable string comparison with constant-time comparison in shared key authentication middleware

## Why

The shared key authentication middleware used the standard `=` operator which terminates on the first mismatch, making it theoretically vulnerable to timing attacks. While exploitation requires high-precision timing measurements and many samples, this is a well-documented vulnerability class.

## How

- Added `constant-time-eq?` helper using `java.security.MessageDigest/isEqual`
- Replaced `=` comparison in `wrap-shared-key-auth` with the new helper

Closes #11121

### DIFF
--- a/backend/src/app/http/middleware.clj
+++ b/backend/src/app/http/middleware.clj
@@ -24,7 +24,8 @@
   (:import
    io.undertow.server.RequestTooBigException
    java.io.InputStream
-   java.io.OutputStream))
+   java.io.OutputStream
+   java.security.MessageDigest))
 
 (set! *warn-on-reflection* true)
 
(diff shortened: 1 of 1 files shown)


EvaMarco (approved):

EvaMarco:
Code is ok, just fix the CI

PR #11125, opened 2026-08-07
:bug: Declare new shape attributes in schemas to match stored files

alvorithm:
_This work was produced with the help of language models._

### Related Ticket

None. Split out of #11101 so it can be judged on its own: it is a defect in `common/` that exists today, independently of the branch that found it.

### Summary

`app.common.types.shape/schema:shape-attrs` is the shape model as *declared*, and it has fallen behind the `Shape` record.

```clojure
(cr/defrecord Shape [id name type x y width height rotation selrect points
                     transform transform-inverse parent-id frame-id flip-x flip-y]
```

`rotation`, `flip-x` and `flip-y` are fields of that record, so they are present on **every shape that exists**, and they are declared nowhere. `rotation` is already named twice in the same namespace, in `allowed-shape-attrs`, and once in `app.common.types.shape.attrs/editable-attrs`, so the schema is the odd one out rather than the data being unusual.

Nothing complains, because the maps are open: an undeclared key validates fine. What breaks is everything reading the model *from the schema* rather than from a live value, such as the generative tests' shape generator, the generated OpenAPI surface, and any consumer reflecting over `schema:shape-attrs`.

**There are four questions for the reviewer below, each marked QUESTION.** In order: whether the optional/nilable rule is the one you meant, whether the `svg-*` group should be typed tighter, whether bool and path should stay silent about geometry, and whether a schema should serve as both read description and write contract.

### Feedback addressed: `{:optional true}` together with `[:maybe …]`

**What @niwinz said.** The first revision marked keys `{:optional true}` **and** `[:maybe …]`. That is fine for `x` and `y`, which are static props that cannot be removed from the type, but other keys are not static props, and marking those `[:maybe]` was not convincing. `rx` and `ry` on `RectAttrs` and `CircleAttrs` were the example.

**Why he is right, and mechanically so.** `app.common.record/defrecord` is not `clojure.core`'s. It cannot remove a base field:

```clojure
;; app.common.record, emit-impl-jvm and emit-impl-js agree
(without     [this key] (case key (:x :y …) (.assoc this key nil) …))
(containsKey [this key] (case key (:x :y …) true …))
```

So the two kinds of key behave differently, and each takes exactly one marker:

| kind of key | absent when unset? | nil when unset? | declaration |
|---|---|---|---|
| `Shape` base field | no, `containsKey` answers true | yes, `without` assocs nil | `[:maybe T]` |
| anything else, in the `$extmap` | yes, dissoc removes it | no, `setup-shape` drops nils through `d/without-nils` | `{:optional true} T` |

Verified in a REPL against this branch:

```clojure
(def r (cts/setup-shape {:type :rect :x 1 :y 2 :width 3 :height 4 :rx 0}))
(cts/shape? (dissoc r :x))                     ;; => true, still a Shape
[(contains? (dissoc r :x) :x)
 (:x (dissoc r :x))]                           ;; => [true nil]
(contains? (dissoc r :rx) :rx)                 ;; => false
(contains? (cts/setup-shape {… :rx nil}) :rx)  ;; => false, the nil was stripped
```

**What changed in this revision.**

| entries | first revision | now |
|---|---|---|
| `rx`, `ry`, `hide-in-viewer`, `use-for-thumbnail`, `svg-attrs`, `svg-defs`, `svg-transform`, `svg-viewbox`, svg-raw `content` | `{:optional true}` + `[:maybe T]` | `{:optional true} T`, the `[:maybe]` dropped |
| `NilableGeometryAttrs`: `x`, `y`, `width`, `height` | `{:optional true}` + `[:maybe T]` | `[:maybe T]`, the `{:optional true}` dropped |
| `rotation`, `flip-x`, `flip-y` | `{:optional true}` + `[:maybe T]` | unchanged, and the next section says why |

A census over 400 generated shapes agrees with the rule: of the nine entries that lost the `[:maybe]`, not one instance is nil, while `flip-x` is nil throughout and bool and path have nil geometry in every sample.

* **QUESTION**: is that the rule you meant? If you would rather keep `{:optional true}` everywhere for uniformity, say so and I will make the geometry group optional again; the cost is only that the schema stops saying a record field is always there.

### `schema:shape-generic-attrs` is also a write contract, which is why three keys keep both markers

`schema:shape-generic-attrs` does two jobs. Inside `schema:shape-attrs` it describes a whole stored shape. Standalone, `check-shape-generic-attrs` validates a **partial update payload**, the `{:blocked true}` that `app.main.data.workspace/update-shape` and `app.main.data.workspace.shapes/update-shape-flags` pass. Every key in it must therefore stay optional, whatever the record says, so `rotation`, `flip-x` and `flip-y` carry `{:optional true}` on top of the `[:maybe]` the record earns them.

`schema:nilable-geom-attrs` has no such second job, so it takes the required form. The general version of the problem is the last question in this description.

### What is declared, measured over a 305-shape corpus

| attribute | shapes carrying it | declared before |
|---|---:|---|
| `rotation` | 305 / 305 | no, record field |
| `flip-x`, `flip-y` | 305 / 305, value nil | no, record fields |
| `hide-in-viewer` | 197 | frames only |
| `svg-attrs`, `svg-defs`, `svg-viewbox` | 101 | no |
| `svg-transform` | 63 | no |
| `x`/`y`/`width`/`height` on bool and path | all, 10 paths hold nil | not in those two branches |

Plus `use-for-thumbnail` on frames, which `app.common.files.migrations` renames `:use-for-thumbnail?` to and `app.common.logic.libraries` reads; `rx`/`ry` on rects and circles, the legacy radii SVG import parses off the element and migration 0003 assocs as `0`; and `content` on svg-raw, which `shapes-builder/create-raw-svg` sets.

Two entries are worth a closer look.

**The `svg-*` group is typed `:map` on purpose**, not `::gmt/matrix` and `::grc/rect`. Legacy files hold `svg-transform` as a plain `{:a … :f}` map and `svg-viewbox` as either a record or a plain map, so a tighter schema would reject files that are otherwise valid. **I would probably like tighter types here**, but I do not know enough about the legacy we need to support for back-compatibility to submit that confidently.

* **QUESTION**: `:map`, or tighter types for `svg-*`?

**`content` on svg-raw is `[:or :map :string]`**, because a bare text node arrives as the string itself: `<text>hi</text>` becomes one svg-raw for the element and another for `"hi"`. `shapes-builder/parse-svg-element` carries a FIXME saying exactly that. I first typed it `:map` and the plugin API suite caught it. I am out of my depth here.

### One thing is tightened now

`schema:nilable-geom-attrs` **requires** `x`, `y`, `width` and `height` on the bool and path branches, nilable. Bool and path are the only two shape types whose geometry can be nil: `make-minimal-shape` gives the four keys a default for every other type and skips those two, whose extent their content and `selrect` imply instead. Those two branches therefore cannot merge `schema:shape-geom-attrs`, which rejects the nil.

Three things bound the risk:

1. `schema:shape` is `[:and [:fn shape?] schema:shape-attrs]`, so every value it validates is a `Shape` record, and `containsKey` answers true for all four keys by construction.
2. Every branch of `schema:shape-attrs` already requires the nine keys of `schema:shape-base-attrs`, so no partial map validated against it before this change either. `check-shape-attrs` throws unconditionally rather than under an assert flag, so a call site passing a partial map would already be failing today.
3. The other seven branches have required those same four keys all along, through `schema:shape-geom-attrs`. Bool and path were the exception, not the rule.

* **QUESTION**: would you like those two branches to stay silent about geometry? In that case I can drop `schema:nilable-geom-attrs` and keep the rest. The cost is that a schema-derived reader continues to see a bool or a path as having no position or size at all. 

### NOTE: we cannot make an analogous change to `ctf/schema:file` (for the moment)

This is mostly a note to future me. The file map carries `:backend`, `:comment-thread-seqn` and `:ignore-sync-until`, none of which `ctf/schema:file` declares. **If we declared them, we would break saving**, which I found out the hard way at 185 failures, mostly `rpc-file-test`.

Why? Because `app.binfile.common/update-file!` derives its UPDATE column list from the keys of a file map, and the `file` table has no `backend` column, since it is synthesized on read.

`check-shape-generic-attrs` above is the same conflation in `common/`, and a milder one: there the write contract only costs us the ability to require a key.

* **QUESTION: should the schema serve as both read description and write contract?** I am very curious to know how you think about this, though it is a discussion we could have separately from this pull request, which solves the current problem within the framework that exists. I can open an issue for it, if desired.

### Verifications

Adding entries changes what `shape-generator` produces, so generative tests begin exercising code paths with these attributes present. That is where a problem would surface.

| check | result |
|---|---|
| common suite, Clojure | 1142 tests, 24702 assertions, 0 failures |
| common suite, ClojureScript | 992 tests, 24017 assertions, 0 failures, 0 errors |
| `cljfmt check --parallel=true src/ test/` in `common/` | all source files formatted correctly |
| `clj-kondo --parallel --lint ../common/src/ src/` | 0 errors, 0 warnings |
| 400 generated shapes against `cts/schema:shape` | all valid |
| Plugin API test suite | passes, including the svg-raw import case |

Every test workflow carries `if: ${{ !github.event.pull_request.draft }}`, so none of them runs while this pull request is a draft. The numbers above are local runs against the pushed tree.

### Checklist

- [ ] Include screenshots or videos, if applicable. *(nothing visible)*
- [ ] Add or modify existing integration tests in case of bugs or new features, if applicable. *(the generative shape tests cover it by construction)*
- [ ] Check CI passes successfully. *(skipped while draft)*

AI-assisted-by: mixed models

### DIFF
--- a/common/src/app/common/types/shape.cljc
+++ b/common/src/app/common/types/shape.cljc
@@ -384,7 +384,10 @@
   (->> (sg/generator schema:shape-base-attrs)
        (sg/mcat (fn [{:keys [type] :as shape}]
                   (sg/let [attrs1 (sg/generator schema:shape-generic-attrs)
-                           attrs2 (sg/generator schema:shape-geom-attrs)
+                           attrs2 (if (or (= type :path)
+                                          (= type :bool))
+                                    (sg/generator schema:nilable-geom-attrs)
+                                    (sg/generator schema:shape-geom-attrs))
                            attrs3 (case type
                                     :text    (sg/generator schema:text-attrs)
                                     :path    (sg/generator schema:path-attrs)
@@ -395,10 +398,7 @@
                                     :bool    (sg/generator schema:bool-attrs)
                                     :group   (sg/generator schema:group-attrs)
                                     :frame   (sg/generator schema:frame-attrs))]
-                    (if (or (= type :path)
-                            (= type :bool))
-                      (merge attrs1 shape attrs3)
-                      (merge attrs1 shape attrs2 attrs3)))))
+                    (merge attrs1 shape attrs2 attrs3))))
        (sg/fmap create-shape)))
 
 (def schema:shape-attrs
--- a/common/src/app/common/types/shape.cljc
+++ b/common/src/app/common/types/shape.cljc
@@ -233,7 +233,50 @@
    [:grow-type {:optional true}
     [::sm/one-of grow-types]]
    [:applied-tokens {:optional true} cto/schema:applied-tokens]
-   [:plugin-data {:optional true} ctpg/schema:plugin-data]])
+   [:plugin-data {:optional true} ctpg/schema:plugin-data]
+
+   ;; `rotation`, `flip-x` and `flip-y` are fields of the `Shape` record (see
+   ;; `cr/defrecord Shape` above) and this schema did not declare them.
+   ;; `rotation` was already named in `allowed-shape-attrs` here and in
+   ;; `app.common.types.shape.attrs/editable-attrs`, so the omission was in this
+   ;; schema and not in the model. Anything reading the model from the schema
+   ;; rather than from a live shape missed all three: the graph projection
+   ;; derives one column per entry (`app.graph.schema.projection`), so shape
+   ;; nodes carried no rotation at all, and a consumer cannot place a shape
+   ;; without it.
+   ;;
+   ;; Nilable, because `app.common.record/defrecord` cannot remove a base
+   ;; field: its `without` assocs nil and its `containsKey` answers true
+   ;; whatever the field holds, so nil is how a record field says "unset".
+   ;; `flip-x` and `flip-y` are nil on every shape `setup-shape` builds, since
+   ;; `make-minimal-shape` gives them no default.
+   ;;
+   ;; Optional as well, unlike the geometry group below, because this schema
+   ;; has a second job: `check-shape-generic-attrs` validates partial update
+   ;; payloads with it, such as the `{:blocked true}` that
+   ;; `app.main.data.workspace/update-shape` passes. A required key here would
+   ;; reject every such payload.
+   [:rotation {:optional true} [:maybe ::sm/safe-number]]
+   [:flip-x {:optional true} [:maybe :boolean]]
+   [:flip-y {:optional true} [:maybe :boolean]]
+
+   ;; Carried on circles, rects and texts too, not only on frames, so it
+   ;; belongs here rather than in `schema:frame-attrs`. Not nilable: the key
+   ;; lives outside the record, `app.common.logic.shapes` dissocs it to unset
+   ;; it, and `setup-shape` drops it when a caller passes nil.
+   [:hide-in-viewer {:optional true} :boolean]
+
+   ;; The SVG provenance an import leaves on a shape. Typed `:map` rather than
+   ;; more precisely on purpose: legacy files hold `svg-transform` as a plain
+   ;; `{:a … :f}` map rather than a `::gmt/matrix` record, and `svg-viewbox` as
+   ;; either a `::grc/rect` record or a plain map, so a tighter schema here
+   ;; would reject files that are otherwise valid. The graph *column* types are
+   ;; tightened separately, where a wrong guess costs a column rather than a
+   ;; rejected file (`app.graph.schema.contract/type-overrides`).
+   [:svg-attrs {:optional true} :map]
+   [:svg-defs {:optional true} :map]
+   [:svg-transform {:optional true} :map]
+   [:svg-viewbox {:optional true} :map]])
 
 (def schema:group-attrs
   [:map {:title "GroupAttrs"}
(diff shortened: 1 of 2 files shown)


alvorithm:
The Backend job here is red and cannot go green on its own: it dies at the Lint step with one `clj-kondo` warning, `develop`'s own unused binding from #11014, before running a single test. #11124 removes it. Nothing in this pull request touches `backend/`.

What this pull request does affect is the common suite, and that is green: 1142 tests, 24702 assertions, 0 failures. Frontend, Library, Bundle, Plugin API and composable are green too.

alvorithm:
Correction to my comment above: the Backend job is green now. #11124 merged into `staging` and reached `develop` in `e1c51442cd`, and this branch is rebased onto that, so the `clj-kondo` warning that was killing the Lint step is gone.

Worth recording why re-running the old job did not show it: GitHub replays the original commit a run was created for, which for a pull request is the merge with the base **as the base stood at the time**. A re-run therefore tests the same tree however far the base has moved. Only a push against a new base produces a meaningful run.

Current state: Backend, Common, Frontend, Library, Bundle, Plugin API and composable all pass. Integration Tests is still running and passed on the previous head.

alvorithm:
@niwinz thanks, that settles two of the questions and improves the generator beyond what I had.

**What your commit does.** Drawing bool and path geometry from `schema:nilable-geom-attrs` rather than skipping it means the generator now produces a real number for those two types roughly 60% of the time. Before, it produced nil every time: the attrs map carried no geometry at all and `create-shape` filled the record fields, so no test ever exercised a bool or a path with a position. Measured over 400 generated shapes, 49 of the 83 bool-or-path shapes now carry a number in `:x`.

Your test also answers two of my questions more clearly than prose would have, so I am treating both as closed: the optional/nilable split is the rule you meant, and bool and path should not stay silent about geometry.

**One fix pushed on top, `51c80644a5`.** `sg/sample` reads `:size`, not `:num`. `:num` is test.check's option and is correct for the `smt/check!` call directly above it, which is where it came from, but `sg/sample` ignores it and falls back to its default of 10 samples. Ten leaves the bool and path assertions vacuous about one run in fourteen: simulated over 200 draws of 10, 14 contained no bool and no path at all, and the median draw held 2. The assertion count shows it, 42 with `:num` against 756 with `:size`, and the common suite goes from 1143 tests and 24744 assertions to 1143 tests and 25458 assertions, no failures either way.

**Two questions are still open**, and neither needs to hold up the merge:

1. `:map`, or tighter types for the `svg-*` group.
2. Whether a schema should serve as both read description and write contract. It bites twice here: `ctf/schema:file` cannot declare `:backend` without breaking saving, and `schema:shape-generic-attrs` cannot require a key because `check-shape-generic-attrs` validates partial update payloads.

Tell me if you would rather have an issue for each than leave them in a merged pull request's description, and I will open them.

niwinz:
> **Two questions are still open**, and neither needs to hold up the merge:
> 
> 1. `:map`, or tighter types for the `svg-*` group.
> 2. Whether a schema should serve as both read description and write contract. It bites twice here: `ctf/schema:file` cannot declare `:backend` without breaking saving, and `schema:shape-generic-attrs` cannot require a key because `check-shape-generic-attrs` validates partial update payloads.
> 
> Tell me if you would rather have an issue for each than leave them in a merged pull request's description, and I will open them.

- The `schema:shape-generic-attrs` can not have required key -> EXACTLY
- The `:backend` key on schema:file i'm not aware of it so open a bug and assign it to me please (feel free to use LLM to try to create a FIX for if you have time and motivation hehe)
- I'm not clearly understand what you mean with `:map` on the first point

PR #11145, opened 2026-08-08
:sparkles: Close search, history and shortcuts panels with Escape

### LINKED ISSUE #9540: Close workspace search with the Escape key

filipsajdak:

Fixes #9540.

## Approach

The key constraint from the issue is that Escape must close a panel **only when the focus is already inside it**, and behave normally otherwise.

`layers.cljs` already had an Escape handler, but it is registered on the `document`:

```clojure
(mf/with-effect []
  (let [key1 (events/listen globals/document "keydown" on-key-down)
        ...
```

A document listener cannot tell where the focus is, so closing the panel from there would have dismissed search even when the user was typing in the layers list. Instead the handler is bound to the panel itself, following the existing `versions.cljs` / `sitemap.cljs` idiom (`kbd/esc?` inside an `:on-key-down`). The event bubbles up from whichever descendant holds the focus, so "focus is inside the panel" comes for free and no containment check is needed.

| Panel | Handler bound to | Closes with |
|---|---|---|
| Search (layers / canvas) | `tool-window-bar` + `replace-wrapper` | `dw/close-layers-search` |
| Shortcuts | panel root | existing `close-fn` |
| History | `right-sidebar-content`, guarded on `is-history?` | `on-close-document-history` |

Two details:

- **Nested dismissal.** If the filters dropdown is open in the search panel, Escape closes that first and leaves the panel open, so the UI unwinds in the order the user expects.
- **Returning focus to the canvas** reuses what the existing close buttons already do — `toggle-search` blurs the focused node before emitting — rather than introducing a new focus target. `stop-propagation` then keeps the document level handler from acting on the same keypress.

Scope kept to the three panels named in the issue; the "other panels should be audited" part is deliberately left out.

## Verification

Live in the devenv (`:main` build completed, 1226 files, 0 warnings):

| Case | Expected | Result |
|---|---|---|
| Search open, focus in the Find input, Escape | panel closes, focus back to canvas | closes, `activeElement` back to `BODY` |
| Search open, **focus moved to the viewport**, Escape | panel stays open | stays open |
| Filters dropdown open, focus in panel, Escape | only the dropdown closes | dropdown closed, panel still open |
| ...then Escape again | panel closes | closes, focus back to `BODY` |
| History open, focus inside, Escape | panel closes | closes, focus back to `BODY` |
| Shortcuts open (`?`), focus inside, Escape | panel closes | closes, focus back to `BODY` |

The second row is the one that matters most: it is exactly the case the document level approach would have broken.

`clj-kondo` 0 errors / 0 warnings, `cljfmt` clean, shadow-cljs `:main` 0 warnings.

No new translation strings, and no change to the plugin API surface, so no `plugins/CHANGELOG.md` entry.

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar.cljs
@@ -42,6 +42,7 @@
    [app.main.ui.workspace.tokens.sidebar :refer [tokens-sidebar-tab*]]
    [app.util.dom :as dom]
    [app.util.i18n :refer [tr]]
+   [app.util.keyboard :as kbd]
    [rumext.v2 :as mf]))
 
 ;; --- Left Sidebar (Component)
@@ -313,7 +314,19 @@
 
         active-tokens-by-type
         (mf/with-memo [active-tokens]
-          (delay (ctob/group-by-type active-tokens)))]
+          (delay (ctob/group-by-type active-tokens)))
+
+        ;; Bound to the sidebar content instead of the document so that
+        ;; Escape only dismisses the history panel when the focus is
+        ;; already inside it.
+        on-history-key-down
+        (mf/use-fn
+         (mf/deps is-history?)
+         (fn [event]
+           (when (and ^boolean is-history? (kbd/esc? event))
+             (dom/stop-propagation event)
+             (some-> (dom/get-target event) (dom/blur!))
+             (on-close-document-history))))]
 
     [:> (mf/provider muc/sidebar) {:value :right}
      [:> (mf/provider muc/active-tokens-by-type) {:value active-tokens-by-type}
@@ -341,7 +354,8 @@
                           :layout layout
                           :page-id page-id}]
 
-       [:div {:class (stl/css :right-sidebar-content)}
+       [:div {:class (stl/css :right-sidebar-content)
+              :on-key-down on-history-key-down}
         (cond
           is-comments?
           [:> comments-sidebar* {}]
--- a/frontend/src/app/main/ui/workspace/sidebar/layers.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/layers.cljs
@@ -371,6 +371,22 @@
                (st/emit! dw/close-layers-search)
                (st/emit! (dw/open-layers-search :find {:force? true}))))))
 
+        ;; Escape is handled here, on the panel itself, rather than in the
+        ;; document listener above: it must only close the search when the
+        ;; focus is already inside the panel, and leave it untouched
+        ;; otherwise. A still open filters dropdown is dismissed first.
+        on-search-key-down
+        (mf/use-fn
+         (mf/deps show-menu? hide-menu)
+         (fn [event]
+           (when (kbd/esc? event)
+             (dom/stop-propagation event)
+             (if ^boolean show-menu?
+               (hide-menu)
+               (do
+                 (some-> (dom/get-target event) (dom/blur!))
+                 (st/emit! dw/close-layers-search))))))
+
         remove-filter
         (mf/use-fn
          (fn [event]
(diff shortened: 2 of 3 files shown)


madalenapmelo-kp:

EvaMarco:
Requesting changes — the core idea (close search/history/shortcuts panels with Escape, scoped to local focus) is a good accessibility fix, but as implemented it introduces a crash and a couple of UX regressions. Details below, with suggested fixes.

🔴 **Blocking: swallows the global Escape shortcut, corrupting in-progress operations — confirmed via both History and Shortcuts panels**

Repro: open the History panel (or the Shortcuts panel), click into it so it has focus, then press P and start drawing a path on the canvas, then press Escape mid-draw.

Result: Assertion Error: expected valid shape (app.common.exceptions.js, from check-shape in [common/src/app/common/types/shape.cljc:497-499](https://github.com/penpot/penpot/blob/develop/common/src/app/common/types/shape.cljc#L497)).

Why: the global Escape shortcut ([shortcuts.cljs:159-163](https://github.com/penpot/penpot/blob/develop/frontend/src/app/main/data/workspace/shortcuts.cljs#L159)) is what properly finishes/cancels in-progress operations — for path drawing, it's what runs clear-drawing + deselect-all + the interrupt flow. It's dispatched from a single handler mounted near the root of the workspace ([shortcuts.cljs:657](https://github.com/penpot/penpot/blob/develop/frontend/src/app/main/ui/shortcuts.cljs#L657)), an ancestor of both the canvas and every sidebar panel.

Each of the three new handlers sits between that root and whatever currently has keyboard focus inside its panel. Just clicking anything inside History or Shortcuts leaves document.activeElement pointed at an element inside that panel — and drawing on canvas with the mouse doesn't move focus away. So when Escape is pressed:

The keydown's real target is still that focused element inside the panel.
The panel's handler fires first (closer ancestor) and calls dom/stop-propagation.
That prevents the event from ever reaching the root shortcuts dispatcher — clear-drawing/deselect-all/the path-interrupt flow never run.
The path is left half-finished; whatever commits it next hits check-shape with an incomplete shape and throws.

This confirms it's not specific to History or to path drawing — it's inherent to the approach in all three files. Any workspace feature relying on the global Escape shortcut (canceling a drag, exiting text edit, closing a color picker, deselecting, etc.) can break the same way, any time focus was last left on any element inside search, history, or shortcuts — even long after the user has moved on to using the mouse elsewhere. The panel doesn't need to be "in use" at the moment of the crash, just previously clicked into.

Suggested fix: don't call stop-propagation unconditionally. Either let the event continue to bubble to the root dispatcher after the panel closes itself (so the global Escape logic still runs), or explicitly check whether a canvas-level operation (drawing, editing, dragging) is in progress before swallowing the key, so the panel's own Escape doesn't take priority over an active tool operation.

🟡 **Nested-dismissal gaps in the History panel**

The search panel correctly handles "dismiss the nested thing first, then close the panel on a second Escape" (the filters dropdown). The History panel doesn't get the same treatment, even though it has its own Escape-consuming children:

Renaming a version: on-name-input-key-down in [versions.cljs:190-200](https://github.com/penpot/penpot/blob/develop/frontend/src/app/main/ui/workspace/sidebar/versions.cljs#L190) cancels the rename on Escape but never calls stop-propagation. The keydown bubbles up to the new panel-level handler too, closing the whole History panel. Before this PR, Escape while renaming just canceled the rename.
The "⋯" version context menu: built with the shared dropdown/dropdown-content* component ([versions.cljs:214](https://github.com/penpot/penpot/blob/develop/frontend/src/app/main/ui/workspace/sidebar/versions.cljs#L214) and [:306](https://github.com/penpot/penpot/blob/develop/frontend/src/app/main/ui/workspace/sidebar/versions.cljs#L306)), whose Escape-to-close is a document-level keyup listener ([dropdown.cljs:40-43](https://github.com/penpot/penpot/blob/develop/frontend/src/app/main/ui/components/dropdown.cljs#L40)) — a different event type entirely, unaffected by stop-propagation on keydown. Dismissing the menu with Escape also closes the whole History panel.

Suggested fix: mirror the search panel's approach — check whether the rename input or the dropdown menu is active before closing the panel, and let Escape dismiss those first.

PR #11146, opened 2026-08-08
:bug: Do not cache the environment generated config.js

### LINKED ISSUE #10556: Self-host: env-generated /js/config.js is cached 7 days and only build-versioned, so PENPOT_FLAGS changes don't reach already-cached browsers

filipsajdak:

Fixes #10556.

## Root cause

`nginx-entrypoint.sh` rewrites `/var/www/app/js/config.js` from the environment on every container start:

```sh
update_flags /var/www/app/js/config.js
update_oidc_name /var/www/app/js/config.js
```

but `nginx.conf.template` matched it with the catch-all static asset rule:

```nginx
location ~* \.(js|css|jpg|png|svg|gif|ttf|woff|woff2|wasm|map)$ {
    add_header Cache-Control "public, max-age=604800" always; # 7 days
}
```

`index.html` requests it as `config.js?version=<APP_BUILD>`, and a flags only restart does not change the build, so the URL is byte for byte identical. A browser that had already loaded the app kept its cached copy for up to seven days, and the new flags were invisible until the cache expired.

## Fix

Give that one file the same `no-store` headers `index.html` already gets — it is the other file whose contents change without its URL changing. An exact match `location = /js/config.js` takes precedence over the regex, so every other static asset keeps its long lived cache.

The file is ~1 KB, so the extra revalidation per load is negligible. Content hashing the URL would also work, but it would mean regenerating `index.html` at container start too; this keeps the change to a single location block.

## Verification

Ran nginx (`nginx:alpine`) against the modified config with a stub `/js/config.js` and `/js/main.js`, before and after the change:

| Request | Before | After |
|---|---|---|
| `/js/config.js` | `public, max-age=604800` | `no-store, no-cache, max-age=0` |
| `/js/main.js` | `public, max-age=604800` | `public, max-age=604800` (unchanged) |
| `/` | `no-store, no-cache, max-age=0` | unchanged |

`nginx -t` reports the configuration valid. The `main.js` row is the regression check: the long lived cache for ordinary assets is untouched.


madalenapmelo-kp:

filipsajdak:
Retargeted this from `develop` to `staging`, and rebased accordingly (single commit, same change).

@luisddm mentioned on #11168 that you are in the feature freeze for 2.18 and asked for that fix to go to `staging`, so I assumed the same applies to this one since it is also a bug fix. Happy to move it back to `develop` if you would rather it waited for the next cycle — just say so.

niwinz:
> Retargeted this from `develop` to `staging`, and rebased accordingly (single commit, same change).
> 
> @luisddm mentioned on #11168 that you are in the feature freeze for 2.18 and asked for that fix to go to `staging`, so I assumed the same applies to this one since it is also a bug fix. Happy to move it back to `develop` if you would rather it waited for the next cycle — just say so.

Nice! Thanks

PR #11151, opened 2026-08-09
:bug: Use a single translation key for the Mixed values label

### LINKED ISSUE #11148: Blur options and numeric inputs show an inconsistent Mixed label in 24 locales
### Description

The design sidebar uses two different translation keys for the same "mixed values" label. Most sections (Fill, Stroke, Export, Border radius, Constraints, Layout, Component) use `settings.multiple`, while the Blur section (`blur.cljs:299`) and the design-system numeric input (`numeric_input.cljs:707`) use `labels.mixed-values`. Both read "Mixed" in English, so the split is invisible in the default locale — but the two keys are translated inconsistently, and the difference surfaces everywhere else.

In 16 locales `labels.mixed-values` has no translation at all while `settings.multiple` does. An empty translation falls back to the default language (`app/util/i18n.cljs:178`), so blur options and every numeric input with mixed values render the English word "Mixed" next to sections that correctly show the localized one: ar, ca, cs, el, eu, fa, ha, hr, id, ms, pl, pt_BR, pt_PT, sr, yo, zh_Hant.

In 8 more locales both keys are translated but with different wording, so the same concept is named two different ways within a single sidebar:

| Locale | `settings.multiple` | `labels.mixed-values` |
|--------|---------------------|-----------------------|
| fr | Divers | Mélangé |
| fr_CA | Multiple | Mixtes |
| he | מעורב | מעורבב |
| ko | 혼합됨 | 혼합 |
| lv | Jaukts | Jauktās |
| ru | Смешаный | Смешать |
| sv | Mixad | Blandad |
| ukr_UA | Змішаний | Змішано |

That is 24 of 49 locales affected. The numeric input is a design-system component reused across the sidebar, so the inconsistency is not limited to the blur options.

### Steps to reproduce

1. Set the interface language to Portuguese (Brazil), or any of the 16 locales listed above.
2. Select two shapes that have different fill colors and different blur values.
3. Open the design sidebar and compare the Fill and Blur sections.

### Expected behavior

Both sections name the mixed state with the same localized word. Fill shows "Misto" in pt_BR, so Blur should show "Misto" as well, rather than falling back to the English "Mixed".

Standardising the two call sites on `settings.multiple` fixes all 24 locales at once, since that key is the one already translated everywhere and already used by the majority of the sidebar. Filling in the missing `labels.mixed-values` translations would instead require 16 new translations and would still leave the 8 locales where the two keys disagree.

### Affected versions

2.17.0

filipsajdak:

Fixes #11148.

## The split

Two keys name the same concept. `settings.multiple` is used across the sidebar (measures, stroke, color, layout, typography, exports — ~18 call sites) and ships a translation in **33** locale files. `labels.mixed-values` had only **2** call sites and exists in **17**.

| | call sites | locale files |
|---|---|---|
| `settings.multiple` | ~18 | 33 |
| `labels.mixed-values` | 2 | 17 |

Because both read "Mixed" in English the split is invisible in the default locale, but elsewhere it shows. Spot-checked against the `.po` files:

- **pl** — `labels.mixed-values` is absent entirely, `settings.multiple` is `"Mieszane"`. The missing string falls back to the default language (`i18n.cljs:178`), so the blur label and every numeric input rendered the English "Mixed" beside sections showing the Polish word.
- **fr** — `"Divers"` vs `"Mélangé"`; **ru** — `"Смешаный"` vs `"Смешать"`. Same sidebar, same concept, two names.

## Fix

Point the two outliers — `ds/controls/numeric_input.cljs` and `options/menus/blur.cljs` — at `settings.multiple`. Nothing else changes, and `labels.mixed-values` is now unused in `src/`.

I did **not** hand-edit the `.po` files to drop the now-unused key: translations are managed externally, so removing the entry belongs to that tooling rather than to this PR. Say the word if you would prefer it stripped here.

## Verification

Live in the devenv with the profile language set to **Polski**, two rectangles selected with different widths so the inputs go mixed — before and after the change in the same session:

| | DS numeric input placeholder |
|---|---|
| before | `Mixed` (English fallback) |
| after | `Mieszane` |

`clj-kondo` 0 errors / 0 warnings, `cljfmt` clean, shadow-cljs `:main` recompiled with 0 warnings.

### DIFF
--- a/frontend/src/app/main/ui/ds/controls/numeric_input.cljs
+++ b/frontend/src/app/main/ui/ds/controls/numeric_input.cljs
@@ -723,7 +723,7 @@
                                 :id id
                                 :class inner-class
                                 :placeholder (if is-multiple?
-                                               (tr "labels.mixed-values")
+                                               (tr "settings.multiple")
                                                placeholder)
                                 :default-value (fmt/format-number (or (mf/ref-val last-value*) value))
                                 :on-blur handle-blur
--- a/frontend/src/app/main/ui/workspace/sidebar/options/menus/blur.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/menus/blur.cljs
@@ -296,7 +296,7 @@
         (if mixed-state
           [:div  {:class (stl/css :first-row)}
            [:span {:class (stl/css :mixed-label)}
-            (tr "labels.mixed-values")]
+            (tr "settings.multiple")]
            [:> icon-button* {:variant "ghost"
                              :aria-label (tr "workspace.options.blur-options.remove-blur")
                              :on-click handle-delete-all


madalenapmelo-kp:

PR #11154, opened 2026-08-10
:sparkles: Add show/hide UI toggle in View Mode

### LINKED ISSUE #10048: Add show/hide UI toggle in View Mode for presenting prototypes without full screen

When presenting a desktop prototype, it would be nice to show the view mode in the browser context without having to go full screen (which doesn't properly emulate normal browser usage.)

Show / hide the penpot UI when in View Mode.  Mimic the current show/ hide ui behaviour in the editor. The top bar could also include a "show / hide UI" button.

The design should then move 'up' to fill the space where the penpot ui was previously shown.

The viewmode UI could be hidden by default, and only appear on mouse-move or click. However I think this would be confusing to new users + still doesn't allow showing of prototypes without the ui outside of fullscreen mode.

### Additional context

_No response_

0xTHAC0:
## Description

Adds a show/hide UI toggle in View Mode, allowing users to present desktop prototypes in the browser without going fullscreen.

Fixes #10048

## Motivation

When presenting a desktop prototype, users want to show the design in a clean view within the browser context — without entering browser fullscreen mode (which changes browser chrome behavior and doesn't properly emulate normal browser usage). The existing fullscreen button requires the browser Fullscreen API; this adds a dedicated in-page toggle.

## Changes

- **State:** Added `:hide-ui?` flag to `viewer-local` default state (defaults to `false`)
- **Event:** Added `toggle-hide-ui` event following the same pattern as `toggle-fullscreen`
- **Shortcut:** Added `\` keyboard shortcut — consistent with the editor's existing hide-ui shortcut
- **UI:** Eye icon toggle button in the header options bar (between the edit and fullscreen buttons). Icon switches between `shown`/`hide` states
- **CSS:** `[data-hide-ui="true"]` rules slide the header up, pagination bar down, and nav arrows off-screen — same CSS transform mechanism as the fullscreen mode
- **Hover reveal:** Header reappears on hover (same 48px invisible hitbox approach as fullscreen)
- **Session persistence:** Toggle state persists during the view session (stored in `viewer-local` atom)
- **Independent:** Works separately from fullscreen mode; both can be active simultaneously
- **i18n:** Added `viewer.header.hide-ui` translation key

## Behavior

1. Click the eye icon in the header or press `\` to hide all UI chrome
2. The header, pagination bar, and frame navigation arrows slide off-screen
3. The design canvas expands to fill the full viewer area
4. Hover over the top edge to temporarily reveal the header
5. Click or press `\` again to restore the full UI

## Testing

- [x] Toggle button shows/hides on click
- [x] `\` keyboard shortcut works
- [x] Chrome hides: header, pagination counter, nav arrows
- [x] Header reappears on hover
- [x] State persists across frame navigation
- [x] Works independently of fullscreen mode
- [x] No regression in View Mode navigation

## Screenshots

_Visual verification requires the Penpot devenv. The implementation follows the same CSS transform patterns as the existing fullscreen mode._

## DCO

I hereby agree to the terms of the DCO.
Signed-off-by: Sumit Ridhal <sridhal@redhat.com>


madalenapmelo-kp:

niwinz:
Triaged. Ready to review. Thanks for the contribution.

EvaMarco:
Hello @0xTHAC0! I found two issues with this PR.

1. Fullscreen throws an error

When I try to enter fullscreen now, I get this error:

Screenshot from 2026-09-23 13-18-22

The problem is the way the refs are dereferenced here. mf/deref is not a plain function call — it wraps React's useSyncExternalStore hook (as confirmed in rumext/v2.cljs), so it must be called on every render, in the same order.

However, or short-circuits: the second argument is only evaluated when the first one is falsy.

So when fullscreen? is false, both refs are dereferenced:
```
(mf/deref fullscreen-ref) ; → false
(mf/deref hide-ui-ref)    ; → false
```
But when entering fullscreen, fullscreen-ref becomes true, so the second deref is skipped:
```
(mf/deref fullscreen-ref) ; → true
;; (mf/deref hide-ui-ref) is not evaluated
```
This means React goes from two hook calls to one between renders, which causes the Rendered fewer hooks than expected error. This also explains why the error happens specifically when toggling into fullscreen.

The fix should be to deref both refs unconditionally and then combine their values:
```
(let [fullscreen? (mf/deref fullscreen-ref)
      hide-ui?    (mf/deref hide-ui-ref)]
  ...
  [:header {:class (stl/css-case :viewer-header true
                                 :fullscreen (or fullscreen? hide-ui?))
            ...}])
```
Also, fullscreen-ref is already dereferenced once at the top of the let (around line 129), so that existing binding can be reused instead of dereferencing it again inline.

2. There is some duplicated CSS

There are several places where the CSS for data-hide-ui is duplicated from the existing fullscreen rules. For example, instead of:
```scss
[data-fullscreen="true"] .viewer-header::after {
  content: " ";
  position: absolute;
  width: 100%;
  height: deprecated.$s-48;
  left: 0;
  top: deprecated.$s-48;
}

/** HIDE UI (without browser fullscreen) */
[data-hide-ui="true"] .viewer-header::after {
  content: " ";
  position: absolute;
  width: 100%;
  height: deprecated.$s-48;
  left: 0;
  top: deprecated.$s-48;
}
```
I think it would be better to share the rules:
```scss
[data-fullscreen="true"] .viewer-header::after,
[data-hide-ui="true"] .viewer-header::after {
  content: " ";
  position: absolute;
  width: 100%;
  height: deprecated.$s-48;
  left: 0;
  top: deprecated.$s-48;
}
```
And the same approach could be used for the other duplicated rules in both SCSS files. This way we avoid having two copies of the same styles that could potentially get out of sync later.

PR #11168, opened 2026-08-10
:bug: Keep comment bubbles from painting over the rulers

### LINKED ISSUE #11163: Comment avatars appear on top of rulers when scrolling canvas
### Steps To Reproduce

1. Create a new file 
2. Add a comment on the canvas
3. Ensure rulers are visible
4. Scroll or pan the canvas
5. Observe the comment avatar circles as they move across the ruler areas

### Expected behavior

Comment avatars should render behind the rulers.

### Actual behavior

Comment avatars appear on top of the rulers, partially or fully covering the ruler ticks and numbers.

### Screenshots or video

Image

Linux  / Chrome

_No response_

PRO

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

filipsajdak:

Fixes #11163.

## Root cause

The comments layer renders inside `.viewport-overlays`, which is `position: absolute` with `z-index: 1` (`viewport.scss:29-35`), and its own container adds `z-index: 1000` (`viewport/comments.scss:9-19`). The rulers are drawn by the canvas underneath. So although the overlays appear *before* the canvas in DOM order, they are lifted above it, and a bubble panned into the ruler bars paints over the ticks and numbers.

Confirmed in the workspace with `document.elementFromPoint` at a point inside the horizontal ruler, with a comment positioned there:

```
main_ui_comments__avatar-mask
 └ main_ui_comments__floating-preview-wrapper
   └ main_ui_workspace_viewport_comments__threads
```

## Fix

Clip the comments container to the area outside the ruler bars while the rulers are shown, reusing `rulers/ruler-area-size` rather than repeating the constant.

This is the approach already used for the same problem elsewhere: `viewport.cljs` defines a `clip-handlers` clip path "so the handlers are not over the rulers", and `rulers-clip-path*` is documented as keeping overlays from painting over the rulers.

Clipping `#comments` specifically — rather than `.viewport-overlays` — matters: the overlays also host the text editing layer, and clipping that could affect editing a text shape near the top-left of the viewport. Only the comments container is touched.

`.threads` itself is a zero-size anchor whose bubbles overflow it, so it is not a usable clip target; `#comments` is the full-viewport, comments-only container.

## Verification

Live in the devenv, comment thread placed so its bubble sits inside the horizontal ruler, same point before and after:

| | topmost element at that point |
|---|---|
| before | `main_ui_comments__avatar-mask` (the comment avatar) |
| after | `rect.cursor-resize-ns-0` inside `g.guides` (the ruler layer) |

The bubble keeps its layout position (unchanged at the same coordinates) — only its painting is clipped, so nothing moves or disappears outside the ruler bars. With rulers hidden no clip is applied at all.

`clj-kondo` 0 errors / 0 warnings, `cljfmt` clean, shadow-cljs `:main` recompiled with 0 warnings.

### DIFF
--- a/frontend/src/app/main/ui/workspace/viewport.cljs
+++ b/frontend/src/app/main/ui/workspace/viewport.cljs
@@ -363,7 +363,8 @@
                                       :page-id page-id
                                       :file-id file-id
                                       :vport vport
-                                      :zoom zoom}])
+                                      :zoom zoom
+                                      :show-rulers show-rulers?}])
 
       (when picking-color?
         [:> pixel-overlay/pixel-overlay* {:vport vport
--- a/frontend/src/app/main/ui/workspace/viewport/comments.cljs
+++ b/frontend/src/app/main/ui/workspace/viewport/comments.cljs
@@ -13,6 +13,7 @@
    [app.main.refs :as refs]
    [app.main.store :as st]
    [app.main.ui.comments :as cmt]
+   [app.main.ui.workspace.viewport.rulers :as rulers]
    [rumext.v2 :as mf]))
 
 ;; Pin transform for the bubble's frame so it follows the frame during a drag,
@@ -70,7 +71,7 @@
 
 (mf/defc comments-layer*
   {::mf/wrap [mf/memo]}
-  [{:keys [vbox vport zoom file-id page-id]}]
+  [{:keys [vbox vport zoom file-id page-id show-rulers]}]
   (let [vbox-x      (dm/get-prop vbox :x)
         vbox-y      (dm/get-prop vbox :y)
         vport-w     (dm/get-prop vport :width)
@@ -114,7 +115,15 @@
       {:id "comments"
        :class (stl/css :workspace-comments-container)
        :style {:width (dm/str vport-w "px")
-               :height (dm/str vport-h "px")}}
+               :height (dm/str vport-h "px")
+               ;; This layer sits above the canvas, so without clipping the
+               ;; bubbles paint over the rulers as they pan past them. Keep
+               ;; them out of the ruler bars, like `clip-handlers` does for
+               ;; the selection handlers.
+               :clip-path (when show-rulers
+                            (dm/fmt "inset(%px 0 0 %px)"
+                                    rulers/ruler-area-size
+                                    rulers/ruler-area-size))}}
       [:div {:class (stl/css :threads)
              :style {:transform (dm/fmt "translate(%px, %px)" pos-x pos-y)}}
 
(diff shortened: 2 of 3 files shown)


madalenapmelo-kp:

luisddm:
@filipsajdak Thanks for your contribution! It looks great. However, I'd like to kindly ask for two little changes before merging:
- Change the base branch to staging; right now we are in the feature freeze phase for the 2.18 release, and we'd like this to be fixed for that version.
- Add the `:show-rulers` parameter to the `comments-layer*` component also in `viewport_wasm.cljs` so it also works with the new WASM render.

filipsajdak:
Thanks @luisddm! Both done:

- Rebased onto `staging` and retargeted the PR there.
- Added `:show-rulers` to the `comments-layer*` call in `viewport_wasm.cljs` too, using the `show-rulers?` binding already present in that component, so the clip applies with the WASM render as well.

Still a single commit, now touching `viewport.cljs`, `viewport_wasm.cljs` and `viewport/comments.cljs`. `clj-kondo` and `cljfmt` clean against `staging`.

One note in case it matters for the WASM path: the clip is applied to the `#comments` container in CSS, so it does not depend on how the rulers themselves are drawn — it just keeps the bubbles out of the ruler bars whenever the rulers are visible. I verified the behaviour on the SVG render locally; I do not have a WASM build here, so a quick check on your side would be welcome.

luisddm:
Thank you for the changes, works as expected. I'll wait for CI checks and then will merge it :)

filipsajdak:
Update on my previous note: I got a WASM build running locally, so I was able to verify that path too — no need for you to check it.

Enabled WebGL rendering in the user settings (so `render-wasm/v1` is active and `viewport.cljs` dispatches to `viewport.wasm/viewport*`), confirmed the workspace is on the WASM renderer (`canvas[data-testid="canvas-wasm-shapes"]`), and placed a comment thread so its bubble sits inside the horizontal ruler.

Result on the WASM render:

- `#comments` receives `clip-path: inset(22px 0px 0px 22px)`, i.e. the `:show-rulers` prop added to the `viewport_wasm.cljs` call site flows through as expected.
- `document.elementFromPoint` at a point inside the ruler returns `rect.cursor-resize-ns-0` (the guides/ruler layer) instead of `main_ui_comments__avatar-mask`.
- Removing the clip in the DOM flips it straight back to the avatar being on top, which confirms the clip is what is doing the work rather than some incidental difference between the two renderers.

The bubble keeps its layout position throughout — only its painting is clipped — and with the rulers hidden no clip is applied at all. Same behaviour as the SVG render.

PR #11170, opened 2026-08-10
:bug: Fix text editor crash when pasting into empty text shapes

### LINKED ISSUE #11149: Text editor crashes when pasting into an empty text shape
### Description

The text editor can crash on the paste path.

Pasting text into a text shape can crash the editor with `Unknown node type`. The paste is lost and the workspace reports an internal error. It happens when the caret is not on a text node: typically an empty text shape that was just focused, or a caret that the browser places on a paragraph/root element instead of a text node (frequent in Firefox).

### Steps to reproduce

Crash on paste:

1. Open a file in the workspace using Firefox.
2. Create a new text shape (or double-click an existing empty one) so the editor gets focus with an empty content.
3. Copy some plain text from anywhere.
4. Press Ctrl+V to paste it into the text shape.

### Steps to reproduce the crash in  a deterministic way:

1. Apply this diff
```
diff --git a/frontend/text-editor/src/editor/clipboard/paste.js b/frontend/text-editor/src/editor/clipboard/paste.js
index 683d40edb0..2bbab420e6 100644
--- a/frontend/text-editor/src/editor/clipboard/paste.js
+++ b/frontend/text-editor/src/editor/clipboard/paste.js
@@ -108,6 +108,11 @@ export function paste(event, editor, selectionController) {
     const forceTextSpan =
       fragment.firstElementChild.dataset.textSpan === "force";
     if (hasOnlyOneParagraph && hasOnlyOneTextSpan && forceTextSpan) {
+      // TEMP DEBUG repro #11149,  remove after testing
+      const para = selectionController.focusParagraph;
+      if (para) {
+        selectionController.setSelection(para, 1, para, 1);
+      }      
       selectionController.insertIntoFocus(fragment.textContent);
     } else {
       selectionController.insertPaste(fragment);
```
2. Crate a text with a single click, write some chars and paste the text (see the attached video)

https://github.com/user-attachments/assets/9a24905b-24f0-4260-92b8-4a33909ac7d4

### Expected behavior

The copied text is inserted at the caret position and the editor keeps working.

### Affected versions

2.17.1-RC5

0xTHAC0:
### Related Ticket

Closes https://github.com/penpot/penpot/issues/11149

### Summary

Three fixes for the text editor paste path:

1. **`SelectionController.insertIntoFocus()`** — handle `isParagraphFocus` and `isRootFocus` to prevent the `"Unknown node type"` crash when the caret lands on a paragraph or root element instead of a text node (common in Firefox and empty text shapes).

2. **`clipboard/paste.js`** — fix missing `selectionController` argument when calling `getFormattedOrPlainFragmentFromClipboardData` in the HTML paste path (`allowHTMLPaste`). Without it, `selectionController` receives `clipboardData` and `clipboardData` is `undefined`, causing HTML paste to always fail silently.

3. **`commands/insertText.js`** — add fallback to `insertIntoFocus()` when the selection is collapsed but focus is not on a text node or line break (paragraph/root focus). Without this, typing into an empty text shape is silently dropped.

### Steps to reproduce

1. Create a text shape
2. Double-click to enter edit mode (shape is empty)
3. Paste text (Ctrl+V / Cmd+V)
4. **Before fix:** Internal error — `Unknown node type`
5. **After fix:** Text is pasted correctly

Also reproducible in Firefox where the browser places the caret on a paragraph element rather than a text node.

### Checklist

- [x] Add a detailed explanation of how to reproduce the issue.

### DIFF
--- a/frontend/text-editor/src/editor/clipboard/paste.js
+++ b/frontend/text-editor/src/editor/clipboard/paste.js
@@ -83,6 +83,7 @@ export function paste(event, editor, selectionController) {
   let fragment = null;
   if (editor?.options?.allowHTMLPaste) {
     fragment = getFormattedOrPlainFragmentFromClipboardData(
+      selectionController,
       event.clipboardData,
     );
   } else {
--- a/frontend/text-editor/src/editor/commands/insertText.js
+++ b/frontend/text-editor/src/editor/commands/insertText.js
@@ -21,6 +21,8 @@ export function insertText(event, editor, selectionController) {
       return selectionController.insertText(event.data);
     } else if (selectionController.isLineBreakFocus) {
       return selectionController.replaceLineBreak(event.data);
+    } else {
+      return selectionController.insertIntoFocus(event.data);
     }
   } else {
     if (selectionController.isMultiParagraph) {
--- a/frontend/text-editor/src/editor/controllers/SelectionController.js
+++ b/frontend/text-editor/src/editor/controllers/SelectionController.js
@@ -1419,8 +1419,23 @@ export class SelectionController extends EventTarget {
         this.focusNode.replaceWith(textNode);
       }
       this.collapse(textNode, newText.length);
+    } else if (this.isParagraphFocus) {
+      const newTextNode = new Text(newText);
+      const newTextSpan = createTextSpan(newTextNode, this.#currentStyle);
+      this.focusNode.replaceChildren(newTextSpan);
+      this.collapse(newTextNode, newText.length);
+    } else if (this.isRootFocus) {
+      const newTextNode = new Text(newText);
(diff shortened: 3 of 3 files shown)


madalenapmelo-kp:

niwinz:
Thank you for the contribution and for investigating issue #11149.

We are going to close this PR as a duplicate of #11150. That PR addresses the same text editor paste crash and includes regression tests covering the affected caret and empty-editor cases.

The HTML paste change included here is outside the current scope of #11149, and the remaining changes overlap with the implementation in #11150. To avoid merging two competing fixes, we will continue with the tested implementation there. The current implementation also needs additional validation, as replacing paragraph or root contents directly could overwrite existing text instead of inserting at the caret position.

We appreciate the time and effort put into diagnosing and addressing this issue.

PR #11184, opened 2026-08-10
fix(exporter): compare gradient type instead of export type in SVG renderer

### LINKED ISSUE #5972: SVG export code is comparing export type with gradient type
### Steps To Reproduce

This is an issue found while reading the SVG export code, not debugging a linear gradient export issue edge case.

### Expected behavior

N/A

### Actual behavior

N/A

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

SVG export code is comparing the export `type` argument of `render` function defined  [here](https://github.com/penpot/penpot/blob/develop/exporter/src/app/renderer/svg.cljs#L112) (`:svg`) with gradient type (`"linear"`) in function `data->gradient-def` [here](https://github.com/penpot/penpot/blob/develop/exporter/src/app/renderer/svg.cljs#L168).

0xTHAC0:
## Summary

`data->gradient-def` in the SVG exporter was comparing the render `type` parameter (`:svg`, `:png`, `:pdf`) against `"linear"` to decide between `linearGradient` and `radialGradient` SVG elements. Since the export type is never `"linear"`, the comparison always fell through to `radialGradient`, causing **all linear gradients to be exported as radial** in SVG output.

## Changes

- `exporter/src/app/renderer/svg.cljs`: Read the gradient type from the data map (`(get-in data ["gradient" "type"])`) instead of the outer `type` binding from `render`'s parameter destructuring.

## How to test

1. Create a shape with a linear gradient fill
2. Export as SVG
3. Previously: the SVG contained a `<radialGradient>` element instead of `<linearGradient>`
4. Now: the SVG correctly contains `<linearGradient>` with the expected attributes

Fixes #5972

### DIFF
--- a/exporter/src/app/renderer/svg.cljs
+++ b/exporter/src/app/renderer/svg.cljs
@@ -176,8 +176,9 @@
                                         "stop-opacity" (get stop-data "opacity")}}))))
 
           (data->gradient-def [id [color data]]
-            (let [id (str "gradient-" id "-" (subs color 1))]
-              (if (= type "linear")
+            (let [id (str "gradient-" id "-" (subs color 1))
+                  gradient-type (get-in data ["gradient" "type"])]
+              (if (= gradient-type "linear")
                 {"type" "element"
                  "name" "linearGradient"
                  "attributes" {"id" id "x1" "0.5" "y1" "1" "x2" "0.5" "y2" "0"}


madalenapmelo-kp:

niwinz:
One small note: the PR title and commit subject do not follow Penpot’s contributing guidelines. They currently use `fix(exporter): ...`, while the project requires an emoji prefix, an imperative and capitalized subject, and a maximum of 70 characters. This is also why the commit-message check is failing. Please update them when convenient. Thank you!

niwinz:
Thanks for the contribution. I reviewed the implementation and confirmed that the reported issue is real: the previous code compared the export format (`:svg`) with `"linear"`, so the linear-gradient branch could never run and linear gradients handled by this path were emitted as radial gradients. Reading the gradient type from the serialized data correctly fixes that behavior.

This path appears to affect gradient-filled text exports rather than every SVG gradient. If possible, please add a focused regression test covering both linear and radial gradients. The implementation itself looks correct and appropriately minimal.

niwinz:
Thanks for the contribution. I added a focused local regression test covering both gradient types:

```clojure
(deftest creates-the-correct-gradient-element
  (doseq [[gradient-type element-name]
          [["linear" "linearGradient"]
           ["radial" "radialGradient"]]]
    (testing gradient-type
      (let [gradient-data {"type" "gradient"
                           "gradient" {"type" gradient-type
                                        "stops" gradient-stops}}
            result (svg-gradient/data->gradient-def
                    "text-id"
                    ["#000001" gradient-data])]
        (is (= element-name (get result "name")))))))
```

The test passes locally with zero failures. It verifies that the serialized gradient type selects the correct SVG element. Would you be open to adding a regression test like this?

niwinz:
I will do it myself, because it requires more plumbing.

niwinz:
Replaced with https://github.com/penpot/penpot/pull/11272

PR #11186, opened 2026-08-10
:bug: Preserve layers panel scroll position across tab switches

### LINKED ISSUE #7440: Preserve scroll on switching between right sidebar tabs

0xTHAC0:
## Problem

Fixes #7440.

When switching between the Layers, Assets, and Tokens tabs in the workspace left sidebar, the active panel component is unmounted and remounted. This causes the scroll position to reset to the top on every tab switch, forcing users to re-scroll each time they return to a panel.

## Solution

Add a module-level `scroll-positions` atom in `layers.cljs`, keyed by `page-id`. 

- A new `on-scroll*` handler saves the current `scrollTop` value into the atom whenever the user scrolls the layers panel.
- The existing scroll handler (`on-scroll`) is composed with `on-scroll*` in `on-scroll-with-save`.
- A `mf/with-effect` keyed on `page-id` restores the saved scroll position whenever `layers-toolbox*` mounts or the user switches pages.

## Changes

- `frontend/src/app/main/ui/workspace/sidebar/layers.cljs` — add `scroll-positions` atom, `on-scroll*` save helper, compose into `on-scroll-with-save`, add `scroll-ref` + restore effect to `layers-toolbox*`

### DIFF
--- a/frontend/src/app/main/ui/workspace/sidebar/layers.cljs
+++ b/frontend/src/app/main/ui/workspace/sidebar/layers.cljs
@@ -39,6 +39,9 @@
    [okulary.core :as l]
    [rumext.v2 :as mf]))
 
+(defonce ^:private scroll-positions
+  (atom {}))
+
 (def ^:private ref:highlighted-shapes
   (l/derived (fn [local]
                (-> local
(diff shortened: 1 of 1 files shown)


madalenapmelo-kp:

niwinz:
Hey @0xTHAC0 

Thanks for your time.

This PR, and many following does not follow any rule defined for contributions: commit format, pr format.
I'm looking on this PR, but many of PR's that does not follow any rule goes to the Low priority list, just FYI.

niwinz:
Additionally, the issue is about preserve scroll on all panels not only the layers. I'm working on it parting from this branch, but the approach of using global atom did not convinced me. So for, now, i say thanks, for the contribution but i will address the issue myself with a different approach, and covering the all tabs.

PR #11187, opened 2026-08-10
fix(rpc): add input validation for upload session and binfile import

### LINKED ISSUE #11103: Upload session accepts zero or negative total-chunks without validation
### Description

The `create-upload-session` RPC method accepts `total-chunks` values of 0 or negative numbers without validation. This creates an inconsistent upload session state that can lead to unexpected behavior in subsequent chunk upload operations.

### Steps to reproduce

1. Call `create-upload-session` with `total-chunks: 0` or `total-chunks: -1`
2. The session is created successfully without any validation error
3. Subsequent operations on this session may behave unexpectedly

### Expected behavior

The `total-chunks` parameter should be validated to ensure it's at least 1. Values of 0 or negative should be rejected with a validation error.

### Affected versions

Current development version

### LINKED ISSUE #11105: import-binfile accepts unsupported version values without validation
### Description

The `import-binfile` RPC method accepts any integer value for the `version` parameter without validation. When an unsupported version (e.g., 2, 0, -1) is provided, the `case` statements throw an unhandled exception instead of returning a clean validation error.

### Steps to reproduce

1. Call the `import-binfile` RPC method with `version: 2` (or any value other than 1 or 3)
2. The request fails with an unhandled exception instead of a validation error

### Expected behavior

The `version` parameter should be restricted to supported values (1 or 3) via schema validation. Unsupported versions should be rejected with a `:validation` error with code `:unsupported-version`.

Additionally, when `version` is not provided, the system should detect the file format from content (magic bytes) rather than defaulting to version 3.

### Affected versions

Current development version

0xTHAC0:
## Summary

Two small schema-level validation fixes.

### Fix #11103 — `create-upload-session` accepts zero/negative chunks

`total-chunks` was typed as `::sm/int` (any integer). Changed to `[:int {:min 1}]` so that sessions with 0 or negative chunk counts are rejected at the schema layer with a clean validation error.

### Fix #11105 — `import-binfile` accepts unsupported version values

`version` was typed as `::sm/int` (any integer). Changed to `[:enum 1 3]` to match the two supported binary-file format versions. Unsupported values (e.g. 2, 0, -1) now return a `:validation` error instead of falling through to an unhandled `case` exception.

## Changes

- `backend/src/app/rpc/commands/media.clj` — `total-chunks` schema: `::sm/int` → `[:int {:min 1}]`
- `backend/src/app/rpc/commands/binfile.clj` — `version` schema: `::sm/int` → `[:enum 1 3]`

Fixes #11103, fixes #11105.

### DIFF
--- a/backend/src/app/rpc/commands/binfile.clj
+++ b/backend/src/app/rpc/commands/binfile.clj
@@ -122,7 +122,7 @@
     [:name [:or [:string {:max 250}]
             [:map-of ::sm/uuid [:string {:max 250}]]]]
     [:project-id ::sm/uuid]
-    [:version {:optional true} ::sm/int]
+    [:version {:optional true} [:enum 1 3]]
     [:file {:optional true} media.v/schema:upload]
     [:upload-id {:optional true} ::sm/uuid]]
    [:fn {:error/message "one of :file or :upload-id is required"}
--- a/backend/src/app/rpc/commands/media.clj
+++ b/backend/src/app/rpc/commands/media.clj
@@ -289,7 +289,7 @@
 
 (def ^:private schema:create-upload-session
   [:map {:title "create-upload-session"}
-   [:total-chunks ::sm/int]])
+   [:total-chunks [:int {:min 1}]]])
 
 (def ^:private schema:create-upload-session-result
   [:map {:title "create-upload-session-result"}


madalenapmelo-kp:

luisddm:
Fixed in #11104

PR #11188, opened 2026-08-10
fix(teams): prevent admin privilege escalation via invitation role

### LINKED ISSUE #11098: Incorrect handling permissions on creating invitation
### Description

The `create-team-invitations` and `update-team-invitation-role` RPC methods allow team admins to grant or elevate invitations to the `:owner` role, bypassing the role-ceiling protection that exists in `update-team-member-role`. This inconsistency allows privilege escalation: an admin (a role explicitly below owner) can grant full ownership control to a third party.

The correct protection already exists in `update-team-member-role` (teams.clj:894-897), which rejects attempts by non-owners to promote anyone to `:owner`. However, this check was not replicated in the invitation methods.

### Steps to reproduce

1. Create a team with an owner and an admin member
2. As the admin, call `create-team-invitations` with `role: :owner`
3. The invitation is created successfully (should be rejected)
4. Alternatively, create an invitation with a lower role, then call `update-team-invitation-role` to elevate it to `:owner`
5. The role update succeeds (should be rejected)

### Expected behavior

Non-owner team members (including admins) should not be able to create or update invitations with the `:owner` role. The system should reject these requests with a `:cant-promote-to-owner` error, matching the behavior of `update-team-member-role`.

### Affected versions

2.17.x

0xTHAC0:
## Problem

Fixes #11098.

`create-team-invitations` and `update-team-invitation-role` only checked that the caller is an admin but did not prevent admins from assigning the `:owner` role to invitations. This is a privilege escalation: a team admin (a role explicitly below owner) could grant full ownership to a third party, bypassing the guard that already exists in `update-team-member-role`.

## Solution

Add the same `cant-promote-to-owner` check to both invitation methods, mirroring the protection that already exists in `update-team-member-role` (teams.clj line 895-897):

- **`create-team-invitations`**: after the `is-admin` check, collect all roles across both parameter formats and raise `:cant-promote-to-owner` if any is `:owner` and the caller is not an owner.
- **`update-team-invitation-role`**: after the `is-admin` check, raise `:cant-promote-to-owner` if `role` is `:owner` and the caller is not an owner.

## Changes

- `backend/src/app/rpc/commands/teams_invitations.clj` — two guard blocks added, one per mutation

Fixes #11098.

### DIFF
--- a/backend/src/app/rpc/commands/teams_invitations.clj
+++ b/backend/src/app/rpc/commands/teams_invitations.clj
@@ -475,6 +475,15 @@
       (ex/raise :type :validation
                 :code :insufficient-permissions))
 
+    ;; Only owners may grant the :owner role (matches update-team-member-role behaviour).
+    (let [roles (if using-emails-format?
+                  #{role}
+                  (into #{} (map :role) (:invitations params)))]
+      (when (and (not (:is-owner perms)) (contains? roles :owner))
+        (ex/raise :type :validation
+                  :code :cant-promote-to-owner
+                  :hint "only owners can invite with the :owner role")))
+
     (when (> invitation-count max-invitations-by-request-threshold)
       (ex/raise :type :validation
                 :code :max-invitations-by-request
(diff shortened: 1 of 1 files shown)


madalenapmelo-kp:

luisddm:
Already fixed in #11099

PR #11189, opened 2026-08-10
fix(media): check source file read permissions in clone-file-media-object

### LINKED ISSUE #11087: clone-file-media-object allows cloning media objects from files without read access
### Description

The `clone-file-media-object` RPC command only verifies edit permissions on the destination `file-id`. The source media object is fetched directly by UUID without checking whether the caller has access to the file that owns it.

As a result, any authenticated user with edit access to at least one of their own files can clone the metadata (and storage reference) of a media object from any other file — including private files from other teams — as long as they know or guess the source UUID. The cloned record is created under the attacker's file but reuses the original `media-id`, giving them a persistent copy of the content.

### Steps to reproduce

1. User A creates a file and uploads a media object (image).
2. User B creates their own file in a different team.
3. User B calls `clone-file-media-object` with their `file-id` as destination and User A's media object `id` as source.
4. The clone succeeds — User B now has a copy of User A's media object in their own file.

### Expected behavior

The command should verify that the caller has at least read permissions on the source file before allowing the clone. If the caller lacks access, the operation should fail with a `:not-found` error (to avoid leaking information about the existence of files/media the caller cannot access).

### Affected versions

Current main branch.

0xTHAC0:
## Problem

Fixes #11087.

`clone-file-media-object` verified edit permissions on the destination `file-id` but fetched the source media object by UUID without any access control. An authenticated user with edit access to any of their own files could clone the metadata (and storage reference) of a media object from a private file in any team, as long as they knew the source UUID.

## Solution

After the existing destination-file edition check, fetch the source media object and call `files/check-read-permissions!` on its owning file before executing the clone. Returns `:not-found` on missing or inaccessible objects to avoid leaking file/media existence to unauthorized callers.

## Changes

- `backend/src/app/rpc/commands/media.clj` — add source read-permission check to `::clone-file-media-object` handler

Fixes #11087.

### DIFF
--- a/backend/src/app/rpc/commands/media.clj
+++ b/backend/src/app/rpc/commands/media.clj
@@ -267,8 +267,15 @@
   {::doc/added "1.17"
    ::sm/params schema:clone-file-media-object
    ::db/transaction true}
-  [{:keys [::db/conn] :as cfg} {:keys [::rpc/profile-id file-id] :as params}]
+  [{:keys [::db/conn] :as cfg} {:keys [::rpc/profile-id file-id id] :as params}]
   (files/check-edition-permissions! conn profile-id file-id)
+  ;; Verify the caller can read the source media object's file before
+  ;; allowing the clone.  Return :not-found on missing/inaccessible objects
+  ;; to avoid leaking information about files the caller cannot access.
+  (let [mobj (db/get-by-id conn :file-media-object id)]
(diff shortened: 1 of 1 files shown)


madalenapmelo-kp:

luisddm:
Fixed in #11090

PR #11190, opened 2026-08-10
fix(media): scope assemble-chunks lookup to the requesting profile

### LINKED ISSUE #11011: Assemble-chunks session lookup ignores profile-id scope
### Description

The `assemble-chunks` function that stitches chunked uploads into a final file looks up the upload session using only the session-id, without verifying that the session belongs to the requesting profile. This lets a request reference another profile's session and have its chunks assembled.

The fix adds a profile-id parameter to `assemble-chunks` and scopes the `upload_session` lookup with it, matching the pattern already used by `upload-chunk`. All three callers (`assemble-file-media-object`, `create-font-variant`, `import-binfile`) are updated to pass the authenticated profile-id.

### Steps to reproduce

1. Profile A starts a chunked upload, creating an upload session, and uploads at least one chunk into it.
2. Profile B sends a request that triggers assembly (`assemble-file-media-object`, `create-font-variant`, or `import-binfile`) passing Profile A's session-id.
3. The assembly succeeds even though the session belongs to Profile A.

### Expected behavior

Assembly fails with a not-found error when the session does not belong to the requesting profile.

0xTHAC0:
## Problem

Fixes #11011.

`assemble-chunks` looked up the upload session using only `session-id`, without verifying that the session belongs to the authenticated profile. A user could trigger assembly of another profile's upload session by supplying a foreign `session-id`.

## Solution

Add `profile-id` as a required parameter to `assemble-chunks` and scope the `upload_session` lookup with `{:id session-id :profile-id profile-id}` — the same pattern already used in `upload-chunk`.

Update all three callers:
- **`assemble-file-media-object`** (media.clj) — passes `::rpc/profile-id`
- **`prepare-font-data-from-uploads`** (fonts.clj) — extracts `::rpc/profile-id` from params
- **`import-binfile`** (binfile.clj) — passes `profile-id` via `db/tx-run!` varargs

The database lookup will now return nil for sessions that don't belong to the caller, causing an automatic `:not-found` error before any assembly takes place.

Fixes #11011.

### DIFF
--- a/backend/src/app/rpc/commands/binfile.clj
+++ b/backend/src/app/rpc/commands/binfile.clj
@@ -152,7 +152,7 @@
 
         params
         (if (some? upload-id)
-          (let [file (db/tx-run! cfg media-cmd/assemble-chunks upload-id)]
+          (let [file (db/tx-run! cfg media-cmd/assemble-chunks upload-id profile-id)]
             (assoc params :file file))
           params)
 
--- a/backend/src/app/rpc/commands/fonts.clj
+++ b/backend/src/app/rpc/commands/fonts.clj
@@ -118,10 +118,10 @@
   "Assembles each chunked-upload session in `uploads` (a `{mtype →
   session-id}` map) into a temp file, validates the media type and
   size of every entry, and returns a `{mtype → path}` data map."
-  [cfg {:keys [uploads] :as params}]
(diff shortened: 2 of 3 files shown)


madalenapmelo-kp:

luisddm:
Fixed in #11012

Issue #11084, opened 2026-08-05
LDAP authentication stores client-supplied email instead of directory email and does not escape filter special characters


niwinz:
### Description

LDAP authentication stores the client-supplied email instead of the directory-verified email, and does not escape LDAP filter special characters (RFC 4515). An attacker with valid LDAP credentials can authenticate with their own password but obtain a Penpot session with a bogus email (e.g. `*@planetexpress.com`), or create duplicate profiles for the same LDAP user via alternate email addresses.

### Steps to reproduce

1. Configure Penpot with LDAP authentication (e.g. `rroemhild/docker-test-openldap`)
2. Login with `*@planetexpress.com` and a valid LDAP user's password
3. The profile email is stored as `*@planetexpress.com` (the literal wildcard), not the directory email (e.g. `amy@planetexpress.com`)
4. Login with an alternate LDAP email (e.g. `hubert@planetexpress.com` for the professor user) — creates a separate profile instead of reusing the existing one

### Expected behavior

1. LDAP filter special characters (`*`, `(`, `)`, `\`, NUL) must be escaped per RFC 4515 before building the search filter
2. The profile email must come from the LDAP directory attribute (`attrs-email`), not from the client-supplied value

### Affected versions

All versions with LDAP authentication enabled (`:login-with-ldap` flag)

PR #11192, opened 2026-08-10
Fix LDAP filter injection and use directory email (#11084)

0xTHAC0:
## Problem

Two bugs in the LDAP authentication backend (`backend/src/app/auth/ldap.clj`):

**1. LDAP filter injection** — the user-supplied email address was interpolated directly into the LDAP search filter template via string substitution without escaping RFC 4515 special characters.  A crafted login email such as `*@example.com` or `admin)(uid=*)` could match unintended LDAP entries and allow authentication bypass or information disclosure.

**2. Client-supplied email stored instead of directory email** — `retrieve-user` returned `:email email` (the value typed by the user at the login form) instead of reading the email attribute from the LDAP search result.  Users who authenticated with an alias, a differently-cased variant, or an injected value would have that string stored as their Penpot profile email rather than their canonical directory address.

## Fix

- Add `escape-ldap-filter-value` which encodes the five RFC 4515 reserved characters (`\` → `\5c`, NUL → `\00`, `(` → `\28`, `)` → `\29`, `*` → `\2a`) before the value is substituted into the filter template.
- Apply the escape inside `search-user` before calling `replace-several`.
- In `retrieve-user`, read `:email` from the LDAP entry using the configured `attrs-email` key (`(get user (-> cfg :attrs-email keyword))`); fall back to the supplied email only when the directory attribute is absent.

## Testing

- Logging in with `*@example.com` should now fail with "not-found" rather than matching arbitrary entries.
- After a successful LDAP login the stored profile email should match the directory attribute value, not the literal string the user typed.


madalenapmelo-kp:

PR #11194, opened 2026-08-10
Use circular indicator for mutually exclusive export options (#8083)

### LINKED ISSUE #8083: bug: UI "Download as Penpot file" export incorrectly uses checkboxes for select-one options, should use radiobuttons.
### Steps To Reproduce

Have a file using shared libraries. Then click .../File/Download/Download Penpot file (.penpot), get a popup with the title "EXPORT FILES", saying "One or more files that you want to export are using shared libraries..."

### Actual behavior

Image

There are three options that are mutually exclusive:
* Export shared libraries
* Include shared library assets in file libraries
* Treat shared library assets as basic objects

The dialog uses checkboxes. Checkboxes are for options that are independend of each other. 

### Expected behavior

Mutually exclusive options should be seletable via dropdowns or the round "radio buttons". In this case, radio buttons would be the right solution (according to interface standards, that are there since at least the late 80s and shared across operating systems)

0xTHAC0:
## Problem

The \"Download as Penpot file\" export dialog presents three mutually exclusive library-handling options:

- Export shared libraries
- Include shared library assets in file libraries
- Treat shared library assets as basic objects

These are mutually exclusive (the UI already uses `<input type="radio">` under the hood), but the custom visual indicator inherited its style from the `%input-checkbox` SCSS placeholder, which has a `border-radius: 4px` — making the indicator appear square/rectangular, indistinguishable from a checkbox.

Standard UX convention (since the 1980s) is that checkboxes are square (for independent toggles) and radio buttons are circular (for mutually exclusive single-select groups).

## Fix

Override `border-radius: 50%` on the indicator `<span>` within `.export-option` so it renders as a circle, visually communicating the single-select/radio semantics.

The underlying `<input type="radio">` and all behaviour are unchanged; only the appearance of the custom indicator is corrected.

Fixes #8083

### DIFF
--- a/frontend/src/app/main/ui/exports/files.scss
+++ b/frontend/src/app/main/ui/exports/files.scss
@@ -223,6 +223,12 @@
   label {
     align-items: flex-start;
 
+    // Override square checkbox shape: radio options are mutually exclusive
+    // so their indicator must be circular, not square.
+    span {
+      border-radius: deprecated.$br-circle;
+    }
+
     .modal-subtitle {
       @include deprecated.body-large-typography;
 


madalenapmelo-kp:

luisddm:
@0xTHAC0 You're right that we shouldn't be using radio buttons that look like checkboxes, and this needs to be fixed. But the code in this PR is not enough, since it does not follow the platform UI conventions -- a selected radio button contains a bullet in its center, not a tick. We'll try to find a more consistent solution. However, thanks for your help!

PR #11195, opened 2026-08-10
Don't show duplicate cursor when selection cannot be alt-duplicated (#11165)

### LINKED ISSUE #11165: Alt + drag to duplicate doesn't work on library items
### Steps To Reproduce

1. Install library (https://penpot.app/penpothub/libraries-templates/minimalistic-wireframing-kit)
2. Place library component on canvas
3. Press alt and drag component to duplicate

### Expected behavior

I expect this to work like it does for default components

### Actual behavior

Shows duplication cursor when alt+dragging but component doesn't get duplicated, just moved

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

0xTHAC0:
## Problem

When pressing Alt and dragging a shape that is **inside a component copy** (but is not the copy root), Penpot shows the `:duplicate` cursor — indicating that alt-drag will clone the shape. However, `duplicate-shapes` filters those shapes out via `ctk/allow-duplicate?` (inner shapes of component copies are intentionally excluded to protect component structure), so the operation silently falls back to a plain move. The cursor was lying.

## Root cause

`setup-cursor` showed the duplicate cursor whenever:
- Alt is held
- Not path-editing
- Not read-only workspace

It had no knowledge of whether the current selection could actually be duplicated.

## Fix

- In `viewport.cljs`: compute `can-alt-duplicate?` using `ctk/allow-duplicate?` against the current selection and base-objects. It is truthy when at least one selected shape is eligible for duplication.
- Pass `can-alt-duplicate?` to `setup-cursor` as a new parameter.
- In `hooks.cljs`: gate the `:duplicate` cursor branch on `can-alt-duplicate?`. When the entire selection is non-duplicatable (e.g. selecting an inner shape of a library component copy), the cursor falls through to `:pointer-inner`, honestly indicating that only a move will happen.

No behaviour change to duplication itself — only the cursor state is corrected.

Fixes #11165

### DIFF
--- a/frontend/src/app/main/ui/workspace/viewport.cljs
+++ b/frontend/src/app/main/ui/workspace/viewport.cljs
@@ -130,6 +130,12 @@
                                (into [] (keep (d/getf objects-modified)))
                                (not-empty))
 
+        ;; True when at least one selected shape can be alt-duplicated.
+        ;; Inner shapes of component copies are excluded by ctk/allow-duplicate?
+        can-alt-duplicate? (mf/with-memo [selected base-objects]
+                             (some #(ctk/allow-duplicate? base-objects %)
+                                   (map (d/getf base-objects) selected)))
+
         ;; STATE
         alt?               (mf/use-state false)
         shift?             (mf/use-state false)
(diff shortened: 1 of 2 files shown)


madalenapmelo-kp:

niwinz:
Thanks :heart:

PR #11214, opened 2026-08-11
:bug: Fix nitrate sso failure message

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot-nitrate/issue/26682
https://tree.taiga.io/project/penpot-nitrate/issue/26740

#### What

- Bad organization SSO issuer URLs no longer surface as an unhandled/internal error
- Users see the SSO error UI instead when discovery/JWKS/SSRF fails

##### Why

- Network/SSL/SSRF failures during OIDC discovery were not mapped to a controlled error
- `prepare-oidc-provider` also raised with a broken `:type`/`:code`, so the HTTP layer treated it as unhandled

#### How

- Wrap discovery/JWKS failures as `:validation` / `:invalid-sso-config`
- Use strict JWKS checks for organization SSO only
- Map `:invalid-sso-config` to the existing SSO error page on the frontend

### Checklist

### DIFF
--- a/backend/src/app/auth/oidc.clj
+++ b/backend/src/app/auth/oidc.clj
@@ -42,31 +42,47 @@
 ;; OIDC PROVIDER (GENERIC)
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(defn- raise-invalid-sso-config
+  "Raise a controlled validation error for OIDC provider configuration failures."
+  [& {:keys [hint cause] :as params}]
+  (throw (ex-info (or hint "invalid-sso-config")
+                  (-> params
+                      (dissoc :cause)
+                      (assoc :type :validation
+                             :code :invalid-sso-config))
+                  cause)))
+
 (defn- discover-oidc-config
   [cfg {:keys [base-uri skip-ssrf-check?] :as provider}]
-  (let [uri (u/join base-uri ".well-known/openid-configuration")
-        rsp (http/req cfg {:method :get :uri (dm/str uri)} {:skip-ssrf-check? skip-ssrf-check?})]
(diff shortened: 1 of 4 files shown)


niwinz:
### 1. [Low] Frontend handler does not defensively handle missing `organization-id`

`frontend/src/app/main/errors.cljs:392-393`

```clojure
(= code :invalid-sso-config)
(st/async-emit! (rt/assign-exception (assoc error :type :sso-error)))
```

The only code path that propagates `:invalid-sso-config` to this handler is `build-organization-sso-auth-redirect-uri`, which always enriches the error with `:organization-id` (line 934). So this is not a bug today. However, the handler doesn't check for `:organization-id` — if a future code path sends `:invalid-sso-config` without it, `sso-error-section*` would show "Unknown Organization". Consider adding a defensive fallback or a test verifying it.

### ~~2. [Medium] Commit message typo~~

~~"nitrste" → "nitrate"~~

### 3. [Medium] `populate-jwks` strict guard is too broad

`backend/src/app/auth/oidc.clj:152-159`

```clojure
(if (:strict-jwks? provider)
  (if (ex/error? cause)
    (throw cause)
    (raise-invalid-sso-config ...))
```

This re-throws *any* `ex/error?`, not just `:invalid-sso-config`. A different error type would escape unwrapped. The other catch blocks (`prepare-oidc-provider`, `prepare-gitlab-provider`) correctly check the `:code`. Align this one:

```clojure
(if (and (ex/error? cause)
         (= :invalid-sso-config (:code (ex-data cause))))
  (throw cause)
  (raise-invalid-sso-config ...))
```

### 4. [Low] Missing test for non-200 discovery response

New tests cover network failures (SSL, SSRF, JWKS) but not the case where the OIDC discovery endpoint returns a non-200 status (404, 500). The `discover-oidc-config` branch at line 75 is untested. Consider adding a test with `{:status 404}`.

niwinz (approved):

PR #11229, opened 2026-08-13
:zap: Skip save_layer for plain image fills

superalex:

## What

- Avoids an offscreen `save_layer` buffer for every `Fill::Image` during tile walks when no shape-level image filter is active.
- Skips redundant container clipping for axis-aligned rects and frames without corner radii.

## Why

Photo-heavy boards were paying one offscreen layer per image fill on every tile pass, even when blur was not involved and the shape did not need a clip before `draw_image_rect`.

## How

- **render-wasm/fills.rs:** only call `save_layer` when a shape image filter is present; use direct draw for plain axis-aligned rects/frames, otherwise `save` + clip without an extra layer.
- **memory:** document the raster image fill fast path in `ffi-rendering-subtleties`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.serena/memories/render-wasm/ffi-rendering-subtleties.md
+++ b/.serena/memories/render-wasm/ffi-rendering-subtleties.md
@@ -17,6 +17,8 @@
 
 ## Tile/render behavior
 
+- Raster `Fill::Image`: skip `save_layer` unless the shape has an image filter; plain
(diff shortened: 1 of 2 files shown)


superalex:
Reopened as a new PR under the renamed branch `superalex-skip-save-layer-plain-image-fills` (this PR closed when the old head branch was deleted).

PR #11234, opened 2026-08-13
:bug: Preserve public URI subpath in asset download URLs

UniversalWill:

## What

- Preserve the configured public URI subpath in temporary export download URLs.
- Apply the same behavior to binary `.penpot` file downloads.
- Add regression coverage for both URL generation paths.

## Why

Asset download paths were joined using an absolute `/assets/by-id/` path. This replaced the path component of `PENPOT_PUBLIC_URI`, causing installations hosted under a subpath such as `https://example.com/penpot` to return `https://example.com/assets/by-id/{id}` instead of `https://example.com/penpot/assets/by-id/{id}`.

## How

- Ensure the configured public URI ends with a slash.
- Join `assets/by-id/` as a relative path so the configured subpath is retained.
- Verify both temporary exporter assets and binary file exports with subpath-based public URIs.

### DIFF
--- a/backend/src/app/rpc/commands/binfile.clj
+++ b/backend/src/app/rpc/commands/binfile.clj
@@ -62,7 +62,8 @@
                                      :bucket "tempfile"})]
 
         (-> (cf/get :public-uri)
-            (u/join "/assets/by-id/")
+            (u/ensure-path-slash)
+            (u/join "assets/by-id/")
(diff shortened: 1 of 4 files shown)


madalenapmelo-kp:

niwinz:
Thanks :heart:

PR #11246, opened 2026-08-14
:bug: Fix export_shape SVG writing null bytes

### LINKED ISSUE #10878: MCP: export_shape with format "svg" writes 2 null bytes instead of the SVG
Self-hosted 2.17.0, MCP server in single-user mode (no `--multi-user`, so filesystem access is enabled).

`export_shape` with `format: "svg"` and a `filePath` reports success and writes a **2-byte file** containing `00 00`. PNG export of the same shape works correctly.

### Cause

The plugin returns a `Uint8Array` for both formats (`PenpotUtils.exportImage`, which calls `shape.export({ type: "svg" })`), and `ExecuteCodeTaskHandler.encodeBytesAsBase64Envelope` wraps every `Uint8Array` result as `{ __type: "base64", data: <base64> }`.

On the server side, `ImageContent.byteData` recognises that envelope — but the SVG path calls `TextContent.textData`, which does not. It falls through to `String.fromCharCode(...Object.values(envelope))`, and the two values are the strings `"base64"` and the base64 payload, so both become `NaN`, producing two null characters.

- `ExportShapeTool.ts:174` — SVG file path calls `TextContent.textData(imageData)`
- `ToolResponse.ts:15-22` — `TextContent.textData` handles only a string or a numeric-keyed object
- `ToolResponse.ts:49-60` — `ImageContent.byteData` does handle the envelope

The envelope was introduced in #9431 and `byteData` was updated for it; `textData` was not.

### Workaround

The data itself is fine — exporting the same shape through `execute_code` and converting the bytes to a string inside the plugin returns valid SVG:

```js
const bytes = await shape.export({ type: "svg" });
let out = "";
for (let i = 0; i < bytes.length; i += 0x8000) {
  out += String.fromCharCode.apply(null, bytes.subarray(i, i + 0x8000));
}
return out;
```

(`TextDecoder` is not available in the plugin sandbox, hence the manual conversion.)

hermz580:
The plugin returns a Uint8Array wrapped in a base64 envelope ({ __type: "base64", data: <base64> }) for both PNG and SVG exports. The PNG path decodes it via ImageContent.byteData(), but the SVG path passed the raw envelope to TextContent.textData(), which doesn't recognize the envelope â€” String.fromCharCode(...Object.values(envelope)) turns the two strings into NaN, producing a 2-byte file of null bytes.

Decode the envelope with ImageContent.byteData() and convert the bytes to UTF-8 text for both the direct-return and file-write SVG paths, so SVG export behaves the same as PNG export.

Closes #10878

### Checklist


madalenapmelo-kp:

opcode81:
Thanks for your PR, but this does not fix the issue at the right level; superseded by #11422.

PR #11258, opened 2026-08-17
:bug: Fix not being able to select right or center-aligned text in v3

### LINKED ISSUE #11244: Fix not being able to select text in auto-width shapes aligned to the right or centered
STR:

1. Ensure text editor v3 is enabled
2. Select or create a new text shape with auto-width
3. Set text alignment to right or center
4. Try to select a word with double click or by clicking & dragging

Expected results: The word is selected as usual

Actual results: Unable to select any inner text.

belen-albeza:
### Related Ticket

Fixes #11244 

### Summary

We were using `f32:MAX` to lay out right-aligned or centered text, which messed up hit test calculations.

Screenshot 2026-08-17 at 5 09 10 PM

### Steps to reproduce 

See #11244 

### Checklist

- [x] ~~Refactor any modified SCSS files following the refactor guide.~~

### DIFF
--- a/render-wasm/src/shapes/text.rs
+++ b/render-wasm/src/shapes/text.rs
@@ -707,16 +707,35 @@ impl TextContent {
         &self,
         use_shadow: Option<bool>,
     ) -> Vec<ParagraphBuilderGroup> {
+        self.paragraph_builders(use_shadow, false, None)
+    }
+
+    /// Creates paragraph builders with always-opaque paint (BLACK @ alpha 255).
+    /// Used as a clip mask for inner stroke rendering.
+    pub fn paragraph_builder_group_opaque(&self) -> Vec<ParagraphBuilderGroup> {
+        self.paragraph_builders(None, true, None)
+    }
+
+    fn paragraph_builders(
+        &self,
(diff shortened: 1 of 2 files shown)


render-wasm/src/shapes/text.rs
```diff
@@ -894,11 +888,15 @@ impl TextContent {
     }
 
     pub fn update_layout(&mut self, selrect: Rect) -> TextContentSize {
```
elenatorro, line 894: There's a minor thing with this change: fonts resolve async, and when they resolve later than once rendered and we need to update them, the force layout won't work for those shapes with auto width directly.

So, since we're checking this here: `self.layout_version == self.content_version`, the way to actually fix this could be to add this setting on `force_next_layout_update`: `self.content_version = self.content_version.wrapping_add(1);`

It's possible to reproduce the issue without adding this line by loading this penpot file: [get-file-text-emoji-board-auto-width.zip](https://github.com/user-attachments/files/31179424/get-file-text-emoji-board-auto-width.zip). Before a relayout, it's cut:

Image

For context, this fix was introduced here: https://github.com/penpot/penpot/pull/10504

PR #11259, opened 2026-08-17
:zap: Improve shape attrs parsing performance

### LINKED ISSUE #11240: Performance issue with WebGL render
### Steps To Reproduce

* Insert this SVG in your document: https://github.com/user-attachments/assets/10e0a59b-b2da-4c7a-90fc-027a7edc1ea1
* Try re-scaling it by dragging its corner gizmos.
* Now try the same with WebGL enabled.

While it is not fast with WebGL disabled, when WebGL is enabled the performance is prohibitive.

### Expected behavior

The same if not a better performance than the non-WebGL renderer.

### Actual behavior

Prohibitively slow performance.

### Desktop

OS: Mac Book Pro, Apple M3 Pro, MacOS: Version 26.5.2 (25F84)
Browser: Google Chrome Version 151.0.7922.109 (Official Build) (arm64)

### Environment

Both design.penpot.app and own instance.

elenatorro:
### Related Ticket

https://github.com/penpot/penpot/issues/11240

### Summary

Testing a group shape of 4k shapes resulted in a very slow performance on load and interactions. There were several things going on:
* The design tab reads `:fill`, `:stroke` and `:text` from every descendant of the group
* Memoize token-attrs (get-attrs is called once per shape per atribute)
* Get attrs also recalculated things that depended only on the shape type. This improves merging not necessary token attributes when not present.

### Steps to reproduce 

1. Create a new document
2. Load the svg file from the issue
3. Check and compare the load time as well as pan and zoom

Note: **resize** not addressed in this PR, there will be a follow-up PR

### DIFF
--- a/common/src/app/common/types/token.cljc
+++ b/common/src/app/common/types/token.cljc
@@ -423,11 +423,8 @@
     :stroke-width :strokes
     token-attr))
 
-(defn shape-attr->token-attrs
-  "Returns the token-attr affected when a given attribute in a shape is changed.
-   The sub-attr is for attributes that may have multiple values, like strokes
-   (may be width or color) and layout padding & margin (may have 4 edges)."
-  ([shape-attr] (shape-attr->token-attrs shape-attr nil))
+(defn- shape-attr->token-attrs*
+  ([shape-attr] (shape-attr->token-attrs* shape-attr nil))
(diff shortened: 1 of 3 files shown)


superalex (approved):
<3

PR #11265, opened 2026-08-18
:sparkles: Add sso events

cocotime:
### Related Ticket

https://tree.taiga.io/project/penpot-nitrate/task/26656

### Checklist

### DIFF
--- a/backend/src/app/auth/oidc.clj
+++ b/backend/src/app/auth/oidc.clj
@@ -771,6 +771,82 @@
 ;; ORG SSO HELPERS
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(defn- organization-sso-oauth-failure-reason
+  [error]
+  (case (d/name error)
+    "access_denied" "access-denied"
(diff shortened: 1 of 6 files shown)


mavalroot (approved):

backend/src/app/auth/oidc.clj
```diff
@@ -771,6 +771,82 @@
...
+(defn- submit-organization-sso-oauth-failed-event
+  [cfg request state-token error]
+  (try
+    (let [state (tokens/verify cfg {:token state-token :iss "oidc"})]
+      (when (:dest-url state)
+        (submit-organization-sso-auth-event
+         cfg request (some-> (session/get-session request) :profile-id)
+         (:organization-id state) "organization-sso-auth-failed"
+         :failure-reason (organization-sso-oauth-failure-reason error))))
+    (catch Throwable _ nil)))
```
niwinz, line 848: This catch shallows all exceptions, is there a reason for doing that? If yes, prefer Exception instead of Throwable

cocotime: The catch is needed to ensure that the user receives the same error as before. We only send the event if we can read the state, and we don't want that to interrupt the flow.

I'll change Throwable to Exception as you suggested.

PR #11282, opened 2026-08-19
:bug: Gate MCP REPL server behind isDevEnv check

### LINKED ISSUE #11283: MCP REPL server exposes unauthenticated code execution endpoint without devenv check
### Description

The MCP server's REPL HTTP endpoint (`POST /execute`) starts unconditionally on every server instance, regardless of the `PENPOT_MCP_DEVENV` flag. The endpoint has no authentication middleware — any local process on the host machine can send arbitrary JavaScript code to the connected Penpot browser plugin for execution.

This enables cookie/session theft, DOM manipulation, and data exfiltration from active Penpot sessions by any local user or process.

### Steps to reproduce

1. Start the MCP server without setting `PENPOT_MCP_DEVENV=true`
2. Send `POST http://localhost:4403/execute` with `{"code": "document.cookie"}`
3. The server forwards the code to the connected plugin without any authentication check

### Expected behavior

The REPL server should only start when `PENPOT_MCP_DEVENV=true` is set, consistent with how `CljsReplTool` and other developer tools are gated.

### Affected versions

2.17.0

niwinz:

## What

The MCP `ReplServer` exposed an unauthenticated `POST /execute` endpoint that started unconditionally on every server instance. Any local process could send arbitrary JavaScript to the connected Penpot plugin for execution.

## How

- **Gate ReplServer behind `isDevEnv()`** — creation, startup, and shutdown are now conditional on `PENPOT_MCP_DEVENV=true`, consistent with how `CljsReplTool` and other dev tools are already protected
- **Consolidate dev-env check** — `isDevEnv()` delegates to a single static `isDevEnvEnabled()` method, eliminating duplicate logic
- **Log REPL-disabled info** — server logs when the REPL is not started, guiding users who expect it to be available
- **Add `PluginBridge.close()`** — proper WebSocket server cleanup on shutdown (was previously missing)
- **Regression tests** — construct `PenpotMcpServer` with and without `PENPOT_MCP_DEVENV` and verify `hasReplServer()` returns the correct value

Closes #11283

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -265,6 +265,7 @@ The Penpot MCP server can be configured using environment variables.
 | `PENPOT_MCP_SERVER_PORT`                         | Port for the HTTP/SSE server                                               | `4401`         |
 | `PENPOT_MCP_WEBSOCKET_PORT`                      | Port for the WebSocket server (plugin connection)                          | `4402`         |
 | `PENPOT_MCP_REPL_PORT`                           | Port for the REPL server (development/debugging)                           | `4403`         |
+| `PENPOT_MCP_REPL_ENABLE`                         | Explicitly enable/disable the REPL server. Set to `true` to enable. When unset, defaults to the value of `PENPOT_MCP_DEVENV`. | (unset)        |
 | `PENPOT_MCP_REMOTE_MODE`                         | Enable remote mode (disables file system access). Set to `true` to enable. | `false`        |
 | `PENPOT_MCP_DEVENV`                              | Enable Penpot development environment tools. Set to `true` to enable.      | `false`        |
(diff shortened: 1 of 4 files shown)


opcode81 (commented):
The REPL server can be useful outside of devenv settings (for local MCP server usage/experimentation), so I would recommend to add a new env variable `PENPOT_MCP_REPL_ENABLE`  (alongside the existing `PENPOT_MCP_REPL_PORT`, documented in `mcp/README.md`), such that ...
* if the variable is set, its value determines whether the server is enabled (`true` enables it, other values disable it)
* if the variable is not set, we default to `isDevEnv` as implemented in this PR.

niwinz:
@opcode81 addressed

opcode81 (approved):

PR #11284, opened 2026-08-19
:bug: Enforce share-link page scope in get-page and get-file-fragment RPC commands

### LINKED ISSUE #11281: Share-link holder reads pages outside authorized scope via get-page RPC command
### Description

A Penpot share-link is scoped to specific pages via the `:pages` set. The `get-view-only-bundle` RPC command correctly filters pages based on this scope, but the `get-page` RPC command does not enforce the same restriction. A share-link holder can read the complete shape/design data of any page in the file by calling `get-page` with an out-of-scope page-id and the share-id, bypassing the intended page-level access control.

The `get-file-fragment` RPC command has the same issue - it accepts a share-id but returns fragments without validating they belong to authorized pages.

### Steps to reproduce

1. Create a file with two pages: Page A (public) and Page B (secret)
2. Create a share-link scoped to Page A only
3. As a share-link holder, call `get-page` with the share-id and Page B's id
4. Observe that Page B's complete content is returned

### Expected behavior

The `get-page` command should validate that the requested page-id is within the share-link's authorized `:pages` set and return a `:not-found` error if it's not, matching the behavior of `get-view-only-bundle`.

The `get-file-fragment` command should deny share-link access to fragments entirely, as fragments don't have a direct page-id mapping and the simpler/safer approach is to restrict access.

### Affected versions

<= 2.15.4 (current codebase as of 2026-08-19)

### Security advisory

GHSA-x9xg-qphx-8grr

### Credits

Vulnerability reported by 5ud0 / Tarmo Technologies

niwinz:

## What

- Share-link holders could read pages outside the authorized scope by calling `get-page` with an out-of-scope page-id
- The `get-file-fragment` RPC command had the same issue, returning fragments without validating page scope
- This bypassed the intended page-level access control that `get-view-only-bundle` correctly enforces

## Why

The `get-page` command only checked that SOME permission existed but didn't validate that the requested page was within the share-link's `:pages` set. The `get-view-only-bundle` command already had the correct implementation using `remove-not-allowed-pages`, but this logic wasn't applied to `get-page` or `get-file-fragment`.

## How

- Added page scope validation to `get-page` that rejects requests for pages not in the share-link's `:pages` set with a `:not-found` error
- Denied share-link access to `get-file-fragment` entirely, as fragments lack direct page-id mapping (simpler/safer approach per security advisory recommendation)
- Added comprehensive tests covering both the security fix and backward compatibility with normal team member access

Closes #11281

### DIFF
--- a/backend/src/app/rpc/commands/audit.clj
+++ b/backend/src/app/rpc/commands/audit.clj
@@ -183,6 +183,7 @@
 
 (sv/defmethod ::get-enabled-flags
   {::audit/skip true
+   ::rpc/auth false
    ::doc/skip true
    ::doc/added "1.20"}
   [_cfg _params]
--- a/backend/src/app/rpc/commands/files.clj
+++ b/backend/src/app/rpc/commands/files.clj
@@ -245,6 +245,10 @@
   [cfg {:keys [::rpc/profile-id file-id fragment-id share-id]}]
   (db/run! cfg (fn [cfg]
                  (let [perms (perms/get-file-read-permissions cfg profile-id file-id share-id)]
+                   (when (= :share-link (:type perms))
+                     (ex/raise :type :not-found
+                               :code :object-not-found
+                               :hint "object not found"))
                    (check-read-permissions! perms)
                    (-> (get-file-fragment cfg file-id fragment-id)
                        (rph/with-http-cache long-cache-duration))))))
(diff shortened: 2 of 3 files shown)


niwinz:
Found some regressions related to permissions

elenatorro (approved):

PR #11296, opened 2026-08-20
:tada: Implement export jobs to process export requests

### LINKED ISSUE #11133: Implement export jobs to process export requests

/export` API is sync, and:
- `wait=true` holds the HTTP connection for the whole render
- `wait=false` returns fast but still starts the work immediately

There's no way to check process, cancel them or limit them

Simple job-based API:

 - `POST /export`  -> creates job
 - `GET /export/:job-id` -> job status + progress
- `DELETE /export/:job-id` -> cancels job

Also, since wasm render blocks the event loop, manage a pool of forked workers to run N exports in parallel. Note: Only the headless wasm path goes to workers (everything else stays on the existing path and the browser pool). But both kinds of export get a job id, only the executor differs.

_No response_

### Additional context

_No response_

elenatorro:
### Related Ticket

https://github.com/penpot/penpot/issues/11133

### Summary

This PR introduces a new mechanism for render exports through **export jobs**. These jobs are created over HTTP, admitted by a scheduler with bounded concurrency, persisted in Redis, can be queried to check its status, and can be cancelled.
Screenshot 2026-08-27 at 15-20-48 brand - Penpot

⚠️ The current `POST /api/export` API is untouched and still runs unscheduled exports that use the browser pool. The export jobs are only triggered from the frontend when the `wasm-export/v1` and `render-wasm/v1` feature flags are enabled on the team **and** the exporter has `enable-wasm-export` in `PENPOT_FLAGS`. Without all three, exports fall back to the current endpoint.

These are the changes introduced:

#### Exporter Changes

1. New Job model, stored on redis

* Job status: queued, running, ended, error, cancelled
* State transitions send the `:export-update` message so websocket clients keep working with the same logic
* Each job is saved on Redis and has a defined TTL
* The job owns the temp files and deletes them once it finishes (if there's any temp file left, it's deleted on boot)

2. Scheduler

The Scheduler handles jobs execution over a queue. It handles errors and job states. Wasm jobs lease a render worker for their whole run, so they are only admitted when a worker is free: `PENPOT_WASM_WORKER_POOL_MAX` is the real limit for them.

Cancel job example:

[screen-recorder-thu-aug-27-2026-15-20-56.webm](https://github.com/user-attachments/assets/ab718a11-c440-430c-9f67-0233d526aba9)

Job waiting for other two jobs, that starts as soon as the queue is free:

[screen-recorder-thu-aug-27-2026-15-22-44.webm](https://github.com/user-attachments/assets/d4fa31df-176b-437c-abe8-ec8f3d9200b4)

3. Workers

Export renders are executed in a worker thread. This is because Skia calls are synchronous and block the HTTP server, so we use a pool of workers. A job keeps one worker for all its renders.

4. Rest API

| Method | Path | Description |
|---|---|---|
| `POST` | `/` | Current endpoint (no queue, no scheduler) |
| `POST` | `/jobs` | Creates an export job |
| `GET` | `/jobs/{id}` | Gets job record |
| `DELETE` | `/jobs/{id}` | Requests cancellation |

⚠️ NOTE: API works through auth session (cookie) only, as the current endpoint. Ownership now comes from the session instead of `:profile-id` in the request body: the job API can read and cancel work, so it rejects anonymous callers and checks the owner on every job. The current endpoint keeps the body value as fallback, unchanged.

⚠️ NOTE: This only applies for multiple shape files. Single shapes are being rendered directly in the client, no API call.

#### Frontend Changes

When the feature flag is enabled, each export calls the new jobs API to create export jobs (and can cancel them from the UI too)

#### Backend Changes

There's a new section on `/dbg` panel for purely debug/info options that reads from Redis

image

#### Deployment

Redis was already being used for websocket notifications in the exporter, so there's no change here. The exporter container now also receives `PENPOT_FLAGS`, which is what enables the headless render.

Review worker pool. Each render worker boots its own render-wasm module, default is 2 workers with a 128MB image cache budget each on top of that. This can be changed through envvars.

⚠️ Single exporter instance only: on boot the exporter cancels every non-terminal job record it finds, since a queue lives in the memory of the process that owned it. With several exporters behind a load balancer this would cancel a sibling's running jobs.

Optional variables:

| Variable | Default |
|---|---|
| `PENPOT_EXPORT_JOB_TTL` | `3600` |
| `PENPOT_EXPORT_MAX_CONCURRENT_JOBS` | `4` |
| `PENPOT_EXPORT_MAX_JOBS_PER_PROFILE` | `2` |
| `PENPOT_EXPORT_QUEUE_MAX` | `64` |
| `PENPOT_WASM_WORKER_POOL_MAX` | `2` |
| `PENPOT_WASM_WORKER_POOL_MIN` | `1` |
| `PENPOT_WASM_RENDER_IDLE_TIMEOUT` | `300` |
| `PENPOT_WASM_IMAGE_CACHE_MB` | `128` |

### Steps to reproduce

Activate the `wasm-export/v1` and `render-wasm/v1` flags on team and also include `enable-wasm-export` on env `PENPOT_FLAGS`.


backend/src/app/http/debug.clj
```diff
@@ -51,11 +48,53 @@
...
+        (recur cursor found)))))
+
+(defn- get-export-jobs
+  [cfg job-id]
+  (let [filtered? (not (str/empty-or-nil? job-id))
+        job-uuid  (when filtered? (parse-uuid job-id))]
+    (if (and filtered? (nil? job-uuid))
+      []
+      (try
+        (let [pattern (str (cf/get :tenant) ".export.job." (or job-uuid "*"))]
```
niwinz: `penpot.exporter.<tenant>.job....`

exporter/README.md
```diff
@@ -0,0 +1,98 @@
...
+|---------------------------------------|---------|--------------------------------------|
+| `PENPOT_REDIS_URI`                    | `redis://redis/0` | Job store and cancel topic |
+| `PENPOT_TENANT`                       | `default` | Key and topic prefix               |
+| `PENPOT_EXPORT_JOB_TTL`               | `3600`  | Lifetime of a job record, in seconds  |
+| `PENPOT_EXPORT_MAX_CONCURRENT_JOBS`   | `4`     | Admission limit                       |
+| `PENPOT_EXPORT_MAX_JOBS_PER_PROFILE`  | `2`     | Per-profile admission limit           |
+| `PENPOT_EXPORT_QUEUE_MAX`             | `64`    | Queue cap; over it, `429 :queue-full` |
+| `PENPOT_WASM_WORKER_POOL_MAX`         | `2`     | Headless render worker threads; min 1 |
+| `PENPOT_WASM_WORKER_POOL_MIN`         | `1`     | Workers kept warm; clamped to the max |
+| `PENPOT_WASM_RENDER_IDLE_TIMEOUT`     | `300`   | Silence before a worker is terminated, in seconds |
```
niwinz: PENPOT_WASM_WORKER_IDLE_TIMEOUT

exporter/README.md
```diff
@@ -0,0 +1,98 @@
...
+## Configuration
+
+| Variable                              | Default | Description                          |
+|---------------------------------------|---------|--------------------------------------|
+| `PENPOT_REDIS_URI`                    | `redis://redis/0` | Job store and cancel topic |
+| `PENPOT_TENANT`                       | `default` | Key and topic prefix               |
+| `PENPOT_EXPORT_JOB_TTL`               | `3600`  | Lifetime of a job record, in seconds  |
+| `PENPOT_EXPORT_MAX_CONCURRENT_JOBS`   | `4`     | Admission limit                       |
+| `PENPOT_EXPORT_MAX_JOBS_PER_PROFILE`  | `2`     | Per-profile admission limit           |
+| `PENPOT_EXPORT_QUEUE_MAX`             | `64`    | Queue cap; over it, `429 :queue-full` |
```
niwinz: `PENPOT_EXPORTER_` para todas las keys de `PENPOT_EXPORT_`

exporter/README.md
```diff
@@ -0,0 +1,98 @@
...
+| `PENPOT_WASM_WORKER_POOL_MAX`         | `2`     | Headless render worker threads; min 1 |
+| `PENPOT_WASM_WORKER_POOL_MIN`         | `1`     | Workers kept warm; clamped to the max |
+| `PENPOT_WASM_RENDER_IDLE_TIMEOUT`     | `300`   | Silence before a worker is terminated, in seconds |
+| `PENPOT_WASM_IMAGE_CACHE_MB`          | `128`   | Per-worker image cache budget          |
```
niwinz: `PENPOT_WAMS_WORKER_IMAGE_CACHE_SIZE` (en bytes, o con soporte para parsear unidades) empezaria con bytes (es como lo usamos en otras keys asi de esta manera es uniforme

exporter/src/app/redis.cljs
```diff
@@ -8,47 +8,162 @@
...
   (swap! client (fn [client]
                   (when client (.quit ^js client))
                   nil)))
 
 (def ^:private tenant (cf/get :tenant))
 
+(defn ->key
+  "Namespaces `parts` under the tenant, the same prefix `pub!` uses for topics."
+  [& parts]
+  (str tenant "." (apply str parts)))
```
niwinz: The key should be `penpot.exporter.<tentant>....`

exporter/src/app/redis.cljs
```diff
@@ -8,47 +8,162 @@
...
+            (->> (with-client-lenient (fn [^js client] (.scan client cursor "MATCH" pattern "COUNT" 200)))
+                 (p/mcat (fn [result]
+                           (if (nil? result)
+                             (p/resolved found)
+                             (let [next-cursor (aget result 0)
+                                   found       (into found (aget result 1))]
+                               (if (= "0" next-cursor)
+                                 (p/resolved found)
+                                 (step next-cursor found))))))))]
+    (step "0" [])))
```
niwinz, line 177: The problem with this approach is that it scans all result in memory instead if consuming it like a stream o sequence. With promise approach there are noway to do it in other way in any case. I suggest just adding a comment on it right now, becuse for the current aproach it may work but we need to rething how we access redis, and implement maybe or RX or Streaming Reader interfaces in future.

PR #11300, opened 2026-08-20
:bug: Freeze viewport gestures during WASM page transition

superalex:
## What

- While the WASM page-transition blur is up, pan/zoom no longer drives `render_from_cache` against an empty tile atlas (blank workspace).
- Any local viewport change that still lands in the store is flushed once when `tiles-complete` ends the overlay.

## Why

Changing page and panning before the first full tile frame finished enabled fast-mode cache compositing with nothing useful in the atlas, so the canvas went blank until the gesture ended.

## How

- No-op `set-view-box` / `view-interaction-start!` while `page-transition?` is true; mark a dirty flag instead.
- Disable pointer events on the viewport SVG for the same window.
- On `end-page-transition!`, sync the deferred workspace zoom/vbox into WASM if dirty.


superalex:
Reopening from renamed branch `superalex-block-viewport-during-page-transition`.

PR #11313, opened 2026-08-21
:zap: Present viewport before interest and clamp paint to atlas

superalex:
## What

- Make zoom settle show a sharp viewport sooner, without waiting for interest-ring tile work to finish.
- Avoid painting tiles larger than the atlas slot (e.g. 1024 into 512 slots at DPR 2) only to downscale them away.

## Why

- Visible tiles could be ready while the renderer kept working on the interest ring, so the viewport stayed soft longer than necessary.
- Oversized paint at high DPR wasted GPU work that was discarded when copying into atlas slots.

## How

- **ViewportReady frame:** present once visible tiles are done, keep requesting frames for deferred interest-ring fill, and skip a second present on the final Full.
- **Atlas-sized paint:** clamp tile rasterization to the atlas slot size so HiDPI does not overpaint.

### DIFF
--- a/frontend/src/app/render_wasm/api.cljs
+++ b/frontend/src/app/render_wasm/api.cljs
@@ -389,6 +389,19 @@
 (def ^:const FRAME_TYPE_NONE 0)     ;; This type should never "leak".
 (def ^:const FRAME_TYPE_PARTIAL 1)  ;; A frame needs more render calls to end.
 (def ^:const FRAME_TYPE_FULL 2)     ;; A frame was full.
+(def ^:const FRAME_TYPE_VIEWPORT_READY 3) ;; Viewport presented; interest tiles may still be pending.
+
+(defn- needs-more-render-frames?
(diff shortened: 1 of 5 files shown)


frontend/src/app/render_wasm/api.cljs
```diff
@@ -2378,9 +2410,7 @@
 
      (set! wasm/context-initialized? false)
 
-     ;; Cancel any pending animation frame to prevent race conditions.
-     (when wasm/internal-frame-id
-       (timers/cancel-af! wasm/internal-frame-id))
+     (stop-progressive-render!)
 
      ;; Reset render flags to prevent new renders from being scheduled.
      (reset! pending-render false)
```
elenatorro: is this reset neded after (stop-progressive-render!)? it's already handled there

superalex: Thank you very much!, you are right, it's not needed here

Issue #11147, opened 2026-08-08
Show a content summary on collapsed sections in the design sidebar


yuricoppe:
### Description

Collapsed sections in the design sidebar show only their title, so there is no way to tell whether a section holds content — or how much — without expanding each one in turn. This adds a compact inline summary next to the title of every collapsed section, so the state of the whole sidebar can be read at a glance.

The summary takes one of two forms depending on what the section holds:

- Color sections (Fill, Stroke, Selected colors) show up to three mini color bullets, followed by a `+N` counter when there are more colors than fit.
- Count sections (Shadow, Blur, Export, Guides) show a badge with the number of items.
- When the selection spans shapes with differing values, the summary reads `Mixed` instead.

Sections with no content show no summary, so an empty sidebar looks exactly as it does today. The summary only renders while the section is collapsed; expanding it restores the current layout unchanged.

### Use case

Designers working on a shape with several fills, strokes, shadows and exports have to expand sections one by one to find where the content is, or to confirm that a section they collapsed still holds what they left there. The summary makes that state visible without any interaction, which matters most on shapes with many properties, where the sidebar is usually kept collapsed to stay navigable.

It also helps when moving between shapes: the collapsed sidebar immediately shows which properties the newly selected shape carries, instead of requiring the same expand-and-check pass on every selection.

### Affected versions

2.17.0

yuricoppe:
1. Light theme, collapsed sections
<img width="312" height="768" alt="Image" src="https://github.com/user-attachments/assets/00fa5308-4f97-43c5-b84b-1a650e7e9507" />
<img width="314" height="247" alt="Image" src="https://github.com/user-attachments/assets/1c9b85d6-1164-4007-9313-fc29884ac607" /> 

2. Dark theme, same sections
<img width="312" height="417" alt="Image" src="https://github.com/user-attachments/assets/d755f99f-4ee0-4d96-b07d-b5c095743a84" />
<img width="320" height="92" alt="Image" src="https://github.com/user-attachments/assets/4de4f477-f170-4639-9231-5e07520c3cc5" />

3. Selected colors
<img width="318" height="46" alt="Image" src="https://github.com/user-attachments/assets/ecd77e8a-7985-468b-8922-abbb4c032eff" />
<img width="312" height="53" alt="Image" src="https://github.com/user-attachments/assets/39199385-cf14-42d6-953f-e044f3196bcd" />

4. Guides
[...]

PR #11321, opened 2026-08-21
:sparkles: Show a count summary on collapsed sidebar sections

yuricoppe:

## What

Collapsed sections in the design sidebar now show a compact summary of what they hold, so the panel can be scanned without expanding section by section.

This first PR adds the mechanism and wires the four count-based sections — Blur, Shadow, Export and Guides — which show an item count, or `Mixed` when the selection has differing values. The color sections (Fill, Stroke, Selected colors) follow in a second PR, since they carry more design surface.

## Why

A collapsed section is a bare title: there is no way to tell an empty Fill from one with four fills, or a board without guides from one with three, other than expanding it and collapsing it again. On a busy selection that is a lot of clicking to answer a question the sidebar could just show.

## How

- **Mechanism.** An optional `summary` prop on the shared `title-bar*`, rendered next to the title only while the section is collapsed. Sections opt in by passing a summary, so every section that passes nothing renders exactly as it does today.
- **Polymorphic by design.** A number renders as a counter badge, a string as a text label, anything else is rendered as-is. Keeping that decision in one place lets the color sections pass a bullet-list element through the same slot with no extra plumbing, and keeps the call sites trivial.
- **Accessibility.** The summary is `aria-hidden` with a `title` tooltip. What it summarizes is reachable by expanding the section, and appending it to the toggle button would turn the accessible name of the Fill control into "Fill +1" — noise for a screen reader.
- **No new derivation.** Each section reuses state it already computes: Blur its `mixed-state` binding, Shadow and Export their existing `:multiple` sentinel, Guides its `has-frame-grids?`, which already folds the `:multiple` case in.
- **`Mixed` label.** Reuses `settings.multiple`, the key the rest of the sidebar already uses, rather than adding a new one (see #11148).

Relates to #11147 — closed by the follow-up PR that wires the color sections.

Thanks @filipsajdak for the prototype work and the review of the approach on the issue; several details here (notably the `:multiple` sentinel notes and the `color-bullet*` `mini` reuse landing in the follow-up) come from that discussion.

## Screenshots

Guides, collapsed, with two grids:

Guides section collapsed showing a count badge of 2

Blur on a multi-selection with differing values:

Blur section collapsed showing the Mixed label

The whole sidebar collapsed — Shadow, Blur and Export carry the badges added here; the Fill and Stroke bullets in the same shot come from the follow-up PR:

Design sidebar with all sections collapsed showing count badges and color bullets

Empty sections are untouched — no badge, and the section stays non-collapsable:

Design sidebar with every section empty and no summaries

### DIFF
--- a/frontend/src/app/main/ui/components/title_bar.cljs
+++ b/frontend/src/app/main/ui/components/title_bar.cljs
@@ -12,7 +12,7 @@
    [rumext.v2 :as mf]))
 
 (mf/defc title-bar*
-  [{:keys [class collapsable collapsed title children
+  [{:keys [class collapsable collapsed title children summary
            btn-icon btn-title add-icon-gap
            title-class on-collapsed on-btn-click] :rest props}]
   (let [props (mf/spread-props props {:class (stl/css :icon-text-btn)
@@ -26,7 +26,25 @@
            [:> icon* {:icon-id icon-id
                       :size "s"
                       :class (stl/css :icon)}]
-           [:div {:class (stl/css :title)} title]])]
+           [:div {:class (stl/css :title)} title]
+           ;; Compact inline representation of the section content; only
+           ;; rendered while collapsed, where the content is not visible.
+           ;; It is a purely visual hint (the content itself is reachable
+           ;; by expanding the section), so it is hidden from assistive
+           ;; technologies to keep the button label clean.
+           (when (and ^boolean collapsed (some? summary))
+             [:div {:class (stl/css :title-summary)
+                    :aria-hidden true
+                    :title (when (string? summary) summary)}
+              (cond
+                (number? summary)
+                [:span {:class (stl/css :title-summary-counter)} summary]
+
+                (string? summary)
+                [:span {:class (stl/css :title-summary-text)} summary]
+
+                :else
+                summary)])])]
 
        [:div {:class [(stl/css-case :title-only true
                                     :title-only-icon-gap add-icon-gap)
--- a/frontend/src/app/main/ui/components/title_bar.scss
+++ b/frontend/src/app/main/ui/components/title_bar.scss
@@ -55,6 +55,44 @@
   color: var(--title-foreground-color-hover);
 }
 
+.title-summary {
+  display: flex;
+  align-items: center;
+  flex: 0 1 auto;
+  gap: deprecated.$s-4;
+  min-width: 0;
+  max-width: deprecated.$s-120;
+  margin-inline-start: deprecated.$s-8;
+  overflow: hidden;
+  color: var(--color-foreground-secondary);
+}
+
+.title-summary-text {
+  @include t.use-typography("body-small");
+
+  min-width: 0;
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+}
+
+.title-summary-counter {
+  @include t.use-typography("body-small");
+
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  min-width: deprecated.$s-16;
+  height: deprecated.$s-16;
+  padding: 0 deprecated.$s-6;
+
+  // Half of the height, so the badge stays fully rounded (pill shaped)
+  // also when the count has more than one digit.
+  border-radius: deprecated.$s-8;
+  background-color: var(--color-background-quaternary);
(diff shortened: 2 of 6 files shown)


madalenapmelo-kp:

filipsajdak (commented):
Reviewed by checking this branch out and running it, since I had built the same thing and can compare directly. It behaves as described and I have no blocking comments.

**Verified locally** (devenv, `develop` + this branch):

| case | result |
|---|---|
| Blur with one blur | counter badge `1` |
| Shadow with two | counter badge `2` |
| Mixed multi-selection | `Mieszane` (localized) via the text span, `title` set |
| Section emptied | no summary at all — importantly, **no `0` badge** |

`clj-kondo` 0 errors / 0 warnings, `cljfmt` clean, shadow-cljs `:main` 47 files 0 warnings.

## Things I think you got right, having done it the other way

- **Computing `summary` per section as a guarded binding** (`(when (seq blur-values) …)`, `(when has-exports? …)`) rather than deriving it inside the slot. I passed `(count …)` inline and leaned on `collapsable` being falsy to hide the empty case — which works, but only by accident of a second condition. Yours makes "empty renders nothing" explicit and local, and it is why the `0` badge case above is clean.
- **`:title` only when the summary is a string.** A badge reading `3` does not need a tooltip that also says `3`, whereas a truncated label does. I had set it for both.
- **`max-width` + ellipsis on `.title-summary`.** I missed this entirely. `settings.multiple` is short in English and Polish but not everywhere, and the summary competes with the section title for a resizable sidebar — `flex: 0 1 auto` with `min-width: 0` is the right shape.
- **`color-bullet-list*` living in `color_bullet.cljs`** (over in #11322) rather than a new namespace. I had put mine in a separate sidebar namespace; yours keeps the bullet and its list together, which is one less place for styling to drift.

## On the `role="button"` question

Agreed it is not a blocker here, and I would not hold the PR for it: the bullets are `div`s with no `tabindex`, so they are not focusable, and the whole subtree is `aria-hidden` — nothing reaches the accessibility tree. Worth noting for anyone auditing later that axe's `nested-interactive` rule only fires on *focusable* nested interactive elements, so this should not trip it either.

One correction though: the follow-up is cheaper than you estimated — it needs **no call-site changes at all**. `color-bullet*` already derives the state it needs:

```clojure
(let [read-only? (nil? on-click) ...]
  ...
  :role "button"
  :data-readonly (str read-only?)
```

so `:role (when-not read-only? "button")` is a one-line change inside the component. Every call site that passes `on-click` keeps the role; the ones that do not are already flagged `data-readonly="true"`.

The reason to still do it separately is not the call sites but the blast radius: it would drop the role from existing read-only bullets elsewhere (inspect, libraries). That is arguably a correctness fix — they are not buttons — but it is a behaviour change beyond this PR's scope, so a follow-up is the right call.

## One thing I did not check

I verified at the default sidebar width only. Since the sidebar is resizable and the summary shares a row with the title, it may be worth a glance at the narrowest width to confirm the title truncates before the summary does. The `flex`/`min-width` values look correct for that; I just did not exercise it.

## Note

I had a complete implementation of the same feature and have not opened a PR for it — no competing PR from me, and I have folded the two ideas of yours I preferred into my local branch only. Reviewing #11322 next.

yuricoppe:
Following up on the sidebar-width question from review: I checked, and the Design panel where these sections live is not actually resizable in normal use — `right-sidebar-default-width` is passed as both the initial *and* minimum value to `use-resize-hook`, and resizing is only enabled at all on the Inspect → Code sub-tab. So the panel is always exactly 318px here; there's no narrower width to test against.

While digging into that I found a related gap: the Inspect → Styles tab renders the same kind of collapsible sections (Fill, Stroke, Shadow, Blur, ...), but through a separate, older header component (`style-box*` in `inspect/styles/style_box.cljs`) that never got a summary slot — collapsing a section there shows nothing today, unlike the Design panel after this change. Filed as a follow-up: #11323.

yuricoppe:
Follow-up for the Inspect tab's Styles panel (separate, older collapsible header component) tracked in #11323.

niwinz:
Do we have issue for this PR?

filipsajdak:
> Do we have issue for this PR?

@niwinz this is part of https://github.com/penpot/penpot/issues/11147

PR #11322, opened 2026-08-21
✨ Show a color summary on collapsed sidebar sections

yuricoppe:

## What

Wires the three color sections of the design sidebar — Fill, Stroke and Selected colors — to the collapsed-section summary added in the previous PR. While collapsed they now show up to three color bullets plus a `+N` counter, or `Mixed` when the selection has differing values.

Completes #11147.

## Why

Follow-up to #11321, the count sections, split out because the color summaries carry the design detail (bullet size, overflow counter, placement) that the counters do not.

## How

- **Reuses `color-bullet*` with its `mini` variant** rather than a new swatch, in a small `color-bullet-list*` wrapper. This matters beyond deduplication: a naive hex swatch silently drops gradients, images and opacity, all of which the existing bullet already handles.
- **Passes color maps, not hex.** Fill uses the color from the fills metadata and Stroke uses `ctc/stroke->color`, the same conversion `stroke-row*` itself does, so the bullets match the rows the section shows when expanded.
- **Selected colors** builds its summary from `all-colors`, which includes token colors. That is wider than `has-colors?`, which counts only plain and library colors — but the summary is guarded by `has-colors?` and renders only while collapsed, so a selection with token colors alone never reaches it.
- Goes through the same polymorphic `summary` slot as the counters, so no new plumbing in `title-bar*`.

## Screenshots

Fill and Stroke collapsed, with the overflow counter on four fills:

Fill and Stroke sections collapsed showing color bullets and a plus-one counter

Selected colors, light and dark:

Selected colors section collapsed showing three bullets and a plus-four counter

Selected colors section collapsed in dark theme

Fill on a multi-selection with differing values:

Selection fill section collapsed showing the Mixed label

Expanded, the sections are unchanged:

Fill and Stroke sections expanded, identical to the current layout

Empty sections are untouched — no bullets, and the section stays non-collapsable:

Design sidebar with every section empty and no summaries

## Known trade-off

`color-bullet*` sets `role="button"` unconditionally, including when it has no `on-click`, so the summary puts elements with an interactive role inside the section's toggle `<button>`. In practice the bullets are not focusable and the whole summary is `aria-hidden`, so they never reach the accessibility tree — but dropping the role when the bullet is read-only would touch every call site of the shared component, which felt out of scope here. Happy to do it as a follow-up if you would rather have it clean.

## Note for review

Stacked on top of #11321, which is not merged yet, so its two commits still appear here. Only the last two commits belong to this PR.

### DIFF
--- a/frontend/src/app/main/ui/components/color_bullet.cljs
+++ b/frontend/src/app/main/ui/components/color_bullet.cljs
@@ -7,6 +7,8 @@
 (ns app.main.ui.components.color-bullet
   (:require-macros [app.main.style :as stl])
   (:require
+   [app.common.data :as d]
+   [app.common.data.macros :as dm]
    [app.common.math :as mth]
    [app.config :as cfg]
    [app.util.color :as uc]
@@ -110,6 +112,22 @@
                    :style {:background (uc/color->background (assoc color :opacity 1))}}]
             [:div {:class (stl/css :color-bullet-right)
                    :style {:background (uc/color->background color)}}]])]))))
+
+(mf/defc color-bullet-list*
+  "Compact row of mini color bullets with an overflow counter. Used as
+  an inline summary of colors (e.g. on collapsed sidebar sections)."
+  [{:keys [colors max-colors]}]
+  (let [max-colors (d/nilv max-colors 3)
+        visible    (take max-colors colors)
+        remaining  (- (count colors) (count visible))]
+    [:div {:class (stl/css :color-bullet-list)}
+     (for [[index color] (d/enumerate visible)]
+       [:> color-bullet* {:key (dm/str "color-" index)
+                          :color color
+                          :mini true}])
+     (when (pos? remaining)
+       [:span {:class (stl/css :color-bullet-list-more)}
+        (dm/str "+" remaining)])]))
 
 (mf/defc color-name*
   [{:keys [color size on-click on-double-click origin]}]
--- a/frontend/src/app/main/ui/components/color_bullet.scss
+++ b/frontend/src/app/main/ui/components/color_bullet.scss
@@ -82,6 +82,19 @@
   border: $b-2 solid var(--color-accent-primary);
 }
 
+.color-bullet-list {
+  display: flex;
+  align-items: center;
+  flex-shrink: 0;
+  gap: var(--sp-xs);
+}
+
+.color-bullet-list-more {
+  @include use-typography("body-small");
+
+  color: var(--color-foreground-secondary);
+}
+
 .color-text {
   @include two-line-text-ellipsis;
   @include use-typography("body-small");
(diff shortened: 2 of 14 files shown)


madalenapmelo-kp:

filipsajdak (commented):
Checked this one out and ran it too. No blocking comments.

**Verified locally** (devenv, full clean rebuild of `:main` — 1228 files, 768 compiled, **0 warnings**; `clj-kondo` 0/0, `cljfmt` clean):

| case | result |
|---|---|
| Fill, four fills | three bullets in the correct colours + `+1` |
| Stroke | bullet renders the stroke colour |

## On my `color_selection.cljs` note — you were right, I was wrong

I flagged that `all-colors` is wider than `has-colors?`. Having read the expanded body properly, that concern does not hold: the section renders `library-colors`, `colors` **and** `token-colors` when open (there is a dedicated `expand-token-color` state), so `all-colors` is exactly what the section shows. Using it for the summary is the consistent choice, not an over-count. Thanks for documenting it rather than just changing it.

One adjacent observation, and to be clear it is **pre-existing and not this PR's business**: `has-colors?` is `(or (seq colors) (seq library-colors))` and deliberately excludes `token-colors`. So a selection whose colours are *only* token colours gets no summary — and is not collapsible either, since `has-colors?` also drives `:collapsable` — while the expanded body would still list those token colours. That asymmetry is in `has-colors?` today, untouched by you. Might be worth a separate issue; definitely not something to fix here.

## Small things I liked

- `color-bullet-list*` living in `color_bullet.cljs` rather than a new namespace — the list and the bullet it composes stay together, and `max-colors` defaulting via `d/nilv` keeps the call sites bare.
- `ctc/stroke->color` for the stroke colours: the same helper `stroke-row*` itself uses, so the summary cannot drift from the row beneath it.
- `(mapv (comp :color meta) fills)` reads the colour the fills pipeline already computed, rather than re-deriving it.

## Follow-ups, both non-blocking

- The `role="button"` question — answered on #11321. Short version: not a blocker (the bullets are not focusable and the subtree is `aria-hidden`), and the follow-up is cheaper than you thought, since `read-only?` is already derived inside `color-bullet*`, so `:role (when-not read-only? "button")` needs no call-site changes. The reason to keep it separate is blast radius on existing read-only bullets, not churn.
- As on the other PR, I only exercised the default sidebar width. Worth a glance at the narrowest width, since the bullet list is `flex-shrink: 0` inside a `max-width` summary and competes with the section title.

Nice work slicing it this way — the count PR being reviewable without any of the colour detail made both much easier to read.

yuricoppe:
Following up on the sidebar-width question from review: I checked, and the Design panel where these sections live is not actually resizable in normal use — `right-sidebar-default-width` is passed as both the initial *and* minimum value to `use-resize-hook`, and resizing is only enabled at all on the Inspect → Code sub-tab. So the panel is always exactly 318px here; there's no narrower width to test against.

While digging into that I found a related gap: the Inspect → Styles tab renders the same kind of collapsible sections (Fill, Stroke, Shadow, Blur, ...), but through a separate, older header component (`style-box*` in `inspect/styles/style_box.cljs`) that never got a summary slot — collapsing a section there shows nothing today, unlike the Design panel after this change. Filed as a follow-up: #11323.

PR #11337, opened 2026-08-24
:bug: Evict multi-scale tile cache on shape edits

### LINKED ISSUE #11302: Large text rotation leaves "partial ghost cache" impression while zooming in and out (MS Windows)
### Steps To Reproduce

1. create a sufficiently large text (like 5000 words) with a particularly weird shape (thing and long) 
2. rotate it
3. zoom in, zoom out, etc, to make sure Penpot creates the appropriate tiles :)
4. select the text and reset angle to 0 (this is arbitrary, just a way to make the change obvious)
5. Slowly zooming in and out will still show fragments of the old rotated text

I have only been able to reproduce on Windows (Chrome) and not always so this could be nothig :( Of course, after a while, it disappears.

### Expected behavior

Rotation and manipulation of big objects like texts, that requiere extensive tiling should immediately discard any previous tiles when zooming in or out.

### Actual behavior

What is shown in the video (sorry, I don't use Windows, I recorded it with my phone)

- OS: MS Windows 10
- Browser: Chrome 151.0.7922.140 (Official Build) (64-bit)

_No response_

- Host hourly.penpot.dev

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

superalex:
Solves #11302 

## What

- After rotating or moving a shape, zooming out could briefly show stale fragments of the previous geometry from the tile cache.
- Zoom settle can present the visible viewport before the interest-ring tiles finish, so the canvas turns sharp sooner.

## Why

- Shape-edit invalidation only covered the new bounds (and interest-clipped tiles at the current scale). Low-zoom cached tiles outside that coverage kept pre-edit pixels, and `render_from_cache` could overlay them after zoom-out.
- Waiting for the full interest ring before presenting made zoom settle feel soft longer than necessary.

## How

- **Tile cache keys:** Key cached tile textures by scale so multi-zoom overlays stay aligned and zoom no longer wipes the whole cache.
- **Edit dirty region:** Snapshot the pre-edit extent on first touch and invalidate `old ∪ new ∪ indexed` coverage so every scale that still holds those tiles is evicted.
- **ViewportReady:** Present visible tiles early, defer interest-ring work to follow-up frames, and paint into atlas slot size so HiDPI does not rasterize oversized tiles only to downscale them.

### DIFF
--- a/render-wasm/src/render/surfaces.rs
+++ b/render-wasm/src/render/surfaces.rs
@@ -92,6 +92,29 @@ pub enum SurfaceId {
     TileAtlas = 0b100_0000_1000,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct TileCacheKey {
+    pub tile: Tile,
+    pub scale_bits: u32,
+}
+
+impl TileCacheKey {
+    pub fn new(tile: Tile, scale: f32) -> Self {
+        Self {
+            tile,
+            scale_bits: scale.to_bits(),
+        }
+    }
+
+    pub fn matches_scale(&self, scale: f32) -> bool {
+        self.scale_bits == scale.to_bits()
+    }
+
+    pub fn scale(&self) -> f32 {
+        f32::from_bits(self.scale_bits)
+    }
+}
+
 pub struct DocAtlas {
     // Persistent 1:1 document-space atlas that gets incrementally updated as tiles render.
     // It grows dynamically to include any rendered document rect.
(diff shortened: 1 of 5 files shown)


render-wasm/src/render/surfaces.rs
elenatorro, line 1737: There's something here that already happens on production which is that we never "clear" the removed set unless we call `repack()`. Now that keys are TileCacheKey this it could be that it never stops growing (scale is a float so it's rarely going to repeat). Since we remove here the tiles from grid, we could also clean these from `removed`, so we could do instead: `for tile in self.removed.drain() {`

render-wasm/src/render/surfaces.rs
```diff
@@ -1818,53 +1926,83 @@ impl TileTextureCache {
...
+    }
+
+    pub fn remove_at(&mut self, tile: Tile, scale: f32) {
+        let key = TileCacheKey::new(tile, scale);
+        if let Some(tile_ref) = self.grid.get(&key) {
             if tile_ref.index < self.textures.len() {
                 self.textures[tile_ref.index].set_empty();
             }
         }
+        self.removed.insert(key);
```
elenatorro, line 2004: invalidate_cached_tiles_intersecting calls remove_at once per matching tile_doc_rects key, so this is what fills the "removed" set mentioned in the other comment. We're saving here the key even if it's not on the grid. Maybe we don't need to insert it always here, right? also, we wouldn't need to set "is_updated = true" in that case, so we wouldn't need to do more snapshots in that case

render-wasm/src/render/surfaces.rs
superalex: Makes sense!, thank you very much!

render-wasm/src/render/surfaces.rs
superalex: Updated! thanks!

PR #11345, opened 2026-08-25
:sparkles: Add storage object status lifecycle and verified dedup

### LINKED ISSUE #11344: Make backend storage resilient to interrupted writes, missing files and stalled cleanup
### Description

When saving data to object storage fails halfway through — a crash, a restart, or a transient S3/filesystem error — the database can be left with records pointing to files that were never fully stored, plus temporary files that nothing cleans up. Users see this as images that fail to load, broken thumbnails, or media that cannot be opened; self-hosted administrators find dangling records and orphaned data that must be cleaned by hand.

Deduplication amplifies the damage: when identical content is saved again, the existing record is reused without confirming its file still exists, so one lost file keeps producing broken references every time the same content reappears.

Garbage collection struggles to reclaim space: a physical deletion that keeps failing (network outage, wrong permissions, misconfigured bucket) was retried indefinitely, letting a single unreachable file block storage reclamation forever.

Storage should behave transactionally from the outside: a record exists only while its content does, reuse of identical content should verify the file exists first and repair itself when it does not, crashed-process leftovers should be swept automatically, and deletion retries should be bounded so background cleanup always makes progress even when some deletions keep failing.

### Affected versions

All released versions; targeted for 2.19.0

niwinz:

## What

- Write-ahead lifecycle for storage objects: records start as `pending`, are promoted to `valid` only after their blob reaches the storage backend, and stay invisible to reads/dedup/GC until then
- New daily maintenance task that reclaims pending rows orphaned by a crash between blob write and promotion
- Deduplication hits verify that the referenced blob physically exists before reuse and self-heal stale references
- Physical blob deletions are tracked per object, deferred on failure, and abandoned after a bounded number of attempts
- Filesystem backend writes blobs atomically via temp file + move, and crashed-process temp files are swept automatically
- Storage GC tasks accept `skip-delay` / `chunk-size` params to exercise the whole deletion cascade immediately
- Ships migration `0153-add-storage-object-status-and-deletion-attempts`

Closes #11344

## Why

Until now, a crash or storage error between the database insert and the blob write left a row without its data behind; deduplication reused such broken references indefinitely; and failed physical deletions were retried in an unbounded loop that pinned garbage-collection progress. This PR implements the backend-side fix described in the linked issue.

### Write-ahead lifecycle

- `put-object!` inserts the row with `status='pending'`, writes the blob through the resolved backend, then flips the row to `'valid'`; if anything fails mid-way the row stays pending by design
- Pending rows are excluded from every read path: dedup lookup SQL, `get-object`, upload-chunk assembly queries, and both touched/deleted GC scans
- New `storage-pending-gc` integrant task (daily cron): selects pending rows older than 24h with `FOR UPDATE SKIP LOCKED` in chunks, deletes the rows in one committed transaction, then removes the orphaned blobs best-effort so blob failures never block row reclamation

### Verified deduplication

- On a content-hash hit, `exists-object?` runs against the backend before reuse (fs stat / S3 `headObject`)
- Missing blob: the stale row is repaired in place — it keeps its id and gets flagged so `put-object!` rewrites the incoming content under the same id, restoring data for every existing reference to that object; a failed repair leaves the row live and valid so a later matching upload retries the heal
- S3 implementation distinguishes definitive misses (`NoSuchKeyException`, answered immediately) from transient errors (retried with linear backoff, rethrown when exhausted)

### Bounded deletion retries

- `del-objects-in-bulk` contract change: backends now return the exact set of failed ids — fs collects per-file exceptions; S3 maps partial `deleteObjects` errors per key and treats a network failure as all-failed instead of raising
- `gc-deleted` restructured into ~25-row chunks, each in its own short transaction (previously one transaction spanned the entire run over a cursor)
- Per chunk: succeeded ids get their rows deleted; failed ids get `deletion_attempts` incremented and `deleted_at` deferred by one day; at 7 attempts the row is dropped and the orphan blob accepted, with a warning log

### Atomic fs writes

- Blob content goes to a tempfile created in the target directory (same filesystem), registered in the cleanup queue with min-age 1h so leftovers from crashed JVMs are swept later, then moved into place with `ATOMIC_MOVE` (fallback to replace when the filesystem does not support it); partial blobs can never appear at the final path

### Simplified GC delays

- The touched GC no longer stacks an extra deletion-delay on top of the domain-level delay — marking `deleted_at = now` is safe because the domain reference scan already ran; the deleted GC picks objects up on its next pass
- Upload chunks encode freshness as a future `touched-at` (+1h, aligned with the upload-session TTL) instead of a special-cased deletion delay for the tempfile bucket
- Task handlers read run props: `storage-gc-touched` honors `:skip-delay` (processes everything regardless of min-age) and `objects-gc` honors `:chunk-size` and `:skip-delay` — enabling the full cascade to be triggered on demand via `run-task!`

### Plumbing

- Storage instances hold the pool directly; `sto/resolve cfg ::db/reuse-conn true` attaches the ambient transaction connection only where atomicity with the surrounding transaction matters (file-overwrite import thumbnail invalidation, snapshot persistence, thumbnail delete RPCs)
- Thumbnail invalidation helper moved from binfile common into the v3 importer, its only consumer; thumbnail deletion RPCs return a clean `nil` payload instead of a JDBC update result

### DIFF
--- a/.serena/memories/backend/storage.md
+++ b/.serena/memories/backend/storage.md
@@ -30,10 +30,46 @@
 - `objects-gc` removes deleted domain rows and touches their storage object IDs.
 - Use `::db/reuse-conn true` with `sto/resolve` inside a database transaction.
 
+## Connection Reuse Details
+
+### `app.storage/resolve` patterns:
+
+**1. Pool mode (default)** - `(sto/resolve cfg)`
+- Returns storage abstraction from config
+- Uses whatever database pool is available
+- **Safe to call outside transaction context**
+- Used in: `rpc/commands/media.clj:363`, `rpc/commands/auth.clj:327`, `rpc/commands/profile.clj:362`
+
+**2. Connection reuse mode** - `(sto/resolve cfg ::db/reuse-conn true)`
+- Internally calls `db/get-connection cfg` to obtain connectable
+- Configures storage with the specific connection from config
+- **Must be paired with transaction that owns this connection**
+- Used in: `features/fdata.clj:100`, `rpc/commands/media.clj:425`, `rpc/commands/files_thumbnails.clj:307,319`, `binfile/v3.clj:722`
+
+**3. Explicit configuration** - `(sto/configure storage conn)`
+- Sets `::db/conn` on storage map directly
+- Asserts `db/conn? connection` (storage.clj:349)
+- Used inside `db/tx-run!` blocks where `conn` is already available
+- Used in: `tasks/file_gc.clj:256`, `rpc/commands/files_thumbnails.clj:347,371`
+
+### Key Warning (from function notes):
+
+The improved note in `import-storage-objects` and `handle-persistence` warns:
+**Do not reuse the main database connection for storage operations within a transaction.** The storage upload process can fail mid-operation, leaving orphaned objects on the backend. If the outer transaction aborts, pending storage objects become unreconciliable because the storage subsystem registers its pending state in separate transactions.
+
+### Rule of Thumb for `sto/put-object!`:
+
+Since `put-object!` uses backend-specific operations (`impl/resolve-backend` + `impl/put-object`) and does not directly use `::db/conn` or `::db/pool`, **all usage of `put-object!` will never run inside a common transaction** (if configured at all). The storage backend operations are independent of the database transaction boundary.
+
 ## Deduplication
 
 - Deduplication requires `::sto/deduplicate?`, a content hash, and bucket metadata.
 - The lookup matches hash, bucket, backend, and `deleted_at IS NULL`.
+- The lookup only considers rows with `status='valid'`; pending rows are invisible.
+- A hit whose blob is missing is repaired in place: the same row/id is kept,
+  the row is flagged `::blob-missing?`, and `put-object!` rewrites the blob
+  under that id. This heals all existing references to the object. If the
+  rewrite fails, the row is left live and valid for a later retry.
 - The lookup does not include file ID, profile ID, team ID, or organization ID.
 - Objects can therefore share content across users and files within one bucket.
 - Deleted objects are not reused.
--- a/.serena/memories/backend/storage.md
+++ b/.serena/memories/backend/storage.md
@@ -67,9 +67,9 @@ Since `put-object!` uses backend-specific operations (`impl/resolve-backend` + `
 - The lookup matches hash, bucket, backend, and `deleted_at IS NULL`.
 - The lookup only considers rows with `status='valid'`; pending rows are invisible.
 - A hit whose blob is missing is repaired in place: the same row/id is kept,
-  the row is flagged `::blob-missing?`, and `put-object!` rewrites the blob
-  under that id. This heals all existing references to the object. If the
-  rewrite fails, the row is left live and valid for a later retry.
+  and `put-object!` rewrites the blob under that id. This heals all existing
+  references to the object. If the rewrite fails, the row is left live and
+  valid for a later retry.
 - The lookup does not include file ID, profile ID, team ID, or organization ID.
 - Objects can therefore share content across users and files within one bucket.
 - Deleted objects are not reused.
--- a/backend/src/app/storage.clj
+++ b/backend/src/app/storage.clj
@@ -70,7 +70,7 @@
   [:map {:title "storage"}
    [::backends schema:backends]
    [::backend [:enum :s3 :fs]]
-   ::db/connectable])
+   ::db/pool])
 
 (def valid-storage?
   (sm/validator schema:storage))
@@ -96,7 +96,7 @@
     (-> (d/without-nils cfg)
         (assoc ::backends backends)
         (assoc ::backend backend)
-        (assoc ::db/connectable pool))))
+        (assoc ::db/pool pool))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Database Objects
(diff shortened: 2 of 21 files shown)


backend/src/app/storage/gc_deleted.clj
```diff
@@ -47,66 +57,103 @@
...
-              (+ total (or deleted 0))))
-          0
-          (get-buckets conn)))
+  [cfg]
+  (loop [total 0]
+    (let [chunk (get-deleted-chunk cfg chunk-size)]
+      (if (seq chunk)
+        (let [by-backend (group-by-backend chunk)
+              deleted    (reduce-kv (fn [acc backend-id ids]
+                                      (+ acc (process-chunk! cfg backend-id ids)))
```
superalex: process-chunk! returns nil when lock-ids finds nothing (SKIP LOCKED). clean-deleted! then does (+ acc nil) and the task crashes. The previous code used (or deleted 0). Suggest returning 0 from process-chunk! when nothing was locked.

backend/src/app/storage/gc_deleted.clj
```diff
@@ -47,66 +57,103 @@
...
-
-(defn- get-buckets
-  [conn]
-  (let [now (ct/now)]
-    (sequence
-     (comp (partition-all 25)
-           (mapcat group-by-backend))
-     (db/cursor conn [sql:get-deleted-sobjects now]))))
+(def ^:private sql:get-deleted-chunk
+  "SELECT id, backend
```
superalex, line 126: sql:get-deleted-chunk has no FOR UPDATE SKIP LOCKED. If another worker holds the locks, we re-select the same eligible rows forever once process-chunk! returns 0. Either lock in this query (like pending-gc) or break the loop when a full chunk makes zero progress.

backend/src/app/storage/gc_deleted.clj
```diff
@@ -47,66 +57,103 @@
...
+  (db/tx-run! cfg
+              (fn [{:keys [::db/conn]}]
+                (when-let [locked-ids (lock-ids conn ids)]
+                  ;; Attempt blob deletion (catch Throwable — if it throws, treat all as failed)
+                  (let [fail-ids (try
+                                   (-> (impl/resolve-backend (::sto/storage cfg) backend-id)
+                                       (impl/del-objects-in-bulk locked-ids))
+                                   (catch Throwable _
+                                     (do
+                                       (l/err :hint "error on physical deletion, will retry"
```
superalex: Should we pass :cause to the log (same pattern as pending-gc/delete-blobs!)?. Same in s3 del-objects-in-bulk

backend/src/app/storage.clj
```diff
@@ -209,29 +176,93 @@
...
+    (if-some [hit (when (and (::deduplicate? params)
+                             (:hash mdata)
+                             (:bucket mdata)
+                             (not= "tempfile" (:bucket mdata)))
+                    (get-database-object-by-hash pool backend
+                                                 (:bucket mdata)
+                                                 (:hash mdata)))]
+
+      ;; PHASE 2: an existing reference is found: reuse or repair it.
+      (if (impl/exists-object? backend' hit)
```
superalex, line 220: Have we done some texts about the extra latency about this?

backend/src/app/storage/gc_deleted.clj
```diff
@@ -19,8 +19,18 @@
    [app.db :as db]
    [app.storage :as sto]
    [app.storage.impl :as impl]
+   [clojure.set :as set]
    [integrant.core :as ig]))
 
+(def ^:private max-attempts
```
superalex, line 25: Should we have tests forcing del-objects-in-bulk to fail (or returns fail-ids) after max attempts ?

.serena/memories/backend/storage.md
```diff
@@ -30,10 +30,46 @@
...
+
+Since `put-object!` uses backend-specific operations (`impl/resolve-backend` + `impl/put-object`) and does not directly use `::db/conn` or `::db/pool`, **all usage of `put-object!` will never run inside a common transaction** (if configured at all). The storage backend operations are independent of the database transaction boundary.
+
 ## Deduplication
 
 - Deduplication requires `::sto/deduplicate?`, a content hash, and bucket metadata.
 - The lookup matches hash, bucket, backend, and `deleted_at IS NULL`.
+- The lookup only considers rows with `status='valid'`; pending rows are invisible.
+- A hit whose blob is missing is repaired in place: the same row/id is kept,
+  the row is flagged `::blob-missing?`, and `put-object!` rewrites the blob
```
superalex: ::blob-missing? isn't used in the rest of the PR

backend/test/backend_tests/storage_test.clj
```diff
@@ -368,27 +379,479 @@
...
+
+      ;; without skip-delay the future deleted row is not processed
+      (let [res (th/run-task! :objects-gc {})]
+        (t/is (= 0 (:processed res))))
+
+      ;; with skip-delay it is processed immediately
+      (let [res (th/run-task! :objects-gc {:skip-delay true})]
+        (t/is (= 1 (:processed res)))))))
+
+(t/deftest put-object-write-failure-leaves-no-row
```
superalex: The test name says leaves-no-row, but the body asserts a pending row remains

backend/src/app/storage.clj
```diff
@@ -209,29 +176,93 @@
...
+          (row->storage-object (cond-> hit touch (assoc :touched-at touched-at))))
+
+        ;; PHASE 2b: the referenced blob is missing (a stale/broken row).
+        ;; Repair the reference in place: rewrite the incoming content
+        ;; under the same id, restoring the blob for all existing
+        ;; references to it. If the write fails, the exception propagates
+        ;; and the row stays live and valid, so a later matching upload
+        ;; retries the heal.
+        (let [object (row->storage-object hit)]
+          (l/wrn :hint " blob not found on reusing storage object"
```
superalex: Probably there is an extra space before "blob"

backend/src/app/storage.clj
niwinz: the extra latency should be negligible  in most cases, only when S3 performance is degraded. For now i consider this leave as is.

PR #11389, opened 2026-08-27
:whale: Migrate media-processor images to DHI

bameda:
## What and why

The media-processor module had a Dockerfile but no way to build it through manage.sh, so neither its bundle nor its image could be produced. This adds both, and migrates the image to Docker Hardened Images along the way.

`scripts/build` now assembles the bundle under `target/` — following the same convention as the other modules — rather than leaving it to manage.sh. This is needed because esbuild leaves the runtime dependencies external (sharp, pino, pino-pretty, pino-loki), so the bundle has to ship package.json / pnpm-lock.yaml / pnpm-workspace.yaml next to dist/, plus a setup script for the image to run at build time. The generated setup mirrors the one mcp/scripts/build produces.

Also fixed while testing: the first bundle build on a fresh devenv hung on corepack's interactive "do you want to download pnpm?" prompt (the module pins pnpm via `packageManager`). `scripts/build` now sets
`COREPACK_ENABLE_DOWNLOAD_PROMPT=0` and calls `corepack install` explicitly, which would otherwise deadlock a CI build. Note it deliberately does *not* call `corepack enable` (unlike mcp/scripts/build): that writes shims into root-owned /opt/node/bin while manage.sh runs the build as the penpot user, and the devenv image already creates them.

**CI is intentionally untouched.** The module is still work in progress and its images are not published yet, so `build-docker.yml` gets no media-processor step — this PR only makes local builds possible.

## How to build and test it

DHI images require an authenticated pull, so log in first with a (free) Docker ID — sign up at https://app.docker.com/signup if you don't have one:

```bash
docker login dhi.io   # or docker login
```

Build the bundle and the image:

```bash
./manage.sh build-media-processor-bundle
./manage.sh build-media-processor-docker-image --tag dhi-test
```

The bundle should come out like this, with `setup` executable:

```bash
find bundles/media-processor -maxdepth 2 | sort
ls -l bundles/media-processor/setup
```

Run the image — the service listens on 6065:

```bash
docker run --rm -p 6065:6065 penpotapp/media-processor:dhi-test
```

Extra test: check that the Ubuntu → Debian base swap didn't break the processing tooling (sharp's native bindings, fontforge and woff2):

```bash
docker run --rm --entrypoint sh penpotapp/media-processor:dhi-test -c '
  node -e "
    const sharp = require(\"sharp\");
    console.log(\"sharp OK\", sharp.versions);
    sharp({create:{width:8,height:8,channels:4,background:{r:255,g:0,b:0,alpha:1}}})
      .png().toBuffer()
      .then(b => console.log(\"png encode OK\", b.length, \"bytes\"))
      .catch(e => console.log(\"FAIL\", e.message));
  "
  echo "--- fontforge ---"; fontforge --version 2>&1 | head -2
  echo "--- woff2 ---";     woff2_compress 2>&1 | head -1
'
```

### Testing notes

All of the above was run on this branch:

- Bundle produced with `dist/index.js`, the three manifests, an executable
  `setup`, `version.txt` and `LICENSE`.
- Image built cleanly and the container starts and listens on 6065.
- sharp loads and encodes a PNG (96 bytes); `fontforge --version` and
  `woff2_compress` both respond.

## Breaking changes

None. New commands, and an image that isn't built by CI yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,6 +92,7 @@ opencode.json
 /playwright/.cache/
 /render-wasm/target/
 /media-processor/dist/
+/media-processor/target/
 /**/node_modules
 /**/.yarn/*
 /.pnpm-store
--- a/docker/images/Dockerfile.media-processor
+++ b/docker/images/Dockerfile.media-processor
@@ -1,17 +1,16 @@
-FROM ubuntu:26.04
+# syntax=docker/dockerfile:1
+FROM dhi.io/node:24.18.0-debian13-dev
 LABEL maintainer="Penpot <docker@penpot.app>"
 
 ENV LANG=en_US.UTF-8 \
     LC_ALL=en_US.UTF-8 \
-    NODE_VERSION=v24.18.0 \
-    DEBIAN_FRONTEND=noninteractive \
-    PATH=/opt/node/bin:$PATH
+    DEBIAN_FRONTEND=noninteractive
 
+# passwd provides useradd, not preinstalled on the DHI base image.
 RUN set -ex; \
-    useradd -U -M -u 1001 -s /bin/false -d /opt/penpot penpot; \
-    mkdir -p /etc/resolvconf/resolv.conf.d; \
-    echo "nameserver 127.0.0.11" > /etc/resolvconf/resolv.conf.d/tail; \
     apt-get -qq update; \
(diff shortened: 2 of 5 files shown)


niwinz:
@copilot resolve the merge conflicts in this pull request

PR #11402, opened 2026-08-27
:sparkles: Add account lockout after failed login attempts

### LINKED ISSUE #11397: Add account lockout after repeated failed login attempts
### Description

Penpot currently has no account lockout mechanism after repeated failed login attempts. An attacker can make unlimited password-guessing attempts against any account without triggering any defensive response. The only existing protection is IP-based rate limiting (opt-in via the  flag), which is ineffective against attackers who rotate IP addresses.

This change adds a per-account brute-force protection: after N failed login attempts within a time window, the account is temporarily locked out and further login attempts are rejected with a 429 response containing a JSON body with error details (including the remaining lockout duration).

### Use case

Protects all Penpot accounts from online brute-force password attacks, especially important for:
- Public-facing instances
- Accounts with weak or reused passwords
- Compliance with security audit recommendations (Criptored T4-N2-09)

### Affected versions

2.19.0

niwinz:

## What

- Adds per-account brute-force protection: after 5 failed login attempts within 15 minutes, the account is temporarily locked out.
- All login attempts (including with the correct password) are rejected with a 429 response containing a JSON body with error details while the lockout is active.
- The frontend displays a coherent error message ("Too many failed attempts. Your account has been locked for X minutes.") instead of the generic "Something wrong has happened".

image

## Why

Penpot had no account lockout mechanism after repeated failed login attempts. An attacker could make unlimited password-guessing attempts against any account. The only existing protection was IP-based rate limiting (opt-in), which is ineffective against attackers who rotate IPs.

## How

- **Redis-backed counter** (`app.auth.login-lockout`): atomic Lua script increments a per-profile counter with a TTL window. No schema migration needed; auto-expiry handles lockout duration naturally.
- **Fail-open on Redis errors**: if Redis is unavailable, logins are still allowed (avoid self-DoS).
- **Configurable**: `:login-lockout-max-attempts` (default 5) and `:login-lockout-window` (default 15m) via config; gated by `:account-lockout` flag.
- **Integration points**: `login-with-password` (check lockout before password, record failure after), `recover-profile` (clear counter on reset), `login-with-ldap` (clear counter on success).
- **Error response**: 429 body includes `{:type :rate-limit :code :account-locked :hint "..." :ttl <seconds>}` so the frontend can display a meaningful message with the remaining lockout duration.
- **Frontend i18n**: new `errors.account-locked` key in EN/ES, falls back to EN for other locales.

Closes #11397

### DIFF
--- a/backend/scripts/_env
+++ b/backend/scripts/_env
@@ -48,6 +48,7 @@ export PENPOT_FLAGS="\
        enable-user-feedback \
        disable-secure-session-cookies \
        enable-smtp \
+       enable-account-lockout \
        enable-prepl-server \
        enable-urepl-server \
        enable-nrepl-server \
--- a/backend/src/app/auth/login_lockout.clj
+++ b/backend/src/app/auth/login_lockout.clj
@@ -1,0 +1,132 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+;;
+;; Copyright (c) KALEIDOS SUBSIDIARY SL
+
+(ns app.auth.login-lockout
+  "Brute-force protection: per-account failed login counter backed by Redis.
+   Uses an atomic Lua script for the increment operation to prevent
+   concurrent login bypass. Stores count:window_start in the value;
+   time is passed from Clojure (ct/now) via ARGV, keeping it injectable
+   for tests. When the counter reaches the configured threshold within
+   the time window, the account is locked out until the window expires."
+  (:require
+   [app.common.generic-pool :as gpool]
+   [app.common.logging :as l]
+   [app.common.time :as ct]
+   [app.config :as cf]
+   [app.redis :as rds]
+   [app.redis.script :as-alias rscript]
+   [clojure.string :as str])
+  (:import
+   java.lang.AutoCloseable))
+
+(def ^:private key-prefix "penpot.login-lockout.")
+
+(def ^:private lockout-script
+  {::rscript/name ::login-lockout
+   ::rscript/path "app/auth/login_lockout.lua"})
+
+(defn- get-pool
+  [cfg]
+  (:app.redis/pool cfg))
+
+(defn- with-conn
+  [cfg f]
+  (let [pool (get-pool cfg)
+        conn (gpool/get pool)]
+    (try
+      (f @conn)
+      (finally
+        (.close ^AutoCloseable conn)))))
+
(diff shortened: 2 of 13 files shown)


niwinz:
## Code review fixes applied

Addressed all review findings:

### High priority
- **retry-after header on locked path**: `locked?` now returns `:ttl` (seconds remaining) by querying Redis TTL. The `Retry-After` header is now correctly populated on every locked attempt, not just the first one that crosses the threshold.

### Medium priority
- **LDAP lockout bypass**: `login-with-ldap` now checks `locked?` before attempting LDAP auth and records failed attempts on authentication failure. Profile lookup by email is done before `ldap/authenticate` to obtain the profile-id for lockout tracking.

### Low priority
- **Dead branch removed**: `get-pool` simplified to `(:app.redis/pool cfg)`.
- **Config validation**: Runtime assertions ensure `login-lockout-max-attempts >= 1` and `login-lockout-window >= 60000ms`.

### Tests
- Added integration test verifying `retry-after` header is present and positive on the already-locked path.
- Added unit tests for `locked?` returning `:ttl` when locked vs `nil` when not locked.

### Not addressed (deferred)
- **Redis calls inside DB transaction**: The lockout check/record still runs inside `db/tx-run!`. This is a larger refactor that should be done separately to avoid introducing risk.
- **Redis-down fail-open test**: Cannot easily mock private functions from test namespace. The fail-open behavior is documented in code and tested implicitly by the disabled-flag tests.

All tests pass: 19 lockout tests + 47 profile tests, lint clean, format clean.

Alotor (approved):

niwinz:
Follow-up note on the 429 response contract.

- `Retry-After` and the restored `x-rate-limit-remaining` / `x-rate-limit-reset` headers are follow-up work. The merged change replaced the header-forwarding 429 handler with a body-only one, so the existing RPC rate limits stopped sending their headers. A follow-up preserves `::http/headers` again and adds `Retry-After: <ttl seconds>` when the exception carries a non-nil `:ttl`.
- The account-lockout 429 keeps its current JSON body (`{:type :rate-limit :code :account-locked :hint ... :ttl <seconds>}`) and gains `Retry-After`.
- Activation stays backend-only: `enable-account-lockout` in `PENPOT_FLAGS` (internal key `:account-lockout`), no frontend flag. Defaults are 5 attempts / 15 minutes, configurable with `PENPOT_LOGIN_LOCKOUT_MAX_ATTEMPTS` and `PENPOT_LOGIN_LOCKOUT_WINDOW`.

Docs (live with the follow-up): https://help.penpot.app/technical-guide/configuration/#account-lockout

PR #11422, opened 2026-08-28
:bug: Fix text response construction from base64 data

### LINKED ISSUE #10878: MCP: export_shape with format "svg" writes 2 null bytes instead of the SVG

opcode81:
Fixes #10878 (SVG exports of shapes)

### DIFF
--- a/mcp/packages/server/src/ToolResponse.ts
+++ b/mcp/packages/server/src/ToolResponse.ts
@@ -10,12 +10,19 @@ export class TextContent implements TextItem {
     constructor(public text: string) {}
 
     /**
-     * @param data - Text data as string or as object (from JSON representation where indices are mapped to character codes)
+     * @param data - Text data as string or as object (wrapped base64 data or JSON representation of array where
+     * indices are mapped to character codes)
      */
     public static textData(data: string | object): string {
         if (typeof data === "object") {
-            // convert object containing character codes (as obtained from JSON conversion of string) back to string
-            return String.fromCharCode(...(Object.values(data) as number[]));
+            // check for wrapped base64 data
+            if ("__type" in data && data["__type"] === "base64" && "data" in data && typeof data["data"] === "string") {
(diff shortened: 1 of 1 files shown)


opcode81:
Hi @Alotor this one seems to have fallen by the wayside.

PR #11443, opened 2026-09-01
:zap: Batch WASM shape upload to speed up page switches

superalex:
## What                
- Upload structural shape attrs via multi-shape `_set_shapes_batch` FFI in chunks of 512
- Apply host-specific attrs (fills, strokes, text, grid, path) per shape after the batch upload
## Why                                                                                         
Page switches with thousands of shapes spent too long in per-shape WASM FFI calls during `set-objects`, blocking the main thread before the viewport could render.
## How                                                                                                                                                            
- **Rust:** new `upload_batch.rs` handler; batch structural props in `_set_shapes_batch`
- **Common:** `render_wasm/api/upload.cljs` serializes shape payloads for batch upload  
- **Frontend:** `set-objects` / `process-shapes-chunk` use batch upload then per-shape host attrs


common/src/app/common/render_wasm/api/upload.cljs
```diff
@@ -0,0 +1,456 @@
...
+    (.setFloat32 dview (+ offset 64) te true)
+    (.setFloat32 dview (+ offset 68) tf true)
+    (.setFloat32 dview (+ offset 72) sx1 true)
+    (.setFloat32 dview (+ offset 76) sy1 true)
+    (.setFloat32 dview (+ offset 80) sx2 true)
+    (.setFloat32 dview (+ offset 84) sy2 true)
+    (.setFloat32 dview (+ offset 88) r1 true)
+    (.setFloat32 dview (+ offset 92) r2 true)
+    (.setFloat32 dview (+ offset 96) r3 true)
+    (.setFloat32 dview (+ offset 100) r4 true)
```
niwinz: Is is ok but would be nice to use app.common.buffer api for it instead of direct interop

render-wasm/src/wasm/shapes/upload_batch.rs
```diff
@@ -0,0 +1,490 @@
...
+        };
+        with_current_shape_mut!(state, |shape: &mut Shape| {
+            shape.set_bool_type(bool_type);
+        });
+    }
+
+    if mask & SECTION_GROW_TYPE != 0 {
+        let raw_grow = cur.u8()?;
+        let _ = cur.take(3)?;
+        let grow_type = match raw_grow {
```
elenatorro: Same here: 

```
text_content.set_grow_type(RawGrowType::from(raw_grow).into())
```

render-wasm/src/wasm/shapes/upload_batch.rs
```diff
@@ -0,0 +1,490 @@
...
+        let _ = cur.take(3)?;
+        with_current_shape_mut!(state, |shape: &mut Shape| {
+            shape.set_masked(masked);
+        });
+    }
+
+    if mask & SECTION_BOOL_TYPE != 0 {
+        let raw_bool = cur.u8()?;
+        let _ = cur.take(3)?;
+        let bool_type = match raw_bool {
```
elenatorro: We already have an enum for BoolTypes, I think we can just do: 

```
shape.set_bool_type(RawBoolType::from(raw_bool).into());
```

render-wasm/src/wasm/shapes/upload_batch.rs
```diff
@@ -0,0 +1,490 @@
...
+const STROKE_ALIGN_OUTER: u8 = 2;
+
+fn decode_optional_f32(value: f32) -> Option<f32> {
+    if value.is_finite() && value >= 0.0 {
+        Some(value)
+    } else {
+        None
+    }
+}
+
```
elenatorro: We already have a `decode_optional` function that does the same on strokes.rs, we can unify this into utils.rs and reuse it

common/src/app/common/render_wasm/api/upload.cljs
superalex: Thanks!

elenatorro (approved):

Issue #11374, opened 2026-08-26
Add Content-Security-Policy and Strict-Transport-Security headers to the frontend image


bameda:
### Is your feature request related to a problem? Please describe.

The frontend image ships four security headers from `docker/images/files/nginx-security-headers.conf` (`X-Content-Type-Options`, `Referrer-Policy`, `Permissions-Policy`, `X-Frame-Options`) and does not include a Content Security Policy or `Strict-Transport-Security`. Administrators who want either one have to add it in front of the container and work out a suitable policy themselves, which for an application with a WebAssembly render engine, web workers and blob-based thumbnails is not obvious to derive from the outside.

### Describe the solution you'd like.

Ship both headers from the same file, so every deployment built from the official images inherits a sensible default rather than each installation reinventing one.

The CSP would ship in report-only mode, where browsers report violations to the developer console without blocking anything, making it safe to enable everywhere while the policy is refined against real usage. A reasonable starting policy is same-origin throughout plus `'wasm-unsafe-eval'` for the render engine, `'unsafe-inline'` in `style-src` for the inline style attributes the UI emits, `worker-src 'self'`, and `blob:` in `img-src` and `connect-src`.

`Strict-Transport-Security` only makes sense when the deployment is served over HTTPS, which the container can infer from `PENPOT_PUBLIC_URI`, so it can default to enabled for https deployments and be absent otherwise.

Both should be overridable through environment variables, so that deployments with different needs — plugins enabled, an intranet without TLS, or a stricter policy — can adjust without rebuilding the image.

### Describe alternatives you've considered.

**Shipping the CSP in enforcing mode from the start.** Two aspects of the current design would need decisions first, so this would either break deployments or force those decisions prematurely. `frontend/resources/templates/index.mustache` contains an inline `<script type="importmap">` and two inline `<script type="module">` blocks; nonces are not usable because nginx serves the built `index.html` statically, so these need build-time hashes emitted into the nginx configuration. Separately, the plugin runtime evaluates plugin code through a SES compartment (`compartment.evaluate` in `plugins/libs/plugins-runtime/src/lib/create-sandbox.ts`), which is how SES provides its hardened execution environment and which requires `'unsafe-eval'`; plugins also load from arbitrary hosts declared in their manifest and render their UI in an iframe pointing at that host, so `connect-src` and `frame-src` need to stay open where plugins are used.

[...]

PR #11463, opened 2026-09-01
:whale: Cover the inline scripts of the served pages with CSP hashes

bameda:
## What

The frontend build now emits the sha256 hashes of the inline scripts of the pages the container serves, and the entrypoint splices them into the default `script-src`. This makes `PENPOT_CSP_MODE=enforce` usable for the first time.

It also stops the external locations from passing through the security headers of their upstreams.

Follow-up to #11374, which added the headers themselves.

## Why

The policy added in #11374 permits `'self'` for scripts, which covers the external scripts of the served pages but not their inline ones. Every page the container serves carries inline scripts, so under report-only they are reported on every page load, and under enforcing the application would not load at all.

Hashes rather than nonces because nginx serves the built pages as static files: a nonce has to change per response, which would mean rewriting the HTML on the fly. The hashes are computed on the rendered output rather than on the mustache templates, since the digest covers the exact bytes served between the script tags.

## How

`generateCspHashes` runs in `frontend/scripts/_helpers.js` right after each page is written, accumulating into a set. All four pages under `resources/public/` contribute: `index.html`, `challenge.html`, `render.html` and `rasterizer.html`. The storybook previews are excluded because they go to `.storybook/` and that container does not serve them.

The build writes `csp-script-hashes.txt`, the image moves it out of the document root, and the entrypoint reads it. A bundle predating this change yields no hashes and the policy stays as it was, so older bundles keep building.

## The upstream headers part

While testing this I found that the three locations of `nginx-external-locations.conf` were passing through whatever security headers their upstreams send:

```bash
$ curl -s -D - -o /dev/null http://localhost:9001/github/penpot-files/README.md | grep -i "content-security|strict-transport" Content-Security-Policy: default-src 'none'; style-src 'unsafe-inline'; sandbox Strict-Transport-Security: max-age=31536000 Content-Security-Policy: default-src 'self'; base-uri 'self'; ... (ours)
```

Browsers enforce the intersection of every Content-Security-Policy they receive, so GitHub's `default-src 'none'; sandbox` takes precedence over ours on those responses. The HSTS one is worse: it lands on our own host, which means a deployment that deliberately sets `PENPOT_HSTS_VALUE` to empty gets HSTS set anyway, for a year, by a third party.

Three `proxy_hide_header` directives per location fix it. This extends to the security headers the same treatment the existing directives already give to CORS and caching headers.

## Testing

Built the image and ran the full stack with `PENPOT_CSP_MODE=enforce`, in a clean browser profile.

| Path | How it was exercised | Result |
|---|---|---|
| `index.html` | dashboard and workspace | loads, no violations |
| `challenge.html` | `/challenge.html?redirect=/dashboard` | redirects, no violations |
| `render.html` | PNG and SVG export | both complete, exporter logs clean |
| `rasterizer.html` | initialised by the frontend | `[app.rasterizer] initialized` |

The export one matters because the exporter loads `render.html` in a headless browser, so a failure there would not show up in the editor at all.

Also verified that every inline script of every served page has its hash emitted, that the six hashes reach the header on all nine routes, that `/csp-script-hashes.txt` is not web accessible, and that only our own CSP now reaches the proxied responses.

## What this does not do

`enforce` is still not the default. The plugin runtime calls `repairIntrinsics` as an import side effect in `plugins/libs/plugins-runtime/src/index.ts`, so SES initialises on every page load whether or not a plugin is opened, and needs `'unsafe-eval'`. Under the default policy those calls are blocked: the application still loads, but the plugin system is degraded. That is the one remaining blocker and it belongs to the plugin runtime rather than to packaging.

The admin console, OIDC single sign-on and the billing flow were checked and need no exception. The admin console loads its scripts from the same origin, and both single sign-on and billing navigate away from Penpot, which no directive of this policy governs. For OIDC in particular, discovery and the token exchange happen on the backend rather than in the browser.


frontend/scripts/_helpers.js
```diff
@@ -412,12 +413,37 @@ async function generateSvgSprites() {
   );
 }
 
+// Collect the CSP hashes of the inline scripts of a rendered template into
+// the given set. The hash covers the exact bytes between the script tags, so
+// it has to be computed on the rendered output and never on the mustache
+// source. Scripts carrying a src attribute are external and are covered by
+// 'self' instead.
+function collectCspHashes(html, hashes) {
+  const pattern = /<script\b(?![^>]*\bsrc=)[^>]*>([\s\S]*?)<\/script>/gi;
```
sancfc: The current regexp treats `data-src` as a real `src` attribute and does not detect `src = "..."`. The first case could leave an inline script
  without a CSP hash and cause it to be blocked under `enforce`.

  Could we use `\ssrc\s*=` in the negative lookahead?

  ```js
  const pattern = /<script\b(?![^>]*\ssrc\s*=)[^>]*>([\s\S]*?)<\/script>/gi;

bameda: Nice catch!

| Example | Wanted  | Current | @sancfc's proposal |
|---|---|---|---|
| `<script>A</script>` | inline | ok | ok |
| `<script type="module">A</script>` | inline | ok | ok |
| `<script type="importmap">…</script>` | inline | ok | ok |
| `<script src="/x.js">` | externo | ok | ok |
| `<script src = "/x.js">` | externo | **fail** | ok |
| `<script type="module" src="/x.js">` | externo | ok | ok |
| `<script\n  src="/x.js">` | externo | ok | ok |
| `<script SRC="/x.js">` | externo | ok | ok |
| `<script data-src="/x.js">A</script>` | inline | **fail** | ok |
| `<script foo-src="/x.js">A</script>` | inline | **fail** | ok |

I've already apply your proposal, @sancfc.

PR #11467, opened 2026-09-02
:bug: Export multiple solid fills to WASM SVG

### LINKED ISSUE #11466: Export multiple fills to SVG
### Description

WASM SVG export renders shapes with more than one fill as a black shape (or with no `fill` attribute). The exporter composes multiple fills through `merge_fills`, which builds a Skia shader; `SkSVGDevice` does not serialize shader paints to native SVG (`fill`, `fill-opacity`, or `<linearGradient>` / `<radialGradient>`).

Single solid fills work because Skia writes a plain `fill="#…"` attribute. Multiple solid fills, gradients, and mixed fill stacks need one geometry draw per fill (matching the classic SVG exporter and the planned image-fill path in #11383).

### Use case

Shapes with stacked fills (e.g. a semi-transparent red under a solid blue) should export with the same appearance as on the canvas. Gradients should emit native SVG gradient definitions instead of being dropped.

### Affected versions

develop / `wasm-export` with `render-wasm/v1`

Relates to #10546

superalex:
## What

- Shapes with multiple solid fills (rect, circle, path, bool, frame background) export with correct `fill` / `fill-opacity` attributes instead of rendering black.
- Text with multiple solid fills exports one `<text>` element per fill layer, stacked in the right order.

Closes #11466

## Why

`merge_fills` composes stacked fills into a single Skia shader. `SkSVGDevice` does not serialize shader paints to SVG attributes, so exported geometry had no `fill` and browsers defaulted to black.

## How

- **Vector shapes:** add `compose_fills` to `VectorRenderer`. SVG export (`compose_fills: false`) draws each fill with `fill.to_paint()` in reverse order (bottom layer first). PDF keeps the merged-shader path.
- **Text:** paint one paragraph pass per fill layer via `to_style_for_fill_layer`, reusing the same bottom-to-top ordering as the classic ClojureScript exporter.
- **Tests:** insta snapshots for rect and text with two solid fills (red 50% over blue 100%).

Image fills are out of scope for this PR.

### DIFF
--- a/render-wasm/src/render/vector.rs
+++ b/render-wasm/src/render/vector.rs
@@ -22,14 +22,24 @@ pub(super) struct VectorRenderer<'a> {
     canvas: &'a Canvas,
     shared: &'a mut RenderResources,
     scale: f32,
+    /// When `true`, multiple fills are composited into a single shader (PDF).
+    /// When `false`, each fill is drawn separately so SkSVGDevice can emit
+    /// `fill` attributes (SVG export).
+    compose_fills: bool,
 }
 
 impl<'a> VectorRenderer<'a> {
-    pub fn new(canvas: &'a Canvas, shared: &'a mut RenderResources, scale: f32) -> Self {
+    pub fn new(
+        canvas: &'a Canvas,
+        shared: &'a mut RenderResources,
+        scale: f32,
+        compose_fills: bool,
+    ) -> Self {
         Self {
             canvas,
             shared,
             scale,
+            compose_fills,
         }
     }
 }
@@ -40,9 +50,9 @@ impl ShapeRenderer for VectorRenderer<'_> {
             return Ok(());
         }
 
-        // Handle image fills individually
         let has_image_fills = fills.iter().any(|f| matches!(f, Fill::Image(_)));
-        if has_image_fills {
+        if !self.compose_fills || has_image_fills {
+            // fills[0] is the topmost layer; draw bottom → top (matches GPU + classic SVG).
             for fill in fills.iter().rev() {
                 match fill {
                     Fill::Image(image_fill) => {
@@ -845,7 +855,7 @@ fn render_frame(
     if !element.fills.is_empty() {
         canvas.save();
         canvas.concat(&matrix);
-        let mut renderer = VectorRenderer::new(canvas, shared, scale);
+        let mut renderer = VectorRenderer::new(canvas, shared, scale, true);
         renderer.draw_fills(element, &element.fills)?;
         renderer.draw_fill_inner_shadows(element)?;
         canvas.restore();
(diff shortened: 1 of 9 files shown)


render-wasm/src/render/vector.rs
```diff
@@ -40,6 +50,25 @@ impl ShapeRenderer for VectorRenderer<'_> {
...
+                        if let Some(filter) = shape.image_filter(1.) {
+                            paint.set_image_filter(filter);
+                        }
+                        draw_shape_geometry(self.canvas, shape, &paint);
+                    }
+                }
+            }
+            return Ok(());
+        }
+
```
elenatorro: This is the very same code as the one in lines 71-85 (!self.compose_fills check), can we refactor this? maybe checking the condition is enough:

```
  if !self.compose_fills || has_image_fills {
     // dup code here
      return Ok(());
  }
```

render-wasm/src/shapes/text.rs
```diff
@@ -716,6 +716,60 @@ impl TextContent {
...
+    /// Builds paragraph builders that paint a single fill layer per span for SVG
+    /// export. `layer_from_bottom` is 0 for the bottommost fill (fills[last]).
+    pub fn paragraph_builder_group_for_fill_layer(
+        &self,
+        layer_from_bottom: usize,
+    ) -> Vec<ParagraphBuilderGroup> {
+        self.paragraph_builders_for_fill_layer(layer_from_bottom)
+    }
+
+    fn paragraph_builders_for_fill_layer(
```
elenatorro: This method is pretty much the same as `paragraph_builders`, we can reuse it by adding a "fill_layer" parameter maybe, what do you think?

```rs
    fn paragraph_builders(
        &self,
        use_shadow: Option<bool>,
        opaque: bool,
        align_override: Option<skia::textlayout::TextAlign>,
        fill_layer: Option<usize>,
    ) -> Vec<ParagraphBuilderGroup> {
        let fonts = get_font_collection();
        let fallback_fonts = get_fallback_fonts();
        let mut paragraph_group = Vec::new();

        for paragraph in self.paragraphs() {
            let mut paragraph_style = paragraph.paragraph_to_style();
            if let Some(align) = align_override {
                paragraph_style.set_text_align(align);
            }
            let mut builder = ParagraphBuilder::new(&paragraph_style, fonts);
            let mut has_text = false;
            for span in paragraph.children() {
                let remove_alpha =
                    opaque || (use_shadow.unwrap_or(false) && !span.is_transparent());
                let text_style = span.to_style_with_paint(
                    &self.bounds(),
                    fallback_fonts,
                    remove_alpha,
                    paragraph.line_height(),
                    fill_layer,
                );
                let text: String = span.apply_text_transform();
                if !text.is_empty() {
                    has_text = true;
                }
                builder.push_style(&text_style);
                add_text_with_tabs(&mut builder, &text, span.font_size);
            }
            if !has_text {
                builder.add_text(" ");
            }
            paragraph_group.push(vec![builder]);
        }

        paragraph_group
    }
```

calling it like: 

```
self.paragraph_builders(None, false, None, Some(layer_from_bottom))
```

And I think we might not need `to_style_for_fill_layer` then.

elenatorro (changes requested):

render-wasm/src/render/vector.rs
superalex: Ups!

render-wasm/src/shapes/text.rs
superalex: Thank you very much!

elenatorro (approved):

PR #11471, opened 2026-09-02
:zap: Reduce WASM text selrect churn after batch page load

superalex:
## What

- Fix WASM auto-width text dimensions so they match the HTML renderer (1px paragraph margin).
- Run post-font text relayout only after all pending image/font fetches finish, not on every stream emission.
- Scope forced relayout and stale-selrect sync to text shapes that actually use not-yet-ready font faces.
- Defer stale-selrect resize until after the first tiles-complete pass, scheduled on idle.
- Chunk stale text resize into batches of 64 with short delays on large pages.
- Coalesce `request-render` after incremental WASM attr updates into one call per batch.

## Why

After #11443 landed batch structural upload, page switches on text-heavy files still hit main-thread stalls and incorrect auto-width selrects. WASM `longest_line()` ignored the 1px right margin the HTML renderer applies, so `text-selrect-stale?` fired and triggered resize work too early and too broadly. Relayout ran on every pending fetch emission and for every text shape once any font was pending.

## How

- **Rust:** add `PARAGRAPH_SET_MARGIN_RIGHT` to auto-width size reporting in `text.rs`.
- **Frontend:** track font-face keys per shape at upload; build an index of shapes per pending face for targeted relayout.
- **Frontend:** move `relayout-after-fonts!` to the `on-complete` path of pending fetches; defer stale-selrect sync until tiles render + idle; partition resize into timed chunks in `workspace.cljs`.
- **Frontend:** batch `request-render` in `process-shape-changes!` via `rx/reduce`.

### DIFF
--- a/frontend/src/app/render_wasm/api.cljs
+++ b/frontend/src/app/render_wasm/api.cljs
@@ -1522,6 +1522,60 @@
   [content]
   (or content (tc/v2-default-text-content)))
 
+(def ^:private empty-text-font-state
+  {:font-index {} :pending-faces #{}})
+
+(defn- text-layout-fonts
+  "Content and fallback faces (emoji, Noto, ...) that `set-shape-text-content`
+  uploads, must match that path for pending-face tracking."
+  [content]
+  (into #{} (concat (f/get-content-fonts content)
+                    (fonts-from-text-content content false))))
+
+(defn- text-font-face-keys-state
+  "All font-face keys for a text content, and the subset not WASM-ready yet."
+  [content]
+  (reduce
+   (fn [acc font]
+     (let [font-data (f/make-font-data font)
+           key       (f/font-data-key font-data)
+           pending?  (not (f/font-ready? font-data))]
+       (-> acc
+           (update :font-face-keys conj key)
+           (cond-> pending?
+             (update :pending-font-face-keys conj key)))))
+   {:font-face-keys #{} :pending-font-face-keys #{}}
+   (text-layout-fonts content)))
+
+(defn- acc-text-font-state
+  [{:keys [font-index pending-faces]} id font-face-keys pending-font-face-keys]
+  {:font-index (reduce (fn [idx face]
+                         (update idx face (fnil conj #{}) id))
+                       font-index
+                       (or font-face-keys #{}))
+   :pending-faces (into (or pending-faces #{})
+                        (or pending-font-face-keys #{}))})
+
+(defn text-font-state-for-shape
+  "Build the font-face index for a single text shape (incremental updates)."
+  [shape]
+  (if (cfh/text-shape? shape)
+    (let [content (ensure-text-content (:content shape))
+          {:keys [font-face-keys pending-font-face-keys]}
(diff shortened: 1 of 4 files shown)


frontend/src/app/main/data/workspace.cljs
```diff
@@ -507,11 +507,20 @@
...
-                              (dwwt/resize-wasm-text-all ids))))
+                    (rx/mapcat
+                     (fn [{:keys [ids]}]
+                       (->> (partition-all 64 ids)
+                            (map-indexed
+                             (fn [idx chunk]
+                               (if (zero? idx)
+                                 (rx/of (dwwt/resize-wasm-text-all chunk))
+                                 (->> (rx/of (dwwt/resize-wasm-text-all chunk))
+                                      (rx/delay (* idx 50))))))))))
```
elenatorro: mapcat needs an observable, not a list of them, so given a list it emits the observable objects as values, potok drops them silently, and the chunks never run. We can use rx/merge, that returns a single observable

```suggestion
                       (apply rx/merge
                              (map-indexed
                               (fn [idx chunk]
                                 (cond->> (rx/of (dwwt/resize-wasm-text-all chunk))
                                   (pos? idx) (rx/delay (* idx 50))))
                               (partition-all 64 ids))))))
```

frontend/src/app/render_wasm/api.cljs
```diff
@@ -1610,6 +1666,33 @@
...
+(defonce ^:private stale-selrect-sync-token (atom 0))
+
+(defn- flush-stale-selrect-sync!
+  []
+  (let [ids (swap! pending-stale-selrect-ids
+                   (fn [pending]
+                     (reset! pending-stale-selrect-ids #{})
+                     pending))]
+    (when (seq ids)
+      (st/emit! (ptk/data-event ::stale-text-selrects {:ids (vec ids)})))))
```
elenatorro: with this, every flush emits again all previous ids, including shapes from pages already visited, so we can read and clear the pending ids with reset-vals! instead of a reset! inside a swap!, because swap! puts back the inner reset

we can do this instead:

```suggestion
  (when-let [ids (seq (first (reset-vals! pending-stale-selrect-ids #{})))]
    (st/emit! (ptk/data-event ::stale-text-selrects {:ids (vec ids)}))))
```

frontend/src/app/render_wasm/api.cljs
```diff
@@ -1561,6 +1613,8 @@
           (set-layout-data shape)))
 
     (let [is-text? (= type :text)
+          font-face-keys (when is-text? (text-font-face-keys content))
+          pending-font-face-keys (when is-text? (text-pending-font-face-keys content))
```
elenatorro: since we have two different methods for this, this means that we're calling `text-layout-fonts` twice, scanning fonts etc. Maybe we can do this calculation in a single method that returns both `font-face-keys` and `pending-font-face-keys`, so we don't need to scan content twice, what do you think? (I might be missing something here)

PR #11474, opened 2026-09-02
:bug: Enforce SSRF checks and add timeouts to HTTP client

pabloalba:
### Related Ticket

https://tree.taiga.io/project/penpot-nitrate/issue/26776

### Summary
* Removed skip-ssrf for nitrate sso
* Added request-timeout and pool limit to http client

### Checklist

### DIFF
--- a/backend/src/app/http/client.clj
+++ b/backend/src/app/http/client.clj
@@ -34,7 +34,9 @@
 
 (defmethod ig/init-key ::client
   [_ _]
-  (http/build-client {:connect-timeout 30000
+  (http/build-client {:connect-timeout 30000 ;; max connection time
+                      :request-timeout 30000 ;; max read time
+                      :executor (java.util.concurrent.Executors/newFixedThreadPool 32) ;; limit pool
                       :follow-redirects :never}))
 
 (defn send!
--- a/backend/src/app/http/client.clj
+++ b/backend/src/app/http/client.clj
@@ -17,12 +17,17 @@
    [app.util.ssrf :as ssrf]
    [cuerdas.core :as str]
    [integrant.core :as ig]
-   [java-http-clj.core :as http])
+   [java-http-clj.core :as http]
+   [promesa.exec :as px])
   (:import
    java.net.http.HttpClient
    java.net.URI))
 
 (def default-max-redirects 5)
+(def default-request-timeout 30000)
+(def default-executor-parallelism 32)
+
+(defonce ^:private executor (atom nil))
 
 (defn client?
   [o]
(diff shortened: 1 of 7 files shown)


niwinz:
### 1. Must fix: media-processor timeout silently drops from 120s to 30s

`app/media/remote.clj` passes `:timeout` in the options map of `http/req`, but that map only carries `:response-type` into `send!` — the value never reaches `http/send` (in java-http-clj, `:timeout` is a request-map key; send options only accept `:as`, `:client`, `:raw?`).

Before this PR the dropped value meant no timeout; now the 30s default applies instead of the value computed from `PENPOT_MEDIA_PROCESSING_SERVICE_TIMEOUT` (default 120000). The media processor does synchronous processing and only responds when done, so any job taking 30–120s before responding now fails with `:media-processor-timeout`.

**Before** (`app/media/remote.clj`):

```clojure
(let [resp (http/req client
                     {:method method
                      :uri uri
                      :body body
                      :headers headers}
                     {:response-type :input-stream
                      :skip-ssrf-check? true
                      :timeout timeout})
```

**After** — move the timeout into the request map:

```clojure
(let [resp (http/req client
                     {:method method
                      :uri uri
                      :body body
                      :headers headers
                      :timeout timeout}
                     {:response-type :input-stream
                      :skip-ssrf-check? true})
```

The request-map `:timeout` then overrides the 30s client default, because the default is the merge base in `send!`.

### 2. Executor: reuse `::wrk/executor` instead of a dedicated pool

Pass the system executor as an integrant ref, the same way `msgbus`, `rpc/climit` or `storage/tmp` already do. Lifecycle is already handled by `app.worker.executor`, so the component loses the `defonce` atom, the `halt-key!` and the parallelism constant. Parallelism is not important here: `send!` blocks on the calling thread, the executor only serves HttpClient internals, and with the new 30s request timeout hung tasks no longer accumulate.

**Before** (`app/http/client.clj`):

```clojure
(def default-max-redirects 5)
(def default-request-timeout 30000)
(def default-executor-parallelism 32)

(defonce ^:private executor (atom nil))

(defmethod ig/init-key ::client
  [_ _]
  (let [exec (px/fixed-executor :parallelism default-executor-parallelism
                                :prefix "penpot/http-client/")]
    (reset! executor exec)
    (http/build-client {:connect-timeout default-request-timeout
                        :executor exec
                        :follow-redirects :never})))

(defmethod ig/halt-key! ::client
  [_ _]
  (when-let [exec @executor]
    (px/shutdown! exec)
    (reset! executor nil)))
```

**After** (`app/http/client.clj`) — the whole executor block collapses to:

```clojure
(def default-max-redirects 5)
(def default-request-timeout 30000)

(defmethod ig/init-key ::client
  [_ {:keys [::wrk/executor]}]
  (http/build-client {:connect-timeout default-request-timeout
                      :executor executor
                      :follow-redirects :never}))
```

Namespace changes that come with it: drop `[promesa.exec :as px]` and add `[app.worker :as-alias wrk]` (the same idiom `rpc/climit.clj` uses to destructure the key).

**Before** (`app/main.clj`):

```clojure
::http.client/client
{}
```

**After** (`app/main.clj`):

```clojure
::http.client/client
{::wrk/executor (ig/ref ::wrk/executor)}
```

`app.main` already requires `[app.worker.executor]`, so no other wiring changes are needed.

### 3. SSRF: operator-facing note

Removing the nitrate SSRF skip is a breaking config change for self-hosted installs whose `admin-console-uri` resolves to a private or loopback address: `ssrf/validate-uri` rejects those targets, and the only escape is `PENPOT_SSRF_ALLOWED_HOSTS`. Since `add-nitrate-licence-to-profile` rejects requests when nitrate is unreachable, affected installs fail hard on licensed features.

Suggested text for the PR description and `CHANGES.md`:

> If `PENPOT_ADMIN_CONSOLE_URI` points to an internal address (private IP or internal hostname), add the host to `PENPOT_SSRF_ALLOWED_HOSTS` (space-separated, e.g. `PENPOT_SSRF_ALLOWED_HOSTS: "internal-admin.example.internal"`). Without it, admin-console requests fail with `ssrf-blocked-target`.

### 4. Minor

- `default-request-timeout` is used in two places with two different meanings: as `:connect-timeout` in `build-client` (time to open the connection) and as `:timeout` in `send!` (time until the response arrives). The value works for both today, but two constants (`default-connect-timeout`, `default-request-timeout`) would let each knob be tuned independently and make each call site read honestly.

### 5. Tests

CI is green, but the nitrate tests mock `nitrate/call`, so none of the new behavior runs anywhere. Worth adding:

- `send!` injects the 30s default when the request has no `:timeout`, and preserves a caller-supplied value.
- `media/remote` puts the configured timeout in the request map.
- nitrate raises `:ssrf-blocked-target` for a private `admin-console-uri`, and proceeds once the host is allowlisted.

niwinz (approved):

PR #11487, opened 2026-09-03
:bug: Fix typography sample overflow in non-Latin locales

### LINKED ISSUE #11486: Typography sample overflows its row in locales with a wider sample (ko, ar)
### Steps To Reproduce

1. Set the UI language to **Korean** (Profile → Settings → Language), or any locale whose typography sample is wider than two Latin glyphs.
2. Open a file that has typography assets, and open the **Assets** tab in the left sidebar.
3. Expand **Typographies**.

The same happens on the dashboard: open a team whose project has a shared library containing typographies, and look at the library card in the grid.

### Expected behavior

The sample glyph next to each typography name renders on a single line inside its row, the way `Ag` does in English.

### Actual behavior

The sample is stacked vertically, one character per line, and overflows the row into the rows above and below.

`workspace.assets.typography.sample` is `"Ag"` in `en.po`, but `"가나다"` in `ko.po` and `"أسلوب خط النص"` in `ar.po`. The element it renders into is sized for two narrow glyphs:

```scss
// frontend/src/app/main/ui/workspace/sidebar/options/menus/typography.scss
.typography-selection-wrapper {
  grid-template-columns: $sz-24 auto 1fr;   // the sample gets a fixed 24px column
}
.typography-sample {
  min-inline-size: $sz-24;
  font-size: px2rem(16);
  block-size: $sz-32;                        // no white-space: nowrap
}
```

Korean text breaks between syllables, so `가나다` (≈48px at 16px) wraps to three lines inside a 24×32px box.

Measured on the shipped 2.17.2 stylesheet, with the real component markup:

| | box | content height |
|---|---|---|
| `Ag` | 24 × 32 | 32 (1 line) |
| `가나다` | 24 × 32 | **50 (overflows by 18px)** |
| dashboard card, `가나다` | 20 × 20 | **43 (overflows by 23px)** |

Three places are affected, all in the same way:

- `.typography-sample` in the assets list row (`.typography-selection-wrapper`)
- `.typography-sample` in the detail panel (`.typography-name-wrapper`)
- `.typography-sample-input` in the rename row (`.font-name-wrapper`), which uses a hard `inline-size: $sz-24`
- `.library-typography-sample` in the dashboard library card (`frontend/src/app/main/ui/dashboard/grid.scss`), `inline-size: px2rem(20)`

### Screenshots or video

Assets panel — left column is current `develop`, right column is with the fix:

assets panel

Dashboard library card:

dashboard library card

### Desktop

- OS: macOS 26.6.2
- Browser & version: Chromium 148.0.7778.96

### Environment

- Host: local instance (docker compose, `penpotapp/frontend:2.17.2`)
- Image version: 2.17.2
- Also present on `develop` — `ko.po` still carries `"가나다"` and the SCSS above is unchanged.

### Notes

I'd rather not fix this by shortening the translations: they are managed on [Weblate](https://hosted.weblate.org/projects/penpot/frontend/ko/), so an edit to `ko.po` in this repo would be overwritten on the next sync, and any future locale could reintroduce the same overflow. A PR making the sample column resilient to any sample string follows.

desty:
## What and why

Closes #11486.

The typography sample glyph is laid out for the English sample `Ag`: it lives in a fixed `$sz-24` grid column and is free to wrap. Locales that translate `workspace.assets.typography.sample` to something wider therefore stack it and overflow the row — `ko` is `"가나다"`, `ar` is `"أسلوب خط النص"`.

This makes the sample column size to its content and keeps the sample on one line:

- `.typography-selection-wrapper` / `.typography-name-wrapper` — the sample track becomes `minmax($sz-24, auto)`
- `.typography-sample` (both rules) — `white-space: nowrap`
- `.typography-sample-input` — hard `inline-size: $sz-24` becomes `min-inline-size: $sz-24`, plus `nowrap`
- `.library-typography-sample` (dashboard library card) — same change, `inline-size: px2rem(20)` becomes `min-inline-size`

The `$sz-24` / `px2rem(20)` floor is kept, so a sample that already fits is unaffected.

I deliberately did not shorten the `ko` / `ar` strings: translations come from [Weblate](https://hosted.weblate.org/projects/penpot/frontend/ko/), so a `.po` edit here would be overwritten on the next sync and would not stop a future locale from reintroducing the overflow.

## Screenshots

Left column is current `develop`, right column is this branch. Rendered with the shipped `main.css` from a 2.17.2 instance and the markup from `typography.cljs` / `grid.cljs`, at the three sample strings that exist today.

Assets panel — list row, detail panel and rename row:

assets panel

Dashboard library card:

dashboard library card

## Testing notes

Measured the rendered boxes in Chromium 148 before and after the change:

| case | before (box / content height) | after |
|---|---|---|
| `Ag`, assets row | 24 × 32 / 32 | 24 × 32 / 32 |
| `가나다`, assets row | 24 × 32 / **50** | 41.5 × 32 / 32 |
| `أسلوب خط النص`, assets row | 24 × 32 / **50** | 94.7 × 32 / 32 |
| `가나다`, dashboard card | 20 × 20 / **43** | 31.1 × 20 / 20 |

No regression for locales that keep `Ag`: the sample box stays 24px wide and the name column still starts at the same offset (36px from the row edge) before and after.

`pnpm run lint:scss` and `pnpm run check-fmt:scss` pass on both changed files.

**Review follow-up (ad6b1b0):** letting the sample grow removed the fixed-width slack that separated it from the name, so:

- `.library-name-block` (dashboard card) drops the hard-coded `calc(...)` width for `flex: 1; min-inline-size: 0`, and the sample gets `flex-shrink: 0`
- `.typography-selection-wrapper`, `.typography-name-wrapper` get `column-gap: var(--sp-xs)`; `.font-name-wrapper` gets `gap: var(--sp-xs)` and its input can shrink
- the Libraries "Updates" tab sample (`libraries.cljs`) gets a `nowrap` class. It was the one spot the first commit missed.

Details and screenshots: https://github.com/penpot/penpot/pull/11487#issuecomment-5740165591

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change

No breaking changes; CSS plus one added class in `libraries.cljs`, no translation changes.

### DIFF
--- a/frontend/src/app/main/ui/dashboard/grid.scss
+++ b/frontend/src/app/main/ui/dashboard/grid.scss
@@ -421,7 +421,8 @@ $thumbnail-default-height: px2rem(168);
 .library-typography-sample {
   block-size: px2rem(20);
   margin-inline-end: var(--sp-xs);
-  inline-size: px2rem(20);
+  min-inline-size: px2rem(20);
+  white-space: nowrap;
 }
 
 // ─── MISC ──────────────────────────────────────
--- a/frontend/src/app/main/ui/workspace/sidebar/options/menus/typography.scss
+++ b/frontend/src/app/main/ui/workspace/sidebar/options/menus/typography.scss
@@ -61,7 +61,7 @@ $font-preview-box-height: 28px;
 
 .typography-selection-wrapper {
   display: grid;
-  grid-template-columns: $sz-24 auto 1fr;
+  grid-template-columns: minmax($sz-24, auto) auto 1fr;
   flex: 1;
   block-size: 100%;
   inline-size: 100%;
@@ -78,6 +78,7 @@ $font-preview-box-height: 28px;
   align-items: center;
   min-inline-size: $sz-24;
   block-size: $sz-32;
+  white-space: nowrap;
   color: var(--color-foreground-secondary);
 }
 
@@ -116,9 +117,10 @@ $font-preview-box-height: 28px;
   display: flex;
   justify-content: center;
   align-items: center;
-  inline-size: $sz-24;
+  min-inline-size: $sz-24;
   block-size: 100%;
   font-size: px2rem(16);
+  white-space: nowrap;
   color: var(--color-foreground-primary);
 }
 
@@ -156,7 +158,7 @@ $font-preview-box-height: 28px;
   @extend %asset-element;
 
   display: grid;
-  grid-template-columns: $sz-24 auto 1fr $sz-28;
+  grid-template-columns: minmax($sz-24, auto) auto 1fr $sz-28;
   flex: 1;
   block-size: $sz-32;
   inline-size: 100%;
@@ -172,6 +174,7 @@ $font-preview-box-height: 28px;
   min-inline-size: $sz-24;
   font-size: px2rem(16);
   block-size: $sz-32;
+  white-space: nowrap;
   padding: 0;
   color: var(--color-foreground-primary);
 }


desty:
@EvaMarco thanks for the thorough review and the screenshots. All four spots are addressed in ad6b1b0:

1. **Dashboard library card**: `.library-name-block` no longer uses the hard-coded `calc(100% - var(--sp-xxl) - var(--sp-s))`. It is now `flex: 1; min-inline-size: 0`, so it takes whatever space the sample leaves and the name still ellipsizes. `.library-typography-sample` gets `flex-shrink: 0`, so it keeps its `var(--sp-xs)` margin instead of being squeezed underneath the name. The colour/component rows that share the class already sized from the same flex/grid parent, so they render the same.
2. **Sidebar typography detail view**: `.typography-name-wrapper` gets `column-gap: var(--sp-xs)`.
3. **Rename / advanced-edit row**: `.font-name-wrapper` gets `gap: var(--sp-xs)`, `.typography-sample-input` gets `flex-shrink: 0`, and `.adv-typography-name` gets `min-inline-size: 0`. The input now shrinks instead of pushing into the sample or the action buttons.
4. **Libraries "Updates" tab**: the sample div now has a `typography-sample` class (`white-space: nowrap`) in `libraries.scss`.

I also added `column-gap: var(--sp-xs)` to `.typography-selection-wrapper` as you suggested, so the gap is explicit rather than coming from `.name-block`'s padding.

Before/after, rendered in Chrome from a static harness that copies these rules and the markup structure (not a full Penpot build), with the three sample strings shipped today:

review follow-up

For `Ag` locales the only visible change is the new 4px gap in the sidebar rows. `prettier -c` and `stylelint` pass on the three changed SCSS files.

On the Arabic sample string: agreed, I'll report that on Weblate separately rather than touch it here.

EvaMarco:
Thanks @desty we will merge this in our next development cicle.

PR #11520, opened 2026-09-05
:bug: Fix MCP plugin reconnect loop on duplicate connection

### LINKED ISSUE #11510: MCP keeps trying to reconnect endlessly in inactive tab eventually freezing the browser
When two tabs of the same Penpot environment are open and the MCP is connected in one of them (green dot), reloading the other tab, where the MCP is not actively connected (grey dot), triggers an infinite loop of reconnection attempts/messages in the console for that inactive tab. If left open long enough, this ends up freezing/crashing the browser, forcing the user to close and reopen it. The issue is easily reproducible and has been observed in both design.penpot.dev and tokens.penpot.dev

Image

### Steps To Reproduce

- Open two tabs pointing to the same Penpot environment (observed in tokens.penpot.dev and pre).
- Connect the MCP in one of the tabs (its status indicator turns green).
- In the second tab, where the MCP is not actively connected (status indicator grey), reload the page (F5).
- Observe the browser console: it starts continuously printing connection/disconnection attempt messages, without stopping.
- Leave that second tab open for a while without interacting with it.

### Expected behavior

The MCP connection/reconnection logic should not enter an infinite retry loop when a duplicate/inactive tab of the same environment is reloaded. The inactive tab should either not attempt to reconnect repeatedly, or should fail gracefully without consuming increasing resources, and it should never lead to the browser becoming

### Actual behavior

The inactive tab (grey dot) continuously attempts to reconnect to the MCP after a page reload, flooding the console with connection/disconnection messages.
If the tab is left open long enough in this state, the browser ends up freezing ("se tuesta") and needs to be force-closed and reopened.
Switching to the active tab and clicking "Connect here" stops the loop, but the behavior recurs whenever this tab configuration happens again.

Image

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

breken-ai:
Fixes #11510

**Problem**
With MCP connected in one tab, reloading a second tab triggers an endless connect/disconnect loop in the reloaded tab (~1 reconnect per second, console flood, eventually freezes the browser).

**Root cause**
- The MCP server (`mcp/packages/server/src/PluginBridge.ts:145`) rejects a second plugin connection for the same user token with `ws.close(1008, "Duplicate connection for given user token; close previous connection first.")`.
- The plugin (`mcp/packages/plugin/src/main.ts`) treats **every** close as retryable: `ws.onclose` unconditionally calls `scheduleReconnect()`, and each brief `onopen` resets the backoff via `cancelReconnect()`. So a rejected tab reconnects at ~1s intervals forever and can never win - the other tab already holds the token.
- The app-level coordination (`mcp.cljs` force-disconnect broadcast) never engages for the reloaded tab, because its connection is rejected before it can report `connected` and broadcast.

**Fix**
In the plugin's `ws.onclose`, treat close code `1008` (policy violation - used by the server for duplicate-token and missing-token rejections) as terminal: stop auto-reconnecting and stay disconnected until the user explicitly reconnects. Non-1008 closes keep the existing capped-backoff retry behavior, and the close reason is surfaced in the status label as before.

**Expected after fix**: the reloaded tab fails once, shows the duplicate-connection reason, and stays grey. No retry storm.

Note: `main.ts` is not covered by the plugin package's existing node:test suite (only ErrorUtils/PenpotUtils are); the change is a small guard in the close handler, verified by type-checking (only pre-existing build-define errors for `style.css`/`PENPOT_MCP_WEBSOCKET_URL`).

> Built by breken, your AI support engineer - breken.ai - this one's on us.

### DIFF
--- a/mcp/packages/plugin/src/main.ts
+++ b/mcp/packages/plugin/src/main.ts
@@ -255,6 +255,15 @@ function connectToMcpServer(baseUrl?: string, token?: string): void {
                 updateCurrentTask(null);
             }
             ws = null;
+            if (event.code === 1008) {
+                // Policy violation (e.g. duplicate connection for the same user
+                // token - another tab already holds the connection). Retrying
+                // would be refused again immediately, so stay disconnected
+                // until the user explicitly reconnects.
+                shouldReconnect = false;
+                cancelReconnect();
+                return;
+            }
             scheduleReconnect();
         };
 


niwinz:
The fix looks nice, but it misses tests and documentation. I have addressed this on this PR: https://github.com/penpot/penpot/pull/11906

niwinz:
I merging this as is and the, rebase the other. Thanks

PR #11552, opened 2026-09-08
:bug: Fix missing text in legacy SVG board thumbnails

### LINKED ISSUE #10182: Text only show on hover
### Steps To Reproduce

1. Open attached file
2. Hover over the boards

[Bug.zip](https://github.com/user-attachments/files/28950150/Bug.zip)

### Expected behavior

You should see all the texts without hovering

### Actual behavior

You only can see one text at the time, hovering over it

### Screenshots or video

_No response_

_No response_

_No response_

PRO

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

EvaMarco:
This bug only happens with the old render.

Board thumbnails rendered by frame-imposter used React's renderToStaticMarkup, a synchronous pass with no live DOM to measure text against. Text shapes without a persisted position-data value therefore rendered as nothing, so a cached board thumbnail silently lost its text until the board was hovered, selected, or the canvas was zoomed past 130%, all of which bypass the cached thumbnail in favor of live content.

frame-imposter now provides the same is-render? context the standalone exporter already sets, so text without position-data falls back to the synchronous foreignObject renderer instead of rendering nothing.

Thumbnails cached before this fix stay broken until something regenerates them, so on each page load, board thumbnails containing text are opportunistically regenerated once per browser (tracked via local-storage) so existing files self-heal without requiring an edit.

AI-assisted-by: claude-sonnet-5

### Related Ticket

This PR closes https://github.com/penpot/penpot/issues/10182

### Summary

Self healing thumbnails

[Screencast from 2026-09-08 11-50-25.webm](https://github.com/user-attachments/assets/50e1abbb-c966-4aba-b9a7-cb8a5fed51a9)

### Checklist

### DIFF
--- a/frontend/src/app/main/data/workspace/thumbnails.cljs
+++ b/frontend/src/app/main/data/workspace/thumbnails.cljs
@@ -12,6 +12,7 @@
    [app.common.thumbnails :as thc]
    [app.common.time :as ct]
    [app.common.types.component :as ctc]
+   [app.common.types.shape-tree :as ctt]
    [app.common.uuid :as uuid]
    [app.main.data.changes :as dch]
    [app.main.data.helpers :as dsh]
@@ -23,6 +24,7 @@
    [app.main.render :as render]
    [app.main.repo :as rp]
    [app.util.queue :as q]
+   [app.util.storage :as storage]
    [app.util.timers :as tm]
    [app.util.webapi :as wapi]
    [beicon.v2.core :as rx]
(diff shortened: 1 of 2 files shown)


frontend/src/app/main/data/workspace/thumbnails.cljs
```diff
@@ -289,13 +291,55 @@
...
+                      :objects)
+        frame-ids (ctt/get-root-frames-ids objects)]
+    (->> (rx/from frame-ids)
+         (rx/filter #(frame-has-text? objects %))
+         (rx/map (fn [frame-id] [frame-id (thc/fmt-object-id file-id page-id frame-id "frame")]))
+         (rx/filter (fn [[_ object-id]] (unhealed-text-thumbnail? state object-id)))
+         (rx/tap (fn [[_ object-id]] (mark-thumbnail-healed! object-id)))
+         (rx/map (fn [[frame-id _]]
+                   (update-thumbnail file-id page-id frame-id "frame" "heal-stale-text-thumbnails"))))))
+
```
niwinz, line 340: This function can be defined using transducers instead of rx stream chain. Example:

```clojure
(defn- heal-stale-text-thumbnails
  "Emits an `update-thumbnail` for every board on the page that has text
  content and hasn't already been healed (see `healed-storage-key`) in this
  browser."
  [state file-id page-id]
  (let [objects   (-> (dsh/lookup-file-data state file-id)
                      (dsh/get-page page-id)
                      :objects)
        frame-ids (ctt/get-root-frames-ids objects)
        xf        (comp
                    (filter #(frame-has-text? objects %))
                    (keep
                      (fn [frame-id]
                        (let [object-id (thc/fmt-object-id
                                          file-id page-id frame-id "frame")]
                          (when (unhealed-text-thumbnail? state object-id)
                            [frame-id object-id])))))]
    (->> (rx/from (eduction xf frame-ids))
         (rx/map
           (fn [[frame-id object-id]]
             (mark-thumbnail-healed! object-id)
             (update-thumbnail
               file-id
               page-id
               frame-id
               "frame"
               "heal-stale-text-thumbnails"))))))
```

PR #11560, opened 2026-09-08
:bug: Fix problems with pixel precission in WASM renderer

### LINKED ISSUE #10751: When I move an element it shouldn't change its size
### Steps To Reproduce

1. Create a shape that is 12,5 x 21,4 px
2. move it to the right with the mouse
3. see how because it wants to snap to the pixel grid, it resizes it to a round number.

### Expected behavior

Its good that it snaps the starting point to the grid, when moving it, but it shouldnt resize it if im just moving it

### Actual behavior

Same thing when I1m resizing it, for example if i edit the width, the height should not change at all.

Also when i slect several eleemtns and move them around, the snapping should only occur with the starting point, the rest of the elements should move as if grouped. This means that they should mantain their decimals relative to the group

### Screenshots or video

_No response_

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

Alotor:
### Related Ticket

Closes #10751 

### Summary

Things that have changed:

- Move a shape: position snaps to the grid, size stays exactly as it was. A 100.5 × 50.3 rect stays 100.5 × 50.3.
- Drag the bottom-right corner: the size snaps, the top-left corner doesn't budge. Before, the whole shape shifted.
- Drag a single edge: only that dimension snaps; the other one keeps its decimals.
- Drag the top-left corner: position and size both snap, since both actually change.
- Shift-locked drag: the locked axis stays where it was instead of drifting up to half a pixel off.
- Rotate: nothing snaps at all, same as the old renderer.
- Tiny shapes: resizing one never shrinks it below 1px (they used to collapse to nothing).
- The flicker while dragging — shape jumping between rounded and unrounded positions frame to frame — is gone.

### Checklist

### DIFF
--- a/render-wasm/src/state/shapes_pool.rs
+++ b/render-wasm/src/state/shapes_pool.rs
@@ -403,6 +403,20 @@ impl ShapesPoolImpl {
     /// gone, but if we don't touch their tiles they keep pointing at the
     /// previous modified position and the tile texture cache may serve stale
     /// pixels.
+    /// Drops the transform modifiers, keeping structure and scale-content
+    /// entries, so the pool serves committed geometry again. Called before
+    /// propagating a new set of transforms, which are relative to that
+    /// geometry.
+    pub fn clear_transform_modifiers(&mut self) {
+        if self.modifiers.is_empty() {
+            return;
+        }
+
+        self.clean_shape_cache();
+        self.modifiers = HashMap::default();
+        self.modifier_uuids.clear();
+    }
+
     pub fn clean_all(&mut self) -> Vec<Uuid> {
         self.clean_shape_cache();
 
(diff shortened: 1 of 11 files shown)


render-wasm/src/state/shapes_pool.rs
```diff
@@ -403,6 +403,20 @@ impl ShapesPoolImpl {
...
+    pub fn clear_transform_modifiers(&mut self) {
+        if self.modifiers.is_empty() {
+            return;
+        }
+
+        self.clean_shape_cache();
+        self.modifiers = HashMap::default();
+        self.modifier_uuids.clear();
+    }
+
```
elenatorro, line 419: nit comment, but we should move this after `clean_all`, cause previous comment belongs to `clean_all`

elenatorro (approved):

PR #11594, opened 2026-09-09
:bug: Show loader instead of empty state while libraries load

### LINKED ISSUE #11452: Add loading message to libraries section

The Libraries section takes a long time to load.

While loading, a message is displayed as if there are no libraries in the team, which causes confusion since actually there are libraries in the team:

Image

Add a loading message so the user doesn't think there are no libraries in the team.

Use the current pencil loading.

_No response_

### Additional context

_No response_

ShreyashAgare26:

## What

- The Libraries section showed the "you have no shared libraries" placeholder while the team's shared files were still loading, so users with libraries were told they had none.
- It now shows the pencil loader until the files arrive.

**Before:**

01-before-empty-state-while-loading

**After:**

02-after-loader-while-loading

## Why

- `libraries-page*` derived its file list straight from `:shared-files` in the app state. Before the fetch resolved, that key is absent, and the filter over it produced an empty sequence.
- A team with no libraries produces the same empty sequence, so `grid*` had no way to tell "still loading" from "nothing to show" and fell through to the empty placeholder in both cases.

## How

- Hold the derived list at `nil` until `:shared-files` is present in the state.
- `grid*` already renders `loading-placeholder*` for a `nil` file list, so no change was needed there. The loading and empty branches now diverge on the same condition the grid was already checking.
- A team with zero shared libraries still reaches the empty placeholder: `shared-files-fetched` writes an empty map rather than leaving the key absent, so the list becomes an empty sequence rather than `nil`. Verified against a freshly created team with no libraries.

Closes #11452


niwinz:
Looks nice, thanks. Waiting CI to merge.

PR #11597, opened 2026-09-09
:sparkles: Add size limits to profile props and plugin registry writes

### LINKED ISSUE #11592: Add size limits to profile props and plugin registry writes
### Description

Profile settings storage (`profile.props`, including the installed plugin registry) currently accepts writes of unbounded size through `update-profile-props` and the dedicated `add-profile-plugin` RPC. Each write rewrites the whole settings document, so repeated or oversized writes can grow a single database row without any controlled error (production p99 sits at ~3 KB, while the only real caps are the 350 MiB HTTP body limit and an internal database limit that surfaces as an uncontrolled error).

This change adds a configurable total size limit (2 MiB default) checked on the merged result before persisting, per-field caps on the plugin registry schema (code size, entry count, name/host/description lengths), and restores rate limiting on the plugin RPCs so the install/uninstall cycle matches the rest of the profile mutations. Oversized growth is rejected with a controlled `:props-too-large` validation error, while shrinking or steady updates on already-large profiles keep working.

### Use case

Operators get a predictable ceiling on per-profile settings storage with clear validation errors instead of database-level failures, and sustained write churn on the profile row is throttled like other profile operations.

### Affected versions

develop

ShreyashAgare26:
### Related Ticket

Closes #11592

### Summary

`profile.props` (which also holds the installed plugin registry) accepted writes of unbounded size through `update-profile-props` and the dedicated plugin RPCs. Every write rewrites the whole settings document, so repeated or oversized writes could grow a single database row until the internal database limit surfaced as an uncontrolled error.

- **Total size limit.** A new `persist-props!` helper in `app.rpc.commands.profile` encodes the merged props and rejects the write with a `:validation` / `:props-too-large` error when the encoded document exceeds `PENPOT_PROFILE_PROPS_MAX_SIZE` (2 MiB by default, `:profile-props-max-size` in config). Only growth is rejected, so shrinking or steady updates on an already oversized profile keep working.
- **Shared by all writers.** `update-profile-props`, `add-profile-plugin` and `remove-profile-plugin` persist through that helper, so the plugin endpoints cannot bypass the ceiling.
- **Plugin registry caps.** `schema:registry-entry` caps code (64 KiB), plugin id and name (250), description (1000), host and icon (2048); the registry is capped at 100 entries, enforced both in the schema and on the add path with a `:too-many-plugins` error.
- **Rate limiting.** `add-profile-plugin` and `remove-profile-plugin` are back in the `:profile-mutations` bucket, so the install/uninstall cycle is throttled like the other profile mutations.

No UI change and no migration needed.

### Steps to reproduce

Automated:

```
cd backend
clojure -M:dev:test --focus backend-tests.rpc-profile-test --focus backend-tests.rpc-plugins-test
```

Two new tests in `backend_tests/rpc_profile_test.clj`: an oversized write is rejected with `:props-too-large` and nothing is persisted, and a profile whose props are already over the limit can still be updated as long as the document shrinks. Reverting `app/rpc/commands/profile.clj` to `develop` makes the first one fail, with the 3 MiB value persisted.

Results on this branch:

```
clojure -M:dev:test --focus backend-tests.rpc-profile-test    # 50 tests, 260 assertions, 0 failures
clojure -M:dev:test --focus backend-tests.rpc-plugins-test    # 7 tests, 0 failures
(common) clojure -M:dev:test                                  # 1218 tests, 25760 assertions, 0 failures
```

clj-kondo and cljfmt are clean on both `backend` and `common`.

Manually, logged in, from the browser console:

```js
fetch('/api/rpc/command/update-profile-props', {
  method: 'POST',
  headers: {'content-type': 'application/json'},
  body: JSON.stringify({props: {'custom-shortcuts': {big: {key: 'a'.repeat(3*1024*1024)}}}})
}).then(async r => console.log(r.status, await r.text()));
```

Returns `400` with `props-too-large`, while normal settings changes still save.

### Checklist

- [ ] Include screenshots or videos, if applicable. Backend only, nothing user visible.
- [ ] Refactor any modified SCSS files following the refactor guide. No SCSS touched.

### DIFF
--- a/backend/resources/rlimit.edn
+++ b/backend/resources/rlimit.edn
@@ -223,7 +223,9 @@
    :main/update-profile-notifications
    :main/delete-profile
    :main/delete-profile-photo
-   :main/request-email-change}
+   :main/request-email-change
+   :main/add-profile-plugin
+   :main/remove-profile-plugin}
  [[:profile-mutations :bucket "30/15/1m"]]
 
  ;; ═══════════════════════════════════════════════
--- a/backend/src/app/config.clj
+++ b/backend/src/app/config.clj
@@ -72,6 +72,7 @@
 
    :telemetry-uri "https://telemetry.penpot.app/"
 
+   :profile-props-max-size (* 1024 1024 2) ; 2MiB
    :media-max-file-size (* 1024 1024 30) ; 30MiB
    :font-max-file-size  (* 1024 1024 30) ; 30MiB
 
@@ -135,6 +136,7 @@
     [:auto-file-snapshot-every {:optional true} ::sm/int]
     [:auto-file-snapshot-timeout {:optional true} ::ct/duration]
 
+    [:profile-props-max-size {:optional true} ::sm/int]
     [:media-max-file-size {:optional true} ::sm/int]
     [:font-max-file-size  {:optional true} ::sm/int]
 
--- a/backend/src/app/rpc/commands/plugins.clj
+++ b/backend/src/app/rpc/commands/plugins.clj
@@ -45,10 +45,15 @@
         plugins (-> plugins
                     (update :ids #(vec (distinct (conj % plugin-id))))
                     (assoc-in [:data plugin-id] plugin))]
-    (db/update! conn :profile
-                {:props (db/tjson (assoc (:props profile) :plugins plugins))}
-                {:id profile-id}
-                {::db/return-keys false})
+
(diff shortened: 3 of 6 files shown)


niwinz:
Hey @Shreyashagare 

Thanks for the contribution, but this is already covered by an existing and previous PR https://github.com/penpot/penpot/pull/11596

PR #11610, opened 2026-09-10
:bug: Keep text image fills aligned during drag

superalex:

## What

- Text shapes with image (or gradient) fills keep the fill aligned when dragging/moving the text
- Resize already looked correct; move left the image behind because of stale absolute shaders in the layout cache

## Why

After #11355, plain-text paint reuses cached Skia paragraphs across move clones. Those paragraphs bake absolute image/gradient shaders at layout time. On drag, glyphs paint at the new selrect while the fill shader stays at the old origin.

## How

- Record `layout_paint_origin` when the layout (and fill paints) are built
- When painting from cache, draw at that origin and translate the canvas by the selrect delta so glyphs and shaders move together
- Sync text bounds before `update_layout` / in `set_selrect` so fills bake against the current container
- Unit tests cover origin recording, move offset, and origin surviving clone+translate

### DIFF
--- a/render-wasm/src/shapes.rs
+++ b/render-wasm/src/shapes.rs
@@ -403,8 +403,8 @@ impl Shape {
         self.invalidate_extrect();
         self.selrect.set_ltrb(left, top, right, bottom);
         if let Type::Text(ref mut text) = self.shape_type {
+            // `update_layout` syncs bounds via set_xywh before baking fill paints.
             text.update_layout(self.selrect);
-            text.set_xywh(left, top, self.selrect.width(), self.selrect.height());
         }
     }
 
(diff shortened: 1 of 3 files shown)


render-wasm/src/shapes.rs
```diff
@@ -403,8 +403,10 @@ impl Shape {
         self.invalidate_extrect();
         self.selrect.set_ltrb(left, top, right, bottom);
         if let Type::Text(ref mut text) = self.shape_type {
-            text.update_layout(self.selrect);
+            // Bounds must match selrect before layout so absolute fill shaders
+            // (image/gradient) bake against the current container origin.
             text.set_xywh(left, top, self.selrect.width(), self.selrect.height());
+            text.update_layout(self.selrect);
```
elenatorro, line 407: update_layout already calls set_xywh with selrect options (set on `self.selrect.set_ltrb(left, top, right, bottom);` above), so we can just remove `text.set_xywh` here, right?

superalex: Thanks!

PR #11635, opened 2026-09-10
:bug: Fix chunked upload storage amplification and cap chunk size

### LINKED ISSUE #11634: Chunked upload stores duplicate objects when the same chunk index is sent twice
### Description

Repeated uploads of the same chunk index in one upload session each create a new stored object, so a one-chunk session can accumulate many times its declared data in temporary storage.
### Steps to reproduce

1. Create an upload session with `totalChunks: 1`.
2. Call `upload-chunk` four times with the same session id and `index: 0`.
3. Inspect live `storage_object` rows for the session: four distinct objects and four times the declared bytes exist.
### Expected behavior

A repeated session and chunk index is rejected or atomically replaces the earlier object, keeping one stored chunk and no extra bytes.
### Affected versions

2.17.2 (verified on 980ccf15fa0e1ece7130db48a6a6defce50045fd; still present on current staging).

niwinz:

## What

- Chunked uploads no longer store duplicate objects when the same chunk index is sent twice; the repeat is rejected with `duplicate-chunk-index` (GHSA-qvq5-c536-pmx8).
- New `upload-max-chunk-size` setting (default 30 MiB) rejects oversize chunks before anything is stored.

## Why

- `upload-chunk` only checked index bounds and wrote a fresh object per request, so one authenticated user could pile unbounded temp storage onto a tiny session; assembly noticed the mismatch only after allocation. It also counted chunks without checking which indices arrived, so duplicated indices could assemble a corrupt file.
- Chunks had no size cap of their own: the 30 MiB limits applied to the assembled file while each request was bounded only by the 350 MiB HTTP body limit.

## How

- **Upload path:** `upload-chunk` now runs in a transaction, locks the session row, and rejects already-stored indices; `assemble-chunks` requires exactly indices `0..n-1`, closing the corrupt-assembly gap for media, fonts, and binfile through the shared helper.
- **Size cap:** per-chunk check against the new setting, before any storage write; app clients slice at 25/10 MiB, so no frontend change was needed.
- **Tests:** four new backend tests (duplicate rejected, index-set assembly guard, dupe-then-assemble, session stays usable, oversize rejected); media, font, and binfile suites green (49 tests, 415 assertions), lint and format clean.

Closes #11634

### DIFF
--- a/backend/src/app/rpc/commands/media.clj
+++ b/backend/src/app/rpc/commands/media.clj
@@ -339,6 +339,8 @@
 
 ;; --- Chunked Upload: Upload a single chunk
 
+(declare get-upload-chunk)
+
 (def ^:private schema:upload-chunk
   [:map {:title "upload-chunk"}
    [:session-id ::sm/uuid]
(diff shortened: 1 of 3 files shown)


backend/src/app/rpc/commands/media.clj
```diff
@@ -354,9 +357,31 @@
   {::doc/added "2.17"
    ::sm/params schema:upload-chunk
    ::sm/result schema:upload-chunk-result}
-  [{:keys [::db/pool] :as cfg}
-   {:keys [::rpc/profile-id session-id index content] :as _params}]
-  (let [session (db/get pool :upload-session {:id session-id :profile-id profile-id})]
+  [cfg {:keys [::rpc/profile-id session-id index content]}]
+  (let [session (db/tx-run! check-upload-chunk-slot session-id profile-id index content)]
```
Alotor: tx-run! should receive cfg as first parameter i think

Alotor (approved):

PR #11651, opened 2026-09-11
:sparkles: Add indirection for upload-chunks storage via new table

### LINKED ISSUE #11644: Add dedicated upload_session_chunk table with its own storage bucket for chunked uploads
### Description

Chunked-upload chunks are stored directly as `storage_object` rows in the generic `tempfile` bucket, with the chunk metadata (upload id, chunk index) embedded in the object metadata. This change introduces an `upload_session_chunk` table with `ON DELETE RESTRICT` foreign keys to `storage_object` and `upload_session`, so chunk existence and counting no longer require scanning `storage_object`, and chunks get their own `upload-session` storage bucket with dedicated garbage-collection handling in `touched-gc` and `deleted-gc`. Assembling a session marks it as consumed (`deleted_at`) instead of deleting it, and the `objects-gc` task procedurally purges consumed and stalled sessions (touching referenced storage objects first). Re-uploading an already stored chunk index now returns a validation error.

### Use case

Cheaper and clearer chunked-upload bookkeeping: session chunk state is queryable without touching the large `storage_object` table, chunk lifecycle is independent from other temporary files, and referential integrity prevents orphaned or prematurely deleted chunk data.

### Affected versions

2.19.0

niwinz:

## What

- Chunked-upload chunks move out of `storage_object` metadata into a dedicated `upload_session_chunk` table, so chunk existence and counting no longer scan `storage_object`.
- Chunks get their own `upload-session` storage bucket with dedicated garbage-collection handling.
- Re-uploading an already stored chunk index returns a validation error instead of duplicating rows.

## Why

Chunk state was embedded in the generic `tempfile` object metadata, which forced metadata scans to answer per-session questions, shared lifecycle with unrelated temporary files, and allowed silent duplicates that broke assembly counts. Referential integrity now prevents orphaned or prematurely deleted chunk data.

## How

- **Schema (migration 0154):** `upload_session_chunk` with non-deleting FKs (`NO ACTION DEFERRABLE`, restrict semantics) to `storage_object` and `upload_session`; `deleted_at` marker on `upload_session`; profile FK moved from `CASCADE` to non-deleting.
- **Write/read path:** `upload-chunk` validates the live session, rejects duplicate indexes, and stores objects in the new bucket with no session metadata; chunk reads go through a `JOIN`.
- **Lifecycle:** assemble removes mappings and marks the session consumed; `objects-gc` procedurally purges consumed, stalled, and profile-owned sessions (touching objects first); `touched-gc`/`deleted-gc` handle the new bucket; `upload-session-gc` is removed and consumed sessions no longer count against quota.
- **Tests:** full backend suite green (844 tests), plus lint and format checks.

Closes #11644


backend/src/app/storage/gc_touched.clj
```diff
@@ -162,6 +162,7 @@
     (= bucket "profile")               (process-objects! conn has-profile-refs? bucket objects)
     (= bucket "file-data")             (process-objects! conn has-file-data-refs? bucket objects)
     (= bucket sto/tempfile-bucket)     (process-objects! conn (constantly false) sto/tempfile-bucket objects)
+    (= bucket sto/upload-session-bucket) (process-objects! conn (constantly false) sto/upload-session-bucket objects)
```
superalex, line 165: It's very minor but the indentation is broken here, shouldn't we add a space to the rest of the lines?

backend/src/app/rpc/commands/media.clj
```diff
@@ -353,35 +352,75 @@
...
+                                      ::sto/touched-at   (ct/in-future {:hours 1})
+                                      :content-type      (:mtype content)
+                                      :bucket            sto/upload-session-bucket})
+                    (catch Throwable cause
+                      (db/delete! pool :upload-session-chunk
+                                  {:session-id session-id :chunk-index index})
+                      (throw cause)))
+          linked  (-> (db/exec-one! pool [sql:link-upload-session-chunk
+                                          (:id object) session-id index])
+                      (db/get-update-count))]
```
superalex: If the UPDATE fails after a put-object! ok,  will the object_id be NULL and the retry will match against :chunk-already-exists?

niwinz: yes, this is expected, the session considered stalled and the gc will clean it

backend/src/app/storage/gc_touched.clj
niwinz: fixed

PR #11657, opened 2026-09-11
:zap: Fetch only caller share-link in view-only bundle

### LINKED ISSUE #11633: Fetch only the caller's share-link in the view-only bundle
### Description

Follow-up hardening to #11285 / #11286 (GHSA-w8mf-4x22-24gg, duplicate GHSA-6fj3-8xh8-x8h3): `get-view-only-bundle` (`backend/src/app/rpc/commands/viewer.clj`) returns the correct rows since #11286, but on the anonymous path it still runs `SELECT * FROM share_link WHERE file_id = ?`, loads every sibling secret token into memory, and only then drops the siblings with `filterv #(= (:id %) share-id)`.

The proposal is predicate pushdown: when `(:type perms)` is `:share-link`, fetch at most one row with `db/get* conn :share-link {:id share-id :file-id file-id}` so sibling tokens never leave Postgres; the member/owner path keeps the full `db/query {:file-id file-id}` list that the share-management dialog needs.

### Use case

Defense in depth for share-link secrecy (a later log, error dump, or refactor touching `links` before the filter cannot re-expose sibling tokens), plus a smaller read on the hot anonymous viewer path; no user-visible behavior change.

### Proposed change

Single `let` binding in `get-view-only-bundle`, one file (`backend/src/app/rpc/commands/viewer.clj`), mirroring the precedent in `get-file-permissions` (`backend/src/app/binfile/common.clj:391`), which already resolves the same credential with `db/get* {:id share-id :file-id file-id}`.

Before:

```clojure
links (cond->> (->> (db/query conn :share-link {:file-id file-id})
                    (mapv (fn [row] (-> row (update :pages db/decode-pgarray #{}) (dissoc :flags)))))
        (= :share-link (:type perms))
        (filterv #(= (:id %) share-id)))
```

After (sketch):

```clojure
decode-link (fn [row] (-> row (update :pages db/decode-pgarray #{}) (dissoc :flags)))
links (if (= :share-link (:type perms))
        (->> (db/get* conn :share-link {:id share-id :file-id file-id}) (some-> decode-link vector) (or []))
        (->> (db/query conn :share-link {:file-id file-id}) (mapv decode-link)))
```

Behavior cases: (a) anonymous holder of restrictive L1 (`share-id: aaa`, file `F`) gets `[aaa]` with SQL `WHERE id = 'aaa' AND file_id = 'F'` instead of loading `[aaa, bbb]` then filtering; (b) owner/member (`:type :membership`) still gets the full list so `frontend/src/app/main/ui/viewer/share_link.cljs` (`d/seek` over `slinks`, `assoc :share-id (:id slink)`) keeps working; (c) cross-file replay (`bbb` from `F` used against `G`) still fails closed via the composite predicate (`nil` → `:not-found` / `[]`); (d) defensive nil `share-id` yields `[]`, never falls back to all rows. The own `:id` stays in the response because the caller already sent it and the viewer dialog matches on it; shape (`:share-links` vector of decoded rows) is unchanged.

### Acceptance criteria

1. Anonymous `get-view-only-bundle` response is byte-identical to today (only the caller's row).
2. Owner/member response still holds all rows.
3. No sibling `:id` reaches the backend process on the anonymous path (single-row SQL).
4. Existing `share-link-token-disclosure` test in `backend/test/backend_tests/rpc_viewer_test.clj` stays green; suggested addition: cross-file case asserting `:not-found` when `share-id` from file `F` is used with another file `G`.

### Affected versions

main (post 2.18.0 / #11286).

makesomethingshit:

### Related Ticket

Closes #11633

### Summary

`get-view-only-bundle` no longer loads every sibling share-link row for a file and filters in memory. When the caller authenticates via share-link, the handler now resolves only the caller's row with a composite `(id, file-id)` predicate (`db/get*`), then decodes `:pages` and strips deprecated `:flags` exactly as before, wrapped in a one-element vector (empty vector when the row is absent). The membership/owner path still fetches the full share-link list with `db/query {:file-id ...}`, so the share-management UI is unaffected. No permission-model, schema, frontend, migration, or linked-library changes.

### Steps to reproduce

1. Create a file with two share-links L1 and L2.
2. Request the view-only bundle with L1 as an anonymous caller.
3. Response `:share-links` contains only L1 (previously the backend read all sibling rows before filtering).
4. Request the same bundle as the file owner: both L1 and L2 are returned.
5. Replaying L1 against a different file fails closed (`:not-found` / `:object-not-found`); anonymous requests without a share-id also return `not-found`.

### Verification status

- `git diff --check` and `./scripts/check-commit` pass on the head commit.
- `cljfmt check` on the two changed files (`viewer.clj`, `rpc_viewer_test.clj`) passes: "All source files formatted correctly".
- `clj-kondo` lint (same args as the backend script): errors 0, warnings 11 — all pre-existing in unrelated files, 0 in `viewer.clj`.
- Backend suite (`rpc_viewer_test`, including `share-link-token-disclosure` and the two new regression tests) has NOT been executed: no `clojure` CLI, no Postgres credentials (password prompt, not guessed), no Valkey/Redis or docker in the contributing environment. Not claimed as passing.

### Checklist

- [ ] Include screenshots or videos, if applicable. (Not applicable: backend-only change.)
- [x] Add or modify existing integration tests in case of bugs or new features, if applicable. (Extended \share-link-token-disclosure` with a predicate-pushdown regression test and cross-file replay coverage.)`
- [ ] Refactor any modified SCSS files following the refactor guide. (Not applicable.)
- [ ] Check CI passes successfully. (Triage pass; commit-message check skipping. See Verification status for locally-run checks.)

### DIFF
--- a/backend/src/app/rpc/commands/viewer.clj
+++ b/backend/src/app/rpc/commands/viewer.clj
@@ -89,18 +89,27 @@
                      (mapv (fn [{:keys [id] :as lib}]
                              (merge lib (bfc/get-file cfg id)))))
 
-        links   (cond->> (->> (db/query conn :share-link {:file-id file-id})
-                              (mapv (fn [row]
-                                      (-> row
-                                          (update :pages db/decode-pgarray #{})
-                                          ;; NOTE: the flags are deprecated but are still present
-                                          ;; on the table on old rows. The flags are pgarray and
-                                          ;; for avoid decoding it (because they are no longer used
-                                          ;; on frontend) we just dissoc the column attribute from
-                                          ;; row.
-                                          (dissoc :flags)))))
-                  (= :share-link (:type perms))
-                  (filterv #(= (:id %) share-id)))
+        decode-link
+        (fn [row]
+          (-> row
+              (update :pages db/decode-pgarray #{})
+              ;; NOTE: the flags are deprecated but are still present
+              ;; on the table on old rows. The flags are pgarray and
+              ;; for avoid decoding it (because they are no longer used
+              ;; on frontend) we just dissoc the column attribute from
+              ;; row.
+              (dissoc :flags)))
+
+        ;; NOTE: on the share-link path we fetch at most the caller's own
+        ;; row with a composite (id, file-id) predicate, so sibling tokens
+        ;; never leave postgres. The membership path keeps the full list
+        ;; the share-management UI needs. A nil share-id never falls back
+        ;; to the full query; it simply resolves to an empty vector.
+        links   (if (= :share-link (:type perms))
+                  (if-some [row (db/get* conn :share-link {:id share-id :file-id file-id})]
+                    [(decode-link row)]
+                    [])
+                  (mapv decode-link (db/query conn :share-link {:file-id file-id})))
 
         fonts   (db/query conn :team-font-variant
                           {:team-id (:id team)
--- a/backend/test/backend_tests/rpc_viewer_test.clj
+++ b/backend/test/backend_tests/rpc_viewer_test.clj
@@ -206,6 +206,41 @@
         (t/is (some #(= link-a-id (:id %)) share-links))
         (t/is (some #(= link-b-id (:id %)) share-links))))
 
+    (t/testing "share-link viewer does not query sibling share-links"
+      ;; Direct regression test for the predicate-pushdown invariant:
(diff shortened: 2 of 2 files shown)


niwinz:
Thanks :heart:

PR #11664, opened 2026-09-14
:bug: Fix form select label resetting to its default option

### LINKED ISSUE #11663: Access token expiration select always shows Never after picking another option
### Steps to reproduce

1. Go to **Your account > Integrations** and click **Create new access token**.
2. Type a name.
3. Open the **Expiration date** select and pick **30 days** (or any option other than **Never**).

### Expected behavior

The select keeps showing **30 days**.

### Actual behavior

The select immediately shows **Never** again. Picking again does not help: it snaps back on every re-render.

The value that is submitted is actually the picked one (the created token expires in 30 days and the success dialog says so), so only the label is wrong, but it makes users believe they can only create non-expiring tokens.

### Cause

`app.main.ui.ds.controls.select/select*` resets its selected id to `default-selected` whenever `options` changes identity:

```clojure
(mf/with-effect [default-selected options]
  (reset! selected-id* (get-selected-option-id options default-selected)))
```

`rumext.v2/deps` compares vectors by identity, and the access token dialog builds its `options` vector inline, so every re-render re-runs the effect. Picking an option updates the form state, which re-renders the dialog, which resets the label to `default-selected "never"`.

`app.main.ui.forms/form-select*` passes the form value to `select*` as `:value`, but `select*` does not read `value`, so the reset never lands on the form value.

### Environment

- Penpot 2.17.2 (self-hosted, Docker). The same code is on `develop` and `staging`.
- Chromium 145 (headless) and Chrome on macOS.

chulgil:

### Related Ticket

Closes #11663

## What

- On **Your account > Integrations > Create new access token**, picking any expiration other than **Never** made the select show **Never** again right away.
- The submitted value was already the picked one (the token expired as chosen), so only the label was wrong, but it made users believe only non-expiring tokens could be created.

## Why

- `select*` resets its selected id to `default-selected` in `(mf/with-effect [default-selected options] ...)`.
- `rumext.v2/deps` compares vectors by identity, and the access token dialog builds its `options` vector inline, so every re-render re-runs that effect.
- Picking an option updates the form state, which re-renders the dialog, which resets the label to `default-selected "never"`.
- `form-select*` passes the form value to `select*` as `:value`, but `select*` does not read `value`, so the reset never landed on the form value.

## How

- `form-select*` now passes the form value as `:default-selected` when it is a non-blank string, so the reset lands on the picked option.
- While the form value is blank, the caller's own `default-selected` still applies, so selects without an initial value behave as before.
- The fix lives in `form-select*` rather than in the access token dialog, so every form-bound select (e.g. the invite role select) shows the form value consistently.

### Checklist

- [x] Add a detailed explanation of how to reproduce the issue and/or verify the fix, if applicable. (See #11663.)
- [x] Refactor any modified SCSS files following the refactor guide. (No SCSS changes.)

### Verification

- `clj-kondo` (v2026.07.24, repo config): 0 errors, 0 warnings, same as before the change.
- `cljfmt check` (0.16.5, repo config): formatted correctly.
- Behavior was verified on a self-hosted Penpot 2.17.2 by applying the equivalent change to the compiled `form-select*`: picking 30/60/90 days now keeps the label, and the created tokens expire as picked (2026-10-14, 2026-11-13).
- The ClojureScript source change itself was not compiled locally; CI will be the first full build.


niwinz:
Thanks, i will backport it to 2.18.0

PR #11693, opened 2026-09-14
:bug: Harden browser logging against invalid levels

### LINKED ISSUE #11690: Browser app crashes when logging receives an empty level and misses fatal level
### Description

The browser app crashes with an uncaught exception `invalid level provided to level->int function: nil` originating in `app.common.logging/level->int`, reachable from the dashboard bundle. Any `nil`/unknown log level (for example calling the JS-exported `debug.set_logging()` with no argument, or `setup!` with `{:app nil}`) throws instead of being ignored, so a logging call takes down the whole app and files its own error report. Additionally, the `:fatal` level allowed by the log-record schema has no branch in `level->int`/`level->color`/`level->name`, so logging at `:fatal` in the browser crashes the same way.

### Steps to reproduce

1. Open the dashboard in a browser.
2. In the devtools console, run `debug.set_logging()` with no argument (or `debug.set_logging("app", "bogus")`).
3. Observe the uncaught `invalid level provided to level->int function: nil` exception.
4. Separately, evaluate a log call at `:fatal` level in CLJS and observe the same kind of throw with `{:level :fatal}`.

### Expected behavior

Invalid or empty levels warn in the console and are ignored so the app keeps running; `:fatal` works in the browser and behaves like `:error`.

### Affected versions

2.17.2

niwinz:

## What

- The browser app no longer crashes when a log call carries an empty or unknown severity level; invalid levels warn in the console and are ignored so the app keeps running.
- The schema-legal `:fatal` level now works in the browser and behaves exactly like `:error`.
- The `debug.set_logging()` console command validates its arguments instead of throwing on missing or misspelled values.

## Why

- Any `nil`/unknown level reached `level->int`, which throws on anything outside its closed keyword set, turning a single bad log call into an uncaught exception (and its own error report); the most likely trigger is `debug.set_logging()` with no argument, since `(keyword nil)` is `nil`.
- The log-record schema already admitted `:fatal` and the server already handled it, but the browser mappings had no `:fatal` branch.

## How

- **Common logging (browser branch only):** fail-soft `enabled?`/`setup!`/console handler with console warnings; `:fatal` aliased to `:error` (50/`ERR`/red); public `valid-levels`/`valid-level?`/`valid-logger?` validators; safe color/name fallbacks for corrupt records.
- **Debug bridge:** `set-logging` coerces string/keyword input safely, validates before `set-level!`, and stores string logger keys (fixing previously unreachable keyword-key writes).
- **Tests:** new `common-tests.logging-test` (both runtimes) plus a frontend `debug-logging-test` covering the exact issue repro; full common JVM suite green with no regressions.

Closes #11690


niwinz:
@copilot resolve the merge conflicts in this pull request

PR #11694, opened 2026-09-14
:sparkles: Preserve sidebar scroll positions across tab switches

### LINKED ISSUE #7440: Preserve scroll on switching between right sidebar tabs

niwinz:

## What

- Switching sidebar tabs (Layers, Assets, Tokens) no longer resets each panel's scroll position to the top; every panel returns to where the user left it.
- Closes #7440.

## Why

- Each panel unmounts on tab switch, so any in-component scroll state dies with it; positions must live in a store that outlives the panels.

## How

- **Shared helper:** new `app.main.ui.workspace.sidebar.scroll` namespace with a save function and a restore hook used by all three panels — one canonical implementation instead of per-panel copies.
- **Scoped store:** a single position map held in `left-sidebar*` (the component that stays mounted while tabs swap) and threaded to the panels as a prop; it dies with the workspace, so nothing leaks across files.
- **Settle-aware restore:** restore retries on animation frames until content height stabilises, so deep positions in lazily rendered lists (100-row chunks) are not clamped to the first chunk.
- **Per-list keys:** Layers and its search-results list, Assets, and Tokens (per token set) track positions independently.

## Testing

- New `frontend-tests.ui.sidebar-scroll-test`: 6 tests, 15 assertions covering save-per-key, missing-target no-ops, restore writes, and the settle-retry loop — green, plus `lint:clj` (0 errors) and `cljfmt` clean.
- Residual risk: real tab-switch behavior and the exact #7440 repro need a manual browser pass (tab-switch matrix, >100-item deep restore, token-set switches).


EvaMarco:
@niwinz check if filtering layers on and off reset the scroll position.

PR #11699, opened 2026-09-15
:bug: Fix stalled saving states

### LINKED ISSUE #11672: Workspace drops unsaved changes when read-only blocks persistence
### Description

When the workspace enters a read-only state while edits are still pending, those changes can stop flushing to the server. After a reload (or after leaving the mode without a successful `update-file`), the edits are lost even though they were still in memory.

The main trigger observed in production is WebGL context loss: it sets workspace `read-only?`, and persistence treated that flag as “do not save.” The same flag is also used by Inspect mode, so pending saves can stall there too.

Version preview is a related case: it also uses read-only, but it swaps in snapshot data. Entering preview with a non-empty persistence queue can stall the same way, and derived workspace watchers can create local commits against the snapshot that should never be persisted.

### Steps to reproduce

1. Open a file with render-wasm enabled and make an edit.
2. Before autosave finishes (within ~3s), lose the WebGL context (`WEBGL_lose_context.loseContext()` on the canvas).
3. Restore the context (`restoreContext()`), or wait for browser recovery.
4. Observe that no successful `update-file` runs for the pending edit (or the persistence queue stays stuck).
5. Reload the page.

Related checks (same root cause):

1. Make an edit, then switch the sidebar to Inspect before autosave finishes; confirm pending `update-file` still runs.
2. Make an edit, then open History version preview before autosave finishes; confirm live edits flush first and preview does not create commits against the snapshot.

### Expected behavior

Pending live-file edits keep saving when the UI is read-only for Inspect or WebGL context loss. Version preview only blocks persistence after live edits have been flushed, and does not generate commits against snapshot data. A reload after recover keeps the changes.

### Affected versions

Current staging / production with render-wasm.

Alotor:
### Related Ticket

Closes #11672

#### The problem

The application saves your edits in the background, a few seconds at a time. The editor also has moments when it deliberately turns read-only. 

So if the flag got stuck on, saving stopped for good. The indicator sat on "Saving…" forever. And when a save request failed, the editor threw the whole pending queue away — your unsaved work went with it, without a word.

Right now it moves to read only mode when:

- Change to inspect mode
- Preview the restore of a snapshot
- Losing a WEBGL context
- Change team role websocket notification

#### What changed

1. Read-only no longer means "stop saving." The decision moved to the moment an edit is made. Previewing history refuses to create edits at all, so the old version stays untouched. Anything already queued keeps saving, whatever the interface is doing.

2. A failed save keeps your work. The queue is kept, the file is marked as failed, and a message stays on screen asking you not to close the tab. Saving picks up again on its own once the cause clears — for example when someone restores your editing permission.

3. No blind retries. If a save is cut off and we cannot tell whether the server received it, we do not send it again: a repeat could duplicate your shapes. The editor reports it instead.

4. A watchdog. Every thirty seconds the editor checks whether saving is moving. If nothing has moved for five minutes it files one error report with the context needed to diagnose it. Nothing the user sees, nothing touched.

5. Version history waits properly. Creating or restoring a version now waits for your pending changes to reach the server, and fails out loud instead of carrying on quietly. A failed restore puts the live file back on screen rather than leaving you parked in a historic preview.

6. Leftover preview state cleaned up. Leaving the workspace mid-preview used to leave a hidden flag behind that made the next file you opened silently refuse every edit. Fixed.

### Checklist


frontend/src/app/main/data/workspace/versions.cljs
```diff
@@ -386,18 +395,18 @@
...
+           ;; Restore the live file before rejecting the plugin promise.
            (rx/catch (fn [error]
-                       (reject error)
-                       (rx/empty)))))))
-
-
-
+                       (rx/concat
+                        (rx/of (exit-preview))
+                        (->> (rx/of error)
```
niwinz, line 410: observe-on :async

Alotor: done

PR #11720, opened 2026-09-15
:bug: Pass the base branch and the project when opening a PR

alvorithm:

## What

- Pass `--base` and `--project "Main"` when the `create-pr` skill opens a pull request.
- Warn that a draft PR runs no checks, and ask before opening one.
- Repeat the `AI-assisted-by:` trailer in the PR body, once per model.

## Why

- The skill resolves the target branch with `scripts/detect-target-branch` in step 1 and then never passes it, so `gh` opens the PR against the repository default. A branch cut from `staging` therefore targets `develop`. `mem:workflow/creating-prs` also requires `--project "Main"`, which the command omitted, so the two disagreed.
- Every test workflow is gated on `if: ${{ !github.event.pull_request.draft }}`, so a draft PR runs no checks at all. An agent that opens a draft PR might report a pull request that doesn't imply CI has run.
- A reviewer reads the PR before the commits, so the assistance belongs where the discussion is. The branch commits keep their own trailers, and a squash merge carries every one of them into the landed message.

## How

- The `gh pr create` example now reads `--base "<BASE>"` and `--project "Main"`, with one sentence naming where `<BASE>` comes from.
- Step 4 states the draft cost and asks for the trailer in the body.

## Review

Two claims in the first commit were wrong, and the second commit removes them.

- A behind-the-base check. The rationale said the checks would judge a tree nobody ships. They would not: every test workflow runs `actions/checkout` with no `ref`, so a `pull_request` event builds `refs/pull/N/merge`, which is head already merged into base. The check only forced a rebase for no gain.
- The trailer rationale said a trailer on branch commits disappears at merge. It does not. GitHub's squash body lists the commit messages, and `91b433985a` carries three `AI-assisted-by` lines from the two commits of #11677.

AI-assisted-by: claude-opus-5


.agents/skills/create-pr/SKILL.md
```diff
@@ -75,10 +80,22 @@ description structure, writing principles) and `mem:workflow/creating-commits`
 (commit type emojis). Derive the title and body from the commits and, when
 there is one, from the issue body. Reference the issue with `Closes #NNNN`.
 
+Put the `AI-assisted-by:` trailer in the body, not only on the commits. Penpot
+squash-merges, so the landed commit message is built from the PR title and the
+body, and a trailer that lives only on branch commits disappears at merge.
+
```
niwinz, line 82: This is is not true, in fact on merge squash, it preserves all AI-assisted-by  trailers on the commit. In any case i'm ok to put it on PR, just want to say it may have multiple values.

.agents/skills/create-pr/SKILL.md
```diff
@@ -61,6 +61,11 @@ Run all checks before reporting, then report every failure together:
    branch was never pushed, or has commits the remote does not have. Never
    push yourself; ask the user to push and to run `/create-pr` again
    afterwards, then stop.
+6. **Branch behind the base.** `git rev-list --count HEAD..origin/<base>` is
+   not zero — the checks would run against head merged into base, so they
+   would report on a tree nobody will ship. Ask the user to rebase onto the
+   base and push, then stop. Rebasing before the PR exists is allowed.
+   Force-pushing during review is not.
```
niwinz: The skill already stops on this without this point. It is covered by point 5, i mean the point 5 checks:

> he remote-tracking ref `origin/<branch>` is missing, or `git rev-parse origin/<branch>` differs from HEAD

So i don't think this point 6 adds real value.

alvorithm: You are right, withdrawn. Then from my side only wondering if the skill should check a conflicting merge (GH cannot compute the merge ref, git merge-tree --write-tree origin/<base> HEAD for the submitter). Not sure if this happens often.

.agents/skills/create-pr/SKILL.md
niwinz: Example: 

Image

.agents/skills/create-pr/SKILL.md
niwinz: did not experimented, in any case i think is not the responsibility of the skill doing this, but is my opinion

niwinz:
i merged it but i will backport it to staging manually  :D

Issue #11724, opened 2026-09-15
Bound frontend error report amplification and handle transient failures safely


niwinz:
### Description

On 2026-09-13, between 07:00 and 08:00 UTC, a single client produced 46,177 `handled-exception` audit events — 98% of that day's total and roughly 348 MB of error reports — while the backend was healthy (344 profiles active and normal event volume). The client entered a loop of network-level RPC failures (`update-file` and `check-nitrate-sso`) and the frontend generated, formatted, stored and reported a full error report (~7.5 KB with stack trace and the last 50 events), plus a toast, for every single failure. There is no deduplication, no grouping window and no per-session cap, so a stuck client can flood audit/telemetry and stay degraded while it generates reports.

This epic tracks the improvements that bound that amplification, stop treating transient network failures as actionable errors, and make persistence safe when the network drops.

### Expected behavior

- A client repeating the same failure emits at most one report per error fingerprint per 2 minutes, and never more than 1000 reports per browser session.
- Errors are grouped, not silently dropped: the next emitted report includes how many occurrences were suppressed.
- Transient failures (`:network`, `:offline`) produce a compact report without stack trace or last-events dump.
- Reporting happens outside the error-handling call stack, so it cannot re-trigger the error or break the handler.
- A transient failure while saving does not escalate to the global error screen; the client retries safely and does not lose or duplicate changes.
- Audit buffer overflows are counted and observable instead of silent.

### Scope

Each item below becomes a sub-issue of this epic.

- [ ] Add an error report governor in the frontend: fingerprint, 2-minute window, 1000-per-session cap, occurrence counts.
- [ ] Emit compact reports for `:network`/`:offline` and add the missing `:offline` error handler.
- [ ] Make `flash` report generation and emission asynchronous and failure-proof.
- [ ] Stop escalating transient persistence failures to the global error handler.
- [ ] Make `update-file` idempotent by `commit-id` so retries cannot apply changes twice.
- [ ] Keep the persistence queue and retry transient failures with backoff.
- [ ] Show a single "connection lost, retrying" notice instead of one toast per failure.
- [ ] Count and surface audit buffer overflows.

### Out of scope

- Server-side rate limiting for `push-audit-events` and redefining the backend `exception-event?` contract.
- E2E (Playwright) coverage unless explicitly requested.

### Risks and notes

- The `update-file` idempotency change is the highest-risk item: it touches the apply transaction and the response shape (`:revn`, `:lagged`). It must land before any automatic retry is enabled.
[...]

PR #11740, opened 2026-09-15
:bug: Bound error report amplification and report environment failures

### LINKED ISSUE #11726: Bound error report amplification and report environment failures
### Description

Frontend error reporting had no volume control and did not distinguish application defects from environment failures. On 2026-09-13 a single client produced 46,177 `handled-exception` events in one hour (~348 MB of reports) while the backend was healthy.

This sub-issue of #11724 covers both fixes:

- A report governor in `app.main.errors`: each error gets a stable fingerprint, the first occurrence is reported, and repeated occurrences of the same fingerprint are counted but not emitted for 2 minutes. The next emitted report includes how many occurrences were suppressed. It applies to `handled-exception`, `unhandled-exception` and `exception-page`. Suppressed occurrences do not pay the cost of building the full report, and the fingerprint cache is bounded (oldest entry evicted) so memory cannot grow without limit. There is no per-session emission cap: volume is governed by the 2-minute window and by new fingerprints appearing.
- Connectivity and gateway failures (`:network`, `:offline`, `:bad-gateway`, `:service-unavailable`, `:nitrate-unavailable`, `:nitrate-not-configured`) are not application defects. `:offline` fell through to `:default` and 502/503 rendered `exception-page`, so they reached the internal error reports and Mattermost alerts with the full payload. They are now audit-only `handled-exception` events with a compact report, and `:network`/`:offline` show a specific `errors.connection-error` toast.

### Use case

A client stuck in an error loop (network outage, crash loop, repeated assertion) can flood audit-log/telemetry and, for unhandled exceptions, the error reports and Mattermost alerts. Grouping identical errors keeps the signal (occurrence counts) while bounding the noise and the client-side cost. Separately, a connectivity problem on the user's side should not file an application-defect report or alert anyone, while still leaving aggregate telemetry.

### Affected versions

2.18.0-RC6

niwinz:

## What

- Frontend error reporting had no volume control: a client stuck repeating the same failure emitted one audit/telemetry event for every occurrence. On 2026-09-13 a single client produced 46,177 `handled-exception` events in one hour (~348 MB of reports) while the backend was healthy.
- Repeated errors sharing a fingerprint are now coalesced within a 2-minute window: the first occurrence is emitted, and the next report carries how many occurrences were suppressed as `:occurrences`. It applies to `handled-exception`, `unhandled-exception` and `exception-page`.
- Connectivity and gateway failures (`:network`, `:offline`, `:bad-gateway`, `:service-unavailable` and nitrate configuration) are no longer treated as application defects. `:offline` fell through to `:default` and 502/503 rendered `exception-page`, so they reached the internal error reports and Mattermost alerts with the full payload. They are now audit-only `handled-exception` events with a compact report.
- `:network` and `:offline` show a specific `errors.connection-error` toast instead of the generic "Something wrong has happened." copy.

## Why

- The degraded states that produce the most repeats (network outage, crash loop, repeated assertion) were also the ones paying the highest cost: every occurrence built a full report (stack trace + last 50 events) and pushed a toast. Coalescing keeps the signal (occurrence counts) while bounding the noise and the client-side cost.
- A connectivity problem on the user's side is not an application defect: it should not file an application-defect report or alert anyone, but it is still worth aggregate telemetry. The compact payload drops the stack, the `ex-data` dump (which can include request headers) and the last-events list.

## How

- **Governor:** pure decision function (`reserve-report*`) plus an atomic wrapper (`reserve-report!`). Fingerprint = `report-name|type|code|hint|first stack frame`; window = 2 minutes; `:occurrences` counts occurrences since the previous emission (first = 1); the fingerprint cache is bounded to 2000 entries, evicting the oldest.
- **Cheap suppression:** `flash` reserves the report before generating it, so suppressed occurrences never call `generate-report`; the toast behavior is unchanged.
- **Never drop a reserved report:** `generate-report` is total — if formatting fails it returns a minimal fallback string instead of `nil`, so a slot consumed by `flash` always materializes.
- **Report name is part of the fingerprint:** a `handled-exception` report cannot coalesce with an `unhandled-exception`/`exception-page` report of the same cause, because the latter two are the ones that reach error reports and alerts.
- **Environment failures:** `environment-error?` classifies them; `flash` derives the audit event name and the report format from the cause, so handlers cannot route an environment failure to the internal reports or pay for a full report. `:offline` gets its own handler, and `:bad-gateway`/`:service-unavailable`/`:nitrate-*` keep their page but submit `handled-exception`.
- **Compact reports:** `generate-report` accepts an explicit `{:format :compact|:full}` (default `:full`); `:compact` keeps the context header plus type/code/uri and skips the stack, the `ex-data` dump and the last-events list. Environment fingerprints drop the stack frame, so grouping does not depend on the internal call site.
- **Exception page:** `static.cljs` passes the cause so its report uses a cause-based fingerprint, and chooses the event name and payload format per cause.
- **Docs:** includes a small commit documenting the GitHub sub-issues procedure used to create this task's issue.

Closes #11726
Relates to #11724


.serena/memories/clojure/idioms.md
niwinz, line 1: Do not wrap lines

.serena/memories/frontend/workspace-state-persistence-subtleties.md
niwinz, line 1: do not wrap lines, and document on serena memories maintenance guide that lines should not be wrapped.

PR #11765, opened 2026-09-17
:sparkles: Export WASM SVG background blur via foreignObject

### LINKED ISSUE #11764: Export background blur to SVG with foreignObject backdrop-filter
### Description

Export background blur (frosted glass) to SVG using an XHTML `foreignObject` with CSS `backdrop-filter`, clipped to the shape silhouette (glyph outlines for text), so exported files keep the canvas glass effect in browsers.

### Use case

Shapes with background blur should still look frosted over content behind them when exported to SVG, matching the WASM canvas as closely as browser SVG allows.

### Affected versions

develop / upcoming with `wasm-export` and `render-wasm/v1`

superalex:

## What

- WASM SVG export now approximates **background blur** (frosted glass) so exports keep the canvas effect in browsers.

## Why

- Skia’s SVG backend has no backdrop filter, so background blur was skipped and SVG downloads lost the glass look that the GPU canvas already paints.

## How

- Emit an empty XHTML `foreignObject` with CSS `backdrop-filter: blur(σpx)` (`radius_to_sigma(value * scale)`, same mapping as layer blur).
- Clip on the **HTML div** with `clip-path: path(...)` (fill ∪ stroke silhouette; glyph outlines for text). SVG clip/mask on the FO empties the browser backdrop sample.
- Emit the FO before opacity / blend / layer-blur wrappers so glass ignores shape opacity (GPU order).
- Stacked on #11652 (`superalex-wasm-svg-masks`).

Closes #11764

AI-assisted-by: composer


render-wasm/src/render/svg/background_blur.rs
```diff
@@ -0,0 +1,173 @@
...
+    let y = page.top();
+    let w = page.width();
+    let h = page.height();
+
+    let local_clip = builder
+        .map_path_to_page(clip_path_local, draw_matrix)
+        .make_offset((-x, -y));
+    let d = skia_path_to_svg_d(&local_clip);
+
+    builder.push_raw(&format!(
```
elenatorro: we're always setting here `evenodd`, we should get the rule (evenodd or nonzero) depending on local_clip.fill_type() instead, and apply the rule on the clip-path

render-wasm/src/render/svg/background_blur.rs
```diff
@@ -0,0 +1,173 @@
...
+    let Some(blur) = element.visible_background_blur() else {
+        return Ok(());
+    };
+    if blur.value <= 0.0 {
+        return Ok(());
+    }
+
+    let sigma = radius_to_sigma(blur.value * scale);
+    let draw_matrix = builder.silhouette_draw_matrix(element);
+    let (clip, stroke_outset) = blur_clip_and_outset(element);
```
elenatorro: after this, if clip is empty, we can do an early return here so we don't need to call the emit_foreign_object function, since empty clips should have no blur

render-wasm/src/render/svg/background_blur.rs
```diff
@@ -0,0 +1,173 @@
...
+            None => path,
+            Some(acc) => acc.op(&path, skia::PathOp::Union).unwrap_or(acc),
+        });
+    }
+
+    let Some(base) = combined else {
+        return skia::Path::rect(element.selrect(), None);
+    };
+
+    expand_path_by_stroke_outset(base, stroke_outset)
```
elenatorro: We could reuse here `union_stroke_outset` , `expand_path_by_stroke_outset` is a duplication of this function

elenatorro (changes requested):

elenatorro (approved):

PR #11766, opened 2026-09-17
:zap: Improve canvas snapshot and nested drop shadows calculation

elenatorro:
### Related Ticket

No related ticket

### Summary

This PR improves performance in two different fronts:

1) Canvas snapshot

`createImageBitmap(canvas)` is a synchronous GPU readback in Firefox, not the async GPU-side operation. It ran every 250 ms on tiles-complete, blocking the main thread ~400 ms each time. Now it checks canvas-snapshot-capture-idle?, so: no render pending, no view interaction, no load, no capture in flight.

2) Shadow extrect calculation on nested drop shadows

`descendants_contained_for_frame_shadow` rejected any frame whose descendants had drop shadows, and tested containment against `child.extrect()`, which at the same times checked its own shadows. By calling just `silhouette_rect()`, the purpose is the same, but it's a lighter check: geometry, strokes, and layer blur, no own shadows.

It also adds some profiling settings to be able to check the problematic functions on the perf traces


frontend/src/app/render_wasm/api.cljs
```diff
@@ -623,6 +585,68 @@
...
+       (not (render-pending?))))
+
+(defonce ^:private schedule-canvas-snapshot-capture!
+  (fns/debounce
+   (fn []
+     (when (canvas-snapshot-capture-idle?)
+       (timers/schedule-on-idle
+        (fn []
+          ;; Re-check: a render may have started between the debounce and the
+          ;; idle slot. The next `tiles-complete` re-arms us.
```
superalex: This assumes another tiles-complete will retry us if the idle check fails. That is not always true: ViewportReady notifies once, then the interest-ring Full skips notify_tiles_render_complete! when viewport_presented is already set. If the debounce fires while still busy (pending-render), we drop the capture and never re-arm until some later gesture. On skip, probably we should re-schedule the debounce (or otherwise retry) instead of relying on the next notify.

render-wasm/src/shapes.rs
```diff
@@ -1167,6 +1167,28 @@ impl Shape {
         self.expand_own_bounds(bounds).to_rect()
     }
 
+    /// Geometry, strokes and layer blur in world space, without this shape's own
+    /// drop shadows: those belong to the shape, not to an ancestor's shadow mask.
+    pub fn silhouette_rect(&self) -> math::Rect {
```
superalex, line 1177: This duplicates the own_base_bounds + text selrect join from layer_bounds. Should we add a shared helper (expand without shadows, then world-transform) so those rules cannot drift later?

render-wasm/src/render.rs
```diff
@@ -3142,6 +3142,8 @@ impl RenderState {
 
     /// Renders descendant silhouettes into the current drop-shadow layer.
     #[allow(clippy::too_many_arguments)]
+    // Kept out of line so profiles can separate it from the element's own shadow.
+    #[inline(never)]
```
superalex: Will this apply to release mode too?

render-wasm/Cargo.toml
```diff
@@ -46,6 +46,15 @@ lto = "fat"
...
 
+[profile.profiling]
+inherits = "release"
+strip = false
+# Fat LTO inlines the shadow helpers into their caller, so a profile cannot
+# attribute cost to them. Thin LTO keeps those boundaries; the code is still
+# optimized, so the proportions stay representative.
+lto = "thin"
+codegen-units = 16
+
```
elenatorro, line 57: Note: this is not strictly related to this PR, but I'm using it for debugging and profiling and it can be useful: 

```
BUILD_MODE=profiling ./build frontend
```

PR #11768, opened 2026-09-17
:bug: Fix Japanese IME Enter duplication in comment input

### LINKED ISSUE #11757: Comment input duplicates the text when confirming a Japanese IME composition with Enter
### Steps To Reproduce

1. Use a Japanese IME (tested with the macOS built-in Japanese input source with Live Conversion). Any CJK IME that confirms the composition with Enter should reproduce it.
2. Open a file, switch to the comments tool (workspace or view mode) and click on the canvas to start a new comment (the same happens when replying to or editing an existing comment).
3. Type some Japanese text, e.g. `henkan` → the IME shows the candidate `変換`.
4. Press **Enter once** to confirm the IME composition (this Enter is a composition-confirm keystroke, not a "new line" keystroke; `KeyboardEvent.isComposing` is `true` and `keyCode` is `229`).

### Expected behavior

The composed text `変換` is inserted once and the caret stays at the end of it. Nothing else happens (the browser handles the composition-confirm Enter itself).

### Actual behavior

The text is duplicated and a line break plus a zero-width space are inserted between the copies. The editor ends up with:

```
変換
変換​
```

(`​` = U+200B zero-width space that the editor uses internally).

Cause: `handle-key-down` in `frontend/src/app/main/ui/comments.cljs` (`comment-input*`) treats every Enter as a "insert line break" key and does not check whether an IME composition is in progress:

https://github.com/penpot/penpot/blob/2.17.2/frontend/src/app/main/ui/comments.cljs#L391-L404

```clojure
(kbd/enter? event)
(let [sel (wapi/get-selection)
      range (.getRangeAt sel 0)]
  (dom/prevent-default event)
  (dom/stop-propagation event)
  (let [[span-node offset] (current-text-node node)]
    (.deleteContents range)
    (handle-input)
    (when span-node
      (let [txt (.-textContent span-node)]
        (dom/set-html! span-node (dm/str (dom/escape-html (subs txt 0 offset)) "\n" zero-width-space (dom/escape-html (subs txt offset))))
        (wapi/set-cursor! span-node (inc offset))
        (handle-input)))))
```

While the composition is still active, the handler rewrites the span's HTML (text + `\n` + zero-width space) and moves the caret; the IME then commits its composition text again on top of that, so the text appears twice. The same code is still present on `main`.

Suggested fix (one line): ignore the key event while a composition is in progress, before any of the `cond` branches, e.g.

```clojure
(fn [event]
  (when-not (.-isComposing (.-nativeEvent event))   ; or (= 229 (.-keyCode event)) for older browsers
    ...existing handler...))
```

I verified locally that adding this guard to the compiled bundle fixes the duplication and that a normal (non-composing) Enter still inserts a line break as before. This is the same class of issue that was fixed for the text editor in #10477 ("Fix line break after accepting IME option (Mac, Japanese)"); the comment editor needs the equivalent check.

### Screenshots or video

Rendered result in the comment box after typing `変換` and pressing Enter once to confirm the IME candidate:

```
変換
変換​
```

- OS: macOS 14.6.1 (Sonoma), Japanese input source with Live Conversion enabled
- Browser & version: Google Chrome 153.0.8010.47

- Host: local instance (self-hosted)
- OS Version: macOS 14.6.1 (Docker Desktop)
- Docker / Docker-compose version: Docker 29.4.0 / Docker Compose v5.1.2
- Image version: penpotapp/frontend:2.17 (Penpot 2.17.2, MCP 2.17.0)

### Additional context

The text editor (Draft.js based) had related IME issues fixed in #10469 / #10477 / #6442; the comment input is a separate contenteditable implementation and still lacks the composition check.

makesomethingshit:

### Related Ticket

Closes #11757

### Summary

Fix Japanese IME composition handling in the comment editor.

The comment input previously treated an Enter key used to confirm an active IME composition as a normal Penpot line-break command. This caused the editor to rewrite the content before the IME finished committing, duplicating the composed text and inserting an extra newline and zero-width space.

This PR:

- Ignores custom comment-input keyboard handling while an IME composition is active, using `nativeEvent.isComposing` with `keyCode === 229` as a fallback.
- Applies the same composition guard to the floating comment thread so a composing Escape does not close the thread.
- Preserves the existing behavior for normal Enter, modifier+Enter, Escape, mention navigation, and Backspace.
- Extracts the keyboard handling into testable functions without introducing separate composition state.
- Adds focused regression coverage for composing and non-composing keyboard paths.

### Steps to reproduce 

Before this fix:

1. Enable a Japanese IME, such as the macOS built-in Japanese input source with Live Conversion.
2. Open a file and create, reply to, or edit a comment.
3. Type Japanese text, for example `henkan`, and select/compose `変換`.
4. Press **Enter once** to confirm the IME composition.
5. Observe that the text is duplicated with an extra line break.

Expected result with this PR:

- The composition-confirm Enter is left entirely to the IME and inserts `変換` only once.
- A normal, non-composing Enter still inserts a line break.
- Escape during composition does not close the comment thread.
- Normal Escape still closes the thread.
- Mention keyboard navigation and Backspace behavior remain unchanged.

Automated verification:

- `clj-kondo` on the touched files: 0 errors, 0 warnings.
- `comment-input-ime-test`: regression coverage for composing Enter/Escape/arrows/Backspace, the `keyCode 229` fallback, plain Enter/mod+Enter, mention handling, and handler ordering.
- Existing `comments-clustering-test` and `workspace-comments-test` continue to pass.

A real macOS Japanese IME browser check is still recommended because the available automated tests simulate the relevant keyboard event properties rather than running an actual OS-level IME composition.

### Checklist

- [x] Include screenshots or videos, if applicable. *(N/A: no visual UI change.)*
- [x] Add or modify existing integration tests in case of bugs or new features, if applicable. *(Added focused frontend regression coverage.)*
- [x] Refactor any modified SCSS files following the refactor guide. *(N/A: no SCSS files modified.)*


belen-albeza:
I believe this is unrelated to the render engine or the text editor v3, since it's about comments

makesomethingshit:
You’re right — this is specific to the comment editor and isn’t related to the render engine or text editor v3.

I only referenced the previous IME fix as a precedent for the isComposing / keyCode 229 handling. I’ll remove that wording to avoid implying that the implementations are related.

Alotor:
Thanks for the contribution

PR #11777, opened 2026-09-18
:sparkles: Restrict optional RPC ids to user-provided UUIDs

### LINKED ISSUE #11778: Restrict optional ids to v4, v7 and v8 UUIDs in creation RPC endpoints
### Description

Creation endpoints that accept a client-provided id (files, teams, projects and media objects) now only accept v4, v7 and v8 UUIDs. Ids with any other version, including reserved ones such as v3, are rejected with a validation error. Previously any well-formed UUID was accepted.

### Use case

This keeps the id space consistent with the ids Penpot itself generates (v4 and v8), avoids collisions with reserved UUID versions, and still supports integrations that supply their own time-ordered v7 ids.

### Affected versions

2.18.0

niwinz:

## What

- Optional `:id` params on creation RPC commands now only accept v4, v7 and v8 UUIDs; reserved versions such as v3 are rejected at validation time.

## Why

- The shared validator accepted any well-formed UUID, so clients could create files, teams, projects and media objects with ids Penpot itself can never generate (v4/v8 only), polluting the id space with reserved versions.

## How

- **Common:** new `::sm/user-provided-uuid` schema backed by `uuid/user-provided?`, a version- and variant-aware regex (`[478]` version, `[89ab]` variant); instances only, mirroring `::sm/uuid` (strings decode into instances first).
- **Backend:** the seven creation schemas (`create-file`, `create-team`, `create-team-with-invitations`, `create-project`, three media-object commands) use the new type; the `get-team` read path keeps the lax type so historic data still reads.
- **Tests:** `user-provided?` and schema decode/validate coverage on JVM and JS runtimes; focused suites, cljfmt and clj-kondo green on `common` and `backend`.

Closes #11778


common/src/app/common/uuid.cljc
```diff
@@ -70,6 +70,20 @@
   [v]
   (= zero v))
 
+(def ^:private user-provided-regex
+  #"(?i)^[0-9a-f]{8}-[0-9a-f]{4}-[478][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$")
+
+(defn user-provided?
```
pabloalba, line 76: I don't like the name. This is a valid uuid, or a uuid that follows the RFCs. The fact that is used when a user provides it is anectotical.

Proposal: `valid-uuid?`

pabloalba: (obviously not only here, but on the schema and usages too)

pabloalba (approved):

PR #11784, opened 2026-09-18
:boom: Remove client-provided id from creation RPC commands

### LINKED ISSUE #11783: Remove optional client-provided id from creation RPC endpoints
### Description

The creation RPC commands (`create-file`, `create-project`, `create-team`, `create-team-with-invitations`, `upload-file-media-object`, `create-file-media-object-from-url` and `assemble-file-media-object`) accept an optional client-provided `id`. That parameter should be removed so the server always generates the identifier. Sending `id` to any of these commands will then fail params validation instead of being silently accepted.

### Use case

Nobody uses this parameter: neither the frontend, nor the import flows, nor the plugin harnesses send it, so it only adds attack surface and confusion (duplicate ids fail on some commands and silently return the existing object on others). Removing it simplifies the API contract and closes the door to id-spoofing on creation endpoints. This is a breaking change for third-party integrations that pass `id` today.

### Affected versions

Next release.

niwinz:

## What

- The seven creation RPC commands no longer accept an optional client-provided `id`: the server always generates the identifier and a sent `id` is ignored.
- Affected commands: `create-file`, `create-project`, `create-team`, `create-team-with-invitations`, `upload-file-media-object`, `create-file-media-object-from-url` and `assemble-file-media-object`.
- Closes #11783 (milestone 2.19.0).

## Why

- No in-repo caller ever sent `id` (verified across frontend, backend internals, `.penpot` import flows and plugin harnesses); only backend tests did, so the parameter only added attack surface (id-spoofing on creation) and inconsistent semantics (duplicate fails on some commands, silently returns the existing object on media).
- Sequenced after #11777 (v4/v7/v8 hardening), which ships to production first.

## How

- **Schemas:** deleted the optional `:id` entries from the seven params schemas.
- **Handlers:** malli maps are open and the RPC layer never strips unknown keys, so `create-file` and `create-project` explicitly drop a smuggled `:id` (their downstream honors explicit ids for internal callers); team and media always generate server-side now.
- **Internals preserved:** project duplicate and binfile import still pass remapped ids through `teams/create-project` (fallback kept with an explanatory comment).
- **Tests:** seven new `*-ignores-client-id` boundary tests; reworked the suites that passed `:id` (`files-crud`, `projects-simple-crud`, delete/restore flows, media idempotency tests); deleted the duplicate-id test whose behavior no longer exists.
- **Breaking change:** third-party integrations passing `id` will see it silently ignored instead of honored; announced via the 2.19.0 milestone and the generated changelog.
- **Verification:** backend suites rpc-file/project/team/media/quotes (135 tests, 1121 assertions) plus snapshot/thumbnail/storage/management/assets neighbors (107 tests, 1164 assertions) green; cljfmt and clj-kondo clean.

AI-assisted-by: muse-spark-1.3-contributor


pabloalba:
**F1 (High)** — `backend/test/backend_tests/helpers.clj` (`create-team*`)

`create-team*` now always uses `(uuid/next)` and ignores `:id`, but the test factory still builds a deterministic id and passes it:

```clojure
(let [id       (mk-uuid "team" i)
      features (cfeat/get-enabled-features cf/flags)]
  (teams/create-team {::db/conn conn} {:id id
                                       :profile-id profile-id
                                       :features features
                                       :name (str "team" i)}))
```

That `:id` is dead: callers that assume `(:id (th/create-team* i …))` equals `(mk-uuid "team" i)` get a random UUID instead. Unlike file/project, no production path remaps team ids through `create-team*`, so keep server-only generation and fix the factory.

Suggested change:

```clojure
(defn create-team*
  ([i params] (create-team* *system* i params))
  ([system i {:keys [profile-id] :as params}]
   (us/assert uuid? profile-id)
   (dm/with-open [conn (db/open system)]
     (let [features (cfeat/get-enabled-features cf/flags)]
       (teams/create-team {::db/conn conn} {:profile-id profile-id
                                            :features features
                                            :name (str "team" i)})))))
```

niwinz:
This PR depends on https://github.com/penpot/penpot/pull/11136, is the best approach correctly this. Once merged we just close the schemas and preserve the internal behavior for tests

pabloalba (approved):

PR #11788, opened 2026-09-18
:bug: Keep interactions of flex children when syncing components

### LINKED ISSUE #11746: Toggling the visibility of a child inside a flex-layout main component silently deletes the prototype interactions of its copies
#### Reproduction file
`interaction-loss-repro.zip` (attached) — contains `interaction-loss-repro.penpot` which is a one page penpot project with 10 shapes and interactions already wired.

```
Screen Main          board, flex column
  Ribbon             COPY of component "Ribbon", flex row
    Button           text, sub-shape of the copy   <- click: open overlay -> Popup
    Rectangle        rect, sub-shape of the copy   <- the shape whose visibility is toggled
Popup                board (plain)
  Close popup        text, belongs to NO component <- click: close overlay (self)
Ribbon               MAIN instance of "Ribbon", flex row, on the same page
  Button
  Rectangle
```

The `Ribbon` component is a **flex row**: `Button` and `Rectangle` are laid out side by side, so
hiding `Rectangle` repositions `Button`.

#### Steps to reproduce

1. Import the file and open it.
2. Open View mode on `Flow 1` and confirm the flow works: clicking `Button` opens the `Popup`
   overlay, clicking `Close popup` closes it. Return to the workspace.
3. Select the `Rectangle` inside the **main instance** of `Ribbon` (the bottom one).
4. Toggle its visibility off (eye icon in the Layers panel).
5. Select `Button` inside the `Ribbon` **copy** and look at its interactions: they are all gone, while the original one should still be there.

### Expected behavior

Hiding a shape in a main component propagates the visibility change to its copies and reflows their
layout. The prototype interactions carried by the copies must be left untouched.

### Actual behavior

The `open-overlay` interaction on `Button` is gone. The visibility change itself is propagated and
persisted correctly. Reloading confirms the interaction is permanently lost.

The `close-overlay` interaction on `Close popup` survives — that shape belongs to no component and is
not reflowed.

- Windows 11
- Chrome Version 153.0.8010.37

| | |
|---|---|
| Penpot version | 2.17.2 |
| Deployment | Self-hosted (Docker) |
| Browser | Chrome Version 153.0.8010.37 |
| WebGL rendering (Beta) | Disabled |

WebGL was disabled throughout, so this is unrelated to #11672. Persistence is healthy: `update-file`
requests are observed and design edits from the same session persist correctly.

Additionally to Penpot version 2.17.2, it was also tested on a local deploy from commit 76e9289fcfecb2ec60406c8eac739f4e5f554941. The same bug appears.

#### Summary

When a main component uses a **flex layout**, hiding or showing one of its children reflows its
siblings. That reflow propagates to every copy of the component — and silently deletes the prototype
interactions carried by the sibling shapes in those copies.

The visibility change itself is applied and persisted correctly. Only the interactions are lost. No
warning is shown, nothing appears in the console, and the loss survives a reload: the interactions
are gone from the persisted file.

The same component **without** a flex layout does not exhibit the problem: hiding a child then
changes no sibling geometry, and the interactions survive. The reflow is the trigger.

This is not the documented behaviour of removing interactions when a destination board is deleted
(Taiga #1656): no board is deleted, moved or ungrouped, and the destination board is untouched.

#### Impact

Editing a base component on a file whose prototype is already wired is a routine design-system
operation — for example adjusting a layout detail after a workshop where the prototype was presented
to a client in View mode. Since practically every component built for a real UI uses a flex layout,
almost any such edit silently destroys part of the prototype wiring, with no indication that
anything was lost.

The loss reaches copies on other boards, so its scope cannot be anticipated before making the edit.
The only reliable workaround is to attach interactions exclusively to shapes that belong to no
component, which defeats the purpose of a component library. For now, I just do not use interactions by fear of losing my job at any time.

#### Potentially related issues

- #5143 — Nested flex layouts don't get updated when toggling visibility (same trigger, different symptom)
- #4480 — Interactions are not carried through component copy operations
- #11723 — Component Interactions (Variants): interactions at component level are unsupported
- #3205 — Cannot add an interaction on a text element inside a flex layout
- #11546 — Workspace crash when removing multiple interactions rapidly
- #3512 / #3660 — Component instance synchronisation defects

ShreyashAgare26:

## What

- Hiding or showing a child inside a flex layout main component deleted the prototype interactions on its copies. The loss was saved to the file, so a reload did not bring them back.
- Interactions on copies now stay put when the main component changes.

**Before:**
pr-11788-before

**After:**
pr-11788-after

## Why

- When a flex component syncs, Penpot copies the flex child settings (margins, sizing, z-index...) from the main children to the copy children. It used the `swap-keep-attrs` list for that, and that list also contains `:interactions`.
- `:interactions` has no sync group, so the "don't overwrite what the user changed" check never applied to it. Every sync replaced the copy's interactions with the main's, which are usually empty.
- Components without flex layout never hit this path, which is why only flex components lost interactions.

## How

- The flex child sync now uses the same list without `:interactions`, in both directions (main to copy, and copy to main).
- `swap-keep-attrs` itself is unchanged, so component swaps still keep interactions as before.
- Added tests for main to copy sync, copy to main sync and reset overrides. Each one fails without the fix.

Closes #11746


madalenapmelo-kp:
Thanks for your contribution! We've received your pull request and assigned it for review. It may take us a little while to get to it, but we appreciate your patience and will be in touch as soon as we can!

A quick note: we welcome responsible use of AI tools, but please make sure any AI-assisted contributions are carefully reviewed by a human before submitting. PRs that don't meet our quality standards may be closed without detailed review.

PR #11789, opened 2026-09-18
:bug: Restore the missing handler when aligning a one-handler node

### LINKED ISSUE #11703: One handle, from aligned to equal
### Steps To Reproduce

1. Create a shape
2. Make a corner node a curve node
3. Remove one of the handles, so there is oonly one handle
4. Move the handle around
5. Now change the handle setting from independent to aligned
6. Nothing happens

### Expected behavior

Would like for the mirror handle to be created when this happens

### Actual behavior

Right now, because there is only one handle, nothing happens when changing it to aligned.

**For this to happen, we need to make any node that only has one handle, to be automatically an independent node. This way, whenever a one handled node is changed from independent to either aligned or equal, then a new mirrored handle will appear on the other side.**

This applies also to the shortcuts

### Screenshots or video

https://github.com/user-attachments/assets/c9d56344-568e-4c44-b88c-4386ef9b2618

_No response_

_No response_

_No response_

### Frontend Stack Trace

```console

```

### Backend Stack Trace

```console

```

### Additional context

_No response_

ShreyashAgare26:

## What

- After removing one handle from a curve node, switching the node to aligned or equal did nothing. Switching to equal could even remove the handle that was left.
- Now the node gets a mirrored handle on the other side. If that side is a straight line, the line becomes a curve.
- Removing a handle now marks the node as independent, as the issue asks.

https://github.com/user-attachments/assets/30ca9645-966d-43ca-b633-0d4035ccc42d

## Why

- Removing a handle either leaves it sitting on top of the node, or turns that side into a straight line. In both cases the node has only one real handle.
- Aligned and equal only adjust an existing opposite handle, so with nothing there they had nothing to adjust. Equal also sometimes picked the collapsed handle as its reference and collapsed the other one to match.

## How

- Before applying aligned or equal, a node with a single handle gets its opposite back: a collapsed handle is moved out to the mirrored position, and a straight line on that side is turned into a curve.
- This also works across the start node of a closed path.
- Independent is untouched, and nodes that already have two handles behave as before.
- One case is left as is: a closed path with no explicit closing segment, since fixing it means inserting a new segment.

Closes #11703


madalenapmelo-kp:
Thanks for your contribution! We've received your pull request and assigned it for review. It may take us a little while to get to it, but we appreciate your patience and will be in touch as soon as we can!

A quick note: we welcome responsible use of AI tools, but please make sure any AI-assisted contributions are carefully reviewed by a human before submitting. PRs that don't meet our quality standards may be closed without detailed review.

Alotor:
Thanks for your contribution! I've added a commit with some small fixes, once the CI passes we'll merge it.

